Issue #587, opened 2022-09-24
Dataset Viewer issue for nlphuji/winogavil


yonatanbitton:
### Link

https://huggingface.co/datasets/nlphuji/winogavil

### Description

The dataset viewer is not working for dataset nlphuji/winogavil.

Error details:

```
Error code:   UnexpectedError
Type is not JSON serializable: PngImageFile
```

**TLDR**: Is there an option to ignore a specific dataset field *only* in the Dataset Viewer, and to keep it as the part of the dataset otherwise? 

My dataset includes a field named "candidates" which is a list of strings, where each string is a filename reference. 
In the process script ([here](https://huggingface.co/datasets/nlphuji/winogavil/blob/main/winogavil.py#L116)) I transform each of the filenames to a png. 

This code:
```python
winogavil = load_dataset("nlphuji/winogavil", use_auth_token=auth_token)["test"]
print(winogavil[0]['candidates'])
print(winogavil[0]['candidates_images'])
```
returns the following:
```python
['eagle', 'teepee', 'factory', 'metal', 'ash']
[<PIL.PngImagePlugin.PngImageFile image mode=RGB size=2384x2384 at 0x7F6ABC6AA2D0>, <PIL.PngImagePlugin.PngImageFile image mode=RGB size=1500x1500 at 0x7F6ABC6AA390>, <PIL.PngImagePlugin.PngImageFile image mode=RGB size=1400x1400 at 0x7F6ABC6AAA50>, <PIL.PngImagePlugin.PngImageFile image mode=RGB size=2400x2400 at 0x7F6ABC6AAA10>, <PIL.PngImagePlugin.PngImageFile image mode=RGB size=900x900 at 0x7F6ABC6AA4D0>]
```

This is the intended use for my dataset, and the only gap is that it does not work in the Dataset Viewer.

Ideally, it will show each image name as string, and have an option to "click" it and to show image. 

- Do you have a support for a list of images in the Dataset Viewer?
- If not, is there an option just to eliminate the "candidate_images" field ONLY in the viewer, and keep it as a part of the dataset (with load_dataset)?

[This colab](https://colab.research.google.com/drive/19qcPovniLj2PiLlP75oFgsK-uhTr6SSi?usp=sharing) shows a usage example for my dataset with huggingface, if that helps. 

I managed to reach a solution by defining it as a list of strings rather than images:

```python
#"candidates_images": [img]
"candidates_images": [datasets.Value("string")]
```

[...]

PR #592, opened 2022-09-27
587 fix list of images or audio

severo:


severo:
Fixes #587.

Note that it fixes the error but that further work is needed in the Dataset Viewer on the Hub to show multiple images in the same cell. For now, the cell will be shown as an array of image URLs.

PR #595, opened 2022-09-29
ci: push the images to Docker Hub in the public organization hf

severo:


severo:
The images and the build cache are stored and published on the Docker Hub

Capture d’écran 2022-09-29 à 10 32 31

severo:
Thanks to the cache management provided by the Docker Hub (which was not available on AWS ECR), the build time is greatly reduced. For example, see https://github.com/huggingface/datasets-server/actions/runs/3149854138/jobs/5121957035: nothing had to be rebuilt, so the time is 13s!

PR #600, opened 2022-10-05
feat: 🎸 change the format of the image cells in /first-rows

severo:
Instead of returning a string with the URL, we now return an object with: src: the URL, height and width: the dimensions of the image in pixels.

BREAKING CHANGE: 🧨 the image cell format is now an object {src, height, width}


severo:
Beware: cannot be deployed before https://github.com/huggingface/moon-landing/pull/4031 is merged and deployed in order not to break the dataset viewer on the Hub.

lhoestq (approved):
Sounds good to me :)

Issue #602, opened 2022-10-10
RowsPostProcessingError in the viewer of serveral datasets


albertvillanova:
We find a `RowsPostProcessingError` in the viewer of several datasets:
```
Server error while post-processing the split rows. Please report the issue.

Error code:   RowsPostProcessingError
```
- https://huggingface.co/datasets/multi_woz_v22
- https://huggingface.co/datasets/qasper

severo:
It seems like the issue is that we don't support a Sequence of dicts adequately:

> A [datasets.Sequence](https://huggingface.co/docs/datasets/v2.5.2/en/package_reference/main_classes#datasets.Sequence) with a internal dictionary feature will be automatically converted into a dictionary of lists. This behavior is implemented to have a compatilbity layer with the TensorFlow Datasets library but may be un-wanted in some cases. If you don’t want this behavior, you can use a python list instead of the [datasets.Sequence](https://huggingface.co/docs/datasets/v2.5.2/en/package_reference/main_classes#datasets.Sequence).

https://huggingface.co/docs/datasets/v2.5.2/en/package_reference/main_classes#datasets.Features

severo:
#603 only fixed part of the datasets, ie https://huggingface.co/datasets/qasper

<img width="811" alt="Capture d’écran 2022-10-10 à 22 02 41" src="https://user-images.githubusercontent.com/1676121/194943667-52d3ed5e-668e-4ce9-8f27-3361e0a22c59.png">


But not https://huggingface.co/datasets/multi_woz_v22

<img width="806" alt="Capture d’écran 2022-10-10 à 22 02 58" src="https://user-images.githubusercontent.com/1676121/194943705-263eba24-d951-4d0e-b108-b1d228a410ba.png">


Working on it in #605

severo:
OK, #605 fixed https://huggingface.co/datasets/multi_woz_v22 as well

<img width="806" alt="Capture d’écran 2022-10-10 à 22 19 35" src="https://user-images.githubusercontent.com/1676121/194945962-6d25e8c2-4415-4291-baf4-1e95062121a2.png">

Still 29 datasets failing with `RowsPostProcessingError` -> it seems to be due to another reason: missing fields (columns), or None values, in some rows. https://github.com/huggingface/datasets-server/pull/606 will fix it

PR #603, opened 2022-10-10
Support Sequence of dicts

severo:
See #602

### DIFF
--- a/services/worker/src/worker/features.py
+++ b/services/worker/src/worker/features.py
@@ -152,7 +152,7 @@ def get_cell_value(
         # https://huggingface.co/docs/datasets/v2.5.1/en/package_reference/main_classes#datasets.Features
         if type(fieldType.feature) == dict:
             if type(cell) != dict or any(type(k) != list for k in cell.values()):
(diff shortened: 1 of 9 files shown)


services/worker/src/worker/features.py
```diff
@@ -147,41 +147,49 @@ def get_cell_value(
...
-                    fieldType.feature,
-                    assets_base_url,
-                    json_path + [idx] if json_path else [idx],
-                )
-                for (idx, subCell) in enumerate(cell)
-            ]
-        elif type(cell) == dict:
+        if type(fieldType.feature) == dict:
+            if type(cell) != dict or any(type(k) != list for k in cell.values()):
+                raise TypeError("The value of a Sequence of dicts should have be a dictionary of lists.")
```
albertvillanova, line 155:
```suggestion
                raise TypeError("The value of a Sequence of dicts should be a dictionary of lists.")
```

albertvillanova (approved):
The fix for the target issue seems good to me:
- #602

As you said, the CI failures are due to an unrelated issue: `requests.exceptions.HTTPError: 400 Client Error: Bad Request for url:...`

PR #633, opened 2022-11-08
ci: 🎡 remove the token for codecov since the repo is public

severo:


severo:
Seems to work

https://app.codecov.io/gh/huggingface/datasets-server/tree/fix-codecov-ci?displayType=tree

Capture d’écran 2022-11-08 à 20 28 39

PR #649, opened 2022-11-28
Small tweaks on Helm charts

n1t0:
- Ignore env files in packaged charts
- Add ability to completely specify `ingress.tls`
- Automatic value for `token.secretName` if not specified
- Do not submit `mongodb-migration` job if image not specified
- Add ability to specify `imagePullPolicy`

### DIFF
--- a/chart/templates/ingress.yaml
+++ b/chart/templates/ingress.yaml
@@ -6,12 +6,7 @@ metadata:
   name: {{ include "release" . }}
   namespace: {{ .Release.Namespace }}
 spec:
-  {{- if .Values.ingress.tls.enabled }}
-  tls:
-    - hosts:
-        - {{ .Values.hostname }}
-      secretName: {{ .Values.ingress.tls.secretName }}
-  {{- end}}
+  tls: {{ toYaml .Values.ingress.tls | nindent 4 }}
   rules:
     - host: {{ .Values.hostname }}
       http:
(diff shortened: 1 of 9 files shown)


chart/templates/ingress.yaml
```diff
@@ -6,12 +6,7 @@ metadata:
   name: {{ include "release" . }}
   namespace: {{ .Release.Namespace }}
 spec:
-  {{- if .Values.ingress.tls.enabled }}
-  tls:
-    - hosts:
-        - {{ .Values.hostname }}
-      secretName: {{ .Values.ingress.tls.secretName }}
-  {{- end}}
+  tls: {{ toYaml .Values.ingress.tls | nindent 4 }}
```
severo, line 9: `.values.ingress.tls` is not defined in env/dev.yaml. Is it OK to have an empty `tls:` section? or should we not include it instead?

chart/templates/jobs/mongodb-migration/job.yaml
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
 
+{{- if .Values.dockerImage.jobs.mongodbMigration }}
```
severo, line 4: what is the use case for not applying the migration scripts? my idea was always to check if a migration is needed (the migrations are stored in the database, so we don't apply the same migration twice or if not needed).

severo (approved):
Just two questions

severo:
Thanks @n1t0!

PR #669, opened 2022-12-23
chore: 🤖 speed-up docker build

severo:
in the next builds, the cache will be used if only src/ has been modified, which should help mostly with the workers/datasets_based image, because the poetry install of dependencies takes a lot of time.


severo:
Merging, as the remaining issue is unrelated

PR #676, opened 2023-01-11
fix: 🐛 only check webhook payload for what we are interested in

severo:
checking the payload in details (eg if the URL field is well formed) while not using it afterward is not useful. It even lead to breaking the webhook after https://github.com/huggingface/moon-landing/pull/4477 (internal link) where the "url" field format had changed.

cc @SBrandeis @coyotte508 fyi

### DIFF
--- a/chart/docker-images.yaml
+++ b/chart/docker-images.yaml
@@ -6,7 +6,7 @@
     },
     "services": {
       "admin": "huggingface/datasets-server-services-admin:sha-6a36caa",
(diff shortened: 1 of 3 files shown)


severo:
reference: https://huggingface.co/datasets/AresEkb/prof_standards_sbert_large_mt_nlu_ru/discussions/1#63beb28c20784381e8e25507

coyotte508 (approved):

PR #670, opened 2022-12-23
Create endpoint size

severo:

PR #677, opened 2023-01-13
Create children in generic worker

severo:
Extracting (from https://github.com/huggingface/datasets-server/pull/670) the logic to create the children jobs


severo:
I searched the [other apps](https://github.com/huggingface/moon-landing/blob/defc8225bc204dd69531fee8b78e870965923df7/server/models/App.ts#L12) code; it seems that datasets-server is the only app to use this token in the CI. So: no other impacted apps.

PR #687, opened 2023-01-20
feat: 🎸 quick and dirty POC for the random rows endpoint

severo:


services/api/src/api/routes/rows.py
```diff
@@ -0,0 +1,89 @@
...
+                parquet_files_paths = [
+                    f"{config}/{parquet_file['filename']}"
+                    for parquet_file in content["parquet_files"]
+                    if parquet_file["split"] == split and parquet_file["config"] == config
+                ]
+                # get the list of the first 100 rows
+                REVISION = "refs/convert/parquet"
+                fs = hffs.HfFileSystem(dataset, repo_type="dataset", revision=REVISION)
+                dataset = pq.ParquetDataset(parquet_files_paths, filesystem=fs)
+                table = dataset.read()
```
lhoestq, line 78: This downloads the full dataset and is not efficient.

lhoestq: I did this some time ago that can help:
```python
import numpy as np
import pyarrow as pa
import pyarrow.parquet as pq

def index(sources, fs=None):
    all_pf = [pq.ParquetFile(source, filesystem=fs) for source in sources]
    rg_offsets = np.cumsum([pf.metadata.row_group(i).num_rows for pf in all_pf for i in range(pf.num_row_groups)])
    rg_readers = [lambda: pf.read_row_group(i) for pf in all_pf for i in range(pf.num_row_groups)]
    return rg_offsets, rg_readers

def query(fromRow, toRow, rg_offsets, rg_readers):
    start_rg, end_rg = np.searchsorted(rg_offsets, [fromRow, toRow], side="right")
    pa_table = pa.concat_tables([rg_readers[i]() for i in range(start_rg, end_rg + 1)])
    offset = fromRow - rg_offsets[start_rg - 1] if start_rg else fromRow
    pa_table = pa_table.slice(offset, toRow - fromRow)
    rows = pa_table.to_pylist()
    return rows
```

At query time it downloads only the necessary row groups. Each row group contains 1000 elements by default though (defined in `datasets`), which may create some issues for datasets with big rows.

services/api/src/api/routes/rows.py
```diff
@@ -0,0 +1,89 @@
...
+                rows = table.take(list(range(fromRow, toRow))).to_pylist()
```
lhoestq, line 79: Also just FYI `pa.Table.slice()` is much more efficient than `pa.Table.take` for contiguous data ;)

services/api/src/api/routes/rows.py
lhoestq: Opened https://github.com/huggingface/datasets/pull/5565 to be able to control the row group size.

Right now it's always 1,000 rows - but to optimize random access for datasets with big rows (e.g. with images) we can reduce it

severo:
See #875

services/api/src/api/routes/rows.py
severo: https://github.com/huggingface/datasets-server/pull/875/commits/646501589f4e21644a6bf2db4a792d9831335adc

services/api/src/api/routes/rows.py
severo: Thanks. I started to implement it in #875 (well: to copy your code)

PR #694, opened 2023-01-23
feat: 🎸 replace Queue.add_job with Queue.upsert_job

severo:
upsert_job ensures only one waiting job for the same set of parameters. On every call to upsert_job, all the previous waiting jobs for the same set of parameters are canceled, and a new one is created with a new "created_at" date, which means it will be put at the end of the queue. It will help to fight against datasets that are updated very often (eg, every minute): they will be treated only when there are available workers.

It's a quick PR to fix the issue that the queues are increasing faster than the availability of the workers and that most of these jobs will be later skipped. Better to reduce the number of results in the queries (by reducing the number of waiting jobs).


severo:
See for example: https://huggingface.co/datasets/atokforps/latent_v1_fullrun_alpha3_13/commits/main. Two commits are pushed every minute. The queues contain thousands of waiting jobs.

Capture d’écran 2023-01-23 à 16 18 11

severo:
The issue is effectively fixed.

Capture d’écran 2023-01-23 à 16 46 04

For the "/first-rows" queue, we will have first to wait for the /splits to be run again on them

Issue #701, opened 2023-01-25
Dataset Viewer issue for bigbio/pubmed_qa


lhoestq:
### Link

https://huggingface.co/datasets/bigbio/pubmed_qa

### Description

this config fails to return the split names: pubmed_qa_labeled_fold6_bigbio_qa

but we get an error even though the other configs work

severo:
somewhat related to #208

severo:
We could fix this conceptually by "splitting" the "splits" processing step into:
- "configs", and expose the configs with `/configs` endpoint,
- then launch N "splits" jobs (one by config), and expose `/splits-in-config` endpoint
The thing is that, if we implement this, we don't know how to "fan-in" to get the response of /splits for all the configs, ie: wait for all the `/splits-in-config` jobs to have finished, then create the `/splits` response by concatening them all.
We could simply compute the response dynamically on every call, but the response time would be worse (in that case we could add another "cache" layer in front of the API server - we already had it in nginx, and we removed it later to make everything simpler)

lhoestq:
`/splits?dataset=...` should keep working since there is a default config, but one should be able to specify another config, maybe with `/splits?dataset=...&config_name=...`

severo:
> /splits?dataset=... should keep working since there is a default config

By the way, currently `/splits` gives the list of all the splits of all the configs, eg https://datasets-server.huggingface.co/splits?dataset=duorc gives:

```json
{
  "splits": [
    {
      "dataset": "duorc",
      "config": "SelfRC",
      "split": "train",
      "num_bytes": 239852925,
      "num_examples": 60721
    },
    {
      "dataset": "duorc",
      "config": "SelfRC",
      "split": "validation",
      "num_bytes": 51662575,
      "num_examples": 12961
    },
    {
      "dataset": "duorc",
      "config": "SelfRC",
      "split": "test",
      "num_bytes": 49142766,
      "num_examples": 12559
    },
    {
      "dataset": "duorc",
      "config": "ParaphraseRC",
      "split": "train",
      "num_bytes": 496683105,
      "num_examples": 69524
    },
    {
      "dataset": "duorc",
      "config": "ParaphraseRC",
      "split": "validation",
      "num_bytes": 106510545,
      "num_examples": 15591
    },
    {
      "dataset": "duorc",
      "config": "ParaphraseRC",
      "split": "test",
      "num_bytes": 115215816,
      "num_examples": 15857
    }
  ]
}
```

[...]

PR #702, opened 2023-01-25
Configs and splits

severo:
See https://github.com/huggingface/datasets-server/issues/701

This PR does:
- create a new endpoint `/config-names`. It gives the list of config names for a dataset
- create a new endpoint `/split-names`. It gives the list of split names for a config (not for a dataset, which is the difference with /splits for now). Note that /split-names, as /splits, may depend on the streaming mode and might fail for this reason.
- introduce a new "input-type": "config" in the processing steps. Before, a processing step was "dataset" (the input is a  dataset name) or "split" (the inputs are dataset, config, and split). Now, we enable "config" (the inputs are dataset and config) for the new endpoint `/split-names`.

Once this PR is merged, the plan is to:
- fill the cache for the two new endpoints for all the datasets on the Hub (might take some hours to complete),
- create a PR on moonlanding to use these two new endpoints instead of `/splits`
- remove the `/splits` processing step, delete the cache for this endpoint, make the endpoint an alias to `/split-names` and add the ability for `/split-names` to concatenate the responses for all the configs if the only parameter is `dataset` (no `config` passed)


chart/docker-images.yaml
```diff
@@ -2,14 +2,14 @@
...
     },
     "services": {
-      "admin": "huggingface/datasets-server-services-admin:sha-5364f81",
-      "api": "huggingface/datasets-server-services-api:sha-5364f81"
+      "admin": "huggingface/datasets-server-services-admin:sha-4dc0383",
+      "api": "huggingface/datasets-server-services-api:sha-4dc0383"
     },
     "workers": {
-      "datasets_based": "huggingface/datasets-server-workers-datasets_based:sha-5364f81"
+      "datasets_based": "huggingface/datasets-server-workers-datasets_based:sha-690d1cd"
```
lhoestq, line 12: This is to update which docker image to deploy right ?
I understand you have to update admin, api and datasets_based (you did changes to all of them) but why do you update mongodbMigration ?

chart/env/dev.yaml
```diff
@@ -86,6 +86,22 @@ api:
...
+    limits:
+      cpu: 1
+
+splitsNames:
+  replicas: 1
+  resources:
+    requests:
+      cpu: 0.01
+    limits:
+      cpu: 1
```
lhoestq, line 103: this is in case we want to deploy a dev cluster on ephemeral ?

tools/docker-compose-datasets-server.yml
```diff
@@ -69,6 +69,40 @@ services:
...
+      - splits-modules-cache:${HF_DATASETS_CACHE-/modules-cache}:rw
+      - splits-numba-cache:${NUMBA_CACHE_DIR-/numba-cache}:rw
+    extends:
+      file: docker-compose-base.yml
+      service: datasets-worker
+    environment:
+      DATASETS_BASED_ENDPOINT: "/split-names" # hard-coded
+    depends_on:
+      - mongodb
+    restart: always
```
lhoestq, line 105: we can use this for a local dev env right ?

libs/libcommon/src/libcommon/config.py
```diff
@@ -143,7 +143,9 @@ def from_env() -> "WorkerLoopConfig":
 class ProcessingGraphConfig:
     specification: ProcessingGraphSpecification = field(
         default_factory=lambda: {
-            "/splits": {"input_type": "dataset", "required_by_dataset_viewer": True},
+            "/config-names": {"input_type": "dataset"},
+            "/split-names": {"input_type": "config", "requires": "/config-names"},
+            "/splits": {"input_type": "dataset", "required_by_dataset_viewer": True},  # to be deprecated
```
lhoestq, line 148: This is where you define the dependencies between the processing steps ?

workers/datasets_based/src/datasets_based/workers/split_names.py
```diff
@@ -0,0 +1,132 @@
+# SPDX-License-Identifier: Apache-2.0
```
lhoestq, line 1: (just curious) why did you choose to add a new /split-names while you could have just extended /splits to support a config= query parameter ?

workers/datasets_based/tests/workers/test_config_names.py
```diff
@@ -0,0 +1,112 @@
...
+    assert worker.should_skip_job() is True
+    worker = get_worker(dataset, app_config, force=True)
+    assert worker.should_skip_job() is False
+
+
+def test_process(app_config: AppConfig, hub_public_csv: str) -> None:
+    dataset = hub_public_csv
+    worker = get_worker(dataset, app_config)
+    assert worker.process() is True
+    cached_response = get_response(kind=worker.processing_step.cache_kind, dataset=hub_public_csv)
```
lhoestq, line 50: is it using an actual mongo database for the cache when running tests ?

workers/datasets_based/tests/workers/test_config_names.py
```diff
@@ -0,0 +1,112 @@
...
+        ("gated", True, None, None),
+        ("private", True, None, None),
+        ("empty", False, "EmptyDatasetError", "EmptyDatasetError"),
+        # should we really test the following cases?
+        # The assumption is that the dataset exists and is accessible with the token
+        ("does_not_exist", False, "ConfigNamesError", "FileNotFoundError"),
+        ("gated", False, "ConfigNamesError", "FileNotFoundError"),
+        ("private", False, "ConfigNamesError", "FileNotFoundError"),
+    ],
+)
```
lhoestq, line 82: that's what I call testing <3

lhoestq (commented):
I learnt a lot, thanks ! I added come comments and questions - mostly for my personal knowledge, I don't have any comment on the code itself :)

workers/datasets_based/src/datasets_based/workers/split_names.py
```diff
@@ -0,0 +1,132 @@
...
+            by a `/`.
+        config (`str`):
+            A configuration name.
+        hf_token (`str`, *optional*):
+            An authentication token (See https://huggingface.co/settings/token)
+    Returns:
+        `SplitNamesResponseContent`: An object with the list of split names for the dataset and config.
+    <Tip>
+    Raises the following errors:
+        - [`~workers.splits.EmptyDatasetError`]
```
AndreaFrancis, line 91: Should this be workers.split_names.EmptyDatasetError instead?

workers/datasets_based/src/datasets_based/workers/split_names.py
```diff
@@ -0,0 +1,132 @@
...
+          The dataset is empty.
+        - [`~workers.splits.SplitsNamesError`]
```
AndreaFrancis, line 93: Should this be workers.split_names.SplitsNamesError instead?

workers/datasets_based/src/datasets_based/workers/split_names.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class SplitNamesWorker(DatasetsBasedWorker):
+    @staticmethod
+    def get_job_type() -> str:
+        return "/split-names"
+
+    @staticmethod
+    def get_version() -> str:
+        return "2.0.0"
```
AndreaFrancis, line 119: I am not sure but since this is a new route, why it is not 1.0.0?

AndreaFrancis (commented):
I put a couple of minor comments/questions please review if they are valid or not. Also I noticed there is no new test on e2e for the new routes, are they needed?

workers/datasets_based/src/datasets_based/workers/split_names.py
severo: copy/paste from split.py, and I forgot to change. Thanks

chart/env/dev.yaml
severo: Yes, that's true

tools/docker-compose-datasets-server.yml
severo: yes, you can launch `make start` at the root, and it will use this docker compose file to deploy all the images.

Until now, I've not really used a "local dev env":
- if I want to try something, I generally create a test (be it a unit test if it only affects a service or a worker, or an e2e test if I want to test the relation between them)
- alternately, I just go to one of the services or worker, launch `poetry run python` then interact with the prompt to test the methods.
- if I need to test the whole app manually, I deploy on the ephemeral k8s cluster

But now that we are various people working on the project, the last point is not really a good idea anymore, so, it would be better to launch locally when one wants to test the whole app manually.

libs/libcommon/src/libcommon/config.py
severo: Yes! It's a bit hidden for now, and it's hard-coded even if it seems to be a config option. I always thought of this as a temporal workaround, but I'm unsure how we want this to evolve.

workers/datasets_based/src/datasets_based/workers/split_names.py
severo: I find it a lot easier to create a new endpoint, populate its cache, then switch ([blue/ green method](https://www.redhat.com/en/topics/devops/what-is-blue-green-deployment)) than having to manage two different logics at the same time.

workers/datasets_based/tests/workers/test_config_names.py
severo: Yes:
- we launch `make up` before launching the test, then `make down`
https://github.com/huggingface/datasets-server/blob/e32590450a1f0c33439b41bb5b0b8a44f58fd9db/tools/PythonTest.mk#L4-L7
- which corresponds to launching the docker compose described in the `DOCKER_COMPOSE` variable
https://github.com/huggingface/datasets-server/blob/e32590450a1f0c33439b41bb5b0b8a44f58fd9db/tools/Docker.mk#L1-L7
- which turns to be only a mongo container:
https://github.com/huggingface/datasets-server/blob/e32590450a1f0c33439b41bb5b0b8a44f58fd9db/services/api/Makefile#L6-L7
- see:
https://github.com/huggingface/datasets-server/blob/e32590450a1f0c33439b41bb5b0b8a44f58fd9db/tools/docker-compose-mongo.yml#L1-L10

Every service or worker has a different mongo container with a different exposed port, so that we can run multiple tests at the same time (just in case).

Also: in the tests, we drop the content of the base before each test.

workers/datasets_based/tests/workers/test_config_names.py
severo: Possibly it's too much and redundant. We could surely optimize the testing time.

severo:
Thanks for your comments! I'm merging and deploying.

severo:
In prod:

- https://datasets-server.huggingface.co/config-names?dataset=glue
- https://datasets-server.huggingface.co/split-names?dataset=glue&config=ax

I'll add an admin endpoint to backfill the cache:
- get the list of all the public datasets
- find the cache misses
- launch a dataset update for them

severo:
> Also I noticed there is no new test on e2e for the new routes, are they needed?

I don't know if all the current e2e are needed. I only added to test_11_auth.py, which already checks that all the responses have been created. The unit tests should already cover more details, or if we have integration bugs, we should create a new e2e test for that specifically. But as such, I'm afraid we already have some redundant tests (splits, first rows).

PR #705, opened 2023-01-26
Add priority field to queue

severo:
In this PR:
- the jobs have a new field called `priority`: `normal` (default) or `low`
- the `normal` jobs are fetched before the `low` ones (but respecting `max_jobs_per_namespace`)
- when updating a job, i.e., if a dataset has been updated again, the priority is inherited, never lowered, even if the priority field is set. This way: when launching backfill (with a low priority) on a dataset that was already waiting for update, we don't deprioritize it.
- same logic with the existing "force" field: when updating a job, if force was true, we maintain it.

### DIFF
--- a/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230126164900_queue_job_add_priority.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230126164900_queue_job_add_priority.py
@@ -36,8 +36,8 @@ class MigrationAddPriorityToJob(Migration):
                 raise ValueError("priority should be 'normal'")
 
         check_documents(DocCls=JobSnapshot, sample_size=10, custom_validation=custom_validation)
-        if JobSnapshot.objects(force=False).count() != JobSnapshot.objects.count():
-            raise ValueError('All the objects should have the "force" field, set to False')
+        if JobSnapshot.objects(priority=Priority.NORMAL).count() != JobSnapshot.objects.count():
+            raise ValueError('All the objects should have the "priority" field, set to "normal"')
 
 
 # --- JobSnapshot ---
--- a/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230126164900_queue_job_add_priority.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230126164900_queue_job_add_priority.py
@@ -1,0 +1,123 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 The HuggingFace Authors.
+
+import enum
+import logging
+import types
+from typing import Generic, Type, TypeVar
+
+from mongoengine import Document
+from mongoengine.connection import get_db
+from mongoengine.fields import BooleanField, DateTimeField, EnumField, StringField
+from mongoengine.queryset.queryset import QuerySet
+
+from mongodb_migration.check import check_documents
+from mongodb_migration.migration import Migration
+
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationAddPriorityToJob(Migration):
+    def up(self) -> None:
+        # See https://docs.mongoengine.org/guide/migration.html#example-1-addition-of-a-field
+        logging.info("Add the priority field, with the default value ('normal'), to all the jobs")
+        db = get_db("queue")
+        db["jobsBlue"].update_many({}, {"$set": {"priority": "normal"}})
+
+    def down(self) -> None:
+        logging.info("Remove the priority field from all the jobs")
+        db = get_db("queue")
+        db["jobsBlue"].update_many({}, {"$unset": {"priority": ""}})
+
+    def validate(self) -> None:
+        logging.info("Ensure that a random selection of jobs have the 'priority' field set to 'normal'")
+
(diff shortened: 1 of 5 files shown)


jobs/mongodb_migration/src/mongodb_migration/migrations/_20230126164900_queue_job_add_priority.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def validate(self) -> None:
+        logging.info("Ensure that a random selection of jobs have the 'priority' field set to 'normal'")
+
+        def custom_validation(doc: JobSnapshot) -> None:
+            if doc.priority != "normal":
+                raise ValueError("priority should be 'normal'")
+
+        check_documents(DocCls=JobSnapshot, sample_size=10, custom_validation=custom_validation)
+        if JobSnapshot.objects(force=False).count() != JobSnapshot.objects.count():
```
AndreaFrancis, line 39: Shouldn't it be priority="normal" ?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230126164900_queue_job_add_priority.py
```diff
@@ -0,0 +1,123 @@
...
+            raise ValueError('All the objects should have the "force" field, set to False')
```
AndreaFrancis, line 40: Also here 'All the objects should have the "priority" field, set to "normal"'

AndreaFrancis (commented):
I put a couple of comments regarding migration validation, besides that, everything looks good!

AndreaFrancis (approved):

severo:
I did a rebase on main after #706, hence the force push. And I'll have to update the docker images with the new value ([3de1315](https://github.com/huggingface/datasets-server/pull/705/commits/3de1315b25faa71afbe39b15900cc696803b202f))

AndreaFrancis (approved):

severo:
Thanks for the review!

severo:
Interesting: the db migration job failed, and rolled back to the previous state (I have tests, but hadn't seen it live 😅) it seems to work as intended:

The deployment failed because the migration failed, ie. the new images have not been deployed because the database is not ready:
```
make upgrade-prod
make[1]: Entering directory '/home/slesage/hf/datasets-server/chart'
helm upgrade --install datasets-server-prod . --values docker-images.yaml --values env/prod.yaml -n datasets-server
Error: UPGRADE FAILED: pre-upgrade hooks failed: timed out waiting for the condition
make[1]: *** [Makefile:18: upgrade] Error 1
make[1]: Leaving directory '/home/slesage/hf/datasets-server/chart'
make: *** [Makefile:42: upgrade-prod] Error 2
```

The logs of the migration job:

```
INFO: 2023-01-26 18:22:15,442 - root - Start migrations
INFO: 2023-01-26 18:22:15,596 - root - 3 migrations have already been applied. They will be skipped.
INFO: 2023-01-26 18:22:15,596 - root - Migrate 20230126164900: add to the migrations collection
INFO: 2023-01-26 18:22:15,605 - root - Migrate 20230126164900: apply
INFO: 2023-01-26 18:22:15,606 - root - Add the priority field, with the default value ('normal'), to all the jobs
INFO: 2023-01-26 18:25:00,630 - root - Migrate 20230126164900: validate
INFO: 2023-01-26 18:25:00,630 - root - Ensure that a random selection of jobs have the 'priority' field set to 'normal'
ERROR: 2023-01-26 18:25:54,426 - root - Validation error on document 636e0dc52af91cd735e0ad3b: priority should be 'normal'. Document: {"_id": {"$oid": "636e0dc52af91cd735e0ad3b"}, "type": "/splits", "dataset": "bigscience-data/roots_ar_tashkeela", "unicity_id": "Job[/splits][bigscience-data/roots_ar_tashkeela][None][None]", "namespace": "bigscience-data", "force": false, "priority": "normal", "status": "success", "created_at": {"$date": 1668156869463}, "started_at": {"$date": 1668156870203}, "finished_at": {"$date": 1668156871088}}
ERROR: 2023-01-26 18:25:54,427 - root - Migration failed: priority should be 'normal'
INFO: 2023-01-26 18:25:54,427 - root - Start rollback
INFO: 2023-01-26 18:25:54,427 - root - Rollback 20230126164900: roll back
INFO: 2023-01-26 18:25:54,427 - root - Remove the priority field from all the jobs
INFO: 2023-01-26 18:33:29,814 - root - Rollback 20230126164900: removed from the migrations collection
INFO: 2023-01-26 18:33:29,819 - root - Rollback 20230126164900: done
INFO: 2023-01-26 18:33:29,820 - root - All executed migrations have been rolled back
```

By the way, the bug is in the validation function, not in the migration itself... Nevermind, I'll fix it (comparison to enum object instead of enum value)

Another comment: it takes 3 minutes to add the field to all the jobs in the db! (and 8 minutes to delete!!) It's a lot. It's due to the size of the queue with the past jobs. At one time, I created too many useless jobs that were skipped afterward, and we have several billions of them that we will never use... Maybe just cleaning the queue and starting from scratch would be better.

Capture d’écran 2023-01-26 à 19 33 52

severo:
fixed in https://github.com/huggingface/datasets-server/pull/707

PR #708, opened 2023-01-26
feat: 🎸 add a /backfill admin endpoint

severo:
The logic is very basic: it updates all the datasets of the Hub, with a low priority. Note that most of the jobs will be skipped, because the response will already be in the cache.

We might want to take a more detailed approach later to reduce the number of unnecessary jobs by specifically creating jobs for the missing data only.

Apart of this, the PR also fixes the creation of children jobs: the priority is preserved (ie low priority jobs created low priority children jobs)

### DIFF
--- a/chart/docker-images.yaml
+++ b/chart/docker-images.yaml
@@ -2,14 +2,14 @@
   "dockerImage": {
     "reverseProxy": "docker.io/nginx:1.20",
     "jobs": {
-      "mongodbMigration": "huggingface/datasets-server-jobs-mongodb_migration:sha-48dee06"
+      "mongodbMigration": "huggingface/datasets-server-jobs-mongodb_migration:sha-1c9e36d"
     },
     "services": {
-      "admin": "huggingface/datasets-server-services-admin:sha-3de1315",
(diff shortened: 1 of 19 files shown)


severo:
Merging...

severo:
Followup: it does not work because the loop that creates the jobs one by one takes too long and the nginx reverse-proxy returns a gateway timeout error.

severo:
Alternatives:
- reduce the duration by doing less checks (don't double-check if the dataset is private) and create the jobs as a batch.
- find a way to increase the timeout for this endpoint
- return immediately, and launch the update asynchronously
- make /backfill be another kind of job

I think the first option is the simplest.

severo:
OK. It's still a lot too slow, and it still timeouts:

```
$ curl -H 'Authorization: Bearer hf_...' -X POST "https://datasets-server.huggingface.co/admin/backfill"


<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx/1.20.2</center>
</body>
</html>
```

PR #711, opened 2023-01-27
Update poetry lock file format to 2.0

### LINKED ISSUE #710: Update poetry lock file format to 2.0
Since `poetry` 1.3.0 (9 Dec 2022), a new lock file format is used (version 2.0). See release notes: https://github.com/python-poetry/poetry/releases/tag/1.3.0

We should update our poetry lock files to the new format.

albertvillanova:
This PR locks poetry files with format 2.0.

Close #710.


severo:
Nice!

Three comments:
- by launching `poetry lock`, you updated the format of the lock file (because they were generated with poetry 1.2, and you are using poetry 1.3), but you also updated the minor versions of the dependencies. It should be OK, but it's good to know that side effect of `poetry lock`.
- I didn't find a way to force the use of a given version of poetry (ie >=1.3.2). It's a problem because if I run again poetry lock with an older version, I would revert the change. So: we have to ensure everybody (@huggingface/datasets-server) is using poetry >= 1.3.2. 
- Please, could you also update the poetry version both in github actions and in dockerfiles?

severo (changes requested):
Thanks for the PR

Please change the poetry version in github actions and dockerfiles.

severo:
Also, it's weird that the CI has not been raised by the changes in poetry.lock. I think it's because the PR comes from a fork, hence the github actions that are triggered by `push` are not run:

https://github.com/huggingface/datasets-server/blob/290f5be529d8c040b5638437cf25fa2fc6da0a53/.github/workflows/j-migration-build-docker.yml#L7-L11

We should add `pr:` below `push:` I think

albertvillanova:
Yes, let me fix the triggering. We have it like that also in `datasets`.

severo:
Not sure if we have to copy the list of paths, or if the same list will apply to both `push:` and `pull-request:`

severo:
> This branch is out-of-date with the base branch

Maybe you have to rebase, by the way...

albertvillanova:
@severo do you think it is OK to fix the CI in this PR? Or better in other PR and the merge that into this one?

I ask because the topic of each are different: update poetry lock format, vs. fix GitHub Actions for forks.

severo:
Please to it here too, it's minor, and this way we ensure the CI is run on this PR.

severo:
ah, OK, you can also do another PR on this branch, you're right

severo (changes requested):
Missing: the CI for PR from a fork

severo (approved):

severo:
Closing, as it has been replaced by #714 which is not affected by #712.

PR #713, opened 2023-01-27
Trigger CI by PRs from forks

### LINKED ISSUE #712: Fix CI for PR from a fork
The CI does not run properly when a PR is made from a fork.

albertvillanova:
Fix #712.

@severo could you please check we have all the required workflows (and no more than the required ones) to be triggered by PRs from forks?


albertvillanova:
@severo maybe some workflows should only be triggered when the PR is merged?

severo:
OK, it seems good; But there is an issue with the docker push to docker Hub. Indeed, I'm not sure how we should manage the creation and push of the docker images:
- we need them to run the e2e test
- but I don't know if we want to mix all the docker images from all the PR in the same repo: https://hub.docker.com/r/huggingface/datasets-server-services-admin/tags?

albertvillanova:
Yes... we have to find a solution for this... Let me think...

severo:
> @severo maybe some workflows should only be triggered when the PR is merged?

Yes. But we lose the ability to run e2e for the PR if we build the docker images only when the PR is merged... One way is to build the docker images as part of the e2e tests, without sending to the Docker Hub. The other way is to find a way to send to Docker Hub in this case too.

severo:
The issue is that `secrets.DOCKERHUB_USERNAME` and `secrets.DOCKERHUB_PASSWORD` do not seem to be available for PR from forks

severo:
If we take to option to build the docker images as part of the e2e tests, the code is already here but commented (beware, there might be some missing parts, though): https://github.com/huggingface/datasets-server/blob/6e63de2ba027a593095fda810eba4e672aa09976/tools/docker-compose-datasets-server.yml#L21-L23 for example.

The issue is that the e2e tests would really take a looooog time

severo:
We might also... just put this PR on hold and don't use forks from external repos for now? Then look with more time how to fix the CI?

albertvillanova:
Yes, secrets are not available on `pull_request` triggering for security reasons.

OK, we can put this on hod for the moment, but we need to find a way to allow the community to contribute to this project.

severo:
Could you rebase? It should be good after that. See https://github.com/huggingface/datasets-server/issues/712#issuecomment-1407151098

severo (approved):

albertvillanova:
@severo, I've merged main into this.
Please let me know if some additional workflow should be modified to be triggered by pull_request.
For the moment, I have ignored:
- build_push_docker_hub

severo:
exactly, build_push_docker_hub will only be run when the PR is merged and pushed to main.

Ready to be merged.

severo (changes requested):
Missing:

could you just fix  `j-mongodb_migration.yml` -> `j-migration.yml` in j-migration.yml (twice)

severo (approved):
Nice, thanks

albertvillanova:
@severo again an issue with codecov: red test.
Is it OK to merge anyway?

severo:
Yes, no worry. Maybe we should rev-move the codecov check, or find how to solve it, it's annoying

PR #720, opened 2023-01-30
fix: 🐛 don't check if dataset is supported when we know it is

severo:
As we are running a loop of updates on supported datasets, it's useless to check if the dataset is supported inside the `update_dataset` method.

### DIFF
--- a/.github/workflows/_unit-tests-python.yml
+++ b/.github/workflows/_unit-tests-python.yml
@@ -59,5 +59,5 @@ jobs:
         with:
           working-directory: ${{ inputs.working-directory }}
           files: ./coverage.xml
-          fail_ci_if_error: true
+          fail_ci_if_error: false
           flags: ${{ env.codecov_flag }}
--- a/libs/libcommon/src/libcommon/operations.py
+++ b/libs/libcommon/src/libcommon/operations.py
@@ -26,7 +26,7 @@ def update_dataset(
     hf_token: Optional[str] = None,
     force: bool = False,
     priority: Priority = Priority.NORMAL,
-    skip_check_support: bool = False,
+    do_check_support: bool = True,
 ) -> None:
     """
     Update a dataset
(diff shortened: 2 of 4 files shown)


severo:
codecov upload is failing... it's not related to the PR

AndreaFrancis (approved):
Looks good

libs/libcommon/src/libcommon/operations.py
```diff
@@ -26,6 +26,7 @@ def update_dataset(
     hf_token: Optional[str] = None,
     force: bool = False,
     priority: Priority = Priority.NORMAL,
+    skip_check_support: bool = False,
```
albertvillanova, line 29: We have recently had a discussion in `datasets` about parameters with reversed semantic meaning...

What about reversing the logic? Something like `do_check_support` and being True by default?

albertvillanova (approved):
Thank you for removing the unnecessary check.

libs/libcommon/src/libcommon/operations.py
albertvillanova: Or `support_checking`...

.github/workflows/_unit-tests-python.yml
```diff
@@ -59,5 +59,5 @@ jobs:
         with:
           working-directory: ${{ inputs.working-directory }}
           files: ./coverage.xml
-          fail_ci_if_error: true
+          fail_ci_if_error: false
```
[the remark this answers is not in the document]
severo: Yes, I used a separate PR for that: https://github.com/huggingface/datasets-server/pull/720#issuecomment-1408447822, and merged it here...
Maybe it would have been better to send to main... Well, next time

libs/libcommon/src/libcommon/operations.py
severo: OK, you're right

severo: done with https://github.com/huggingface/datasets-server/pull/720/commits/52d2cf6e58c15443692fd53577408daaf636ad0f

albertvillanova (approved):
Thanks!

severo:
Sorry, didn't see you already approved... Merging

PR #722, opened 2023-01-30
fix: 🐛 add a missing default value for org name in admin/

severo:
Thanks @lhoestq for spotting the bug


albertvillanova (approved):
Thanks @severo for the fix in running docker compose locally and @lhoestq for having found the bug.

PR #728, opened 2023-01-30
feat: 🎸 add indexes, based on recommendations from mongo cloud

severo:


lhoestq (approved):
Cool ! This way we can follow more easily the jobs per dataset ?

severo:
Capture d’écran 2023-01-31 à 13 31 18

PR #732, opened 2023-01-30
Add gradio admin interface

lhoestq:
A simple Gradio app to show the pending jobs.

Contrary to the observable notebook, this can run locally and connect to your local dev environment.

I added a feature to see the number of jobs in the queue, and a simple SQL query feature that uses DuckDB.

The app can be extended to support more admin features e.g.
- refresh jobs
- cancel jobs
- show report

I'll also deploy it on Spaces

image


front/admin_ui/app.py
```diff
@@ -0,0 +1,121 @@
...
+            return {
+                auth_page: gr.update(visible=False),
+                welcome_title: gr.update(value=f"### Welcome {user['name']}"),
+                main_page: gr.update(visible=True)
+            }
+        else:
+            return f"❌ Unauthorized (user '{user['name']} is not a member of '{ADMIN_HF_ORGANIZATION}')"
+
+    def view_jobs(token):
+        global pending_jobs_df
```
freddyaboulton, line 80: I'm not sure if this will work for concurrent users if deployed on spaces. The same df will be shared by everyone. 

Might be best to save this in an invisible `gr.Dataframe` component? That will be isolated per session.

front/admin_ui/requirements.txt
```diff
@@ -0,0 +1,5 @@
+gradio==~3.16.1
```
severo, line 1: I think we should have only one: pyproject.toml or requirements.txt

severo (approved):
Excellent, it's super useful, thanks.

We can see later how to integrate it inside the deployment to get a published version.

front/admin_ui/app.py
lhoestq: It's ok if it's shared by everyone for now.

I tried to store it in a `gr.Dataframe` though but it crashed. The dataframe is 40MB and has string and timestamp types that can be null.

front/admin_ui/requirements.txt
lhoestq: I added requirements.txt to easily deploy it on Spaces - I don't think Spaces support pyproject.toml files.

front/admin_ui/app.py
lhoestq: More specifically I tried to store the pandas dataframe it in a gr.Dataframe with visible=False, and it caused my server to hang

front/admin_ui/requirements.txt
severo: ok!
If needed, we could later generate the file automatically in the CI with https://python-poetry.org/docs/cli/#export, in order to avoid duplicating the dependencies information.

lhoestq:
Merging for now, and later we can improve the dataframe thing and auto deployment if we want

PR #733, opened 2023-01-31
feat: 🎸 adapt number of replicas to flush the queues

severo:
Capture d’écran 2023-01-31 à 09 59 04

Current status: https://grafana.huggingface.tech/d/i7gwsO5Vz/global-view?orgId=1&from=now-15m&to=now&refresh=1m&var-slo_quantile_99_5=0.995&var-slo_quantile_85=0.85&var-slo_period=7d

A lot of jobs in /first-rows (critical for the Hub) and /parquet-and-dataset-info (less critical for now, only for parquet files)

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -276,7 +276,7 @@ firstRows:
 
   nodeSelector:
     role-datasets-server: "true"
-  replicas: 60
+  replicas: 90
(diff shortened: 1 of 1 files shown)


severo:
Merging. We're using the 50 nodes at the moment.

PR #738, opened 2023-01-31
fix: remove mongo migration job execution on pre-install hook

rtrompier:


severo (approved):
Perfect, thanks.

For the record: on install, as the job was run before the MongoDB even created ready (due to pre-install), it would fail and block the deployment. It's normally just needed when upgrading a version of the app.

PR #739, opened 2023-01-31
feat: merge helm lint action with publish

rtrompier:


severo (changes requested):
Hmmm, I think we need to have two different actions, because I would like to keep helm lint on the pull requests. I don't want to merge, then see that the helm chart has an error, then open a new PR, etc.

rtrompier:
You are totally right :)

Issue #731, opened 2023-01-30
Dataset Viewer issue for jonas/undp_jobs_raw


AndreaFrancis:
### Link

https://huggingface.co/datasets/jonas/undp_jobs_raw

### Description

When going to the preview panel, this message is shown:
```
'update' command document too large
Error code:   UnexpectedError
```

Other datasets with same issue:

- SamAct/medium_cleaned split train
- grasshoff--lhc_sents split train
- Sangmun/wiki_doc_preprocessed_withmaxlength split train
- DavidVivancos--MindBigData2022_Imagenet_IN split test and train
- heanu/soda split test and validation
- DavidVivancos/MindBigData_Imagenet_IN split train
- DavidVivancos/MindBigData2022_Imagenet_IN_Spct splitr test

When trying to investigate the issue I saw the following logs:
DEBUG: 2023-01-30 13:16:36,215 - root - the size of the first 10 rows (1087032102) is above the max number of bytes (-7173342), they will be truncated
DEBUG: 2023-01-30 13:16:40,076 - root - the size of the rows is now (11944186) after truncating row idx=0

It shows that the remaining space for the document without considering the rows is negative (-7173342) which means only the total space for columns is more that what is accepted.

When looking at the csv file for **jonas/undp_jobs_raw** dataset it looks like the datase has 103630 columns.

```
>>> import pandas as pd
>>> huge_ds = pd.read_csv('undp_jobs.csv')
>>> len(huge_ds.columns)
103630
>>> huge_ds.head(1)
                                                   0                                                  1  ...           104998                                             104999
0  {'content': ['hiv and sti clinical consultant ...  {'content': ['internship- pacific digital econ...  ...  {'content': []}  {'content': ['deputy head of resident coordina...

[1 rows x 103630 columns]
>>> 
```
Need to find a way to keep the columns and store them in the cache without issues.

AndreaFrancis:
#self-assign

AndreaFrancis:
Based on a couple of suggestions from @severo, I thought this could work:

1. Get the response without rows `{dataset, config, split, features, rows=[]} `
2.  If response size is > : Take only N features (`features_max_number`)
3.  While response size is still  > `rows_max_bytes`: Reduce number of features progressively backwards until response size is < `rows_max_bytes`
4. If new response size < `rows_max_bytes`: Get first M rows (`rows_min_number`) 
[...]

PR #749, opened 2023-01-31
Adding custom exception when cache insert fails because of too many columns

AndreaFrancis:
Adding custom exception to fix scenarios like https://github.com/huggingface/datasets-server/issues/731

### DIFF
--- a/chart/env/dev.yaml
+++ b/chart/env/dev.yaml
@@ -196,6 +196,7 @@ firstRows:
     limits:
       cpu: 1
       memory: "4Gi"
+  columnsMaxNumber: 1_000
 
 parquetAndDatasetInfo:
 
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -226,6 +226,9 @@ firstRows:
   queue:
     # Maximum number of jobs running at the same time for the same namespace
     maxJobsPerNamespace: 4
+  
+  # Max number of columns in the /first-rows endpoint response
+  columnsMaxNumber: 1_000
 
   nodeSelector:
     role-datasets-server: "true"
--- a/chart/templates/worker/first-rows/_container.tpl
+++ b/chart/templates/worker/first-rows/_container.tpl
@@ -35,7 +35,7 @@
   - name: FIRST_ROWS_MIN_NUMBER
     value: {{ .Values.firstRows.minNumber| quote }}
   - name: FIRST_ROWS_COLUMNS_MAX_NUMBER
-    value: {{ .Values.firstRows.columns_max_number| quote }}
+    value: {{ .Values.firstRows.columnsMaxNumber| quote }}
   volumeMounts:
   {{ include "volumeMountAssetsRW" . | nindent 2 }}
   {{ include "volumeMountCache" . | nindent 2 }}
--- a/chart/templates/worker/first-rows/_container.tpl
+++ b/chart/templates/worker/first-rows/_container.tpl
@@ -34,6 +34,8 @@
     value: {{ .Values.firstRows.minCellBytes | quote }}
   - name: FIRST_ROWS_MIN_NUMBER
     value: {{ .Values.firstRows.minNumber| quote }}
+  - name: FIRST_ROWS_COLUMNS_MAX_NUMBER
+    value: {{ .Values.firstRows.columns_max_number| quote }}
   volumeMounts:
   {{ include "volumeMountAssetsRW" . | nindent 2 }}
   {{ include "volumeMountCache" . | nindent 2 }}
--- a/workers/datasets_based/README.md
+++ b/workers/datasets_based/README.md
@@ -44,6 +44,8 @@ Set environment variables to configure the first rows worker (`FIRST_ROWS_` pref
 - `FIRST_ROWS_MAX_NUMBER`: the max number of rows fetched by the worker for the split and provided in the /first-rows endpoint response. Defaults to `100`.
 - `FIRST_ROWS_MIN_CELL_BYTES`: the minimum size in bytes of a cell when truncating the content of a row (see `FIRST_ROWS_ROWS_MAX_BYTES`). Below this limit, the cell content will not be truncated. Defaults to `100`.
 - `FIRST_ROWS_MIN_NUMBER`: the min number of rows fetched by the worker for the split and provided in the /first-rows endpoint response. Defaults to `10`.
+- `FIRST_ROWS_COLUMNS_MAX_NUMBER`: the max numbers of columns allowed to fetch sample data
+
 
 Also, set the assets-related configuration for the first-rows worker. See [../../libs/libcommon/README.md](../../libs/libcommon/README.md).
 
--- a/workers/datasets_based/README.md
+++ b/workers/datasets_based/README.md
@@ -44,7 +44,7 @@ Set environment variables to configure the first rows worker (`FIRST_ROWS_` pref
 - `FIRST_ROWS_MAX_NUMBER`: the max number of rows fetched by the worker for the split and provided in the /first-rows endpoint response. Defaults to `100`.
 - `FIRST_ROWS_MIN_CELL_BYTES`: the minimum size in bytes of a cell when truncating the content of a row (see `FIRST_ROWS_ROWS_MAX_BYTES`). Below this limit, the cell content will not be truncated. Defaults to `100`.
 - `FIRST_ROWS_MIN_NUMBER`: the min number of rows fetched by the worker for the split and provided in the /first-rows endpoint response. Defaults to `10`.
-- `FIRST_ROWS_COLUMNS_MAX_NUMBER`: the max numbers of columns allowed to fetch sample data
+- `FIRST_ROWS_COLUMNS_MAX_NUMBER`: the max number of columns (features) provided in the /first-rows endpoint response. If the number of columns is greater than the limit, an error is returned. Defaults to `1_000`.
 
 
 Also, set the assets-related configuration for the first-rows worker. See [../../libs/libcommon/README.md](../../libs/libcommon/README.md).
--- a/workers/datasets_based/README.md
+++ b/workers/datasets_based/README.md
@@ -46,7 +46,6 @@ Set environment variables to configure the first rows worker (`FIRST_ROWS_` pref
 - `FIRST_ROWS_MIN_NUMBER`: the min number of rows fetched by the worker for the split and provided in the /first-rows endpoint response. Defaults to `10`.
 - `FIRST_ROWS_COLUMNS_MAX_NUMBER`: the max number of columns (features) provided in the /first-rows endpoint response. If the number of columns is greater than the limit, an error is returned. Defaults to `1_000`.
 
-
 Also, set the assets-related configuration for the first-rows worker. See [../../libs/libcommon/README.md](../../libs/libcommon/README.md).
 
 ### Parquet and dataset info worker
--- a/workers/datasets_based/src/datasets_based/workers/first_rows.py
+++ b/workers/datasets_based/src/datasets_based/workers/first_rows.py
@@ -495,7 +495,9 @@ def compute_first_rows_response(
         features = info.features
 
     if len(features) > columns_max_number:
-        raise TooManyColumnsError(f"Too many columns. The maximum supported number of columns is {columns_max_number}.")
+        raise TooManyColumnsError(
+            f"Too many columns. The maximum supported number of columns is {columns_max_number}."
+        )
 
     # get the rows
     try:
--- a/workers/datasets_based/src/datasets_based/workers/first_rows.py
+++ b/workers/datasets_based/src/datasets_based/workers/first_rows.py
@@ -572,7 +572,7 @@ def compute_first_rows_response(
         ) from err
 
     # truncate the rows to fit within the restrictions, and prepare them as RowItems
-    surrounding_json_size = get_json_size(response)
+    surrounding_json_size = get_json_size(response_features_only)
     row_items = create_truncated_row_items(
         rows=transformed_rows,
         min_cell_bytes=min_cell_bytes,
--- a/workers/datasets_based/src/datasets_based/workers/first_rows.py
+++ b/workers/datasets_based/src/datasets_based/workers/first_rows.py
@@ -369,11 +369,6 @@ def to_features_list(features: Features) -> List[FeatureItem]:
     ]
 
 
-def validate_content_size(obj: Any, rows_max_bytes: int):
-    if get_json_size(obj) > rows_max_bytes:
-        raise TooBigContentError("The first rows content after truncation exceeds the maximum size.")
-
-
 class SplitFullName(TypedDict):
     dataset: str
     config: str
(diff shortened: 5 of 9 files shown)


severo:
Thanks! That's not really what I had in mind. I thought we could instead add a new env var to set the maximum number of columns (features) supported in /first-rows, and check this early in the function (as soon as we have the features), and then raise if the number of features is above the limit.

Then, once deployed in prod, we refresh the /first-rows endpoint for the affected datasets -> they should now have this error, which is more explicit.

AndreaFrancis:
Got it, it will first validate if the amount of columns exceeded the limit but what if even having a less number of columns but the content without rows is also too long that exceeds the rows_max_bytes?  I think another kind of error could be thrown for now or just leave it as it is as unexpectederror

severo:
Yes, two cases are missing:
- too many columns, compared to the new configuration parameter that we need to create -> TooManyColumnsError
- truncated content is still too big compared to the max number of bytes of the response -> it's an internal error because it should be the result of an incoherent set of configuration parameters, ie: we could fix it by reducing the max number of column, or by increasing the min size of the response, or else. But you're right, it should be a new error, something like: TooBigTruncatedContent, something like that

chart/templates/worker/first-rows/_container.tpl
```diff
@@ -34,6 +34,8 @@
     value: {{ .Values.firstRows.minCellBytes | quote }}
   - name: FIRST_ROWS_MIN_NUMBER
     value: {{ .Values.firstRows.minNumber| quote }}
+  - name: FIRST_ROWS_COLUMNS_MAX_NUMBER
+    value: {{ .Values.firstRows.columns_max_number| quote }}
```
severo, line 38: columns_max_number -> columnsMaxNumber

And you also have to add the value in values.yaml, in the worker template, and if necessary in prod.yaml and dev.yaml

workers/datasets_based/README.md
```diff
@@ -44,6 +44,8 @@ Set environment variables to configure the first rows worker (`FIRST_ROWS_` pref
 - `FIRST_ROWS_MAX_NUMBER`: the max number of rows fetched by the worker for the split and provided in the /first-rows endpoint response. Defaults to `100`.
 - `FIRST_ROWS_MIN_CELL_BYTES`: the minimum size in bytes of a cell when truncating the content of a row (see `FIRST_ROWS_ROWS_MAX_BYTES`). Below this limit, the cell content will not be truncated. Defaults to `100`.
 - `FIRST_ROWS_MIN_NUMBER`: the min number of rows fetched by the worker for the split and provided in the /first-rows endpoint response. Defaults to `10`.
+- `FIRST_ROWS_COLUMNS_MAX_NUMBER`: the max numbers of columns allowed to fetch sample data
```
severo, line 47:
```suggestion
- `FIRST_ROWS_COLUMNS_MAX_NUMBER`: the max number of columns (features) provided in the /first-rows endpoint response. If the number of columns is greater than the limit, an error is returned. Defaults to `1_000`.
```

workers/datasets_based/README.md
```diff
@@ -44,6 +44,8 @@ Set environment variables to configure the first rows worker (`FIRST_ROWS_` pref
...
+
```
severo, line 48:
```suggestion
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -409,6 +431,8 @@ def compute_first_rows_response(
             The maximum number of rows of the response.
         rows_min_number (`int`):
             The minimum number of rows of the response.
+        columns_max_number (`int`):
+            The maximun number of columns allowed.
```
severo, line 435:
```suggestion
            The maximum number of columns supported.
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -427,6 +451,8 @@ def compute_first_rows_response(
           If the split rows could not be obtained using the datasets library in normal mode.
         - [`~workers.first_rows.RowsPostProcessingError`]
           If the post-processing of the split rows failed, e.g. while saving the images or audio files to the assets.
+        - [`~workers.first_rows.TooManyColumnsError`]
+          If the response cannot be saved because of too many columns in the dataset.
```
severo, line 455:
```suggestion
          If the number of columns (features) exceeds the maximum supported number of columns.
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -480,6 +506,24 @@ def compute_first_rows_response(
             raise FeaturesError("The split features (columns) cannot be extracted.", cause=err) from err
     else:
         features = info.features
+
+    if len(features) > columns_max_number:
```
severo, line 510: beware, features can be None (as for the types). We have to check this before calling len()

```suggestion
    if features and len(features) > columns_max_number:
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -480,6 +506,24 @@ def compute_first_rows_response(
...
+        raise TooManyColumnsError(
+            f"Too many columns. The maximum supported number of columns is {columns_max_number}."
+        )
+
+    # validate size of response without the rows
+    features_list = to_features_list(features=features)
+    response: FirstRowsResponse = {
```
severo, line 517: Good idea.

I would prefer to give it another name though, just in case, to avoid returning this. eg

```suggestion
    response_features_only: FirstRowsResponse = {
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -480,6 +506,24 @@ def compute_first_rows_response(
...
+        "dataset": dataset,
+        "config": config,
+        "split": split,
+        "features": features_list,
+        "rows": [],
+    }
+
+    validate_content_size(response, rows_max_bytes=rows_max_bytes)
```
severo, line 525:
```suggestion
    validate_content_size(response_features_only, rows_max_bytes=rows_max_bytes)
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -101,6 +103,20 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
...
+
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "TooManyColumnsError", cause, True)
+
+
+class TooBigContentError(FirstRowsWorkerError):
+    """Raised when the first rows content exceeded the max size of bytes."""
+
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "TooBigContentError", cause, True)
```
severo, line 117: It's an internal error (we have not been able to correctly set the environment variables that restrict and truncate the size of the response). The user cannot solve it, it's an issue for us.

```suggestion
        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "TooBigContentError", cause, False)
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -353,6 +369,11 @@ def to_features_list(features: Features) -> List[FeatureItem]:
     ]
 
 
+def validate_content_size(obj: Any, rows_max_bytes: int):
+    if get_json_size(obj) > rows_max_bytes:
+        raise TooBigContentError("First rows content exceeded max size bytes.")
```
severo, line 374:
```suggestion
        raise TooBigContentError("The first rows content after truncation exceeds the maximum size.")
```

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -526,24 +570,18 @@ def compute_first_rows_response(
...
-        "dataset": dataset,
-        "config": config,
-        "split": split,
-        "features": features_list,
-        "rows": [],
-    }
-    surrounding_json_size = get_json_size(response)
+
     # truncate the rows to fit within the restrictions, and prepare them as RowItems
+    surrounding_json_size = get_json_size(response)
```
severo, line 575:
```suggestion
    surrounding_json_size = get_json_size(response_features_only)
```

also: maybe try to avoid computing the size twice.

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -526,24 +570,18 @@ def compute_first_rows_response(
...
     row_items = create_truncated_row_items(
         rows=transformed_rows,
         min_cell_bytes=min_cell_bytes,
         rows_max_bytes=rows_max_bytes - surrounding_json_size,
         rows_min_number=rows_min_number,
     )
     response["rows"] = row_items
+    validate_content_size(response, rows_max_bytes=rows_max_bytes)
```
severo, line 583: I think it's not necessary here. In the worst case, `create_truncated_row_items` should return an empty array, which would give the same size as surrounding_json_size.

severo: But we have to change a bit the implementation of `create_truncated_row_items` to effectively return an empty array (if I remember correctly, I forced having at least one row)

workers/datasets_based/tests/workers/test_first_rows.py
```diff
@@ -174,8 +176,15 @@ def test_truncation(
...
-    print(get_json_size(response))
-    assert (get_json_size(response) <= rows_max_bytes) is successful_truncation
+
+    if error_code:
+        with pytest.raises(CustomError) as error_info:
+            worker.compute()
+        assert error_info.value.code == error_code
+    else:
+        response = worker.compute()
+        print(get_json_size(response))
```
severo, line 189:
```suggestion
```

severo (commented):
Good!

I made a lot of small comments, as it's one of your first PR.

In particular, I tried to make the error messages a bit more explicit for the end users, since it will be shown in the dataset viewer.

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -480,6 +506,24 @@ def compute_first_rows_response(
...
+        "split": split,
+        "features": features_list,
+        "rows": [],
+    }
+
+    validate_content_size(response, rows_max_bytes=rows_max_bytes)
+
     # get the rows
     try:
         rows = get_rows(
```
lhoestq, line 529: You should also check the number of columns here, in case `features` is `None`

lhoestq (commented):
Cool ! Just one comment:

workers/datasets_based/src/datasets_based/workers/first_rows.py
AndreaFrancis: Yes, I moved that validation to line 510

chart/templates/worker/first-rows/_container.tpl
AndreaFrancis: Thanks, I will add the it to the templates

workers/datasets_based/src/datasets_based/workers/first_rows.py
AndreaFrancis: @severo  I reviewed the code and couldn't find where the function  `create_truncated_row_items` returns at least one, but anyway I think we should add that validation (_If size is bigger return empty array)_ in the `compute_first_rows_response` function instead on adding that logic in `create_truncated_row_items` since its only responsibility should be only "truncate" rows but not "reduce" the response, that could be  part of the compute logic, what do you think?

AndreaFrancis: Sorry, I just realized that logic is managed into truncate_row_item and features+rows are combined which makes difficult to manage outside in compute_first_rows_response. I removed the extra validation since it is no more needed and handled in another function

chart/env/dev.yaml
```diff
@@ -196,6 +196,7 @@ firstRows:
     limits:
       cpu: 1
       memory: "4Gi"
+  columnsMaxNumber: 1_000
```
severo, line 199: just fyi, if it's the same value as values.yaml, it's not necessary to repeat it. We can let it

chart/env/prod.yaml
```diff
@@ -226,6 +226,9 @@ firstRows:
   queue:
     # Maximum number of jobs running at the same time for the same namespace
     maxJobsPerNamespace: 4
+  
+  # Max number of columns in the /first-rows endpoint response
+  columnsMaxNumber: 1_000
```
severo, line 231: same comment: we don't need to set variables if only set in values.yaml with the same value

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -427,6 +446,8 @@ def compute_first_rows_response(
           If the split rows could not be obtained using the datasets library in normal mode.
         - [`~workers.first_rows.RowsPostProcessingError`]
           If the post-processing of the split rows failed, e.g. while saving the images or audio files to the assets.
+        - [`~workers.first_rows.TooManyColumnsError`]
+          If the number of columns (features) exceeds the maximum supported number of columns.
```
severo, line 450: TooBigContentError is missing in the docstring

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -526,24 +566,19 @@ def compute_first_rows_response(
...
-        "dataset": dataset,
-        "config": config,
-        "split": split,
-        "features": features_list,
-        "rows": [],
-    }
-    surrounding_json_size = get_json_size(response)
+
     # truncate the rows to fit within the restrictions, and prepare them as RowItems
+    surrounding_json_size = get_json_size(response_features_only)
```
severo, line 571: you could assign the variable at line 520, to avoid computing twice the same value

severo (approved):
Just details. Feel free to merge once fixed. Thanks!!!

workers/datasets_based/src/datasets_based/workers/first_rows.py
AndreaFrancis: Done, I added it, thanks!

workers/datasets_based/src/datasets_based/workers/first_rows.py
AndreaFrancis: Thanks! I moved it to 520

workers/datasets_based/src/datasets_based/workers/first_rows.py
severo: > But we have to change a bit the implementation of create_truncated_row_items to effectively return an empty array (if I remember correctly, I forced having at least one row)

Sorry I was wrong... Indeed: we keep at least `min_rows_number` rows, and if needed, we truncate their content until the size is under `max_rows_bytes - surrounding_json_size`. If the parameters `FIRST_ROWS_MAX_BYTES`, `FIRST_ROWS_MIN_CELL_BYTES`, `FIRST_ROWS_MIN_NUMBER` and `FIRST_ROWS_COLUMNS_MAX_NUMBER` are set correctly, truncating should get the size under the threshold. It's not guaranteed because the features, even if their number is under `FIRST_ROWS_COLUMNS_MAX_NUMBER`, can have any size (think of nested dictionaries), but I don't think we should over-engineer here. It would work for most cases.

Maybe a least we should add a check here: https://github.com/huggingface/datasets-server/blob/main/workers/datasets_based/src/datasets_based/workers/first_rows.py#L267. If, after truncating all the cells of the 10 (`min_rows_number`) first rows, the size is still over the maximum, raise a new kind of error. It would avoid an UnknownError.

PR #752, opened 2023-01-31
remove docker-images.yaml, and fix dev.yaml

severo:


severo:
For now, I have an error in the dev environment: https://datasets-server.us.dev.moon.huggingface.tech/valid gives me 502. Waiting until tomorrow, in case there is DNS involved...

rtrompier (approved):

severo:
It still does not work, there is an issue with the connection to the services

rtrompier (approved):

Issue #745, opened 2023-01-31
Improve the error messages in the dataset viewer


severo:
The error messages are shown in the dataset viewer. We should try to improve them:
- help the user understand what is going on
- help the user fix the issue
- show that we understand what is going on, and that it's not a bug.

See https://huggingface.slack.com/archives/C04L6P8KNQ5/p1675171833108199 (internal) and https://github.com/huggingface/datasets/issues/4886.

Also:
- https://github.com/huggingface/datasets-server/issues/731#issuecomment-1410297107
- https://github.com/huggingface/datasets-server/issues/718

severo:
Is the issue solved by #754, @albertvillanova?

Issue #718, opened 2023-01-28
Dataset Viewer issue for rfernand/basic_sentence_transforms


rfernand2:
### Link

https://huggingface.co/datasets/rfernand/basic_sentence_transforms

### Description

The dataset viewer is not working for dataset rfernand/basic_sentence_transforms.

Error details:

```
Error code:   ResponseNotReady

```

rfernand2:
It is working now.  Did it just need a long time to prepare/cache the data somehow?  The data is not that large.

severo:
> Did it just need a long time to prepare/cache the data somehow

Yes. Usually, it should be very quick for small datasets, but we still have issues sometimes when the processing queue gets filled with other datasets. We are working on solutions to reduce this kind of incident.

rfernand2:
The error message hints at that but would be helpful to change it to something for explicit like "Dataset Viewer server is busier than usual; please try again later".   With the current message, I thought it might still be an issue on my side.

Thanks,

- Roland

severo:
Closing, fixed by https://github.com/huggingface/datasets-server/pull/754.

PR #754, opened 2023-02-01
Improve error messages

albertvillanova:
Related to:
- #745
  - #718


workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -494,19 +498,26 @@ def compute_first_rows_response(
...
+                cause=err,
+            ) from err
     else:
         features = info.features
 
     if features and len(features) > columns_max_number:
         raise TooManyColumnsError(
-            f"Too many columns. The maximum supported number of columns is {columns_max_number}."
+            f"The number of columns ({len(features)}) exceeds the maximum supported number of columns"
+            f" ({columns_max_number})."
```
albertvillanova, line 520: Should we suggest to reduce the number of columns?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -521,7 +532,10 @@ def compute_first_rows_response(
 
     surrounding_json_size = get_json_size(response_features_only)
     if surrounding_json_size > rows_max_bytes:
-        raise TooBigContentError("The first rows content after truncation exceeds the maximum size.")
+        raise TooBigContentError(
+            f"The size of the first rows content after truncation ({surrounding_json_size} B) exceeds the maximum"
+            f" supported size ({rows_max_bytes} B)."
+        )
```
albertvillanova, line 539: The same, should we suggest to reduce the size of the rows?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -459,12 +459,12 @@ def compute_first_rows_response(
...
     except _EmptyDatasetError as err:
-        raise EmptyDatasetError("The dataset is empty.", cause=err) from err
+        raise EmptyDatasetError(f"The dataset '{dataset}' is empty.", cause=err) from err
     except Exception as err:
-        raise SplitsNamesError("Cannot get the split names for the dataset.", cause=err) from err
+        raise SplitsNamesError(f"Cannot get the split names for the dataset '{dataset}'.", cause=err) from err
     # ^ can raise DatasetNotFoundError or SplitsNamesError
     if config not in [split_full_name["config"] for split_full_name in split_full_names]:
-        raise ConfigNotFoundError(f"config {config} does not exist for dataset {dataset}")
+        raise ConfigNotFoundError(f"The config '{config}' does not exist for dataset '{dataset}'")
```
albertvillanova, line 467: I saw `{dataset}` was passed in the error message here, so I included it in other error messages. But I would like to be sure this is right.
- For a specific user, it does not make sense to tell them the name of their dataset
- But for our cache, it might be useful information
To be confirmed.

workers/datasets_based/src/datasets_based/workers/first_rows.py
severo: Yes, good idea!

Generally, it comes from a flaw in the design of the dataset (not "tidy data"), but anyway, people have the right to do what they want.

So, I think we just want to say it's a limitation of the dataset viewer, and that by reducing the number of columns, it would be working again

workers/datasets_based/src/datasets_based/workers/first_rows.py
severo: Hmmm, in that case, it's an issue with our configuration (see https://github.com/huggingface/datasets-server/pull/749#discussion_r1095587218), and the details will not be shown to the user, just the error message.

You're right, it should be clear that the user has nothing special to do. Note that the viewer will automatically propose to create an issue here.

workers/datasets_based/src/datasets_based/workers/first_rows.py
severo: Nope, it's not necessary since the cache entry in mongo already contains dataset/config/split: https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/simple_cache.py#L69

This means that giving the config is also redundant.

albertvillanova: @severo I agree with dataset. On the contrary, the config is redundant for our cache, but it might be informative for the user getting it in the error message, no?

severo: Oh, yes, you're right

chart/static-files/openapi.json
```diff
@@ -98,11 +98,11 @@
         },
         "examples": {
           "SplitsResponseNotReadyError": {
-            "summary": "The list of splits is not ready yet. Please retry later.",
+            "summary": "The server is busier than usual and the list of splits is not ready yet. Please retry later.",
             "value": "SplitsResponseNotReadyError"
           },
           "SplitsNamesError": {
-            "summary": "Cannot get the split names for the dataset.",
+            "summary": "Cannot get the split names for the dataset. Please fix your loading script.",
```
severo, line 105: btw, is it possible to get this error even if the dataset does not contain a loading script?

chart/static-files/openapi.json
```diff
@@ -196,19 +196,19 @@
             "value": "FirstRowsResponseNotReady"
           },
           "InfoError": {
-            "summary": "The info cannot be fetched for the dataset config.",
+            "summary": "The info cannot be fetched for the config of the dataset. Please fix your loading script.",
```
severo, line 199: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -468,12 +468,14 @@ def compute_first_rows_response(
     try:
         split_full_names = get_dataset_split_full_names(dataset=dataset, use_auth_token=use_auth_token)
     except _EmptyDatasetError as err:
-        raise EmptyDatasetError("The dataset is empty.", cause=err) from err
+        raise EmptyDatasetError("The dataset is empty. Please fix your loading script.", cause=err) from err
```
severo, line 471: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -491,7 +493,10 @@ def compute_first_rows_response(
             use_auth_token=use_auth_token,
         )
     except Exception as err:
-        raise InfoError("The info cannot be fetched for the dataset config.", cause=err) from err
+        raise InfoError(
+            f"The info cannot be fetched for the config '{config}' of the dataset. Please fix your loading script.",
```
severo, line 497: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -503,19 +508,27 @@ def compute_first_rows_response(
                 use_auth_token=use_auth_token,
             )
             if not isinstance(iterable_dataset, IterableDataset):
-                raise TypeError("load_dataset should return an IterableDataset")
+                raise TypeError("load_dataset should return an IterableDataset. Please fix your loading script.")
```
severo, line 511: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -503,19 +508,27 @@ def compute_first_rows_response(
...
             iterable_dataset = iterable_dataset._resolve_features()
             if not isinstance(iterable_dataset, IterableDataset):
-                raise TypeError("load_dataset should return an IterableDataset")
+                raise TypeError("load_dataset should return an IterableDataset. Please fix your loading script.")
```
severo, line 514: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -503,19 +508,27 @@ def compute_first_rows_response(
...
             features = iterable_dataset.features
         except Exception as err:
-            raise FeaturesError("The split features (columns) cannot be extracted.", cause=err) from err
+            raise FeaturesError(
+                (
+                    f"Cannot extract the features (columns) for the split '{split}' of the config '{config}' of the"
+                    " dataset. Please fix your loading script."
```
severo, line 520: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -554,7 +570,10 @@ def compute_first_rows_response(
             )
         if info.size_in_bytes is None or info.size_in_bytes > MAX_SIZE_FALLBACK:
             raise StreamingRowsError(
-                "Cannot load the dataset split (in streaming mode) to extract the first rows.",
+                (
+                    "Cannot load the dataset split (in streaming mode) to extract the first rows. Please make your"
+                    " loading script support streaming."
```
severo, line 575: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -568,7 +587,10 @@ def compute_first_rows_response(
             )
         except Exception as err:
             raise NormalRowsError(
-                "Cannot load the dataset split (in normal download mode) to extract the first rows.",
+                (
+                    "Cannot load the dataset split (in normal download mode) to extract the first rows. Please fix"
+                    " your loading script."
```
severo, line 592: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/parquet_and_dataset_info.py
```diff
@@ -522,9 +523,11 @@ def compute_parquet_and_dataset_info_response(
             for config in get_dataset_config_names(path=dataset, revision=source_revision, use_auth_token=hf_token)
         )
     except _EmptyDatasetError as err:
-        raise EmptyDatasetError("The dataset is empty.", cause=err) from err
+        raise EmptyDatasetError("The dataset is empty. Please fix your loading script.", cause=err) from err
```
severo, line 526: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/parquet_and_dataset_info.py
```diff
@@ -522,9 +523,11 @@ def compute_parquet_and_dataset_info_response(
...
     except Exception as err:
-        raise ConfigNamesError("Cannot get the configuration names for the dataset.", cause=err) from err
+        raise ConfigNamesError(
+            "Cannot get the config names for the dataset. Please fix your loading script.", cause=err
```
severo, line 529: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/split_names.py
```diff
@@ -103,9 +103,12 @@ def compute_split_names_response(
             for split in get_dataset_split_names(path=dataset, config_name=config, use_auth_token=use_auth_token)
         ]
     except _EmptyDatasetError as err:
-        raise EmptyDatasetError("The dataset is empty.", cause=err) from err
+        raise EmptyDatasetError("The dataset is empty. Please fix your loading script.", cause=err) from err
```
severo, line 106: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/split_names.py
```diff
@@ -103,9 +103,12 @@ def compute_split_names_response(
...
     except Exception as err:
-        raise SplitNamesError("Cannot get the split names for the dataset and config.", cause=err) from err
+        raise SplitNamesError(
+            f"Cannot get the split names for the config '{config}' of the dataset. Please fix your loading script.",
```
severo, line 109: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/splits.py
```diff
@@ -118,9 +118,11 @@ def compute_splits_response(
     try:
         split_items = get_dataset_split_full_names(dataset=dataset, use_auth_token=use_auth_token)
     except _EmptyDatasetError as err:
-        raise EmptyDatasetError("The dataset is empty.", cause=err) from err
+        raise EmptyDatasetError("The dataset is empty. Please fix your loading script.", cause=err) from err
```
severo, line 121: same question: is this error only possible if a loading script exists?

workers/datasets_based/src/datasets_based/workers/splits.py
```diff
@@ -118,9 +118,11 @@ def compute_splits_response(
...
     except Exception as err:
-        raise SplitsNamesError("Cannot get the split names for the dataset.", cause=err) from err
+        raise SplitsNamesError(
+            "Cannot get the split names for the dataset. Please fix your loading script.", cause=err
```
severo, line 124: same question: is this error only possible if a loading script exists?

severo (commented):
Very nice, thanks!

Just one comment (that I put everywhere to be sure no to forgot one instance): are we sure the dataset contains a loading script, when we ask the user to fix it?

If the dataset has no loading script, maybe it's better not mentioning it.

Also: there are two possible users:
- the user is the maintainer of the dataset: they will be able to apply the advice
- the user is seeing the page of a dataset they do not own: they should open a discussion, so that the maintainer can fix it (note that the viewer already directs to opening a discussion in that case).
Which one are we directing the messages to?

albertvillanova:
@severo:
- I am addressing the error messages to the creator/owner of the dataset: they are responsible for their dataset, it is them who are interested in having the viewer working on their dataset and it is them who must eventually merge the fixes (either implemented by themselves or by some other community member proposing a PR)
- You are right that in many cases the error can appear either when there is a script or when there are only data files
  - I am removing the mention to the script
  - As a follow-up work, I am investigating these cases to see if we could raise a specific error for each case from the datasets library

workers/datasets_based/src/datasets_based/workers/config_names.py
```diff
@@ -89,9 +89,11 @@ def compute_config_names_response(
             for config in sorted(get_dataset_config_names(path=dataset, use_auth_token=use_auth_token))
         ]
     except _EmptyDatasetError as err:
-        raise EmptyDatasetError("The dataset is empty.", cause=err) from err
+        raise EmptyDatasetError("The dataset is empty. Please fix your loading script.", cause=err) from err
     except Exception as err:
-        raise ConfigNamesError("Cannot get the config names for the dataset.", cause=err) from err
+        raise ConfigNamesError(
+            "Cannot get the config names for the dataset. Please fix your loading script.", cause=err
```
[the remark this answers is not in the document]
albertvillanova: For this very specific case, I think this can only be raised when using a loading script. When using data files, the config name is just the dataset id with double hyphens (set by HubDatasetModuleFactoryWithoutScript):
```python
"config_name": self.name.replace("/", "--"),
```

severo (approved):
Thanks!

severo:
Feel free to merge!

PR #758, opened 2023-02-01
ci: 🎡 only run on PR and on main

severo:
currently, the actions are run twice in the PRs. See https://github.com/huggingface/datasets-server/pull/757/commits/c80eeacfdb2839149ad8b7b81cdaf6b0b4fcb944 for example.

### DIFF
--- a/.github/workflows/chart.yml
+++ b/.github/workflows/chart.yml
@@ -5,6 +5,8 @@ name: chart
 on:
   workflow_dispatch:
   push:
+    branches:
+      - main
     paths:
       - 'chart/**'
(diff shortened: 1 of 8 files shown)


albertvillanova (approved):
Thanks for the fix?

What about also requiring to run pull requests only if the base branch is main?

severo:
> What about also requiring to run pull requests only if the base branch is main?

Sometimes we do PR against a branch used in another PR (which might be a bad idea and we should instead do the PR against main and rebase). In that case the CI would not run.

I think we can let it that way, not sure if it's really an issue to run the CI on other branches

Issue #736, opened 2023-01-31
Avoid creating jobs as much as possible


severo:
We are often facing a full queue that blocks the viewer on the Hub from being available (https://github.com/huggingface/datasets-server/issues/725, https://github.com/huggingface/datasets-server/issues/704, etc.)

It's generally the /first-rows queue.

There are several ideas to improve this. One of them is to detect as soon as possible which jobs are useless because they will be skipped when processed by the worker. I suspect that a large number of the processed jobs is skipped (to do: get numbers), and that being able to remove that unnecessary burden on the workers would help.

This means, when adding a job, test if the job would be skipped or not. If it would be skipped, I think that we cannot just avoid creating it and forget, because a dependent processing step could still need to be computed (new step, error, outdated). I think that we should effectively avoid creating the job, but still run the post-processing action to try to add the dependent jobs (and do the same).
I think what this means is separating the "event" that triggers a possible refresh for a dataset, and the "materialization" step (see https://docs.dagster.io/concepts/assets/asset-materializations) where we look at which operations are needed, and run them.

severo:
As a first step, we might generate methods and an endpoint that exposes the current state of a dataset for each step.

severo:
It's the same issue as #764

PR #761, opened 2023-02-02
update the logic to skip a job

severo:
Instead of retrying for any non-successful response in the cache, we
only retry if the error is in the list of "retry-able" errors. Also:
refactor the logic and add complete tests

### DIFF
--- a/workers/datasets_based/src/datasets_based/worker.py
+++ b/workers/datasets_based/src/datasets_based/worker.py
@@ -12,6 +12,7 @@ from libcommon.exceptions import CustomError
 from libcommon.processing_graph import ProcessingStep
 from libcommon.queue import JobInfo, Priority, Queue, Status
 from libcommon.simple_cache import (
+    DoesNotExist,
     SplitFullName,
     delete_response,
     get_response,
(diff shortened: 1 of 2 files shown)


severo:
This kind of error should lead to retry the job.

Capture d’écran 2023-02-02 à 09 44 02

workers/datasets_based/src/datasets_based/worker.py
```diff
@@ -237,17 +244,26 @@ def should_skip_job(self) -> bool:
...
-                and cached_response["worker_version"] is not None
-                and self.compare_major_version(cached_response["worker_version"]) == 0
-                and cached_response["dataset_git_revision"] is not None
-                and cached_response["dataset_git_revision"] == dataset_git_revision
-            )
         except Exception:
+            # an exception occurred while getting the git revision from the Hub - the job will fail anyway, but we
+            # process it to store the error in the cache
             return False
+        return dataset_git_revision is not None and cached_response["dataset_git_revision"] == dataset_git_revision
```
lhoestq, line 265: Can we also make sure to not **emit** jobs if this condition is met ? This could help making the queue less full of unnecessary jobs

lhoestq (approved):
Cool !

workers/datasets_based/src/datasets_based/worker.py
severo: Yes, sure, I'd planned to do it for #736. But I'm trying to do small PRs, I think it will work better now that we work as a team

lhoestq: nice !

PR #766, opened 2023-02-02
Locally use volumes for workers code

lhoestq:
This way in a local dev env, one can restart a single worker container to apply their code changes instead of running `make stop` and `make start` again


severo (changes requested):
Good idea!

The CI is failing because the e2e tests use the docker compose, and it doesn't work for some reason with this new setup.

lhoestq:
I ended up defining a separate docker-compose file for dev.

This way the e2e tests can use the actual docker image that we may deploy.

I introduced
```bash
make dev-start
```

severo (approved):
OK, nice!

Two small comments:
- it's not DRY (but it's not really a problem, I just read https://gordonc.bearblog.dev/dry-most-over-rated-programming-principle/ :P)
- it's only to make it easier to dev on the worker. We could extend this to have a dev environment on the services too

lhoestq:
Yea we'll just have to remember to update both when modifying something, but it should be fine since they're next to each other.

PR #768, opened 2023-02-02
Create doc for every PR

lhoestq:
This is what is done in the other HF repos on github.

This way the Delete doc github action has something to delete.
If the doc doesn't exists, the job fails


severo (approved):
THANKS

lhoestq:
Arf the job failed anyway, will take a look

Issue #755, opened 2023-02-01
Fill /split-names and /first-rows from /parquet-and-dataset-info if needed


severo:
/split-names and /first-rows are created using streaming. They might fail for this reason while /parquet-and-dataset-info works for the same dataset.

In this case, we should fill /split-names (from /dataset-info) and /first-rows (from /parquet).

Note that we would generate a concurrence between two execution paths for the same result: we need to manage this by avoiding launching one of them if we already have the result, and avoiding updating the result if it has already been computed (in the case of concurrent jobs)

severo:
We prepared the switch by deprecating the `FIRST_ROWS_FALLBACK_MAX_DATASET_SIZE` parameter: it's now hard-coded, and that part of the code will be deleted once this issue is fixed:

https://github.com/huggingface/datasets-server/blob/7188c0a4b84b10b88e39ff9f3ad88ae04843402e/workers/datasets_based/src/datasets_based/workers/first_rows.py#L539L561

See #769/#771 and #773

AndreaFrancis:
From https://github.com/huggingface/datasets-server/pull/797#issuecomment-1425445279 comment
**First:** I will first allow processing steps to support cache-kind  different value than endpoint, that way I will be able to create the new worker for split-names-db.  
**Second:** Create new worker for split-names-db, but I have a question here:
> Note that we would generate a concurrence between two execution paths for the same result: we need to manage this by avoiding launching one of them if we already have the result, and avoiding updating the result if it has already been computed (in the case of concurrent jobs)

Don't we already have that logic in should_skip_job worker method? Or is other kind of consideration to have in mind?

**Third:** Create new worker for first-rows ( Maybe /first-rows-parquet)
Does this sounds like a plan?

severo:
> Don't we already have that logic in should_skip_job worker method

yes

> Or is other kind of consideration to have in mind?

yes: imagine the result from split-names-db is already there, and is a success, and that later split-names-streaming finished and is an error: we don't want to store that error. So we have to put a control at the moment where we store the value in the cache to manage the concurrent access.

severo:
I have a proposal.

The same endpoint could have multiple associated processing steps, along two dimensions:

- multiple [cache kinds](https://github.com/huggingface/datasets-server/blob/fb3399aa3a96abe0f46b1c6a9ee8ac9f3358f394/libs/libcommon/src/libcommon/processing_graph.py#L49),
[...]

PR #773, opened 2023-02-06
refactor: 💡 hard-code the value of the fallback

severo:
The fallback will be removed once
https://github.com/huggingface/datasets-server/issues/755 is implemented. Meanwhile, we hide the parameter to prepare deprecation.

### DIFF
--- a/workers/datasets_based/src/datasets_based/workers/first_rows.py
+++ b/workers/datasets_based/src/datasets_based/workers/first_rows.py
@@ -391,7 +391,6 @@ def compute_first_rows_response(
     assets_base_url: str,
     hf_token: Optional[str],
     min_cell_bytes: int,
-    max_size_fallback: int,
     rows_max_bytes: int,
     rows_max_number: int,
     rows_min_number: int,
@@ -418,8 +417,6 @@ def compute_first_rows_response(
             The Hub endpoint (for example: "https://huggingface.co")
         hf_token (`str` or `None`):
             An authentication token (See https://huggingface.co/settings/token)
-        max_size_fallback (`int`):
-            The maximum number of bytes of the split to fallback to normal mode if the streaming mode fails.
         rows_max_bytes (`int`):
             The maximum number of bytes of the response (else, the response is truncated).
         rows_max_number (`int`):
@@ -534,7 +531,8 @@ def compute_first_rows_response(
             use_auth_token=use_auth_token,
         )
     except Exception as err:
-        if info.size_in_bytes is None or info.size_in_bytes > max_size_fallback:
+        MAX_SIZE_FALLBACK = 100_000_000
+        if info.size_in_bytes is None or info.size_in_bytes > MAX_SIZE_FALLBACK:
             raise StreamingRowsError(
                 "Cannot load the dataset split (in streaming mode) to extract the first rows.",
                 cause=err,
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks for the fix, @severo.

What about to allow passing the parameter, raising a deprecation warning saying we are using constant value instead?

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -534,7 +535,8 @@ def compute_first_rows_response(
             use_auth_token=use_auth_token,
         )
     except Exception as err:
-        if info.size_in_bytes is None or info.size_in_bytes > max_size_fallback:
+        MAX_SIZE_FALLBACK = 100_000_000
```
albertvillanova, line 538: After having imported warnings, something like:
```suggestion
        if max_size_fallback:
            warnings.warn(<A WARNING MESSAGE HERE>)
        MAX_SIZE_FALLBACK = 100_000_000
```

albertvillanova (approved):
Thanks, just a comment below.

workers/datasets_based/src/datasets_based/workers/first_rows.py
albertvillanova: I think you should fix your last commit... :sweat_smile:

albertvillanova (changes requested):
Sorry, mu suggestion code above was not functional...

workers/datasets_based/src/datasets_based/workers/first_rows.py
severo: yes, I know, I just wanted to have you as a co-author in the squashed commit

workers/datasets_based/src/datasets_based/workers/first_rows.py
```diff
@@ -535,9 +535,12 @@ def compute_first_rows_response(
...
         )
     except Exception as err:
-        if max_size_fallback:
-            warnings.warn(<A WARNING MESSAGE HERE>)
         MAX_SIZE_FALLBACK = 100_000_000
+        if max_size_fallback:
+            logging.warning(
+                f"The parameter 'max_size_fallback' is deprecated. The hard-coded value `{MAX_SIZE_FALLBACK}`"
+                " will be used instead."
+            )
```
albertvillanova, line 543: Do you prefer logging rather than `warnings.warn("MESSAGE", FutureWarning)`?

severo: oh, I thought it was the same.

albertvillanova: Not exactly... https://docs.python.org/3/howto/logging.html#when-to-use-logging

albertvillanova: > Issue a warning regarding a particular runtime event
> - warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> - logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

severo: just discovered https://docs.python.org/3/library/warnings.html 😅

albertvillanova (approved):
Thanks!!!

albertvillanova:
Please note currently there are connectivity issues with http://azure.archive.ubuntu.com mirrors...

severo:
Thanks for the help with that @albertvillanova!

Issue #765, opened 2023-02-02
/sizes doesn't work in local dev env 


lhoestq:
it always return
```
{
    "error": "The response is not ready yet. Please retry later."
}
```

severo:
Weird

lhoestq:
This comes from parquet-and-dataset-info which fails:
```

DEBUG: 2023-02-03 14:19:56,849 - root - [/parquet-and-dataset-info] job assigned: JobRunner(job_id=63dd17fdd561bb64690362c7 dataset=lhoestq/demo1 config=None split=None force=True)
INFO: 2023-02-03 14:19:56,849 - root - [/parquet-and-dataset-info] compute JobRunner(job_id=63dd17fdd561bb64690362c7 dataset=lhoestq/demo1 config=None split=None force=True)
DEBUG: 2023-02-03 14:19:56,856 - urllib3.connectionpool - Starting new HTTPS connection (1): huggingface.co:443
DEBUG: 2023-02-03 14:19:57,255 - urllib3.connectionpool - https://huggingface.co:443 "GET /api/datasets/lhoestq/demo1 HTTP/1.1" 200 407
DEBUG: 2023-02-03 14:19:57,261 - root - Directory created at: /datasets-cache/2023-02-03-14-19-57--parquet-and-dataset-info-lhoestq-demo1-034f45a5
DEBUG: 2023-02-03 14:19:57,263 - root - datasets data cache set to: /datasets-cache/2023-02-03-14-19-57--parquet-and-dataset-info-lhoestq-demo1-034f45a5
INFO: 2023-02-03 14:19:57,263 - root - get parquet files and dataset info for dataset=lhoestq/demo1
DEBUG: 2023-02-03 14:19:57,265 - urllib3.connectionpool - Starting new HTTPS connection (1): huggingface.co:443
DEBUG: 2023-02-03 14:19:57,759 - urllib3.connectionpool - https://huggingface.co:443 "POST /datasets/lhoestq/demo1/ask-access HTTP/1.1" 401 15499
```

lhoestq:
It seems we miss the PARQUET_AND_DATASET_INFO_COMMITTER_HF_TOKEN locally

lhoestq:
actually even with a token I get a 401 error at POST huggingface.co/datasets/lhoestq/demo1/ask-access

lhoestq:
This raises a DatasetNotFoundError, even if the dataset does exist and I have a valid token

[...]

PR #774, opened 2023-02-06
Use hub-ci locally

lhoestq:
I switched the HF endpoint to the hub-ci one and added the corresponding tokens, and fixed the docker-compose to use the right base.
Now the local dev environment from `make dev-start` works correctly.

Close https://github.com/huggingface/datasets-server/issues/765

### DIFF
--- a/workers/datasets_based/src/datasets_based/worker_loop.py
+++ b/workers/datasets_based/src/datasets_based/worker_loop.py
@@ -4,6 +4,7 @@
 import logging
 import random
 import time
+import os
 from dataclasses import dataclass
 
 from libcommon.config import WorkerLoopConfig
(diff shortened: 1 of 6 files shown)


workers/datasets_based/src/datasets_based/worker_loop.py
```diff
@@ -97,6 +98,8 @@ def sleep(self) -> None:
         time.sleep(duration)
 
     def loop(self) -> None:
+        if os.environ.get("HF_ENDPOINT"):
```
severo, line 101: We don't access the os environment variables outside of the config.py files. Do you think it's possible to keep on with this practice and move the log there?

severo (approved):
OK

I will let you define what to do with the comment about accessing the env vars.

workers/datasets_based/src/datasets_based/worker_loop.py
lhoestq: I kept the logging here and got the endpoint value from worker_factory.app_config.common

severo: Perfect, thanks!

PR #775, opened 2023-02-06
ci: 🎡 the e2e tests must now be run on any code change

severo:


albertvillanova (commented):
Some questions @severo:
- If this PR makes e2e tests run on any code change, what was the case before this PR?
- What is the purpose of `e2e.yml` then?
- I remember there were some issues with PRs from forks and e2e tests: will these changes include changes introduced by fork PRs? What about the associated docker building issue?

severo:
> If this PR makes e2e tests run on any code change, what was the case before this PR?

we had a two-steps process:
- change in any of the jobs/, services/ or workers/ -> unit tests, code quality, and build a docker image
- change the docker images in the Helm chart -> e2e tests
Now, the e2e tests runner builds the docker images itself, but I had forgotten to change this (in https://github.com/huggingface/datasets-server/pull/716)

> What is the purpose of e2e.yml then?

It's still useful, I think,
- to check the code quality of the e2e tests,
- to re-run the e2e tests if e2e.yml or the docker compose for example, have changed

> I remember there were some issues with PRs from forks and e2e tests: will these changes include changes introduced by fork PRs? What about the associated docker building issue?

Hmmm, I don't think it's an issue anymore. We add an action, the trigger sources have not changed.

albertvillanova (commented):
Just one more question: jobs run in parallel, so e2e-tests will run at the same time as quality and unit-tests. Is this expected?

EDIT: I mean, the Docker image will be built (by e2e) even if eventually quality or unit tests fail.

severo:
Yes. Do you think it's a problem? We can also make it depend on the other two ones, as you prefer

albertvillanova (commented):
I mean do you want to build the Docker images if quality/unit tests have failed?

We have two options here:
- Either making e2e dependent on the others: but this will make the total CI time longer because e2e will not start until the others have finished.
- Or make e2e immediately fail when some of the others have failed: CI will not increase, but maybe the Docker image is already built when some of the others fail (and make e2e fail).
  - EDIT: Maybe this is not possible for jobs outside a matrix though :(

severo:
> I mean do you want to build the Docker images if quality/unit tests have failed?

Note that the images are built, but not sent to docker.io. So, we have to decide: reduce the CI duration (current) or reduce the wasted resources (depend e2e on the success of the other actions)

severo:
OK, merging as such, and we can change later if needed

severo:
OK, it was a bad idea to do this way, because we have several actions doing the same e2e tests at the same time, I hadn't understood...

Capture d’écran 2023-02-08 à 10 01 50

I will send a new PR to fix that.

Issue #737, opened 2023-01-31
Use a generic worker


severo:
All the workers use the same codebase, but we assign only one processing step to each worker at startup. We could instead allow a worker to handle all (or part of) the processing steps and let the queue manager give it the most adequate job.

By the way, the current way of doing is a particular case, where the set of supported processing steps for each worker only contain one element.

In the transition, we could have part of the workers working as of now, while some other workers are "generic", or manage a set of several steps...

The benefit would be to make it a lot easier to handle the deployment and scaling of workers, and avoiding lazy workers.

PR #778, opened 2023-02-06
Fix CI mypy error: "WorkerFactory" has no attribute "app_config"

albertvillanova:
Fix type checking on WorkerLoop.loop method, when tryin to access the worker_factory's attribute app_config.

This PR fixes an issue introduced by:
- #774

```
src/datasets_based/worker_loop.py:100: error: "WorkerFactory" has no attribute "app_config"
```

### DIFF
--- a/workers/datasets_based/src/datasets_based/worker_loop.py
+++ b/workers/datasets_based/src/datasets_based/worker_loop.py
@@ -10,7 +10,7 @@ from libcommon.config import WorkerLoopConfig
 from libcommon.queue import EmptyQueueError, Queue
 from psutil import cpu_count, disk_usage, getloadavg, swap_memory, virtual_memory
 
-from datasets_based.worker_factory import DatasetBasedWorkerFactory
+from datasets_based.worker import WorkerFactory
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


albertvillanova:
The root error raised by mypy is that `WorkerLoop.loop()` is trying to get the attribute `self.worker_factory.app_config` on an instance of `WorkerFactory`, whereas the attribute `app_config` is defined only for its child `DatasetBasedWorkerFactory(WorkerFactory)`.

albertvillanova:
Another approach would be to define the attribute `app_config` in the interface `WorkerFactory`...

severo:
The bug has been introduced by:

https://github.com/huggingface/datasets-server/pull/774/files#diff-4326852d55eab85b019c702a4579f696cb429aaf13586d8e3a1824581f8c9e34R100

@lhoestq I would vote to remove this log or move it elsewhere because we will soon implement https://github.com/huggingface/datasets-server/issues/737, which will allow a worker loop to manage various endpoints.

albertvillanova:
I am making the corresponding changes and you review them.

lhoestq (approved):
thanks - it's fine to remove it

severo (approved):
Thanks!!

severo:
(I'm merging to fix the CI on other PRs)

PR #779, opened 2023-02-07
Pass processing step to worker

severo:


severo:
The CI error is treated here: https://github.com/huggingface/datasets-server/pull/778. We will have to wait for it to be fixed before merging this PR.

workers/datasets_based/src/datasets_based/worker_factory.py
```diff
@@ -17,34 +21,68 @@
...
+
+    @abstractmethod
+    def _create_worker(self, job_info: JobInfo) -> Worker:
+        pass
+
+
+class WorkerFactory(BaseWorkerFactory):
+    def __init__(self, app_config: AppConfig, processing_graph: ProcessingGraph) -> None:
         self.app_config = app_config
+        self.processing_graph = processing_graph
```
albertvillanova, line 42: There is one thing I don't understand: the `processing_graph` continues being part of the `app_config`? So you are passing it here twice (separate and within the app_config)?

albertvillanova: I mean, if you are trying to decouple them, then `ProcessingGraph` shouldn't be removed from `AppConfig` (in workers/datasets_based/src/datasets_based/config.py)?

albertvillanova (approved):
Awesome!!! Thanks!

Definitely it makes much more sense to have the `WorkerLoopConfig` defined within `workers/`, where `WorkerLoop` is defined as well.

Also, as I commented in other PR (https://github.com/huggingface/datasets-server/pull/778#issuecomment-1419444822):
> Maybe, we could use `BaseWorkerFactory` to define the interface, and rename `DatasetBasedWorkerFactory` to `WorkerFactory`...
I think it is much clearer this way.

And I find the introduction of the `processing_graph` attribute in the `WorkerFactory` will help to be more flexible and generalize to more complex processings...

Just some questions below.

lhoestq (approved):
better this way, thanks :)

workers/datasets_based/src/datasets_based/worker_factory.py
severo: oh, you're right.

This PR is a part I extracted from a bigger PR where I'm moving processing_graph outside of the app_config and I forgot to remove it. I'll fix this for now, for the sake of coherence

AndreaFrancis (approved):

Issue #770, opened 2023-02-03
Add a check in /first-rows worker if truncating didn't succeed


severo:
See https://github.com/huggingface/datasets-server/pull/749#discussion_r1095587218

AndreaFrancis:
Done in https://github.com/huggingface/datasets-server/pull/780

Issue #762, opened 2023-02-02
Handle the case where the DatasetInfo is too big


severo:
In the /parquet-and-dataset-info processing step, if DatasetInfo is over 16MB, we will not be able to store it in MongoDB (https://pymongo.readthedocs.io/en/stable/api/pymongo/errors.html#pymongo.errors.DocumentTooLarge). We have to handle this case, and return a clear error to the user.

See https://huggingface.slack.com/archives/C04L6P8KNQ5/p1675332303097889 (internal). It's a similar issue to https://github.com/huggingface/datasets-server/issues/731 (should be raised for that dataset, btw)

AndreaFrancis:
I think we could have validation size at worker level when getting the content https://github.com/huggingface/datasets-server/blob/main/workers/datasets_based/src/datasets_based/worker.py#L282 and if response size exceeds a MAX_CONTENT_SIZE then raise an exception. This would help us handle size issue not only for dataset-info but for all other workers types.

severo:
Yes, it's a good idea.

I'm wondering what the best way is:
1. adding a MAX_CONTENT_SIZE that could be lower than 16MB and raising if any response is above this size
2. catching the pymongo.errors.DocumentTooLarge exception (i.e. hard-coding MAX_CONTENT_SIZE to 16MB) and raising a new exception with a clear message

Any thoughts?

AndreaFrancis:
I think having max content validation in worker let us decoupled from the database engine, lets say in the future we support another type, we wont depend on mongodb.
Also having the validation at worker level would fix https://github.com/huggingface/datasets-server/issues/770

severo:
OK, let's go with MAX_CONTENT_SIZE then

severo:
Done with #780

PR #780, opened 2023-02-07
Dataset info big content error

AndreaFrancis:
Adding content size validation before trying to insert/update in db.
Fix for https://github.com/huggingface/datasets-server/issues/762 and  https://github.com/huggingface/datasets-server/issues/770 
New `DATASETS_BASED_CONTENT_MAX_BYTES` configuration is added to limit the size of the content result of a worker compute process.


libs/libcommon/src/libcommon/config.py
```diff
@@ -39,13 +39,15 @@ def from_env() -> "AssetsConfig":
 COMMON_HF_ENDPOINT = "https://huggingface.co"
 COMMON_HF_TOKEN = None
 COMMON_LOG_LEVEL = logging.INFO
+COMMON_CONTENT_MAX_SIZE = 1_0000_000
```
lhoestq, line 42:
```suggestion
COMMON_CONTENT_MAX_SIZE = 10_000_000
```

.github/workflows/_e2e_tests.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           ADMIN_UVICORN_NUM_WORKERS: "2"
           ADMIN_UVICORN_PORT: "8081"
           COMMON_LOG_LEVEL: "DEBUG"
+          COMMON_CONTENT_MAX_SIZE : "10000000"
```
severo, line 40: Could you use "10_000_000" instead? It's easier to read and avoids errors

https://peps.python.org/pep-0515/

lhoestq (approved):
Awesome ! thank you :)

libs/libcommon/README.md
```diff
@@ -16,7 +16,7 @@ Set the common environment variables to configure the following aspects:
 - `COMMON_HF_ENDPOINT`: URL of the HuggingFace Hub. Defaults to `https://huggingface.co`.
 - `COMMON_HF_TOKEN`: App Access Token (ask moonlanding administrators to get one, only the `read` role is required) to access the gated datasets. Defaults to empty.
 - `COMMON_LOG_LEVEL`: log level, among `DEBUG`, `INFO`, `WARNING`, `ERROR`, and `CRITICAL`. Defaults to `INFO`.
-
+-  `COMMON_CONTENT_MAX_SIZE`: the maximum size in bytes of the content to be stored in the cache. Defaults to `100_000_000`
```
severo, line 19: Could you move it to the "CACHE_" namespace? It would make more sense

libs/libcommon/src/libcommon/config.py
severo, line 42: 10_000_000

libs/libcommon/src/libcommon/config.py
```diff
@@ -58,6 +60,7 @@ def from_env() -> "CommonConfig":
                 hf_endpoint=env.str(name="HF_ENDPOINT", default=COMMON_HF_ENDPOINT),
                 hf_token=env.str(name="HF_TOKEN", default=COMMON_HF_TOKEN),  # nosec
                 log_level=env.log_level(name="LOG_LEVEL", default=COMMON_LOG_LEVEL),
+                content_max_size=env.int(name="DOCUMENT_MAX_SIZE", default=COMMON_CONTENT_MAX_SIZE),
```
severo, line 63: DOCUMENT / CONTENT

workers/datasets_based/src/datasets_based/worker.py
```diff
@@ -296,6 +311,13 @@ def process(
             try:
                 self.pre_compute()
                 content = self.compute()
+
+                # Validate content size
```
severo, line 315: Maybe move this to the libcommon.cache -> it's more a limitation of the cache backend than of the worker as itself.

severo (changes requested):

libs/libcommon/src/libcommon/config.py
AndreaFrancis: Thanks, I modified the value

.github/workflows/_e2e_tests.yml
AndreaFrancis: Done

libs/libcommon/src/libcommon/config.py
AndreaFrancis: Done

libs/libcommon/src/libcommon/config.py
AndreaFrancis: Fixed

workers/datasets_based/src/datasets_based/worker.py
AndreaFrancis: I moved the validation to simple cache, I will need to pass the max configuration since `CachedResponse`  inherits from Document and couldn't have any env configuration inside at least I wouldn't like to mix db engine with config logic values

libs/libcommon/README.md
AndreaFrancis: Will I have to move also that config to `CacheConfig`? If so, I am afraid to have the `post_init` `connect_to_cache_database` running.
Maybe add another namespace but will have to create another config class no handle it

chart/env/dev.yaml
```diff
@@ -71,6 +71,7 @@ common:
   hfEndpoint: "https://huggingface.co"
   # Log level
   logLevel: "DEBUG"
+  contentMaxSize: 100000
```
severo, line 74:
```suggestion
  contentMaxSize: 100_000
```

chart/env/prod.yaml
```diff
@@ -61,7 +61,8 @@ common:
   hfEndpoint: "https://huggingface.co"
   # Log level
   logLevel: "INFO"
-
+  contentMaxSize: 1000000
```
severo, line 64:
```suggestion
  contentMaxSize: 1_000_000
```

Is it the value we want in production, btw?

libs/libcommon/README.md
severo: I'm currently working on removing all the "post_init" things from the config: we will have a new "Resource" object, that will be allocated and then released (database, log, storage, etc.)
So, if I finish soon enough, we could move this parameter to the cache config, it would be more logical.

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -121,6 +153,13 @@ class CachedResponse(Document):
 CachedResponse.split.required = False  # type: ignore
 
 
+def validate_content_size(content: Any, content_max_size: int) -> None:
```
severo, line 156: I was thinking of calling it inside "upsert_response", as a validation step.

workers/datasets_based/src/datasets_based/worker.py
```diff
@@ -296,6 +297,7 @@ def process(
             try:
                 self.pre_compute()
                 content = self.compute()
+                validate_content_size(content=content, content_max_size=self.common_config.content_max_size)
```
severo, line 300: move to the CachedResponse.upsert_response method?

severo:
I have still changed my mind 😬. Thinking more about it, I now believe that we want:
- to handle the MongoDB 16MB limit nicely (currently, it returns UnexpectedError); here we don't want to reduce this size, why would we limit the storage?
- to limit the size of the REST API responses, at least some of them, to avoid wasting "bandwidth" and having big latencies.
  - for the lists of things (split names, config names, parquet files, etc.) we can paginate. Additionally, if needed, we could use the maximum size (instead of the number of elements) to define where to cut for the next page (not sure if it's a good idea).
  - for the other things (dataset_info, sizes) we should find a solution. It could be:
    - limiting the total size of the response
    - changing the format of the response so that it's small, or it's a list of things that can be paginated. For example: /dataset-info is a nested dict that contains a list of features that can potentially exceed the maximum size. Maybe we should not expose /dataset-info as such, but expose /features and /metadata (the rest)... I don't know. Also: maybe /first-rows could be paginated, that way we can send bigger responses if needed, and the client decides to get more or less rows. Beware: it surely means changing how we store the response in the MongoDB, maybe one response per page?

My ideas on this are not totally clear, so for now (this PR), I think we can:
- add a limit on the response size, ie: in the worker/datasets_based/ directory (sorry to make you change again)

and create issues to: 
- handle the 16MB limit in MongoDB with a dedicated error
- paginate the responses
- change the format of some endpoints

libs/libcommon/README.md
AndreaFrancis: As per last commit, I kept  the logic in the worker

workers/datasets_based/src/datasets_based/worker.py
AndreaFrancis: As per last commit, I kept  the logic in the worker

chart/env/dev.yaml
AndreaFrancis: I fixed it to 10_000_000

chart/env/prod.yaml
AndreaFrancis: I fixed it to 10_000_000

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: As per last comment, I kept it on the worker

workers/datasets_based/src/datasets_based/worker.py
AndreaFrancis: As per last comment, I kept it on the worker

.github/workflows/_e2e_tests.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           ADMIN_UVICORN_NUM_WORKERS: "2"
           ADMIN_UVICORN_PORT: "8081"
           COMMON_LOG_LEVEL: "DEBUG"
+          COMMON_CONTENT_MAX_SIZE : "10_000_000"
```
severo, line 40: could you move it to `DATASETS_BASED_CONTENT_MAX_BYTES`? (here, and elsewhere in the code: chart/, datasets_based/, tools/)

- `COMMON_` -> `DATASETS_BASED_`: what we really want would be `WORKER_CONTENT_MAX_SIZE` (or `WORKER_RESPONSE_CONTENT_MAX_SIZE`), but we will create this "namespace" when working on https://github.com/huggingface/datasets-server/issues/737, for now `DATASETS_BASED` is the namespace for all the workers)
- `SIZE` -> `BYTES`: giving the unit avoids errors when modifying the configurations

chart/values.yaml
```diff
@@ -47,6 +47,7 @@ common:
   hfEndpoint: ""
   # Log level
   logLevel: "INFO"
+  contentMaxSize: 1000000
```
severo, line 50: "10_000_000", no?

chart/env/prod.yaml
```diff
@@ -61,7 +61,8 @@ common:
   hfEndpoint: "https://huggingface.co"
   # Log level
   logLevel: "INFO"
-
+  contentMaxSize: 10_000_000
```
severo, line 64: I think it must be a string, not a number: `"10_000_000"`

chart/env/dev.yaml
```diff
@@ -71,6 +71,7 @@ common:
   hfEndpoint: "https://huggingface.co"
   # Log level
   logLevel: "DEBUG"
+  contentMaxSize: 10_000_000
```
severo, line 74: I think it must be a string, not a number: `"10_000_000"`

libs/libcommon/README.md
```diff
@@ -16,7 +16,7 @@ Set the common environment variables to configure the following aspects:
 - `COMMON_HF_ENDPOINT`: URL of the HuggingFace Hub. Defaults to `https://huggingface.co`.
 - `COMMON_HF_TOKEN`: App Access Token (ask moonlanding administrators to get one, only the `read` role is required) to access the gated datasets. Defaults to empty.
 - `COMMON_LOG_LEVEL`: log level, among `DEBUG`, `INFO`, `WARNING`, `ERROR`, and `CRITICAL`. Defaults to `INFO`.
-
+-  `COMMON_CONTENT_MAX_SIZE`: the maximum size in bytes of the content to be stored in the cache. Defaults to `10_000_000`
```
severo, line 19: (to be moved to workers/datasets_based/README.md)

what about:

the maximum size in bytes of the response content computed by a worker (to prevent returning big responses in the REST API). Defaults to `10_000_000`.

libs/libcommon/src/libcommon/config.py
```diff
@@ -39,13 +39,15 @@ def from_env() -> "AssetsConfig":
 COMMON_HF_ENDPOINT = "https://huggingface.co"
 COMMON_HF_TOKEN = None
 COMMON_LOG_LEVEL = logging.INFO
+COMMON_CONTENT_MAX_SIZE = 10_0000_000
```
severo, line 42: 10_000_000

tools/docker-compose-dev-base.yml
```diff
@@ -6,6 +6,7 @@ services:
       COMMON_HF_ENDPOINT: ${COMMON_HF_ENDPOINT-https://hub-ci.huggingface.co}
       COMMON_HF_TOKEN: ${COMMON_HF_TOKEN-hf_app_datasets-server_token}
       COMMON_LOG_LEVEL: ${COMMON_LOG_LEVEL-INFO}
+      COMMON_CONTENT_MAX_SIZE: ${COMMON_CONTENT_MAX_SIZE-10_00_000}
```
severo, line 9: 10_000_000

workers/datasets_based/src/datasets_based/worker.py
```diff
@@ -104,6 +106,19 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
         )
 
 
+class TooBigContentError(GeneralWorkerError):
+    """Raised when content size in bytest is bigger than the supported value."""
```
severo, line 110:
```suggestion
    """Raised when content size in bytes is bigger than the supported value."""
```

workers/datasets_based/src/datasets_based/worker.py
severo: Yes, thanks

workers/datasets_based/src/datasets_based/worker.py
```diff
@@ -297,6 +312,13 @@ def process(
             try:
                 self.pre_compute()
                 content = self.compute()
+
+                # Validate content size
+                if len(orjson_dumps(content)) > self.common_config.content_max_size:
+                    raise TooBigContentError(
+                        "Could not process content, it exceeds the supported size in bytes"
```
severo, line 319:
```suggestion
                        "The computed response content exceeds the supported size in bytes ({self.common_config.content_max_size})."
```

workers/datasets_based/tests/workers/test__datasets_based_worker.py
```diff
@@ -31,7 +34,7 @@ def compute(self) -> Mapping[str, Any]:
         if self.config == "raise":
             raise ValueError("This is a test")
         else:
-            return {}
+            return {"col1": ["a" * 10 for _ in range(200)]}
```
severo, line 37: maybe something simpler to understand, and that should also be greater than the max size

```suggestion
            return {"col1": "a" * 200}
```

severo (commented):
Thanks, it's better that way.

chart/values.yaml
```diff
@@ -377,3 +377,8 @@ sizes:
     limits:
       cpu: 0
   tolerations: []
+
+# --- datasets_based ---
+datasetsBased:
+  # TODO: This config have to be moved to worker namespace
```
severo, line 383:
```suggestion

```

tools/docker-compose-dev-base.yml
```diff
@@ -26,6 +26,7 @@ services:
     environment:
       # datasets
       DATASETS_BASED_HF_DATASETS_CACHE: ${HF_DATASETS_CACHE-/datasets-cache}
+      DATASETS_BASED_CONTENT_MAX_BYTES: ${CONTENT_MAX_BYTES-10_000_000}
```
severo, line 29:
```suggestion
      DATASETS_BASED_CONTENT_MAX_BYTES: ${DATASETS_BASED_CONTENT_MAX_BYTES-10_000_000}
```

workers/datasets_based/tests/conftest.py
```diff
@@ -51,6 +51,7 @@ def set_env_vars(datasets_cache_directory: Path, modules_cache_directory: Path)
     mp.setenv("PARQUET_AND_DATASET_INFO_COMMITTER_HF_TOKEN", CI_USER_TOKEN)
     mp.setenv("DATASETS_BASED_HF_DATASETS_CACHE", str(datasets_cache_directory))
     mp.setenv("HF_MODULES_CACHE", str(modules_cache_directory))
+    mp.setenv("CONTENT_MAX_BYTES", "10_000_000")
```
severo, line 54:
```suggestion
    mp.setenv("DATASETS_BASED_CONTENT_MAX_BYTES", "10_000_000")
```

severo (approved):
Details, and it's OK

PR #786, opened 2023-02-08
ci: 🎡 run e2e tests only once for a push or pull-request

severo:
see
https://github.com/huggingface/datasets-server/pull/775#issuecomment-1422255785.


albertvillanova (approved):
Thanks, this will avoid duplicate tests.

PR #787, opened 2023-02-08
Fix dockerfiles

severo:
Running `make e2e` locally, I had the following error:

```
 => ERROR [e2e-worker-dataset-info 11/13] RUN poetry install --no-cache                                                                                                                                                                                                    0.4s
------
 > [e2e-worker-dataset-info 11/13] RUN poetry install --no-cache:
#0 0.345 runc run failed: unable to start container process: exec: "/bin/sh": stat /bin/sh: no such file or directory
```

Adding the path fixes the issue.

### DIFF
--- a/jobs/mongodb_migration/Dockerfile
+++ b/jobs/mongodb_migration/Dockerfile
@@ -5,20 +5,21 @@ FROM python:3.9.15-slim
 ENV PYTHONFAULTHANDLER=1 \
     PYTHONUNBUFFERED=1 \
     PYTHONHASHSEED=random \
-    PIP_NO_CACHE_DIR=off \
+    PIP_NO_CACHE_DIR=1 \
     PIP_DISABLE_PIP_VERSION_CHECK=on \
     PIP_DEFAULT_TIMEOUT=100 \
     POETRY_NO_INTERACTION=1 \
     # Versions:
     POETRY_VERSION=1.3.2 \
-    POETRY_VIRTUALENVS_IN_PROJECT=true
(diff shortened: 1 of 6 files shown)


jobs/mongodb_migration/Dockerfile
```diff
@@ -5,20 +5,21 @@ FROM python:3.9.15-slim
...
-    PIP_NO_CACHE_DIR=off \
+    PIP_NO_CACHE_DIR=1 \
     PIP_DISABLE_PIP_VERSION_CHECK=on \
     PIP_DEFAULT_TIMEOUT=100 \
     POETRY_NO_INTERACTION=1 \
     # Versions:
     POETRY_VERSION=1.3.2 \
-    POETRY_VIRTUALENVS_IN_PROJECT=true
+    POETRY_VIRTUALENVS_IN_PROJECT=true \
+    PATH="$PATH:/root/.local/bin"
```
albertvillanova, line 15: What is the executable inside `/root/.local/bin` you are adding to PATH?

albertvillanova (approved):
I don't have too much experience with Dockerfiles. I guess you have tested locally and now works as expected.

jobs/mongodb_migration/Dockerfile
albertvillanova: I mean I guess there is no other executable with the same name already in PATH, because that would have precedence...

severo: poetry

PR #788, opened 2023-02-08
feat: remove job after 5 minutes

rtrompier:
To avoid to block the uninstall process : the PVC wait for job deletion before remove.


severo (approved):
OK.

Two questions:
- when I run one of these migrations, it took a long time, because the database is huge (or because my migration script uses bad queries?). The job timeout-ed (maybe 5 minutes after starting, btw) and the "upgrade" failed. Also note that the migration seems to have worked anyway, after some time. Would it be the same with this change?
- why do you think the job was during so long? Could you share more details?

PR #789, opened 2023-02-08
feat: 🎸 add logs when an unexpected error occurs

severo:


services/admin/src/admin/utils.py
```diff
@@ -57,8 +58,12 @@ def __init__(self, message: str):
...
     """Raised when an unexpected error occurred."""
 
-    def __init__(self, message: str):
-        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "UnexpectedError")
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "UnexpectedError", cause)
+        if cause:
+            logging.exception(message, exc_info=cause)
+        else:
+            logging.error(message)
```
AndreaFrancis, line 66:
```suggestion
            logging.exception(message)
```

AndreaFrancis (approved):
Just one minor observation

services/admin/src/admin/utils.py
severo: thanks!!

services/api/src/api/utils.py
```diff
@@ -59,6 +60,10 @@ class UnexpectedError(ApiCustomError):
 
     def __init__(self, message: str, cause: Optional[BaseException] = None):
         super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "UnexpectedError", cause)
+        if cause:
+            logging.exception(message, exc_info=cause)
+        else:
+            logging.exception(message)
```
albertvillanova, line 66: Note that basically: `logging.exception = logging.error(..., exc_info=True)`
```suggestion
        logging.error(message, exc_info=cause)
```

workers/datasets_based/src/datasets_based/worker.py
```diff
@@ -115,6 +115,10 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
             cause=cause,
             disclose_cause=False,
         )
+        if cause is not None:
+            logging.exception(message, exc_info=cause)
+        else:
+            logging.exception(message)
```
albertvillanova, line 121: The same here:
```suggestion
        logging.error(message, exc_info=cause)
```

services/admin/src/admin/routes/backfill.py
```diff
@@ -54,7 +54,7 @@ async def backfill_endpoint(request: Request) -> Response:
             )
         except AdminCustomError as e:
             return get_json_admin_error_response(e, max_age=0)
-        except Exception:
-            return get_json_admin_error_response(UnexpectedError("Unexpected error."), max_age=0)
+        except Exception as e:
+            return get_json_admin_error_response(UnexpectedError("Unexpected error.", e), max_age=0)
```
albertvillanova, line 58: We used to log the error type and error message as well in `datasets`. What do you think?
```suggestion
            return get_json_admin_error_response(UnexpectedError(f"Unexpected error {type(e)}: {e}", e), max_age=0)
```

If you agree, then you should update the other lines accordingly.

albertvillanova (approved):
Thanks, we can get more insights this way...

services/admin/src/admin/routes/backfill.py
severo: Hmmm, this message will be shown to the users in the dataset viewer, and in the API responses. We don't want to show internal errors, only user-related errors. At least, it's how I've created the error system for now. That's why I don't put details in the message, but the traceback in the logs.

services/api/src/api/utils.py
albertvillanova: https://github.com/python/cpython/blob/5f0b81905084eb5da7131d06a6a4f15beeab460f/Lib/logging/__init__.py#L2117

severo: OHHH. It's perfect

services/admin/src/admin/routes/backfill.py
albertvillanova: I see your point, yes I agree.

PR #790, opened 2023-02-08
feat: 🎸 ensure immutability of the configs

severo:
if we decide to allow changing config parameters later, it will need to be explicit.

Depends on: #784


AndreaFrancis (approved):
Cool! It has more sense for configs

PR #791, opened 2023-02-08
use classmethod for factories instead of staticmethod

severo:
See https://stackoverflow.com/questions/12179271/meaning-of-classmethod-and-staticmethod-for-beginner for example

Depends on: #790


albertvillanova (commented):
I think you need to merge main branch once you merged #790.

AndreaFrancis (approved):
Interesting, It looks better for factory methods

severo:
Merging

PR #793, opened 2023-02-08
Basic stats

lhoestq:
Started a `/basic-stats` endpoints that computes histogram for numerical data using dask.

Not sure if we want to merge this feature right away, I implemented this mostly to trigger some discussions on how to add new data aggregates: does this way sound correct to you ?

Feel free to also comment on how we could expand this to include stats for non-numerical columns. How can we add new stats that would be returned in the same endpoint ?

You can try it on `__DUMMY_DATASETS_SERVER_USER__/letters` for example:

<details>

<summary>Results of `/basic-stats` on `__DUMMY_DATASETS_SERVER_USER__/letters`</summary>

```json
{
    "basic_stats": [
        {
            "dataset": "__DUMMY_DATASETS_SERVER_USER__/letters",
            "config": "__DUMMY_DATASETS_SERVER_USER__--letters",
            "split": "train",
            "column_name": "label",
            "histogram": {
                "hist": [
                    10,
                    0,
                    0,
                    0,
                    0,
                    11,
                    0,
                    0,
                    0,
                    5
                ],
                "bin_edges": [
                    0.0,
                    0.2,
                    0.4,
                    0.6,
                    0.8,
                    1.0,
                    1.2,
                    1.4,
                    1.6,
                    1.8,
                    2.0
                ]
            }
        },
        {
            "dataset": "__DUMMY_DATASETS_SERVER_USER__/letters",
            "config": "__DUMMY_DATASETS_SERVER_USER__--letters",
            "split": "train",
            "column_name": "id",
            "histogram": {
                "hist": [
                    3,
                    2,
                    3,
                    2,
                    3,
                    2,
                    3,
                    2,
                    3,
                    3
                ],
                "bin_edges": [
                    0.0,
                    2.5,
                    5.0,
                    7.5,
                    10.0,
                    12.5,
                    15.0,
                    17.5,
                    20.0,
                    22.5,
                    25.0
                ]
            }
        },
        {
            "dataset": "__DUMMY_DATASETS_SERVER_USER__/letters",
            "config": "__DUMMY_DATASETS_SERVER_USER__--letters",
            "split": "train",
            "column_name": "fr_scrabble_value",
            "histogram": {
                "hist": [
                    10,
                    3,
                    3,
                    3,
                    0,
                    0,
                    0,
                    2,
                    0,
                    5
                ],
                "bin_edges": [
                    1.0,
                    1.9,
                    2.8,
                    3.7,
                    4.6,
                    5.5,
                    6.4,
                    7.3,
                    8.2,
                    9.1,
                    10.0
                ]
            }
        }
    ]
}
```

</details>

TODO:
- [ ] tests
- [ ] docs

cc @severo


libs/libcommon/src/libcommon/config.py
```diff
@@ -120,6 +120,7 @@ class ProcessingGraphConfig:
             "/parquet": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
             "/dataset-info": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
             "/sizes": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
+            "/basic-stats": {"input_type": "split", "requires": "/parquet-and-dataset-info"},
```
severo, line 123: you could directly depend on /parquet since you don't use the dataset-info part in the compute function

workers/datasets_based/src/datasets_based/workers/basic_stats.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+    content = response["content"]
+    try:
+        basic_stats: List[BasicColumnStats] = []
+        parquet_files_urls = [
+            parquet_file["url"]
+            for parquet_file in response["content"]["parquet_files"]
+            if parquet_file["config"] == config and parquet_file["split"] == split
+        ]
+        df = dd.read_parquet(parquet_files_urls)
```
severo, line 125: how would it work for shards? will it read metadata from every shard?

severo:
> does this way sound correct to you ?

yes, perfect

> How can we add new stats that would be returned in the same endpoint ?

Good question. I imagine that we should add them to the response, and then trigger an update of the cache for all the existing entries.

workers/datasets_based/src/datasets_based/workers/basic_stats.py
lhoestq: It loads all the parquet files only when `compute()` is called (and optimize the processing graph to only use the required shards).

For example, `dd.read_parquet` doesn't raise an error if you pass parquet files with different columns, but `compute()` can.

severo: OK, nice!

lhoestq:
Closing this one since it's not the priority for now. It would require to fix too many merge conflicts

Let's keep this PR in mind if we ever want to add these features

Issue #753, opened 2023-02-01
Delete extracted the downloaded files just after being extracted


severo:
```
from datasets import load_dataset, DownloadConfig

download_config = DownloadConfig(delete_extracted=True)
dataset = load_dataset("./codeparrot", split="train", download_config=download_config)
```

internal ref: https://huggingface.slack.com/archives/C04L6P8KNQ5/p1675260906599029?thread_ts=1675260896.849119&cid=C04L6P8KNQ5

It will help to reduce the disk usage consumed by the workers' cache.

PR #794, opened 2023-02-08
Delete extracted downloaded files of a dataset

polinaeterna:
Will close https://github.com/huggingface/datasets-server/issues/753


polinaeterna:
@severo did I find all the place where this change should be applied? Also, do I need to write a test for it?

severo:
> all the place where this change should be applied

maybe also in  splits, config_names and split_names, because the methods of https://github.com/huggingface/datasets/blob/b065547654efa0ec633cf373ac1512884c68b2e1/src/datasets/inspect.py also take DownloadConfig, but I'm not sure it's useful or not to get the list of splits and the list of configs.

> do I need to write a test for it?

I don't think so, as it's already tested in datasets (I imagine)

severo (approved):

PR #797, opened 2023-02-08
Change split names upstream and source

AndreaFrancis:
Partial fix for https://github.com/huggingface/datasets-server/issues/755

- Changing the predecessor of split-names to dataset-info
- Changing source of split-names (db instead of dataset lib in streaming mode)


severo:
Beware: we want to have /split-names as soon as possible, which generally means getting it from datasets using the streaming mode.
But if /dataset-info has been computed quicker, or if the streaming mode is not working, we want to fallback to using /dataset-info as the source.
It's like a race between /split-names and /dataset-info, and the first that returns a successful response is the source (like [Promise.any()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any) if we speak JS)

severo:
But for now, we don't have the concept of double source in ProcessingGraph. I'm not sure how we could implement it.

AndreaFrancis:
In that case I think having two different workers for now would work, lets say `/split-names-db` `/split-names-streaming`

- `/split-names-db` would depend on `/dataset-info` and `/split-names-streaming `on` /config-names`
- Once one of them runs, ask if there exists a result in the db, if so, skip, if not, compute according to its source
- One problem here is, how to identify if the previous existing response in the db was generated because of the other parallel worker or if it is an outdated result?  Maybe we could link a response with a job id but that would increase the db document size
- Also as you mention, I don't know if we support having two ancestor for a worker (In case of /first-rows, it will also have to know which of the split-names worker finished first), at least we implement also two firs-rows workers XD (`/first-rows-streaming` depending on `/split-names-streaming` and `/first-rows-parquet` depending on `/split-names-db`) 
On the other hand, I think having this kind of new requirements also depends on the future implementation we will have (Maybe Airflow/Dagster). Adding extra logic to the processing graph would be obsolete after implementing a new DAG framework but I am not sure (I didn't go deeper on airflow/dagster) I just don't want to have extra logic in our pipeline processing logic to deprecate it later.

severo:
> how to identify if the previous existing response in the db was generated because of the other parallel worker or if it is an outdated result
we store both the commit id and the worker version, which should be enough to say if the result is outdated or not.

> I don't know if we support having two ancestor for a worker

For that, possibly the idea is to make https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/processing_graph.py evolve to break the current constraint where the API endpoint, the job type and the cache kind are equal. For split names, we could have:

```
{endpoint: "/split-names", job_type: "/split-names-streaming", cache_entry: "/split-names"}
{endpoint: "/split-names", job_type: "/split-names-db", cache_entry: "/split-names"}
```

For the API service, it does not change much, because it will still look at the same cache entry.

PR #792, opened 2023-02-08
Check dataset connection before migration job (and other apps)

severo:
Before starting the services (api, admin) and workers, we ensure the database is accessible and the assets directory (if needed) exists.

In the case of the migration job: if the database cannot be accessed, we skip the migration, to avoid blocking Helm

Fixes #763. Replaces #767. Depends on #791.

PR #800, opened 2023-02-09
Move workers/datasets_based to services/worker

severo:
Based on #792.


services/worker/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 authors = ["Sylvain Lesage <sylvain.lesage@huggingface.co>"]
-description = "Worker for processing steps that need the datasets library"
-name = "datasets_based"
+description = "Worker that processes jobs and store the responses in the cache"
```
AndreaFrancis, line 3:
```suggestion
description = "Worker that processes jobs and stores the responses in the cache"
```

services/worker/src/worker/config.py
```diff
@@ -24,15 +24,15 @@ def get_empty_str_list() -> List[str]:
...
     max_load_pct: int = WORKER_LOOP_MAX_LOAD_PCT
     max_memory_pct: int = WORKER_LOOP_MAX_MEMORY_PCT
     sleep_seconds: int = WORKER_LOOP_SLEEP_SECONDS
     storage_paths: List[str] = field(default_factory=get_empty_str_list)
 
     @classmethod
-    def from_env(cls) -> "WorkerLoopConfig":
+    def from_env(cls) -> "LoopConfig":
         env = Env(expand_vars=True)
         with env.prefixed("WORKER_LOOP_"):
```
AndreaFrancis, line 37:
```suggestion
        with env.prefixed("LOOP_"):
```

AndreaFrancis: But I am not sure, we might need to change all other WORKER_ configs as well :/. Maybe keep it as it is

AndreaFrancis (commented):
I put comments for minor changes. Just one question, how do we use that new vendor trec-car?

severo:
> how do we use that new vendor trec-car?

it's not new, it's just moved with the rest of the code. It's a dependency required by some dataset scripts on the Hub: 
https://github.com/huggingface/datasets-server/blob/5267f427dfa5bae2e4b4fba0b8042d4783fc3a39/services/worker/pyproject.toml#L39
I had to put it inside the code because it's not available in pypi or github or anything (as far as I could find).

services/worker/src/worker/config.py
severo: You're right, I have to change in the rest of the code

severo: well, I will let as WORKER_LOOP better in the code as it's more descriptive. It's not coherent, maybe we should let worker_loop.py in the code...

AndreaFrancis (approved):

Issue #806, opened 2023-02-13
Parquet export: ignore ALL datasets bigger than a certain size


lhoestq:
Cases to ignore:
- [x] the dataset repository is >max_size
- [ ] the dataset uses a script that downloads more >max_size of data

To fix the second case I think we can pass a custom download manager to the dataset builder `_split_generators` to record the size of the files to donwload. It can also be implemented by checking the number of fiels passed as `gen_kwargs` and checking their sizes.

Currently max_size=5GB

PR #809, opened 2023-02-13
Ignore big datasets from external files

lhoestq:
Raise an error for big datasets that use a loading script to download data files external to HF.

It was the only case for a dataset to not be ignored. Hopefully it makes the `/parquet-and-dataset-info` job stop wasting too much resources on big datasets.

Close https://github.com/huggingface/datasets-server/issues/806


services/worker/tests/fixtures/hub.py
```diff
@@ -568,4 +597,12 @@ def hub_datasets(
             "first_rows_response": create_first_rows_response(hub_public_big, BIG_cols, BIG_rows),
             "parquet_and_dataset_info_response": None,
         },
+        "external_files": {
+            "name": hub_public_external_files,
+            "config_names_response": create_config_names_response(hub_public_external_files),
+            "split_names_response": create_split_names_response(hub_public_external_files),
+            "splits_response": create_splits_response(hub_public_external_files),
+            "first_rows_response": create_first_rows_response(hub_public_external_files, TEXT_cols, TEXT_rows),
+            "parquet_and_dataset_info_response": None,
```
lhoestq, line 606: not sure if this was needed - I added it anyway

lhoestq:
There's one mypy check that fails - not sure how to fix it since it happens when calling `datasets` code.

severo:
re mypy: you can still add:

```python
        builder._split_generators(mock_dl_manager)  # type: ignore
```

But it would be good to report it to datasets if it's a bug

services/worker/tests/fixtures/files.py
```diff
@@ -64,3 +64,52 @@ def extra_fields_readme(tmp_path_factory: pytest.TempPathFactory) -> str:
...
         f.writelines(f"{line}\n" for line in lines)
     return path
+
+
+DATASET_SCRIPT_WITH_EXTERNAL_FILES_CONTENT = """
+import datasets
+
+_URLS = {
+    "train": [
+        "https://huggingface.co/datasets/lhoestq/test/resolve/main/some_text.txt",
```
severo, line 74: This adds a dependency on the prod Hub for the tests. Why not, we didn't have (for the workers) as for now.

We could generate the repo on the hub-ci, depend on it, and then delete it. Maybe it's too much?

services/worker/tests/fixtures/hub.py
severo: no, not needed since we never access this dataset from the `hub_datasets` fixture. But we could, to ensure we know how to manage datasets with a script (all the other tests use a simple data file).

severo (approved):
Thanks! Some comments... We can address them here or in another issue.

We have to ensure the CI passes, though (see https://github.com/huggingface/datasets-server/pull/809#issuecomment-1429551075)

services/worker/tests/fixtures/files.py
lhoestq: It's simpler to use this one directly. And I already use it elsewhere for tests so it's not going to change.

services/worker/tests/fixtures/hub.py
lhoestq: I'll leave it here in case it can be useful for some tests later if you don't mind

severo: Can you create an issue ([here](https://www.notion.so/huggingface2/Internal-issues-b15b180c923346ceb6893d1e6805f180) or [here](https://github.com/huggingface/datasets-server/)) to ensure we remind it?

lhoestq: done on notion :)

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
```diff
@@ -528,10 +528,28 @@ def raise_if_too_big_from_external_data_files(
...
-        builder._split_generators(mock_dl_manager)
-    except NotImplementedError as err:
-        if "is not implemented in streaming mode." not in str(err):
-            raise
+        builder._split_generators(mock_dl_manager)  # type: ignore
+    except (requests.exceptions.RequestException, NotImplementedError) as error:
+        if isinstance(error, NotImplementedError):
+            # we can ignore the errors from functions not implemented in streaming mode like `.extract()` on TAR files
+            if "is not implemented in streaming mode." not in str(error):
+                raise
```
severo, line 536: as we raise, it will be caught and transformed into an "Unexpected error" when shown to the user in the dataset viewer with the raw message. That's why we always try to catch the exceptions and create a specific message that will help the user understand what they can do on their side to fix the issue or, else, how they can report it.

cc @albertvillanova

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
```diff
@@ -528,10 +528,28 @@ def raise_if_too_big_from_external_data_files(
...
+        elif isinstance(error, requests.exceptions.HTTPError):
+            raise ExternalFilesSizeRequestHTTPError(
+                "Couldn't get the list of external files in `_split_generators` because a request failed"
```
severo, line 539: In the same vein, maybe give a higher level message that a user of the Hub can make sense of.

Also: pass the error to the constructor as `cause`, to get the traceback into the database.

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
lhoestq: Ok, good to know. Should I add an dedicated error here as well then ?

severo: Yes, I think so

severo: We really should have Unexpected errors only for the cases we don't expect and that indicate a bug in the datasets-server code.

severo:
Also: we should increment the version number of the job runner.

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
albertvillanova: I agree, for the end user is better to have actionable error messages.

lhoestq:
I improved the errors and bumped the job runner version :)

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
```diff
@@ -423,6 +447,202 @@ class EmptyFeaturesError(Exception):
     pass
 
 
+class DatasetWithTooManyExternalFilesError(ParquetAndDatasetInfoJobRunnerError):
+    """Raised when the dataset size (sum of config sizes given by the datasets library) is too big."""
+
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "DatasetWithTooManyExternalFilesError", cause, False)
```
severo, line 454: For all these new errors, we should disclose the exception since the dataset maintainer can fix the issue. Same for all the errors below.

```suggestion
        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "DatasetWithTooManyExternalFilesError", cause, True)
```

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
```diff
@@ -423,6 +447,202 @@ class EmptyFeaturesError(Exception):
...
+                        "Couldn't get the list of external files in `_split_generators` because it doesn't support"
+                        f" streaming:\n{error}"
+                    ),
+                    error,
+                )
+        elif isinstance(error, requests.exceptions.HTTPError):
+            raise ExternalFilesSizeRequestHTTPError(
+                (
+                    "Couldn't get the list of external files in `_split_generators` because a request"
+                    f" failed:\n{error}\nPlease consider moving your data fiels in this dataset repository instead"
```
severo, line 555:
```suggestion
                    f" failed:\n{error}\nPlease consider moving your data files in this dataset repository instead"
```

same below

severo (approved):
Some details to fix, but else it's good. Thanks a lot, it will remove a lot of work to the job runners

PR #810, opened 2023-02-13
WIP- Separate endoint from processing step

AndreaFrancis:
Firs part of code for https://github.com/huggingface/datasets-server/issues/755
Since we are going to have two different processing steps for one same endpoint:
- Changed the dependency between endpoint-cache_kind in ProcessingStep
- Moved configuration for endpoints to api service, it should not exist in other layers


chart/env/dev.yaml
```diff
@@ -166,7 +166,7 @@ configNames:
       cpu: 1
       memory: "4Gi"
 
-splitNames:
+splitNamesStreaming:
```
AndreaFrancis, line 169: I know that we should send another PR for chart folder but since this change is dependent on the worker name I just did the change of the names (Only names, nothing related to infrastructure configuration)

severo:
> TODO: Once we delete the relation endpoint-job_type, can we remove the "/" character from cache kind value for all the db? it would be confusing sometimes.

yes, I agree. But it should be a dedicated PR, because migrating the existing db is error-prone.

chart/env/dev.yaml
severo: Hmmm, I think it's better to keep with the logic that: if we don't update the docker images, we don't touch the chart. When we do a new release, the list of changes will make it clear what refactoring should be done in chart/. wdyt?

severo:
In the API, we have to fix https://github.com/huggingface/datasets-server/blob/3b863f523b0f3d0ce12037f0db87db25ae55518b/services/api/src/api/app.py#L74:

The loop should be on the endpoints, not the processing steps (or job types). For a given endpoint, we should find which job type / cache kind to use to get the response, depending on the parameters (ie relating to the "input_type" of the processing steps: dataset -> InputType==dataset, dataset +  config -> InputType ==config, dataset+config+split -> Not implemented). It should be precomputed and given by the processing graph. This also means that when computing the processing graph, we must check that, for each endpoint, we have zero or one step for every InputType.

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230213170000_split_names_streaming.py
```diff
@@ -0,0 +1,35 @@
...
+from mongodb_migration.check import check_documents
+from mongodb_migration.migration import Migration
+
+db_name = "cache"
+
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationUpdateSplitNames(Migration):
+    def up(self) -> None:
+        logging.info(f"Rename {db_name} kind field from /split-name to /split-names-streaming")
```
severo, line 19:
```suggestion
        logging.info(f"Rename cache_kind field from /split-name to /split-names-streaming")
```

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230213170000_split_names_streaming.py
```diff
@@ -0,0 +1,35 @@
...
+        db = get_db(db_name)
+        db["cachedResponsesBlue"].update_many({"kind": "/split-names"}, {"$set": {"kind": "/split-names-streaming"}})
+
+    def down(self) -> None:
+        logging.info(f"Rollback {db_name} kind field from /split-name-streaming to /split-names")
```
severo, line 24:
```suggestion
        logging.info(f"Rollback cache_kind field from /split-name-streaming to /split-names")
```

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -32,4 +35,8 @@ def get_migrations(self) -> List[Migration]:
                 version="20230126164900",
                 description="add 'priority' field to jobs in queue database",
             ),
+            MigrationUpdateSplitNames(
```
severo, line 38: we should also migrate the jobs in the queue database, no?

severo: btw: there are so many past jobs that we don't need anymore that we could delete them periodically. Manually or with a cron job. Otherwise, the migration jobs take forever, and the helm update timeouts and fails. A good reference for that is https://andrewlock.net/deploying-asp-net-core-applications-to-kubernetes-part-7-running-database-migrations/#helm-chart-hooks

> My initial testing of this approach generally worked well, with one exception. If a database migration took a long time, Helm would timeout waiting for the job to complete, and would fail the install. In reality, the job may or may not succeed in the cluster.
> 
> This was a deal breaker for us. Random timeouts, and the reality that production environments (with the larger quantity of data and higher database loads) were likely to be slower for migrations made us look elsewhere. Ultimately we settled on the init container approach I described in the previous section.

Maybe create two issues for that:
1. periodically clean the queue database deleting the old, finished jobs
2. migrate from jobs to init containers to ensure the migration has time to finish without making Helm upgrade timeout

severo: Note that for jobs, we have two fields to update, ideally: type and unicity_id

Capture d’écran 2023-02-14 à 15 11 10

severo:
Maybe this PR could be split in two:
- change the logic: one endpoint, multiple processing steps
- change the name split-names -> split-names-streaming.

It would be easier to review and think of.

severo:
Another thing I thought of: we should add an admin endpoint to query the response of a specific processing step. ie:
/admin/cache/split-names-streaming and /admin/cache/split-names-normal to debug the cached responses without digging into the database.

chart/env/dev.yaml
AndreaFrancis: Got it, I removed this changes

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: Got it, I will send the related code to migration split-names-streaming in another PR

libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -32,6 +33,7 @@ class ProcessingStep:
     - the next steps (the steps which previous step is the current one)
     """
 
+    job_type: str
```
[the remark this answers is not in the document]
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/817/files. I will close this PR since the branch name is obsolete.

Issue #764, opened 2023-02-02
Calling /api endpoints create unnecessary jobs


lhoestq:
For example calling /config-names does create a /config-names job even though the result is cached.

The job is then skipped since the result is cached.

severo:
> For example calling /config-names does create a /config-names job even though the result is cached.

I think the case is not exactly this one.

The logic is: if we call an endpoint, and if no cache entry exists, while it should (the dataset is supported: exists on the Hub and is not private), then we "update" the dataset, ie run the DAG (create a new job for every "init" processing step).

So, as such, calling /config-names should not create a /config-names job when the result is cached, but calling /first-rows will create a /config-names job if /first-rows cache is not filled, even if /config-names already has a cached entry.

Two solutions:
1. remove this behavior (create jobs when an endpoint has no cache entry but should have one) and just wait for the current jobs to finish (because the most typical case is that a previous processing step has not finished yet)
2. run a "backfill" on this dataset: find which cache entries are missing, and create the corresponding jobs. Note that if a previous processing step has not finished yet, we will do nothing, so it's the same as 1. The difference is only if something went wrong on the server, in which case I think it's better to manage this independently, for example by launching a backfill cron job every since and then to fill the missing cache entries if there are some.

severo:
I think we should stick with this for now.

The alternative would change the architecture drastically, which would take a significant amount of time (or it would mean using an external tool for this).

## Proposal

We should add a new service: orchestrator.

On one side, it would receive events:
- webhooks when a dataset has changed (added, updated, deleted, changed to gated, etc.)
- manual trigger: refresh a dataset, update all the datasets for a specific step, etc.

On the other side, it would command the jobs:
- create only the jobs that are needed
- receive the heartbeat to ensure a job is still running
- receive the result of a job and:
  - store it, 
  - create the dependent jobs based on the result

The orchestrator would have access to the queue and to the cache.

This means that the job runners can be a lot dumber:
- they don't need to check if they should skip the job or not: they just run what they are given
- they don't need to create the dependent jobs or even know about the processing graph
- they don't have access to the cache (see #751) and to the queue (but the worker still has to access the queue; we could also make the orchestrator launch the jobs on demand instead of having workers that loop and look for jobs in the queue. It's an unrelated issue, though.)

severo:
I propose to close this issue, and to open a new one to handle the refactoring with an orchestrator. What do you think @lhoestq?

lhoestq:
Sounds good !

[...]

PR #815, opened 2023-02-14
feat: 🎸 add a new admin endpoint: /dataset-status

severo:
While looking at https://github.com/huggingface/datasets-server/issues/764 and https://github.com/huggingface/datasets-server/issues/736#issuecomment-1412242342, I added a new admin endpoint that gives the current status of a dataset. I think it can help to get insights about a dataset when doing support manually.

It takes a `?dataset=` parameter, ie https://dataset-server.huggingface.co/admin/dataset-status?dataset=glue

As with all the admin endpoints, it requires being authenticated as part of the huggingface org.

### DIFF
--- a/libs/libcommon/src/libcommon/queue.py
+++ b/libs/libcommon/src/libcommon/queue.py
@@ -584,6 +584,18 @@ class Queue:
             )
         }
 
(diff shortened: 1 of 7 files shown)


lhoestq (approved):
Awesome ! Will be cool to have it in the admin UI ad well :)

AndreaFrancis (approved):

Issue #741, opened 2023-01-31
Detect the "zombie" jobs, and kill them 🧟 🔫


severo:
Sometimes the pods crash:

```
prod-datasets-server-worker-first-rows-8579994756-vgpkg           0/1     OutOfmemory              0               92m
prod-datasets-server-worker-first-rows-8579994756-vmvk7           0/1     OutOfmemory              0               92m
prod-datasets-server-worker-first-rows-8579994756-vsmmt           1/1     Running                  0               4h27m
prod-datasets-server-worker-first-rows-8579994756-vxtwn           0/1     OutOfmemory              0               3h12m
prod-datasets-server-worker-first-rows-8579994756-vzs6j           0/1     OutOfmemory              0               3h12m
prod-datasets-server-worker-first-rows-8579994756-w2k55           1/1     Running                  2 (3h25m ago)   4h16m
prod-datasets-server-worker-first-rows-8579994756-w5c6m           1/1     Running                  0               4h27m
prod-datasets-server-worker-first-rows-8579994756-w5ks6           1/1     Running                  1 (4h21m ago)   4h27m
prod-datasets-server-worker-first-rows-8579994756-w7ds5           1/1     Running                  0               4h27m
prod-datasets-server-worker-first-rows-8579994756-wbqlq           1/1     Running                  0               4h16m
```

The job that was running then stays forever in the "started" status, what we can call a zombie.

The issue is that, for the rules implemented in the queue logic, it can prevent other jobs for the same dataset, or for the same user, to be processed. It even prevents to re-run the same job.

Ideally, we should detect that the job has failed, change its status to "error" and put an error response in the cache database.

To implement this, an option proposed by @XciD is:
- to have a parallel thread (heartbeat) that will update the job in the database every xxx seconds
- a "garbage collector" loop will look for zombies and finish them as described above

AndreaFrancis:
Why not to handle that validation before worker loop starts? I mean, before https://github.com/huggingface/datasets-server/blob/main/workers/datasets_based/src/datasets_based/main.py#L15  look all "started" jobs and cancel them and create another in the queue (Like in force-refresh).
With this approach we could avoid having parallel extra validation. I assume that while Service is alive and a job is in "started" status, it is because something is being computed and indeed is "alive" but, if service starts and there are "started" jobs, they should be "death".

severo:
[...]

PR #824, opened 2023-02-15
Add heartbeat

lhoestq:
Add heartbeat to workers.

It adds a `last_heartbeat` field to documents in the queue.

The field is not mandatory - it only appears for jobs that are or were running when a heartbeat happens (once per minute by default).

## Implementations details

I added a `WorkerExecutor` that runs the worker loop in a **subprocess**. This way the executor can have its own loop with a heartbeat.

The executor knows about the worker state by reading a JSON file when I store the state of the loop. This is helpful to know the ID of the current Job to update its last_heartbeat field. I used filelock to make sure there's no race conditions when reading/writing this file.

## TODO

- [x] fix merge conflicts
- [x] tests

related to https://github.com/huggingface/datasets-server/issues/741


services/worker/src/worker/config.py
```diff
@@ -47,6 +50,10 @@ def from_env(cls) -> "WorkerConfig":
                 sleep_seconds=env.int(name="SLEEP_SECONDS", default=WORKER_SLEEP_SECONDS),
                 only_job_types=env.list(name="ONLY_JOB_TYPES", default=get_empty_str_list()),
                 storage_paths=env.list(name="STORAGE_PATHS", default=get_empty_str_list()),
+                state_path=env.str(name="STATE_PATH", default=None),
```
AndreaFrancis, line 53: I think we could move the default value WORKER_STATE_PATH here  "worker_state.json", almost all configs had their own default value

libs/libcommon/src/libcommon/queue.py
```diff
@@ -148,6 +148,7 @@ class Job(Document):
     created_at = DateTimeField(required=True)
     started_at = DateTimeField()
     finished_at = DateTimeField()
+    last_heartbeat = DateTimeField()
```
AndreaFrancis, line 151: Can you add a brief description of the field on the args doc? after line 123

AndreaFrancis (commented):
Awesome! 
But after the having last_heartbeat, will we still need another process to CANCEL/RETRY those jobs with a last_heartbeat more than 60 seconds (It could be more)? 
Another dummy question, I assume the json file would persist even if the pod crashes right? If so,I it makes me think again in the logic: On service start, would we read the file and see if there is a curent_job_info with "STARTED" state and CANCEL/RETRY it? Since the file lives in a pod isolated storage I think we could assume that if there is a job in the file it means that one was being processed by the pod that crashed.

services/worker/src/worker/config.py
lhoestq: `WORKER_STATE_PATH ` has no default - it changes for every sessions using a temporary directory.

lhoestq:
this is ready for review :)

services/worker/pyproject.toml
```diff
@@ -40,6 +40,7 @@ transformers = "^4.26.1"
 trec-car-tools = { path = "vendors/trec-car-tools/python3" }
 typer = "^0.4.2"
 wget = "^3.2"
+mirakuru = "^2.4.2"
```
severo, line 43: nit: we try to get the dependencies sorted, it's easier to read

services/worker/src/worker/main.py
```diff
@@ -1,53 +1,90 @@
...
-        LibrariesResource(
-            hf_endpoint=app_config.common.hf_endpoint,
-            init_hf_datasets_cache=app_config.datasets_based.hf_datasets_cache,
-            numba_path=app_config.numba.path,
-        ) as libraries_resource,
-        CacheMongoResource(
-            database=app_config.cache.mongo_database, host=app_config.cache.mongo_url
-        ) as cache_resource,
-        QueueMongoResource(
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
severo, line 77: Are we sure this cannot raise? Depending on how we deploy, maybe we have no access to the directory, and it cannot be created/deleted (see [/chart](https://github.com/huggingface/datasets-server/tree/main/chart)). If we have `WORKER_STATE_FILE_NAME` in the env, we don't need to create a tmp directory.

services/worker/src/worker/config.py
```diff
@@ -19,6 +19,7 @@
 WORKER_MAX_LOAD_PCT = 70
 WORKER_MAX_MEMORY_PCT = 80
 WORKER_SLEEP_SECONDS = 15
+WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS = 60
```
severo, line 22: nit here too: we try to keep the values sorted

services/worker/src/worker/config.py
```diff
@@ -34,6 +35,8 @@ class WorkerConfig:
     only_job_types: list[str] = field(default_factory=get_empty_str_list)
     sleep_seconds: int = WORKER_SLEEP_SECONDS
     storage_paths: List[str] = field(default_factory=get_empty_str_list)
+    state_path: Optional[str] = None
+    heartbeat_time_interval_seconds: int = WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS
```
severo, line 39: nit: shorter name? `WORKER_HEARTBEAT_SECONDS`, or `WORKER_HEARTBEAT_INTERVAL_SECONDS`?

services/worker/src/worker/config.py
severo: Yes, that's already what we do with the assets directory: https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/config.py#L14

Note: I prefer to always indirect via a constant, ie. `WORKER_STATE_PATH: Optional[str] = None`. Not sure if it's an exaggeration, though.

severo: Another comment: I didn't get that STATE_PATH is the state file basename (we create a file by appending `.lock` at the end at https://github.com/huggingface/datasets-server/pull/824/files#diff-643260fba42f231dbf4e91102b52af6acb3c216a0eb6f538ac42bc667ebe5381R145). Maybe the name could be more descriptive, or maybe we could replace by `STATE_FILENAME` and just use it without appending .lock?

services/worker/src/worker/loop.py
```diff
@@ -119,13 +126,23 @@ def process_next_job(self) -> bool:
...
         job_runner = self.job_runner_factory.create_job_runner(job_info)
         finished_status = job_runner.run()
         self.queue.finish_job(job_id=job_runner.job_id, finished_status=finished_status)
+        self.set_worker_state(current_job_info=None)
         logging.debug(f"job finished with {finished_status.value}: {job_runner}")
         return True
+
+    def set_worker_state(self, current_job_info: Optional[JobInfo]) -> None:
+        worker_state: WorkerState = {"current_job_info": current_job_info}
+        if self.worker_config.state_path:
```
severo, line 145: if it set to None, what does it imply?

services/worker/src/worker/main.py
```diff
@@ -1,53 +1,90 @@
-# SPDX-License-Identifier: Apache-2.0
-# Copyright 2022 The HuggingFace Authors.
+import json
```
severo, line 1: you removed the copyright header

services/worker/src/worker/main.py
```diff
@@ -1,53 +1,90 @@
...
-from worker.job_runner_factory import JobRunnerFactory
-from worker.loop import Loop
-from worker.resources import LibrariesResource
+from worker.loop import WorkerState
+
+WORKER_STATE_FILE_NAME = "worker_state.json"
+START_WORKER_LOOP_PATH = start_worker_loop.__file__
+
+
+class WorkerExecutor:
```
severo, line 23: Could we move that to its own file? so that main.py remains as easy to understand as possible?

services/worker/src/worker/config.py
```diff
@@ -47,6 +50,10 @@ def from_env(cls) -> "WorkerConfig":
                 sleep_seconds=env.int(name="SLEEP_SECONDS", default=WORKER_SLEEP_SECONDS),
                 only_job_types=env.list(name="ONLY_JOB_TYPES", default=get_empty_str_list()),
                 storage_paths=env.list(name="STORAGE_PATHS", default=get_empty_str_list()),
+                state_path=env.str(name="STATE_PATH", default=None),
+                heartbeat_time_interval_seconds=env.int(
+                    name="HEARTBEAT_TIME_INTERVAL_SECONDS", default=WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS
+                ),
```
severo, line 56: The two new env vars should be documented in the README.md

services/worker/src/worker/main.py
```diff
@@ -1,53 +1,90 @@
...
+    def _create_worker_loop_executor(self) -> OutputExecutor:
+        banner = self.app_config.worker.state_path
+        if not banner:
+            raise ValueError("Failed to create the executor because WORKER_STATE_PATH is missing.")
+        start_worker_loop_command = [
+            sys.executable,
+            START_WORKER_LOOP_PATH,
+            "--print-worker-state-path",
+        ]
+        return OutputExecutor(start_worker_loop_command, banner, timeout=10)
```
severo, line 36: Do you think we could use [PidExecutor](https://github.com/ClearcodeHQ/mirakuru#pidexecutor) and rely on the existence of the lock file?
I never worked with mirakuru, so I wonder if it would work.

I find the current implementation a bit weird (even if it's surely one of the simplest possible). Btw: why not just printing a static string instead of the path?

services/worker/src/worker/main.py
```diff
@@ -1,53 +1,90 @@
...
+
+    def start(self) -> None:
+        worker_loop_executor = self._create_worker_loop_executor()
+        worker_loop_executor.start()  # blocking until the banner is printed
+        logging.info("Starting heartbeat.")
+        while worker_loop_executor.running():
+            self.heartbeat()
+            time.sleep(self.app_config.worker.heartbeat_time_interval_seconds)
+        worker_loop_executor.stop()
```
severo, line 45: we have to stop it if it's not running? :P

services/worker/src/worker/main.py
lhoestq: PidExecutor is less robust IMO, because we might launch a wrong script and don't know about it. Printing a value confirms that we're launching the right script.

Also printing the worker state path was just a way to confirm that the subprocess received it and is using it

services/worker/src/worker/main.py
lhoestq: Yes, this way it gets the exit code and raises an error if it was 1

services/worker/src/worker/main.py
```diff
@@ -1,53 +1,90 @@
...
+        worker_state_path = self.app_config.worker.state_path
+        if not worker_state_path:
+            raise ValueError("Failed to get worker state because WORKER_STATE_PATH is missing.")
+        if os.path.exists(worker_state_path):
+            with FileLock(worker_state_path + ".lock"):
+                try:
+                    with open(worker_state_path, "r") as worker_state_f:
+                        worker_state = json.load(worker_state_f)
+                        return WorkerState(current_job_info=worker_state.get("current_job_info"))
+                except json.JSONDecodeError:
```
severo, line 57: maybe we want to crash early here

services/worker/src/worker/main.py
```diff
@@ -1,53 +1,90 @@
...
+            self.heartbeat()
+            time.sleep(self.app_config.worker.heartbeat_time_interval_seconds)
+        worker_loop_executor.stop()
+
+    def get_state(self) -> WorkerState:
+        worker_state_path = self.app_config.worker.state_path
+        if not worker_state_path:
+            raise ValueError("Failed to get worker state because WORKER_STATE_PATH is missing.")
+        if os.path.exists(worker_state_path):
+            with FileLock(worker_state_path + ".lock"):
```
severo, line 52: same as above: we could set the filename before, instead of appending .lock everytime (not a performance issue, just simpler code)

services/worker/src/worker/main.py
lhoestq: It's supposed to work - but feel free to connect to a worker and try it out x)

Otherwise we can define a dedicated directory (and we can use the env var `TMPDIR` to tell python to use it)

severo:
Super nice implementation, thanks! I put some comments, but it's so good to finally have the heartbeat!

services/worker/src/worker/main.py
severo: OK, perfect

services/worker/src/worker/main.py
severo: Oh, OK!

services/worker/src/worker/loop.py
severo: Oh, nice.

services/worker/src/worker/main.py
severo: https://github.com/huggingface/datasets-server/pull/850/files#r1118571249

services/worker/src/worker/main.py
severo: OK, let's it like this for now.

PR #825, opened 2023-02-16
Renaming split-names to split-names-from-streaming

AndreaFrancis:
Part of code for https://github.com/huggingface/datasets-server/issues/755
Since we are going to handle two different sources for split-names, the previous worker/processing step/cache-kind have to be changed to /split-names-from-streaming (There already exist /split-names-from-dataset-info)
TODO: Once this PR is merged, a new PR with changes on docker image/chart should be sent in order to change the WORKER_ONLY_JOB_TYPES param as well


lhoestq (approved):
Sounds good to me, thanks :)

I also checked that moon landing was not using this endpoint, and there is nothing to update in the docs

jobs/mongodb_migration/src/mongodb_migration/check.py
```diff
@@ -16,8 +16,8 @@
 
 
 def get_random_oids(collection: Collection, sample_size: int) -> List[int]:
-    pipeline = [{"$project": {"pk": 1}}, {"$sample": {"size": sample_size}}]
-    return [s["pk"] for s in collection.aggregate(pipeline)]
+    pipeline = [{"$project": {"_id": 1}}, {"$sample": {"size": sample_size}}]
```
severo, line 19: why? I changed it to fix a mypy issue (pk is always available)

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -32,4 +38,12 @@ def get_migrations(self) -> List[Migration]:
                 version="20230126164900",
                 description="add 'priority' field to jobs in queue database",
             ),
+            MigrationCacheUpdateSplitNames(
+                version="20230216112500",
+                description="update 'kind' field in cache from /split-names to /split-names-streaming",
```
severo, line 43: shouldn't it be `/split-names-from-streaming`?

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -32,4 +38,12 @@ def get_migrations(self) -> List[Migration]:
...
+            ),
+            MigrationQueueUpdateSplitNames(
+                version="20230216141000",
+                description="update 'type' and 'unicity_id' fields in job from /split-names to /split-names-streaming",
```
severo, line 47: shouldn't it be `/split-names-from-streaming`?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
```diff
@@ -0,0 +1,30 @@
...
+
+db_name = "cache"
+
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationCacheUpdateSplitNames(Migration):
+    def up(self) -> None:
+        logging.info("Rename cache_kind field from /split-name to /split-names-streaming")
+        db = get_db(db_name)
+        db["cachedResponsesBlue"].update_many({"kind": "/split-names"}, {"$set": {"kind": "/split-names-streaming"}})
```
severo, line 20: shouldn't it be `/split-names-from-streaming`?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
```diff
@@ -0,0 +1,30 @@
...
+from mongodb_migration.check import check_documents
+from mongodb_migration.migration import Migration
+
+db_name = "cache"
+
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationCacheUpdateSplitNames(Migration):
+    def up(self) -> None:
+        logging.info("Rename cache_kind field from /split-name to /split-names-streaming")
```
severo, line 18: shouldn't it be `/split-names-from-streaming`?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216141000_queue_split_names_from_streaming.py
```diff
@@ -0,0 +1,48 @@
...
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationQueueUpdateSplitNames(Migration):
+    def up(self) -> None:
+        logging.info(
+            f"Rename unicity_id field from Job[{split_names}][<dataset>][<config>][<split>] to"
+            f" Job[{split_names_from_streaming}][<dataset>][<config>][<split>]"
+        )
+
+        for job in Job.objects(type=split_names):
```
severo, line 25: I don't know if we should use mongoengine in the migration scripts since a mongoengine document is related to one version of the object, while here we are migrating from one version to the other. I tried to use pymongo directly. See also https://docs.mongoengine.org/guide/migration.html

severo: Also, maybe we could update both fields at the same time

jobs/mongodb_migration/src/mongodb_migration/check.py
AndreaFrancis: It throws an error when running migration's validation it didn't find the pk field, did you validated running the migrations with existing data when modified to fix the mypy issue?

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: Thanks, I will send a new PR to fix this

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: Same

jobs/mongodb_migration/src/mongodb_migration/check.py
severo: hmmm, ok, so maybe I'm wrong in the rest of the migrations! We could open a new issue to check that (and add some tests with an extract of the real data - the migrations are not tested, not sure it's overkill, or on the opposite, a good idea)

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216141000_queue_split_names_from_streaming.py
AndreaFrancis: I created another PR to fix these typos and avoid using the documents https://github.com/huggingface/datasets-server/pull/837
One thing I couldn't do:  use functions/aggregations to replace the unicity_id in update_many function, I used the update/save instead

jobs/mongodb_migration/src/mongodb_migration/collector.py
severo: https://github.com/huggingface/datasets-server/pull/837/files

jobs/mongodb_migration/src/mongodb_migration/collector.py
severo: https://github.com/huggingface/datasets-server/pull/837

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
severo: https://github.com/huggingface/datasets-server/pull/837

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
severo: https://github.com/huggingface/datasets-server/pull/837

jobs/mongodb_migration/src/mongodb_migration/check.py
severo: I created https://github.com/huggingface/datasets-server/issues/859.

PR #826, opened 2023-02-17
Kill zombies

lhoestq:
I defined zombies as started jobs with a `last_heartbeat` that is older than `max_missing_heartbeats * heartbeat_time_interval_seconds`

Then I added `kill_zombies` to the worker executor. It runs every `kill_zombies_time_interval_seconds` seconds and set the zombie jobs status to ERROR.

Given that heartbeats happen every 60s, I defined:
- max_missing_heartbeats = 5
- kill_zombies_time_interval_seconds = 10min

## Implementations details

To keep full control on the time interval for every action the executor runs, I switched to `asyncio`

### DIFF
--- a/services/worker/src/worker/config.py
+++ b/services/worker/src/worker/config.py
@@ -20,6 +20,8 @@ WORKER_MAX_LOAD_PCT = 70
 WORKER_MAX_MEMORY_PCT = 80
 WORKER_SLEEP_SECONDS = 15
 WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS = 60
+WORKER_MAX_MISSING_HEARTBEATS = 5
+WORKER_KILL_ZOMBIES_TIME_INTERVAL_SECONDS = 10 * 60
 
 
 def get_empty_str_list() -> List[str]:
@@ -37,6 +39,8 @@ class WorkerConfig:
     storage_paths: List[str] = field(default_factory=get_empty_str_list)
     state_path: Optional[str] = None
     heartbeat_time_interval_seconds: int = WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS
+    max_missing_heartbeats: int = WORKER_MAX_MISSING_HEARTBEATS
+    kill_zombies_time_interval_seconds: int = WORKER_KILL_ZOMBIES_TIME_INTERVAL_SECONDS
 
     @classmethod
     def from_env(cls) -> "WorkerConfig":
@@ -53,6 +57,10 @@ class WorkerConfig:
                 state_path=env.str(name="STATE_PATH", default=None),
                 heartbeat_time_interval_seconds=env.int(
                     name="HEARTBEAT_TIME_INTERVAL_SECONDS", default=WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS
+                ),
+                max_missing_heartbeats=env.int(name="MAX_MISSING_HEARTBEATS", default=WORKER_MAX_MISSING_HEARTBEATS),
+                kill_zombies_time_interval_seconds=env.int(
+                    name="KILL_ZOMBIES_TIME_INTERVAL_SECONDS", default=WORKER_KILL_ZOMBIES_TIME_INTERVAL_SECONDS
                 ),
             )
 
--- a/services/worker/src/worker/main.py
+++ b/services/worker/src/worker/main.py
@@ -1,11 +1,13 @@
+import asyncio
 import json
 import logging
 import os
 import sys
 import tempfile
-import time
-from typing import Optional
+from datetime import timedelta
+from typing import Any, Callable, List, Optional
 
+import pytz
 from filelock import FileLock
 from libcommon.log import init_logging
 from libcommon.queue import Job, Status, get_datetime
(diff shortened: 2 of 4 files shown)


services/worker/src/worker/main.py
```diff
@@ -38,11 +50,14 @@ def _create_worker_loop_executor(self) -> OutputExecutor:
...
+        loop = asyncio.get_event_loop()
         logging.info("Starting heartbeat.")
-        while worker_loop_executor.running():
-            self.heartbeat()
-            time.sleep(self.app_config.worker.heartbeat_time_interval_seconds)
-        worker_loop_executor.stop()
+        loop.create_task(every(self.heartbeat, seconds=self.app_config.worker.heartbeat_time_interval_seconds))
+        loop.create_task(every(self.kill_zombies, seconds=self.app_config.worker.kill_zombies_time_interval_seconds))
+        loop.run_until_complete(
+            every(self.is_worker_alive, worker_loop_executor=worker_loop_executor, seconds=1, stop_on=False)
```
AndreaFrancis, line 59: Do we need this hard-coded? seconds=1

services/worker/src/worker/main.py
```diff
@@ -72,6 +87,33 @@ def heartbeat(self) -> None:
...
+                and get_datetime()
+                >= pytz.UTC.localize(job.last_heartbeat)
+                + timedelta(seconds=max_missing_heartbeats * heartbeat_time_interval_seconds)
+            ]
+        else:
+            return []
+
+    def kill_zombies(self) -> None:
+        zombies = self.get_zombies()
+        Job.objects(pk__in=[zombie.pk for zombie in zombies]).update(status=Status.ERROR)
```
AndreaFrancis, line 108: We might need to updated also finished_at right? or maybe call the `finish_job` function in the queue

AndreaFrancis (approved):
Great! I just put a couple of questions, but it looks good!

services/worker/src/worker/main.py
lhoestq: We want the executor to stop as soon as the worker stopped running, so it's fine to hardcode it IMO

services/worker/src/worker/main.py
lhoestq: I added `finished_at` since `finish_job` is used for a single job and here we may do it on lots of jobs.

AndreaFrancis (approved):

services/worker/src/worker/main.py
severo: as mentioned in https://github.com/huggingface/datasets-server/pull/824#discussion_r1116047384, I would prefer to have this implementation inside queue.py, with unit tests.

services/worker/src/worker/config.py
```diff
@@ -54,6 +58,10 @@ def from_env(cls) -> "WorkerConfig":
                 heartbeat_time_interval_seconds=env.int(
                     name="HEARTBEAT_TIME_INTERVAL_SECONDS", default=WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS
                 ),
+                max_missing_heartbeats=env.int(name="MAX_MISSING_HEARTBEATS", default=WORKER_MAX_MISSING_HEARTBEATS),
+                kill_zombies_time_interval_seconds=env.int(
+                    name="KILL_ZOMBIES_TIME_INTERVAL_SECONDS", default=WORKER_KILL_ZOMBIES_TIME_INTERVAL_SECONDS
+                ),
```
severo, line 64: could we add them, with a description, to README.md?

services/worker/src/worker/config.py
```diff
@@ -20,6 +20,8 @@
 WORKER_MAX_MEMORY_PCT = 80
 WORKER_SLEEP_SECONDS = 15
 WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS = 60
+WORKER_MAX_MISSING_HEARTBEATS = 5
+WORKER_KILL_ZOMBIES_TIME_INTERVAL_SECONDS = 10 * 60
```
severo, line 24: nit: could we sort them alphabetically?

services/worker/src/worker/config.py
```diff
@@ -37,6 +39,8 @@ class WorkerConfig:
     storage_paths: List[str] = field(default_factory=get_empty_str_list)
     state_path: Optional[str] = None
     heartbeat_time_interval_seconds: int = WORKER_HEARTBEAT_TIME_INTERVAL_SECONDS
+    max_missing_heartbeats: int = WORKER_MAX_MISSING_HEARTBEATS
+    kill_zombies_time_interval_seconds: int = WORKER_KILL_ZOMBIES_TIME_INTERVAL_SECONDS
```
severo, line 43: We might find a shorter name

services/worker/src/worker/main.py
```diff
@@ -72,6 +87,50 @@ def heartbeat(self) -> None:
         if current_job:
             current_job.update(last_heartbeat=get_datetime())
 
+    def get_zombies(self) -> List[Job]:
+        started_jobs = Job.objects(status=Status.STARTED)
```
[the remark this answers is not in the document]
severo: For example, it would help managing the time/timezone at one place, which can be a source of errors.

severo:
Yeah! Dead zombies.
Some comments, that I can work on in a new PR.

PR #827, opened 2023-02-20
Set error response when zombie job is killed

lhoestq:
Last step for https://github.com/huggingface/datasets-server/issues/741 regarding zombies

### DIFF
--- a/services/worker/src/worker/main.py
+++ b/services/worker/src/worker/main.py
@@ -41,7 +41,7 @@ async def every(
 
 class WorkerCrashedError(CustomError):
     def __init__(self, message: str, cause: Optional[BaseException] = None):
-        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "WorkerCrashedError", cause, True)
+        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "WorkerCrashedError", cause, False)
 
 
 class WorkerExecutor:
--- a/services/worker/src/worker/main.py
+++ b/services/worker/src/worker/main.py
@@ -39,9 +39,9 @@ async def every(
         await asyncio.sleep(seconds)
 
 
-class JobCrashedError(CustomError):
+class WorkerCrashedError(CustomError):
     def __init__(self, message: str, cause: Optional[BaseException] = None):
-        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "JobCrashedError", cause, True)
+        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "WorkerCrashedError", cause, True)
 
 
 class WorkerExecutor:
@@ -147,7 +147,7 @@ class WorkerExecutor:
                     priority=zombie.priority,
                 )
                 job_runner = self.job_runner_factory.create_job_runner(job_info)
-                error = JobCrashedError("Job crashed and was automatically killed from the queue.")
+                error = WorkerCrashedError("Worker crashed while running this job.")
                 upsert_response(
                     kind=job_runner.processing_step.cache_kind,
                     dataset=job_runner.dataset,
--- a/services/worker/src/worker/main.py
+++ b/services/worker/src/worker/main.py
@@ -5,18 +5,25 @@ import os
 import sys
 import tempfile
 from datetime import timedelta
+from http import HTTPStatus
 from typing import Any, Callable, List, Optional
 
 import pytz
 from filelock import FileLock
+from libcommon.exceptions import CustomError
 from libcommon.log import init_logging
-from libcommon.queue import Job, Status, get_datetime
-from libcommon.resources import QueueMongoResource
+from libcommon.processing_graph import ProcessingGraph
+from libcommon.queue import Job, JobInfo, Status, get_datetime
+from libcommon.resources import CacheMongoResource, QueueMongoResource
+from libcommon.simple_cache import upsert_response
+from libcommon.storage import init_assets_dir
 from mirakuru import OutputExecutor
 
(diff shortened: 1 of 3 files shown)


services/worker/src/worker/main.py
```diff
@@ -32,9 +39,15 @@ async def every(
         await asyncio.sleep(seconds)
 
 
+class JobCrashedError(CustomError):
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "JobCrashedError", cause, True)
```
lhoestq, line 44: Not sure if we should disclose the error to the user here - maybe I should set it to False
```suggestion
        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "JobCrashedError", cause, False)
```

services/worker/src/worker/main.py
```diff
@@ -32,9 +39,15 @@ async def every(
         await asyncio.sleep(seconds)
 
 
+class WorkerCrashedError(CustomError):
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "WorkerCrashedError", cause, True)
```
lhoestq, line 44: Not sure if we should disclose the error to the user here - maybe I should set it to False
```suggestion
        super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "WorkerCrashedError", cause, False)
```

AndreaFrancis: Agree, since it is an internal error that we should handle it is better to False

services/worker/src/worker/main.py
```diff
@@ -123,6 +136,35 @@ def kill_zombies(self) -> None:
...
+                upsert_response(
+                    kind=job_runner.processing_step.cache_kind,
+                    dataset=job_runner.dataset,
+                    config=job_runner.config,
+                    split=job_runner.split,
+                    content=dict(error.as_response()),
+                    http_status=error.status_code,
+                    error_code=error.code,
+                    details=dict(error.as_response_with_cause()),
+                    worker_version=job_runner.get_version(),
```
AndreaFrancis, line 160: Not sure if we should get the current version of the job, I was thinking, what if the service crashed (Included the WorkerExecutor) and we are deploying a new job version?, it will assume the version which the job crashed is the new one and not the previous one. But not sure

AndreaFrancis (approved):
I put two comments (questions) but nothing critical, I think it looks good thanks!

services/worker/src/worker/main.py
lhoestq: Maybe we should add the job version to the `Job` objects in the queue then ? Otherwise we can't know if the zombie job is using the current version or a different one

lhoestq: created https://github.com/huggingface/datasets-server/issues/830

severo: Related: the last dataset commit can also have changed. So: we should specify both in the job, I think.

Beware, it also means that we should add some logic to cancel previous jobs on every new commit to main.

services/worker/src/worker/main.py
```diff
@@ -123,6 +136,35 @@ def kill_zombies(self) -> None:
             Job.objects(pk__in=[zombie.pk for zombie in zombies]).update(
                 status=Status.ERROR, finished_at=get_datetime()
             )
+            for zombie in zombies:
```
severo, line 139: All this logic is the same as in job_runner.py. We should extract it somewhere

services/worker/src/worker/main.py
```diff
@@ -123,6 +136,35 @@ def kill_zombies(self) -> None:
...
+                job_info = JobInfo(
+                    job_id=str(zombie.pk),
+                    type=zombie.type,
+                    dataset=zombie.dataset,
+                    config=zombie.config,
+                    split=zombie.split,
+                    force=zombie.force,
+                    priority=zombie.priority,
+                )
+                job_runner = self.job_runner_factory.create_job_runner(job_info)
```
severo, line 149: it's a bit weird to create a new job runner, just to update the cache?

services/worker/src/worker/main.py
```diff
@@ -140,10 +182,35 @@ def is_worker_alive(self, worker_loop_executor: OutputExecutor) -> bool:
         app_config = AppConfig.from_env()
         init_logging(log_level=app_config.common.log_level)
 
-        with QueueMongoResource(
-            database=app_config.queue.mongo_database, host=app_config.queue.mongo_url
-        ) as queue_resource:
+        init_logging(log_level=app_config.common.log_level)
```
severo, line 185: already above

services/worker/src/worker/main.py
```diff
@@ -140,10 +182,35 @@ def is_worker_alive(self, worker_loop_executor: OutputExecutor) -> bool:
...
+        # ^ set first to have logs as soon as possible
+        assets_directory = init_assets_dir(directory=app_config.assets.storage_directory)
```
severo, line 187: OK. It's weird to have exactly the same code in two files...

severo:
Thanks. I added some comments, but I think all the logic is here.

services/worker/src/worker/main.py
lhoestq: Yea but I need it to be able to call `.get_version()` and `.get_dataset_git_revision()`

services/worker/src/worker/main.py
lhoestq: not sure it's worth factoring this

PR #829, opened 2023-02-22
Update datasets dependency to 2.10.0 version

### LINKED ISSUE #828: Update datasets to 2.10.0
After 2.10.0 `datasets` release, update dependencies on it.

albertvillanova:
Close #828.

### DIFF
--- a/services/worker/poetry.lock
+++ b/services/worker/poetry.lock
@@ -963,19 +963,19 @@ toml = ">=0.10.0,<0.11.0"
 
 [[package]]
 name = "datasets"
-version = "2.9.0"
+version = "2.10.0"
 description = "HuggingFace community-driven open-source library of datasets"
 category = "main"
 optional = false
 python-versions = ">=3.7.0"
 files = [
-    {file = "datasets-2.9.0-py3-none-any.whl", hash = "sha256:f1aa5b98959cddb30f5077448204c8ce4235a4f1c8ec2473920660ebd6fc304f"},
-    {file = "datasets-2.9.0.tar.gz", hash = "sha256:c82458d635539b5a5dbed0fba8837006dfc3c213a5bcc00e18a67789f0f0f16f"},
+    {file = "datasets-2.10.0-py3-none-any.whl", hash = "sha256:42968d0cc88608ebc4a92a2fc26b4cd65e38f3ce3f5bb90a468cae6a6f0de989"},
+    {file = "datasets-2.10.0.tar.gz", hash = "sha256:c2ddab0899e17590fe39bc5783828ac9a7331c74508ab7a345ea0dca3112ead0"},
 ]
 
 [package.dependencies]
 aiohttp = "*"
-dill = "<0.3.7"
+dill = ">=0.3.0,<0.3.7"
 fsspec = {version = ">=2021.11.1", extras = ["http"]}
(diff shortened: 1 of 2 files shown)


services/worker/src/worker/job_runners/parquet_and_dataset_info.py
```diff
@@ -503,7 +503,7 @@ def _request_size(url: str, hf_token: Optional[str] = None) -> Optional[int]:
     return int(size) if size is not None else size
 
 
-class _MockStreamingDownloadManager(StreamingDownloadManager):
+class _MockStreamingDownloadManager(StreamingDownloadManager):  # type: ignore
```
severo, line 506: not sure, but does it mean that mypy will totally ignore that class? If so, I would prefer to keep the previous way, with exceptions only for three lines, instead of a global exception for the class

severo (approved):
Thanks. A small question about mypy

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
albertvillanova: @severo please note this is needed because of `datasets` has become untyped.

You can check it:
- Just before I fix any mypy-related issue, in my second commit named "[Update poetry lock file](https://github.com/huggingface/datasets-server/pull/829/commits/993e6c7ce0012096c9c1054e138e53edf3ff2a04)", the CI already failed for this code line
  - CI: https://github.com/huggingface/datasets-server/actions/runs/4243203372/jobs/7375645159
  - Error:
    > src/worker/job_runners/parquet_and_dataset_info.py:506: error: Class cannot subclass "StreamingDownloadManager" (has type "Any")  [misc]
    As `datasets` is untyped, the `StreamingDownloadManager` has type `Any` and mypy does not allow to subclass from type `Any`

severo: > @severo please note this is needed because datasets has become untyped.

yes, I hadn't understood that. Therefore, we might have more things to change in the code? I'm not sure how mypy manages untyped, or partly typed, dependencies

albertvillanova: Also note that the "previous way" is not just ignoring the three lines above. If you look at the CI link above, you see that once `datasets` is untyped, we need:
- either adding the ignore comment to ALL datasets imports
- or instead (and equivalently), setting to ignore all `datasets` in the config file

severo: Yes, that's what I fear. But why is the CI working, if we are importing `datasets` everywhere? Is it because some of the methods are typed, but not others like `StreamingDownloadManager`?

albertvillanova: This specific mypy error is because we have `disallow_subclassing_any` enabled (by `strict = True` in the config), that is we tell mypy to raise an error if we subclass from a type `Any`.

albertvillanova: To avoid this general setting to be applied to this specific case, we need to put the `# type: ignore` comment here.

severo: OK!

Issue #561, opened 2022-09-15
Filter the list of datasets by feature type (image, audio, class label...)


severo:
The idea is to provide the information to moon-landing to improve the search and add tags to help the faceted search.

It might be done by:
-  adding a filter to /valid (/valid?feature_type=image): filters the valid datasets to only include datasets that have a least one split with at least one Image feature type
- or by adding the list of feature types to the /valid route response (or add a new route for that), so that moon-landing could fetch it regularly and update the search index, creating specific tags for every dataset

wdyt @julien-c @gary149 @lhoestq?

severo:
closing:
- /valid does not exist anymore
- /hub-cache will provide a new field: `tags: list[str]` that will be used by the Hub to associate tags to the datasets. See https://github.com/huggingface/datasets-server/pull/2386.

I created https://github.com/huggingface/datasets-server/issues/2408, in case we want to detect tasks.

PR #833, opened 2023-02-22
Lower parquet row group size for image datasets

lhoestq:
REQUIRES test_get_writer_batch_size to be merged, and to update the `datasets` version to use this feature.

This should help optimize random access to parquet files for https://github.com/huggingface/datasets-server/pull/687/files


services/worker/src/worker/job_runners/parquet_and_dataset_info.py
```diff
@@ -650,6 +652,25 @@ def raise_if_too_big_from_external_data_files(
...
+    Args:
+        ds_config_info (`datasets.info.DatasetInfo`):
+            Dataset info from `datasets`.
+
+    Returns:
+        writer_batch_size (`Optional[int]`):
+            Writer batch size to pass to a dataset builder.
+            If `None`, then it will use the `datasets` default.
+    """
+    return 100 if "Image(" in str(ds_config_info.features) else None
```
severo, line 671: Can we define a constant for `100`?

severo: we can add a constants.py file as in libcommon

services/worker/src/worker/job_runners/parquet_and_dataset_info.py
```diff
@@ -774,12 +795,17 @@ def compute_parquet_and_dataset_info_response(
     parquet_files: List[ParquetFile] = []
     dataset_info: dict[str, Any] = {}
     for config in config_names:
+        ds_config_info = get_dataset_config_info(
```
severo, line 798: `get_dataset_config_info` calls `load_dataset_builder` (that we also call below). Also: it can rely on streaming and fail if the dataset does not support streaming, right? In that case, it defeats the purpose of downloading the dataset. Maybe handle the case with a try/except block?

severo:
Hi, do you want to adapt this PR, now that #875 has been merged to main?

lhoestq:
We'll do a new release of `datasets` tomorrow to add the new parameter to `download_and_prepare`, I'll update this PR after the release

severo:
keep open

lhoestq:
this is ready for review @severo :)

severo (approved):
OK. We will have to force-refresh the image datasets. Should we increment the job runner version? It would refresh everything, instead of just the image datasets...

libs/libcommon/src/libcommon/constants.py
```diff
@@ -26,4 +26,5 @@
 PROCESSING_STEP_SPLIT_NAMES_FROM_STREAMING_VERSION = 3
 PROCESSING_STEP_DATASET_SPLIT_NAMES_VERSION = 2
 
+PROCESSING_STEP_PARQUET_AND_DATASET_INFO_ROW_GROUP_SIZE_FOR_IMAGE_DATASETS = 100
```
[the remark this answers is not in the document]
lhoestq: Even with a row group size of 1000 we were only saving the images of the requested range - not the full row group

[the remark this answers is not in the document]
severo: ah OK, you're right :)

lhoestq:
> Another comment: why don't we do the same for audio datasets?

we should ! let me update this

lhoestq:
Could we hardcode a rule that would not recompute the parquet files if there's no image/audio in it ?

severo:
> Could we hardcode a rule that would not recompute the parquet files if there's no image/audio in it ?

No, I think we should instead find a way to get the list of datasets with a specific type in it (it's useful anyway -> https://github.com/huggingface/datasets-server/issues/561)

lhoestq:
Ok :) I just increased the job version

PR #834, opened 2023-02-22
Endpoint respond by input type

AndreaFrancis:
Part of https://github.com/huggingface/datasets-server/issues/755
This change will allow endpoints respond based on input type having a list of processing steps 
e.g:
- /splits with dataset param will reach out to /splits cache kind
- /splits with config param will reach out to /splits-from-streaming first and then /splits-form-dataset-info

### DIFF
--- a/chart/static-files/openapi.json
+++ b/chart/static-files/openapi.json
@@ -1871,7 +1871,7 @@
                   "missing-dataset": {
                     "summary": "The dataset parameter is missing.",
                     "value": {
-                      "error": "Parameter 'dataset' is required"
+                      "error": "Parameters 'split', 'config' and 'dataset' are required"
                     }
                   },
                   "missing-config": {
@@ -1889,7 +1889,7 @@
                   "empty-dataset": {
                     "summary": "The dataset parameter is empty.",
                     "value": {
-                      "error": "Parameter 'dataset' is required"
+                      "error": "Parameters 'split', 'config' and 'dataset' are required"
                     }
                   },
                   "empty-config": {
--- a/chart/static-files/openapi.json
+++ b/chart/static-files/openapi.json
@@ -1877,7 +1877,7 @@
                   "missing-config": {
                     "summary": "The config parameter is missing.",
                     "value": {
-                      "error": "Parameters 'config' and 'dataset' are required"
+                      "error": "Parameters 'split', 'config' and 'dataset' are required"
                     }
                   },
                   "missing-split": {
--- a/chart/static-files/openapi.json
+++ b/chart/static-files/openapi.json
@@ -1877,13 +1877,13 @@
                   "missing-config": {
                     "summary": "The config parameter is missing.",
                     "value": {
-                      "error": "Parameters 'config' and 'split' are required"
+                      "error": "Parameter 'config' is required"
                     }
                   },
                   "missing-split": {
                     "summary": "The split parameter is missing.",
                     "value": {
-                      "error": "Parameters 'config' and 'split' are required"
+                      "error": "Operation not supported or not found"
                     }
                   },
                   "empty-dataset": {
@@ -1895,13 +1895,13 @@
                   "empty-config": {
                     "summary": "The config parameter is empty.",
                     "value": {
-                      "error": "Parameters 'config' and 'split' are required"
+                      "error": "Parameter 'config' is required"
                     }
                   },
                   "empty-split": {
                     "summary": "The split parameter is empty.",
                     "value": {
-                      "error": "Parameters 'config' and 'split' are required"
+                      "error": "Operation not supported or not found"
                     }
                   }
                 }
--- a/chart/static-files/openapi.json
+++ b/chart/static-files/openapi.json
@@ -1877,13 +1877,13 @@
                   "missing-config": {
                     "summary": "The config parameter is missing.",
                     "value": {
-                      "error": "Parameter 'config' is required"
+                      "error": "Parameters 'config' and 'dataset' are required"
                     }
                   },
                   "missing-split": {
                     "summary": "The split parameter is missing.",
                     "value": {
-                      "error": "Operation not supported or not found"
+                      "error": "Parameters 'split', 'config' and 'dataset' are required"
                     }
                   },
                   "empty-dataset": {
@@ -1895,13 +1895,13 @@
                   "empty-config": {
                     "summary": "The config parameter is empty.",
                     "value": {
-                      "error": "Parameter 'config' is required"
+                      "error": "Parameters 'split', 'config' and 'dataset' are required"
                     }
                   },
                   "empty-split": {
                     "summary": "The split parameter is empty.",
                     "value": {
-                      "error": "Operation not supported or not found"
+                      "error": "Parameters 'split', 'config' and 'dataset' are required"
                     }
                   }
                 }
--- a/e2e/tests/test_11_auth.py
+++ b/e2e/tests/test_11_auth.py
@@ -40,8 +40,6 @@ def test_auth_e2e(
     # asking for the dataset will launch the jobs, without the need of a webhook
     endpoints = [
         f"/config-names?dataset={dataset}",
-        f"/split-names-from-streaming?dataset={dataset}&config={config}",
-        # TODO: ^ this will be changed to /splits in an uncoming PR
         f"/splits?dataset={dataset}",
         f"/first-rows?dataset={dataset}&config={config}&split={split}",
         f"/parquet-and-dataset-info?dataset={dataset}",
--- a/libs/libcommon/src/libcommon/processing_graph.py
+++ b/libs/libcommon/src/libcommon/processing_graph.py
@@ -5,14 +5,9 @@ from __future__ import annotations
 
 import warnings
 from dataclasses import dataclass
-from enum import Enum
 from typing import List, Literal, Mapping, Optional, TypedDict
 
-
-class InputType(str, Enum):
-    DATASET = "dataset"
-    CONFIG = "config"
-    SPLIT = "split"
+InputType = Literal["dataset", "config", "split"]
 
 
 class _ProcessingStepSpecification(TypedDict):
--- a/libs/libcommon/src/libcommon/processing_graph.py
+++ b/libs/libcommon/src/libcommon/processing_graph.py
@@ -5,9 +5,14 @@ from __future__ import annotations
 
 import warnings
 from dataclasses import dataclass
+from enum import Enum
 from typing import List, Literal, Mapping, Optional, TypedDict
 
-InputType = Literal["dataset", "config", "split"]
+
+class InputType(str, Enum):
+    DATASET = "dataset"
+    CONFIG = "config"
+    SPLIT = "split"
 
 
 class _ProcessingStepSpecification(TypedDict):
--- a/services/api/src/api/app.py
+++ b/services/api/src/api/app.py
@@ -24,14 +24,18 @@ from api.routes.webhook import create_webhook_endpoint
 
 def create_app() -> Starlette:
     app_config = AppConfig.from_env()
+    endpoint_config = EndpointConfig.from_env()
+    return create_app_with_config(app_config=app_config, endpoint_config=endpoint_config)
 
+
+def create_app_with_config(app_config: AppConfig, endpoint_config: EndpointConfig) -> Starlette:
     init_logging(log_level=app_config.common.log_level)
     # ^ set first to have logs as soon as possible
 
     prometheus = Prometheus()
 
     processing_graph = ProcessingGraph(app_config.processing_graph.specification)
-    endpoints_definition = EndpointsDefinition(processing_graph, EndpointConfig.from_env())
+    endpoints_definition = EndpointsDefinition(processing_graph, endpoint_config)
     processing_steps_required_by_dataset_viewer = processing_graph.get_steps_required_by_dataset_viewer()
     init_processing_steps = processing_graph.get_first_steps()
 
--- a/services/api/src/api/app.py
+++ b/services/api/src/api/app.py
@@ -85,7 +85,7 @@ def create_app() -> Starlette:
                 max_age_short=app_config.api.max_age_short,
             ),
         )
-        for endpoint_name, steps_by_input_type in endpoints_definition.processing_steps_by_endpoint.items()
+        for endpoint_name, steps_by_input_type in endpoints_definition.steps_by_input_type_and_endpoint.items()
     ]
     to_protect: List[BaseRoute] = [
         # called by the Hub webhooks
--- a/services/api/src/api/app.py
+++ b/services/api/src/api/app.py
@@ -75,7 +75,8 @@ def create_app() -> Starlette:
         Route(
             endpoint_name,
             endpoint=create_endpoint(
-                input_types=input_types,
+                endpoint_name=endpoint_name,
+                steps_by_input_type=steps_by_input_type,
                 init_processing_steps=init_processing_steps,
                 hf_endpoint=app_config.common.hf_endpoint,
                 hf_token=app_config.common.hf_token,
@@ -84,7 +85,7 @@ def create_app() -> Starlette:
                 max_age_short=app_config.api.max_age_short,
             ),
         )
-        for endpoint_name, input_types in endpoints_definition.definition.items()
+        for endpoint_name, steps_by_input_type in endpoints_definition.processing_steps_by_endpoint.items()
     ]
     to_protect: List[BaseRoute] = [
         # called by the Hub webhooks
--- a/services/api/src/api/app.py
+++ b/services/api/src/api/app.py
@@ -75,7 +75,7 @@ def create_app() -> Starlette:
         Route(
             endpoint_name,
             endpoint=create_endpoint(
-                processing_steps=processing_steps,
+                input_types=input_types,
                 init_processing_steps=init_processing_steps,
                 hf_endpoint=app_config.common.hf_endpoint,
                 hf_token=app_config.common.hf_token,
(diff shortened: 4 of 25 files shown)


services/api/src/api/config.py
```diff
@@ -81,21 +81,24 @@ def from_env(cls) -> "AppConfig":
...
+            "/split-names-from-streaming": {"config": ["/split-names-from-streaming"]},
+            "/splits": {
+                "dataset": ["/splits"],
+                "config": ["/split-names-from-streaming", "/split-names-from-dataset-info"],
+            },
+            "/first-rows": {"split": ["/first-rows"]},
+            "/parquet-and-dataset-info": {"dataset": ["/parquet-and-dataset-info"]},
+            "/parquet": {"dataset": ["/parquet"]},
+            "/dataset-info": {"dataset": ["/dataset-info"]},
+            "/sizes": {"dataset": ["/sizes"]},
```
lhoestq, line 101: Maybe add a comment to explain that the lists correspond to the priority list of processing step that may have the response (first = prioritized, and we loop on the list until we get a response) ?

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,77 @@
...
+            return result
+        except DoesNotExist:
+            logging.warning(
+                f"processing_step={processing_step.name} dataset={params.dataset} "
+                f"config={params.config} split={params.split} no entry found"
+            )
+    return None
+
+
 def create_endpoint(
```
lhoestq, line 102: (nit) would be cool to add a test for this one at one point (not particularly in this PR though)

chart/static-files/openapi.json
```diff
@@ -1895,13 +1895,13 @@
...
                     "value": {
-                      "error": "Parameters 'config' and 'split' are required"
+                      "error": "Parameter 'config' is required"
                     }
                   },
                   "empty-split": {
                     "summary": "The split parameter is empty.",
                     "value": {
-                      "error": "Parameters 'config' and 'split' are required"
+                      "error": "Operation not supported or not found"
```
lhoestq, line 1904: why is it different from "empty-config" ?

lhoestq (approved):
Cool thanks !

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,77 @@
...
+
+    split = None
+    split_param = query_params.get("split")
+    if are_valid_parameters([split_param]):
+        input_type = "split"
+        split = split_param
+        if config is None:
+            raise MissingRequiredParameterError("Parameter 'config' is required")
+
+    return InputParams(input_type=input_type, dataset=dataset, config=config, split=split)
```
albertvillanova, line 81: Consequently:
```suggestion
def get_params(query_params: QueryParams) -> InputParams:
    dataset = query_params.get("dataset")

    if not are_valid_parameters([dataset]) or not dataset:
        raise MissingRequiredParameterError("Parameter 'dataset' is required")

    config = None
    config_param = query_params.get("config")
    if are_valid_parameters([config_param]):
        config = config_param

    split = None
    split_param = query_params.get("split")
    if are_valid_parameters([split_param]):
        split = split_param
        if config is None:
            raise MissingRequiredParameterError("Parameter 'config' is required")

    return InputParams(dataset=dataset, config=config, split=split)
```

albertvillanova (commented):
Thanks @AndreaFrancis, great work!

One suggestion that may simplify the code...

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,77 @@
...
+            for endpoint, input_types in endpoint_config.specification.items()
         }
 
 
+@dataclass
+class InputParams:
+    input_type: str
+    dataset: str
+    config: Optional[str]
+    split: Optional[str]
```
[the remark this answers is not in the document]
AndreaFrancis: Good idea, thanks!

services/api/src/api/routes/endpoint.py
AndreaFrancis: Agree, for now, I added test for the functions used inside this one.

chart/static-files/openapi.json
AndreaFrancis: Given the new logic of endpoints responding based on the params to its corresponding processing steps we have 
- Infer the input type from the query params
- Based on it, get the result from the processing step list
For this specific case I had to modify the response of /first-rows case with empty split since as per our configuration on
            "/first-rows": {"split": ["/first-rows"]},
we only support first-rows passing dataset, config and split params, if we pass only dataset and config, the process would assume that the client is trying to request /first-rows only config  input_type which is not supported.
I know that maybe a better message would be "split" param missing but what if in the future we support /first-rows with both input_types (config and split) , that message will never be shown.

services/api/src/api/app.py
```diff
@@ -75,7 +75,7 @@ def create_app() -> Starlette:
         Route(
             endpoint_name,
             endpoint=create_endpoint(
```
severo, line 77: it might be good to pass endpoint_name as a parameter for logging and error messages purposes (see comment below)

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,77 @@
...
+def first_entry_from_steps(processing_steps: List[ProcessingStep], params: InputParams) -> Optional[CacheEntry]:
+    for processing_step in processing_steps:
+        try:
+            result = get_response(
+                kind=processing_step.cache_kind,
+                dataset=params.dataset,
+                config=params.config,
+                split=params.split,
+            )
+            return result
```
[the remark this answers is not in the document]
AndreaFrancis: Done, also I added unit tests for the different scenarios

services/api/src/api/routes/endpoint.py
```diff
@@ -51,34 +109,27 @@ def create_endpoint(
...
+                f"input_type={params.input_type}, dataset={params.dataset}, config={params.config},"
+                f" split={params.split}"
+            )
+
+            if params.input_type not in input_types:
+                raise MissingRequiredParameterError("Operation not supported or not found")
+
+            processing_steps = input_types[params.input_type]
+            if not processing_steps:
+                raise MissingProcessingStepsError("Missing processing steps")
```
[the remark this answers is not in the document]
AndreaFrancis: Done, thanks

services/api/src/api/routes/endpoint.py
```diff
@@ -88,21 +139,23 @@ async def processing_step_endpoint(request: Request) -> Response:
                     return get_json_error_response(
                         content=content, status_code=http_status, max_age=max_age_short, error_code=error_code
                     )
-            except DoesNotExist as e:
-                # maybe the response is in process
+
+            # maybe the response is in process"
```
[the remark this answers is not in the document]
AndreaFrancis: Makes sense, I moved it on the function that validates the first succeeded entry

services/api/src/api/app.py
AndreaFrancis: Done

services/api/src/api/routes/endpoint.py
```diff
@@ -51,60 +156,31 @@ def create_endpoint(
...
-            logging.info(f"{processing_step.job_type}, dataset={dataset}, config={config}, split={split}")
+            # getting input params
+            params: InputParameters = get_input_parameters(request.query_params)
+            logging.info(
+                f"input_type={params.input_type}, dataset={params.dataset}, config={params.config},"
+                f" split={params.split}"
+            )
+
+            if params.input_type not in steps_by_input_type:
+                raise MissingRequiredParameterError("Operation not supported or not found")
```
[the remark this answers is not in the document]
severo: also, I think the pythonic way would be something like

```python
try:
  processing_steps = steps_by_input_type[params.input_type]
except KeyError as e:
  raise MissingRequiredParameterError("Operation not supported or not found") from e
```

instead of

```python
if params.input_type not in steps_by_input_type:
  raise MissingRequiredParameterError("Operation not supported or not found")

processing_steps = steps_by_input_type[params.input_type]
```

@albertvillanova is the pythonic advisor here :P

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,124 @@
...
+                    split=input_parameters.split,
+                    hf_endpoint=hf_endpoint,
+                    hf_token=hf_token,
+                )
+            except (PreviousStepError, DatasetError):
+                raise ResponseNotFoundError("Not found.")
+    return last_result
+
+
+def get_response_from_cache_entry(
```
severo, line 128: This part can stay inside the main method, no need to extract it as a method

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,124 @@
...
+                kind=processing_step.cache_kind,
+                dataset=input_parameters.dataset,
+                config=input_parameters.config,
+                split=input_parameters.split,
+            )
+
+            if last_result["http_status"] == HTTPStatus.OK:
+                return last_result
+        except DoesNotExist:
+            logging.warning(
```
severo, line 109: maybe not a warning, because it will be common to have missing cache entries (the job has not finished yet). Info or debug

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,124 @@
...
+                raise ResponseNotFoundError("Not found.")
+    return last_result
+
+
+def get_response_from_cache_entry(
+    result: Optional[CacheEntry],
+    max_age_long: int = 0,
+    max_age_short: int = 0,
+) -> Response:
+    if result:
```
severo, line 133: detail: I use to return as soon as possible when there is an error, to reduce the number of indented lines in the common case, that generally contains more code:

```python
if not result:
  raise ResponseNotReadyError(
        "The server is busier than usual and the response is not ready yet. Please retry later."
    )
content = result["content"]
...
```

severo (commented):
Thanks! Still some details, but it's much better now

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +30,124 @@
...
+StepsByInputType = Mapping[InputType, List[ProcessingStep]]
+
+StepsByInputTypeAndEndpoint = Mapping[str, StepsByInputType]
+
 
 class EndpointsDefinition:
     """Definition of supported endpoints and its relation with processing steps."""
 
-    processing_steps_by_endpoint: Mapping[str, List[ProcessingStep]]
+    processing_steps_by_endpoint: StepsByInputTypeAndEndpoint
```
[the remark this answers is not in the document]
AndreaFrancis: Done

services/api/src/api/routes/endpoint.py
AndreaFrancis: I separated it in order to unit test it, since async proc has other extra logic and only wanted to test this part, will leave async test for other PR for now

services/api/src/api/routes/endpoint.py
AndreaFrancis: I changed it to debug

services/api/src/api/routes/endpoint.py
AndreaFrancis: Done, thanks!

services/api/src/api/routes/endpoint.py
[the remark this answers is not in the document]
AndreaFrancis: > Is it what we want? (maybe)

The problem is that now that we support having different responses based on query params, we need to infer the input param to use.
What if the inferred input param is not supported?
I thought in these possible scenarios (Almost sure those are the unique ones):
- endpoint supporting only dataset input type
  - request with dataset param-> OK
  - request with dataset and config params-> Ignone the other fields
  - request with dataset, config and split params> Ignone the other fields
  - request with dataset, config, split and somethingelse params-> Ignone the other fields
- endpoint supporting dataset and config input types
  - request with dataset param-> Which of them should we use? dataset -> OK, config-> Throw missing param error
  - request with dataset and config params -> Which of them should we use? it will work for both, for dataset ignoring config and for config using both params 
  - request with dataset, config and split params-> Which of them should we use? It will work for dataset and config but for split-> Missing param error
  - request with dataset, config, split and somethingelse params -> Same as above, we could ignore the somethingelse param
- endpoint supporting dataset, config and split input types
  - request with dataset param-> Which input type should we use? dataset -> OK, config -> Missing param, split -> Missing param
  - request with dataset and config params -> Which input type should we use (dataset/config)? , split -> Missing param
  - request with dataset, config and split params -> Which input type should we use? it could work with all of them
  - request with dataset, config, split and somethingelse params -> Same as above, we could ignore the somethingelse param
  
  As you can see, there are many possibilities, I didn't want to complicate the validation.
  I thought the simple way to validate it could be only look for dataset,config and split param, if there is any other param they will be ignored (Like somethingelse=true ..).
  Based on dataset,config and split get the input type to use and if we dont support that logic, throw error.

What do you think? @huggingface/datasets-server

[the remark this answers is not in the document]
AndreaFrancis: I think I have a better idea, considering that each endpoint should support at least one associated input type:
1. Get the inferred input type from request params
2. Get the supported sorted input type list by the endpoint  (split > config > dataset , e.g if the endpoint supports all the input types, it will return [split, config, dataset], if the endpoint supports config and split, it will return [split,config])
3. Iterate over supported sorted input type list and if matches inferred input type, get the result based on its processing steps, else, store corresponding input type message
   -  split ("Parameters 'split', 'config' and 'dataset' are required")
   -  config ("Parameters 'config' and 'dataset' are required")
   -  dataset ("Parameter 'dataset' is required)
4.  At the end, if no one supported input type matched the inferred input type, return the last error message

e2e/tests/test_11_auth.py
```diff
@@ -40,8 +40,6 @@ def test_auth_e2e(
     # asking for the dataset will launch the jobs, without the need of a webhook
     endpoints = [
         f"/config-names?dataset={dataset}",
-        f"/split-names-from-streaming?dataset={dataset}&config={config}",
-        # TODO: ^ this will be changed to /splits in an uncoming PR
```
AndreaFrancis, line 44: I removed this one since it will be supported by /splits?dataset=...&config=....

services/api/tests/test_app.py
```diff
@@ -91,6 +91,7 @@ def test_get_endpoint(client: TestClient, first_dataset_endpoint: str) -> None:
     assert response.status_code == 422
 
 
+@pytest.mark.skip(reason="no way to test this scenario since there is no endpoint supporting only config")
```
AndreaFrancis, line 94: I marked this to skip because now that splits supports both config and dataset and if we send missing param for `config`, it wont return error it can reach out to `dataset` . 
See previous config in EndpointConfig:

```
            "/config-names": {InputType.DATASET: ["/config-names"]},
            "/splits": {
                InputType.DATASET: ["/splits"],
                InputType.CONFIG: ["/split-names-from-streaming", "/split-names-from-dataset-info"],
            },
            "/first-rows": {InputType.SPLIT: ["/first-rows"]},
            "/parquet-and-dataset-info": {InputType.DATASET: ["/parquet-and-dataset-info"]},
            "/parquet": {InputType.DATASET: ["/parquet"]},
            "/dataset-info": {InputType.DATASET: ["/dataset-info"]},
            "/sizes": {InputType.DATASET: ["/sizes"]},
```

severo: OK. We should keep the test, but create a dedicated processing step, instead of getting it from libcommon. Same for the next test

AndreaFrancis: I added other endpoint config specific for the api app testing

services/api/src/api/routes/endpoint.py
```diff
@@ -51,60 +143,45 @@ def create_endpoint(
...
-                if not are_valid_parameters([config, split]):
-                    raise MissingRequiredParameterError("Parameters 'config' and 'split' are required")
-            logging.info(f"{processing_step.job_type}, dataset={dataset}, config={config}, split={split}")
+
+            config = request.query_params.get("config")
+            split = request.query_params.get("split")
+
+            input_types = get_input_types_by_priority(steps_by_input_type=steps_by_input_type)
+
+            logging.debug(f"endpoint={endpoint_name} dataset={dataset} config={config} split={split}")
```
severo, line 158: it's a lot of if/elif... I'm wondering if we could simplify a bit (not sure how)

libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -5,9 +5,14 @@
...
 from dataclasses import dataclass
+from enum import Enum
 from typing import List, Literal, Mapping, Optional, TypedDict
 
-InputType = Literal["dataset", "config", "split"]
+
+class InputType(str, Enum):
+    DATASET = "dataset"
+    CONFIG = "config"
+    SPLIT = "split"
```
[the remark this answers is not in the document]
severo: Fixed, thanks!

services/api/src/api/routes/endpoint.py
AndreaFrancis: I think a strategy pattern will work in this case, let me try

services/api/src/api/routes/endpoint.py
```diff
@@ -28,21 +28,171 @@
...
+        return last_result
+
+    raise ResponseNotReadyError(
+        "The server is busier than usual and the response is not ready yet. Please retry later."
+    )
+
+
+class InputTypeValidator(ABC):
+    def __init__(self, input_type: InputType) -> None:
+        self.input_type = input_type
```
severo, line 107: Nice (the code is longer, but the logic is good)

nit: input_type should be an abstract field, set directly in the children classes, not a parameter (it's fixed for a given class)

severo (approved):

services/api/src/api/routes/endpoint.py
AndreaFrancis: Makes sense, thanks!

Issue #818, opened 2023-02-14
Periodically clean the queue database deleting the old, finished jobs


AndreaFrancis:
From comment on https://github.com/huggingface/datasets-server/pull/810#discussion_r1105871837

> there are so many past jobs that we don't need anymore that we could delete them periodically. Manually or with a cron job. Otherwise, the migration jobs take forever, and the helm update timeouts and fails.

@severo suggested:
 Periodically clean the queue database deleting the old, finished jobs

PR #837, opened 2023-02-23
Fix typos in split-names-from-streaming

AndreaFrancis:


jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216141000_queue_split_names_from_streaming.py
```diff
@@ -18,29 +18,33 @@
...
+            f" {split_names_from_streaming}"
         )
 
-        for job in Job.objects(type=split_names):
-            job.update(unicity_id=f"Job[{split_names_from_streaming}][{job.dataset}][{job.config}][{job.split}]")
-
-        logging.info(f"Rename type field from {split_names} to {split_names_from_streaming}")
         db = get_db("queue")
-        db["jobsBlue"].update_many({"type": split_names}, {"$set": {"type": split_names_from_streaming}})
+        for job in db["jobsBlue"].find({"type": split_names}):
```
severo, line 27: I didn't see it first, but I'm afraid the loop will take forever (one query per entry, and there are currently 9.971.923 entries in the collection!) and it will make the migration timeout.

I think that we must try to do a update_many query that updates the unicity_id.

If it's not possible (I don't know much about MongoDB queries), we should delete part of the historical jobs (#818) before running the migration.

AndreaFrancis: > I think that we must try to do a update_many query that updates the unicity_id.

Yes, I tried using the $set operator for update the unicity_id with replace and concat (based on dataset/config value) operation but they look to be not working on pymongo, I looked for documentation but I was not able to find a specific one. For mongo the indeed exist doc to update based on functions like $replace $concat but it did'nt worked with pymongo, any ideas? 
I tried this 
` db["jobsBlue"].update_many(
            {"type": "/split-names"},
            {
                "$set": {
                    "unicity_id": {
                        "$replaceOne": {
                            "input": "$unicity_id",
                            "find": "Job[/split-names]",
                            "replacement": "Job[/split-names-from-streaming]",
                        }
                    }
                }
            },
        )` 
But db records were updated with a json in the unicity_id field :/

AndreaFrancis: > I think that we should also add one test for each migration script, using a dump of 100 random entries from prod. WDYT? It might be a new issue/PR

Yes for sure we need to test the migrations simulating existing old data

severo: cc @coyotte508 for https://github.com/huggingface/datasets-server/pull/837#discussion_r1116903403 maybe

AndreaFrancis: Thanks! @coyotte508 it worked

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216141000_queue_split_names_from_streaming.py
```diff
@@ -18,29 +18,55 @@
...
+                        "unicity_id": {
+                            "$replaceOne": {
+                                "input": "$unicity_id",
+                                "find": f"Job[{split_names}]",
+                                "replacement": f"Job[{split_names_from_streaming}]",
+                            }
+                        }
+                    }
+                },
+                {"$set": {"type": split_names_from_streaming}},
```
severo, line 41: maybe it can be inside the same "$set" object as above?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216141000_queue_split_names_from_streaming.py
```diff
@@ -18,29 +18,55 @@
...
+        db["jobsBlue"].update_many(
+            {"type": split_names},
+            [
+                {
+                    "$set": {
+                        "unicity_id": {
+                            "$replaceOne": {
+                                "input": "$unicity_id",
+                                "find": f"Job[{split_names}]",
+                                "replacement": f"Job[{split_names_from_streaming}]",
```
severo, line 36:
```suggestion
                                "replacement": f"Job[{split_names_from_streaming}",
```

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216141000_queue_split_names_from_streaming.py
```diff
@@ -18,29 +18,55 @@
...
-        db["jobsBlue"].update_many({"type": split_names}, {"$set": {"type": split_names_from_streaming}})
+        db["jobsBlue"].update_many(
+            {"type": split_names},
+            [
+                {
+                    "$set": {
+                        "unicity_id": {
+                            "$replaceOne": {
+                                "input": "$unicity_id",
+                                "find": f"Job[{split_names}]",
```
severo, line 35: I think we want to change the prefix, right?

```suggestion
                                "find": f"Job[{split_names}",
```

severo (approved):
Nice. A small change, I think, in the string replacement.

Also, in two other issues or PR:
-> add unit tests for the migration scripts (https://github.com/huggingface/datasets-server/issues/860)
-> fix the issue with config or split equal to `None` (unicity_id includes the string `None` in that case, which is a valid config name or split name). Thanks for finding the bug @AndreaFrancis. https://github.com/huggingface/datasets-server/issues/861

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216141000_queue_split_names_from_streaming.py
severo: > If it's not possible (I don't know much about MongoDB queries), we should delete part of the historical jobs (https://github.com/huggingface/datasets-server/issues/818) before running the migration.

Fixed by https://github.com/huggingface/datasets-server/pull/851. But as the index will be created after the migration job, it will not help here. Maybe we can add it manually to the index in production?

severo: I created the index manually in the prod database. Reduced from 10,000,000 entries to 200,000 (last seven days)

severo:
OK, I think it's ready to be merged

PR #839, opened 2023-02-23
Adding missing function on split-names-from-dataset-info worker

AndreaFrancis:
`get_new_splits` function is used when creating children jobs, this function was missing in the new worker `split-names-from-dataset-info`


severo (approved):
Thanks!

severo:
Do you think we should turn https://github.com/huggingface/datasets-server/blob/ce7ebdfcecbe2980c9d04d72847ae0a1735d71a4/services/worker/src/worker/job_runner.py#L385-L386 into an abstract method to force the job runners to implement it?

AndreaFrancis:
> Do you think we should turn  ...
>  into an abstract method to force the job runners to implement it?

Yes, I think we can add it in another PR

severo:
OK, I created https://github.com/huggingface/datasets-server/issues/854

PR #843, opened 2023-02-24
Add pdb to avoid disruption when we update kubernetes

XciD:


rtrompier (approved):

severo (approved):
Nice!

XciD:
> Should we add one for the reverse proxy too?

Totally, I forgot. Adding it

PR #851, opened 2023-02-24
POC: Adding mongo TTL index to Jobs

AndreaFrancis:
Will fix https://github.com/huggingface/datasets-server/issues/818
From mongo doc [TTL index](https://www.mongodb.com/docs/manual/core/index-ttl/) :

> TTL indexes are special single-field indexes that MongoDB can use to automatically remove documents from a collection after a certain amount of time or at a specific clock time. 

Adding index on Job collection to be removed every 7 days = 604800 seconds
Also mongo support adding filters for this index e.g If we would like to remove only those SUCCEEDED jobs will be able to do it according to https://www.mongodb.com/docs/manual/core/index-partial/
Consider that by default mongod TTL monitor service runs every 60 seconds but we can modify that configuration that is, all jobs which finished_at value are less that current_time - QUEUE_TTL_SECONDS will be removed every 60 secods by default.
Those documents without a value in the field "finished_at" are not removed.


severo:
Oh perfect! Support is here in mongoengine (for reference) http://docs.mongoengine.org/guide/defining-documents.html?highlight=string#indexes

libs/libcommon/src/libcommon/constants.py
```diff
@@ -4,3 +4,4 @@
 ASSETS_CACHE_APPNAME = "datasets_server_assets"
 CACHE_MONGOENGINE_ALIAS = "cache"
 QUEUE_MONGOENGINE_ALIAS = "queue"
+QUEUE_TTL_SECONDS = 604800
```
severo, line 7:
```suggestion
QUEUE_TTL_SECONDS = 604_800  # 7 days
```

severo (approved):
Nice!!!

PR #868, opened 2023-02-28
Add doc, simplify docker, update chart

severo:
- add doc on how to setup the TMPDIR environment variable
 - simplify docker compose
 - update chart to use the four new environment variables
 - upgrade some minor dependencies

(sorry, it's a mix of too many things, should have been various PRs)


tools/docker-compose-datasets-server.yml
```diff
@@ -240,6 +94,7 @@ services:
       PARQUET_AND_DATASET_INFO_SUPPORTED_DATASETS: ${PARQUET_AND_DATASET_INFO_SUPPORTED_DATASETS-}
       PARQUET_AND_DATASET_INFO_TARGET_REVISION: ${PARQUET_AND_DATASET_INFO_TARGET_REVISION-refs/convert/parquet}
       PARQUET_AND_DATASET_INFO_URL_TEMPLATE: ${PARQUET_AND_DATASET_INFO_URL_TEMPLATE-/datasets/%s/resolve/%s/%s}
+      WORKER_STATE_PATH: ${WORKER_STATE_PATH-/state/worker-state}
```
lhoestq, line 97: The WORKER_STATE_PATH needs to be different every time the worker restarts. You shouldn't set a default value for it.

It's important otherwise we'll end up with executors with an outdated state

severo: oh, you're right!

severo: (even if in that case there is only one worker)

severo: updated

chart/templates/_envWorker.tpl
```diff
@@ -4,14 +4,24 @@
...
 - name: WORKER_MAX_LOAD_PCT
   value: {{ .Values.worker.maxLoadPct | quote }}
 - name: WORKER_MAX_MEMORY_PCT
   value: {{ .Values.worker.maxMemoryPct | quote }}
+- name: WORKER_MAX_MISSING_HEARTBEATS
+  value: {{ .Values.worker.maxMissingHeartbeats | quote }}
 - name: WORKER_SLEEP_SECONDS
   value: {{ .Values.worker.sleepSeconds | quote }}
+- name: WORKER_STATE_PATH
+  value: "/tmp/worker-state"
```
lhoestq, line 22: same here

severo: the /tmp directory is only for one pod (container), it's not shared.

AndreaFrancis (approved):

severo:
OK, after talking with @lhoestq, it's probably better to just give a way to setup the tmp directory, but not specifically the state file path. I'm going to add this to the PR

severo:
OK, could you do a last review please?

lhoestq (approved):
LGTM ! Feel free to merge once it's green

chart/templates/_envWorker.tpl
severo: OK, I ensured this by setting `TMPDIR`

PR #869, opened 2023-02-28
ci: 🎡 add missing step to get the source code

severo:


severo:
Merging without review, as it's simple

PR #878, opened 2023-03-01
Renaming split-names response to splits

AndreaFrancis:
Fix content response name in split_names_from_streaming and split_names_from_dataset_info from "split_names" to "splits" so that it is consistent with the response of splits worker


severo (approved):
Thanks

PR #879, opened 2023-03-01
Increase version of workers /split-names from streaming and dataset info

AndreaFrancis:
As per PR https://github.com/huggingface/datasets-server/pull/878, worker versions should have been increased


severo (commented):
Thanks!

Issue #740, opened 2023-01-31
Improve the /backfill endpoint


severo:
See https://github.com/huggingface/datasets-server/pull/708#issuecomment-1406319901 and following comments.

Maybe we should not have this endpoint as such. See the related issue: https://github.com/huggingface/datasets-server/issues/736. Maybe #736 should be fixed first, then see how we implement a backfill trigger.

severo:
Changed the title: what we need is to convert this endpoint to a "cron" job

AndreaFrancis:
Tasks:

- [ ] Make `/backfill` a one time job (like `cache_refresh` and `mongo_migration` https://github.com/huggingface/datasets-server/tree/main/jobs) to run each deploy (keep `/backfill` logic as it is initially). 
- [ ] Make backfill  k8s job to only append Jobs in order to  "complete" missing cache entries (According to current processing graph). This could be done by iterating each supported dataset and verifying its missing cache by processing step (Could be heavy) or we could store somewhere the processing graph steps, and compare it with the current one, in case of new steps or newer job runner version, backfill by step (This could probably work as well to replace [cache-refresh](https://github.com/huggingface/datasets-server/tree/main/jobs/cache_refresh) job).
- [ ] Make backfill cron job that will run every x time

severo:
related issue: https://github.com/huggingface/datasets-server/issues/844
Ideally, the backfill should take care of removing the dangling assets/cache entries/jobs/parquet files

severo:
Done

PR #880, opened 2023-03-01
WIP - Adding new admin enpoint to refresh outdated cache 

AndreaFrancis:
Since #879 increased the workers version but we have existing records in dev/prod, we need a way to process the cache again given a specific processing step.
This solution will get the current version and will append a new Job for every cache entry with a `worker_version` minor than the current worker version.

Pending:
- [ ] Tests


libs/libcommon/src/libcommon/constants.py
```diff
@@ -5,3 +5,14 @@
 CACHE_MONGOENGINE_ALIAS = "cache"
 QUEUE_MONGOENGINE_ALIAS = "queue"
 QUEUE_TTL_SECONDS = 604_800  # 7 days
+
+
+PROCESSING_STEP_CONFIG_NAMES_VERSION = "1.0.0"
```
AndreaFrancis, line 10: Not sure if this is the best way to manage the versions but, I didnt find another way to have them in admin project without importing workers as dependency (I woulndt like to do it since will be having access to extra logic).

services/admin/src/admin/routes/refresh_outdated_cache.py
```diff
@@ -0,0 +1,61 @@
...
+            logging.info(f"/refresh-outdated-cache{processing_step.cache_kind}")
+
+            # if auth_check fails, it will raise an exception that will be caught below
+            auth_check(external_auth_url=external_auth_url, request=request, organization=organization)
+
+            current_version = processing_graph_config.specification[processing_step.cache_kind]["version"]
+
+            logging.info(f"refresh cache entries with worker_version<{current_version}")
+            for cache_info in get_cache_info_for_kind_minor_than_version(processing_step.cache_kind, current_version):
+                Queue().upsert_job(
```
AndreaFrancis, line 43: Not sure if it is enough to upsert a new Job. I saw that the previous outdated cache will remain in db, should we delete it?

libs/libcommon/src/libcommon/constants.py
severo: Good point. You're right, services/admin does not have access to the job runners' versions.

Instead of an admin endpoint, maybe we want an one-time job (like db migration -> cache_refresh?) run before deploying a new version. We would hardcode the current job runner versions there, and new jobs would be created to update the obsolete cache entries. wdyt?

services/admin/src/admin/routes/refresh_outdated_cache.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            logging.info(f"/refresh-outdated-cache{processing_step.cache_kind}")
+
+            # if auth_check fails, it will raise an exception that will be caught below
+            auth_check(external_auth_url=external_auth_url, request=request, organization=organization)
+
+            current_version = processing_graph_config.specification[processing_step.cache_kind]["version"]
+
+            logging.info(f"refresh cache entries with worker_version<{current_version}")
+            for cache_info in get_cache_info_for_kind_minor_than_version(processing_step.cache_kind, current_version):
```
severo, line 42: OK, but we will have the same issue as https://github.com/huggingface/datasets-server/issues/740: the request may last too long and timeout (nginx proxy will emit the timeout error).

We need a better mechanism to run long-lasting administration operations.

libs/libcommon/src/libcommon/constants.py
severo: Good idea to share the versions in libcommon. We will have to ensure updating it on every change in the job runners' code (as of now, where the version is inside the same file... not sure if it changes anything - CI might help here?).

services/admin/src/admin/routes/refresh_outdated_cache.py
severo: No, that's it. We want to serve the outdated cache entry until we have a better one. And in parallel, create jobs with priority=LOW to update them (also force=True? not sure, since they will be updated anyway due to the new major job runner's version)

AndreaFrancis: If we will keep the outdated versions, the validation should be: Look for cache entries with worker_version minor that the current one BUT without a newest cache entry. Otherwise, it will be creating Jobs with LOW priority in each deployment.

libs/libcommon/src/libcommon/constants.py
AndreaFrancis: > Instead of an admin endpoint, maybe we want a one-time job (as database migration -> cache_refresh?) run before (or in parallel to) deploying a new version. We would hardcode the current job runner versions in the script, and new jobs would be created to update the obsolete cache entries. wdyt?

Yeah, I also think that is a better Idea but was not sure if that will increase k8s/docker resources, if that is not a problem, lets do it.

services/admin/src/admin/routes/refresh_outdated_cache.py
severo: I'm not sure to understand. We don't keep the outdated versions, the cache entries will be updated, and the worker_version field with them, so next time, this cache will not be included in the refresh batch

libs/libcommon/src/libcommon/constants.py
severo: no, I don't see why the resources would be a problem.

PR #883, opened 2023-03-02
feat: 🎸 add an index

severo:
Recommended by the managed mongo database service

Capture d’écran 2023-03-02 à 10 39 45


severo:
Merging

severo:
Deployed. The index has been created automatically by mongoengine.

Capture d’écran 2023-03-03 à 10 08 22

Note that mongo atlas still recommends creating it. It could be an issue with their recommendation system since it seems like we made precisely the index he proposes. Let's wait a bit and see if the recommendation disappears.

Capture d’écran 2023-03-03 à 10 08 30

PR #884, opened 2023-03-02
chore: 🤖 use other nodes for the workers

severo:


chart/Chart.yaml
```diff
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.3.1
+version: 1.4.0
```
XciD, line 21: no need to change the chart version when you just update env values

XciD (approved):

chart/Chart.yaml
severo: reverted

PR #886, opened 2023-03-02
feat: 🎸 reduce the requests (and limits) in k8s resources

severo:
It's required for the api/admin services which were absolutely over-provisioned. And for the workers: we reduce a lot for now, since most of the datasets don't require that quantity of RAM, and we now manage the failed jobs due to OOm. We will work on a strategy to relaunch these jobs on "big" workers (to be done)

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -102,7 +102,7 @@ reverseProxy:
   replicas: 2
   resources:
     requests:
-      cpu: 200m
+      cpu: 1
       memory: "64Mi"
     limits:
       cpu: 1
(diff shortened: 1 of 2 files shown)


chart/env/prod.yaml
```diff
@@ -166,7 +166,7 @@ api:
     type: NodePort
   resources:
     requests:
-      cpu: 4
+      cpu: 200m
```
XciD, line 169: I would go on 1

XciD: 200m may stack a lot of pods on node. Same for above and bellow

severo: Ok, changed.

severo:
Fixed. Merging

PR #888, opened 2023-03-03
feat: 🎸 add an index to the jobs collection

severo:
proposed by mongo atlas

Capture d’écran 2023-03-03 à 10 08 30


severo:
Merging, as it's the same as #883

severo:
Deployed. The index has been created, and mongo atlas does not recommend any more index creation.

Capture d’écran 2023-03-03 à 10 51 41
Capture d’écran 2023-03-03 à 10 51 20

PR #890, opened 2023-03-03
Allow to use service nodeport

rtrompier:
More info here : https://github.com/huggingface/private-hub-package/issues/24


severo (approved):
OK. I don't have access to the referred issue, so I'm not sure to understand everything.
But adding the icon is a very nice improvement :P

rtrompier:
Icon is to removing the warning on `helm lint` ;)

Issue #877, opened 2023-03-01
Reduce the endpoints response time


severo:
Many issues (like https://github.com/huggingface/datasets-server/issues/876) are created with `Error code:   ClientConnectionError`. They are due to the Hub backend timing out when requesting a response from the datasets server.

We monitor the endpoints' response time at (internal) https://grafana.huggingface.tech/d/i7gwsO5Vz/global-view

severo:
Just reproduced on https://datasets-server.huggingface.co/first-rows?dataset=wikimedia/wit_base&config=wikimedia--wit_base&split=train. It took 10s to complete

<img width="109" alt="Capture d’écran 2023-03-01 à 17 21 42" src="https://user-images.githubusercontent.com/1676121/222199939-1210c344-727b-4b70-a249-d1a0896ed25f.png">
<img width="1907" alt="Capture d’écran 2023-03-01 à 17 21 47" src="https://user-images.githubusercontent.com/1676121/222199946-c51a1a2b-0471-4ecc-9df7-7fc687449a42.png">

severo:
We have a raise in the response time in the metrics:

<img width="385" alt="Capture d’écran 2023-03-01 à 17 24 19" src="https://user-images.githubusercontent.com/1676121/222200454-a22aefca-8091-46f0-8063-97b904c12e0e.png">

(since several days, so it's not linked to recent changes in the infra or in the code)

severo:
Looking at the metrics, we regularly have periods where the authentication to the Hub (`/api/datasets/%s/auth-check`) takes too long.

Two measures:
- fail early (timeout on the auth check after a short time: 250ms?)
- allow the client to pass a special token to bypass the auth. Indeed, when we show the dataset viewer on the Hub's dataset pages, the Hub's backend sends requests to the datasets-server to get the list of splits and the list of the first rows. But these two requests, in turn, call the Hub's API (powered by the same Hub's backend) for an authentication check. We can avoid the round-trip by allowing the Hub's backend to bypass the authentication step.

severo:
Note that the implementation of the authentication bypass can be more or less complicated:
  - a shared secret both backends, passed via a dedicated header (X-Api-Key), see https://github.com/huggingface/moon-landing/issues/5597#issuecomment-1456256768
[...]

PR #897, opened 2023-03-06
fix: 🐛 add a mechanism with shared secret to bypass auth check

severo:
Option 1 in
https://github.com/huggingface/datasets-server/issues/877#issuecomment-1456474625.

### DIFF
--- a/services/api/src/api/authentication.py
+++ b/services/api/src/api/authentication.py
@@ -32,7 +32,10 @@ class RequestAuth(AuthBase):
 
 
 def auth_check(
-    dataset: str, external_auth_url: Optional[str] = None, request: Optional[Request] = None
+    dataset: str,
+    external_auth_url: Optional[str] = None,
+    request: Optional[Request] = None,
+    external_auth_bypass_key: Optional[str] = None,
 ) -> Literal[True]:
     """check if the dataset is authorized for the request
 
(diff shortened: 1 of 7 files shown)


services/api/src/api/authentication.py
```diff
@@ -42,12 +45,20 @@ def auth_check(
...
+        request (Request | None): the request which optionally bears authentication headers: "cookie",
+          "authorization" or "X-Api-Key"
+        external_auth_bypass_key (str|None): if the request bears this secret key in the "X-Api-Key" header, the
+          external authentication service is bypassed.
 
     Returns:
         None: the dataset is authorized for the request
     """
+    if (
+        external_auth_bypass_key is not None
```
coyotte508, line 57: In the README it says this: 

> Defaults to `""`.

But here it checks against `None`

coyotte508 (approved):

services/api/src/api/authentication.py
coyotte508:
```suggestion
        external_auth_bypass_key
```
```suggestion
        not not external_auth_bypass_key
```
I feel like either would be slightly better, to guard against the env being set to an empty string as well

severo: You're right, thanks!

severo:
See https://github.com/huggingface/datasets-server/pull/898, as an alternative: instead of comparing the value to a shared secret, we check if we can decode the value as a JWT with the public key

severo:
Superseded by #898 (we decided to go with the JWT on the moonlanding side)

PR #898, opened 2023-03-06
Add jwt to bypass auth

severo:
Alternative to https://github.com/huggingface/datasets-server/pull/897, corresponding to option 2 in https://github.com/huggingface/datasets-server/issues/877#issuecomment-1456474625

### DIFF
--- a/.github/workflows/_e2e_tests.yml
+++ b/.github/workflows/_e2e_tests.yml
@@ -72,5 +72,7 @@ jobs:
           ADMIN_UVICORN_PORT: "8081"
           COMMON_LOG_LEVEL: "DEBUG"
           API_HF_TIMEOUT_SECONDS: "10"
+          API_HF_JWT_PUBLIC_KEY_URL: "https://huggingface.co/api/keys/jwt"
+          API_HF_JWT_ALGORITHM: "EdDSA"
         run: |
           poetry run python -m pytest -vv -x tests
--- a/services/api/README.md
+++ b/services/api/README.md
@@ -11,7 +11,7 @@ The worker can be configured using environment variables. They are grouped by sc
 Set environment variables to configure the application (`API_` prefix):
 
 - `API_HF_AUTH_PATH`: the path of the external authentication service, on the hub (see `HF_ENDPOINT`). The string must contain `%s` which will be replaced with the dataset name. The external authentication service must return 200, 401, 403 or 404. Defaults to "/api/datasets/%s/auth-check".
-- `API_HF_AUTH_BYPASS_KEY`: the key to use to bypass the external authentication service. If not set (empty string), the authentication check cannot be bypassed. Defaults to `""`.
+- `API_HF_AUTH_BYPASS_KEY`: the key to use to bypass the external authentication service. If not set, the authentication check cannot be bypassed. Defaults to not set.
 - `API_HF_TIMEOUT_SECONDS`: the timeout in seconds for the requests to the Hugging Face Hub. Defaults to `0.2` (200 ms).
 - `API_MAX_AGE_LONG`: number of seconds to set in the `max-age` header on data endpoints. Defaults to `120` (2 minutes).
 - `API_MAX_AGE_SHORT`: number of seconds to set in the `max-age` header on technical endpoints. Defaults to `10` (10 seconds).
--- a/services/api/README.md
+++ b/services/api/README.md
@@ -11,7 +11,8 @@ The worker can be configured using environment variables. They are grouped by sc
 Set environment variables to configure the application (`API_` prefix):
 
 - `API_HF_AUTH_PATH`: the path of the external authentication service, on the hub (see `HF_ENDPOINT`). The string must contain `%s` which will be replaced with the dataset name. The external authentication service must return 200, 401, 403 or 404. Defaults to "/api/datasets/%s/auth-check".
-- `API_HF_AUTH_BYPASS_PUBLIC_KEY`: the public key used to decode the JWT that can be passed in the 'X-Api-Key' header to bypass the external authentication check. If not set, we don't check if a JWT has been passed in the request. Defaults to empty.
+- `API_HF_JWT_PUBLIC_KEY_URL`: the URL where the "Hub JWT public key" is published. The "Hub JWT public key" must be in JWK format. It helps to decode a JWT sent by the Hugging Face Hub, for example, to bypass the external authentication check (JWT in the 'X-Api-Key' header). If not set, the JWT are ignored. Defaults to empty.
+- `API_HF_JWT_ALGORITHM`: the algorithm used to encode the JWT. Defaults to `"EdDSA"`.
 - `API_HF_TIMEOUT_SECONDS`: the timeout in seconds for the requests to the Hugging Face Hub. Defaults to `0.2` (200 ms).
 - `API_MAX_AGE_LONG`: number of seconds to set in the `max-age` header on data endpoints. Defaults to `120` (2 minutes).
 - `API_MAX_AGE_SHORT`: number of seconds to set in the `max-age` header on technical endpoints. Defaults to `10` (10 seconds).
--- a/services/api/README.md
+++ b/services/api/README.md
@@ -11,6 +11,7 @@ The worker can be configured using environment variables. They are grouped by sc
 Set environment variables to configure the application (`API_` prefix):
 
 - `API_HF_AUTH_PATH`: the path of the external authentication service, on the hub (see `HF_ENDPOINT`). The string must contain `%s` which will be replaced with the dataset name. The external authentication service must return 200, 401, 403 or 404. Defaults to "/api/datasets/%s/auth-check".
+- `API_HF_AUTH_BYPASS_KEY`: the key to use to bypass the external authentication service. If not set (empty string), the authentication check cannot be bypassed. Defaults to `""`.
 - `API_HF_TIMEOUT_SECONDS`: the timeout in seconds for the requests to the Hugging Face Hub. Defaults to `0.2` (200 ms).
 - `API_MAX_AGE_LONG`: number of seconds to set in the `max-age` header on data endpoints. Defaults to `120` (2 minutes).
 - `API_MAX_AGE_SHORT`: number of seconds to set in the `max-age` header on technical endpoints. Defaults to `10` (10 seconds).
--- a/services/api/README.md
+++ b/services/api/README.md
@@ -11,7 +11,7 @@ The worker can be configured using environment variables. They are grouped by sc
 Set environment variables to configure the application (`API_` prefix):
 
 - `API_HF_AUTH_PATH`: the path of the external authentication service, on the hub (see `HF_ENDPOINT`). The string must contain `%s` which will be replaced with the dataset name. The external authentication service must return 200, 401, 403 or 404. Defaults to "/api/datasets/%s/auth-check".
-- `API_HF_AUTH_BYPASS_KEY`: the key to use to bypass the external authentication service. If not set, the authentication check cannot be bypassed. Defaults to not set.
+- `API_HF_AUTH_BYPASS_PUBLIC_KEY`: the public key used to decode the JWT that can be passed in the 'X-Api-Key' header to bypass the external authentication check. If not set, we don't check if a JWT has been passed in the request. Defaults to empty.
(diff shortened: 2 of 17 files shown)


XciD:
You don't even need the pubkey, the hub exposes it

AndreaFrancis (approved):
Great! I like more this approach instead of the shared secret

services/api/README.md
```diff
@@ -11,6 +11,7 @@ The worker can be configured using environment variables. They are grouped by sc
 Set environment variables to configure the application (`API_` prefix):
 
 - `API_HF_AUTH_PATH`: the path of the external authentication service, on the hub (see `HF_ENDPOINT`). The string must contain `%s` which will be replaced with the dataset name. The external authentication service must return 200, 401, 403 or 404. Defaults to "/api/datasets/%s/auth-check".
+- `API_HF_AUTH_BYPASS_PUBLIC_KEY`: the public key used to decode the JWT that can be passed in the 'X-Api-Key' header to bypass the external authentication check. If not set, we don't check if a JWT has been passed in the request. Defaults to empty.
```
coyotte508, line 14: As @XciD pointed out, it should get the JWT public key directly from asking /api/keys/jwt

severo:
> You don't even need the pubkey, the hub exposes it /api/keys/jwt

Well... https://huggingface.co/api/keys/jwt returns:

```
[{"crv":"Ed25519","x":"-RBhgyNluwaIL5KFJb6ZOL2H1nmyI8mW4Z2EHGDGCXM","kty":"OKP"}]
```

while pyjwt [expects a string](https://github.com/Privex/pyjwt/blob/e8e242c593990aa276b32597348427ab3e8d50e7/jwt/contrib/algorithms/py_ed25519.py#L41):

```
"Expecting a PEM-formatted or OpenSSH key."
```

I'm not sure how I would translate one to the other.

I think that `"crv":"Ed25519"` is the algorithm that will be used to generate the JWT (that we can provide in [jwt.decode](https://pyjwt.readthedocs.io/en/stable/api.html#jwt.decode) - translating it to `EdDSA`... maybe it's better to hardcode this part)

I'm not sure if `kty` is important to decode, but I don't find a reference for `OKP` (https://www.rfc-editor.org/rfc/rfc7517#page-6 gives "RSA" and "EC" as examples).

Finally, I understand that `x` is the parameter that defines the key (a coordinate in an elliptic curve)... but I'm not sure how I compute the public key string from this...

If somebody has insights to help me save some investigation time...

coyotte508:
It's in JWK format: https://jwcrypto.readthedocs.io/en/latest/jwk.html

XciD:
I guess this can help you: https://github.com/huggingface/spaces-app-manager/commit/ff4a3c16c57a08ffb42f99b41e3051b6963585f1

services/api/README.md
SBrandeis: You can look here: https://github.com/huggingface/spaces-app-manager/blob/main/sse-api/sse_api/config.py#L24-L51

For an example of how to parse the key from moon

services/api/src/api/authentication.py
```diff
@@ -35,10 +36,25 @@ def __call__(self, r: PreparedRequest) -> PreparedRequest:
...
 
+def is_jwt_valid(request: Optional[Request], public_key: Optional[str]) -> bool:
+    if not request or not public_key:
+        return False
+    encoded = request.headers.get("x-api-key")
+    if not encoded:
+        return False
+    try:
+        jwt.decode(jwt=encoded, key=public_key, algorithms=["RS256"])
+        # ^ we're not interested in the contents of the JWT token, we just want to check that it's valid
```
SBrandeis, line 47: I strongly suggest you also check the JWT content, otherwise any JWT generated by moon-landing will be invalid

SBrandeis: eg, the JWT passed to Spaces' iframes

severo: OK, you mean that somebody could reuse a JWT they got in their space?

severo:
Missing: use type `str` on the key (model: https://github.com/huggingface/spaces-app-manager/commit/7a5746ada2606372acb19ffdae2c8eb8a71f0158) instead of `Any`

services/api/src/api/jwt_token.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    if not public_key or not algorithm:
+        return False
+    try:
+        decoded = jwt.decode(jwt=token, key=public_key, algorithms=[algorithm], options={"require": ["exp", "sub"]})
+    except Exception as err:
+        print(err)
+        return False
+    sub = decoded.get("sub")
+    return isinstance(sub, str) and sub == dataset
```
coyotte508, line 91: Honestly I think we should add more info in the JWT cc @SBrandeis 

For example if a space and dataset have the same name would the same JWT be valid for both?

We should use something like `{repoName: <...>, repoType: "dataset", read: true}` or `{repoId: <content of _id>, read: true}`

severo: Oh, sure, I didn't overthink it, but we have to avoid reusable JWT. `{repoName: <...>, repoType: "dataset", read: true}` seems ok to me.

severo: Done in [3c9da5c](https://github.com/huggingface/datasets-server/pull/898/commits/3c9da5cfe5cf5fd3d2412bc4a40ffffa0f0178a7)

coyotte508 (approved):
LGTM!

services/api/src/api/authentication.py
severo: Done

services/api/README.md
severo: Done

services/api/src/api/jwt_token.py
```diff
@@ -0,0 +1,161 @@
...
+        EllipticCurvePublicKey,
+        EllipticCurvePrivateKey,
+        Ed25519PublicKey,
+        Ed448PublicKey,
+        Ed25519PrivateKey,
+        Ed448PrivateKey,
+        RSAPrivateKey,
+        RSAPublicKey,
+    ]
+) -> Union[EllipticCurvePublicKey, Ed25519PublicKey, Ed448PublicKey, RSAPublicKey]:
```
SBrandeis, line 48: Nitpick: This function does not return a `Union[...]` but a `bool`

services/api/src/api/jwt_token.py
```diff
@@ -0,0 +1,161 @@
...
+    return hasattr(key, "public_bytes")  # type: ignore
+    # ^ type: ignore could be removed by using TypeGuard. But it requires Python 3.10
+
+
+def parse_jwt_public_key(keys: Any, hf_jwt_algorithm: str) -> Any:
```
SBrandeis, line 53:
```suggestion
def parse_jwt_public_key(keys: Any, hf_jwt_algorithm: str) -> str:
```

services/api/src/api/jwt_token.py
```diff
@@ -0,0 +1,161 @@
...
+        ).decode("utf-8")
+    except (jwt.InvalidKeyError, KeyError) as err:
+        raise RuntimeError(f"Failed to parse JWT key: {err.args[0]}") from err
+
+
+def fetch_jwt_public_key(
+    url: str,
+    hf_jwt_algorithm: str,
+    hf_timeout_seconds: Optional[float] = None,
+) -> Any:
```
SBrandeis, line 94:
```suggestion
) -> str:
```

services/api/src/api/jwt_token.py
```diff
@@ -0,0 +1,161 @@
...
+    The JWT is decoded with the public key, and the "sub" claim must be:
+      {"repoName": <...>, "repoType": "dataset", "read": true}
+    where <...> is the dataset identifier.
+
+    Returns True only if all the conditions are met. Else, it returns False.
+
+    Args:
+        dataset (str): the dataset identifier
+        token (Any): the JWT token to decode
+        public_key (Any|None): the public key to use to decode the JWT token
```
SBrandeis, line 141: I think you can constrain the type of `public_key` to `str` here

services/api/src/api/jwt_token.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Returns:
+        bool: True if the JWT is valid for the input dataset, else False
+    """
+    if not public_key or not algorithm:
+        return False
+    try:
+        decoded = jwt.decode(jwt=token, key=public_key, algorithms=[algorithm], options={"require": ["exp", "sub"]})
+    except Exception as err:
+        print(err)
```
SBrandeis, line 152: Is that `print` intended?

services/api/src/api/jwt_token.py
severo: 🤦

services/api/src/api/jwt_token.py
```diff
@@ -0,0 +1,161 @@
...
+        return False
+    sub = decoded.get("sub")
```
SBrandeis, line 153: If I understand correctly, `sub` is expected to be a JSON encoded object, correct?
If true, note that `sub` claims are strings as per the JWT standard.
WDYT about passing this info in the JWT payload directly?

Ref: https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.2

coyotte508: oh yes I agree

coyotte508: Maybe make `sub` `/dataset/user/repo` :thinking: (and extra info in the payload)

coyotte508: > No need for `"repoType": "dataset"`, right?

yes

severo: Done

.github/workflows/_e2e_tests.yml
```diff
@@ -72,5 +72,7 @@ jobs:
           ADMIN_UVICORN_PORT: "8081"
           COMMON_LOG_LEVEL: "DEBUG"
           API_HF_TIMEOUT_SECONDS: "10"
+          API_HF_JWT_PUBLIC_KEY_URL: "https://huggingface.co/api/keys/jwt"
+          API_HF_JWT_ALGORITHM: "EdDSA"
```
coyotte508, line 76: By the way you shouldn't need this, you should be able to read this from the JWK no?

coyotte508 (approved):
A lot of work compared to the API key just for the benefit of asymmetric authentication ^^

.github/workflows/_e2e_tests.yml
severo: I followed the recommendation of https://pyjwt.readthedocs.io/en/stable/algorithms.html#specifying-an-algorithm:

> Do not compute the algorithms parameter based on the alg from the token itself, or on any other data that an attacker may be able to influence, as that might expose you to various vulnerabilities (see [RFC 8725 §2.1](https://www.rfc-editor.org/rfc/rfc8725.html#section-2.1)). Instead, either hard-code a fixed value for algorithms, or configure it in the same place you configure the key. Make sure not to mix symmetric and asymmetric algorithms that interpret the key in different ways (e.g. HS* and RS*).

coyotte508: ok, if not from the JWT you should be able to read it from the JWK no?

severo: If I understand well the code of https://github.com/jpadilla/pyjwt/blob/7665aa625506a11bae50b56d3e04413a3dc6fdf8/jwt/api_jwk.py and https://github.com/jpadilla/pyjwt/blob/7665aa625506a11bae50b56d3e04413a3dc6fdf8/jwt/algorithms.py:
- yes: the algorithm is obtained from `alg` or `crv`
- but: the key (`EdDSA` in our case) is not exposed in [PyJWK](https://github.com/jpadilla/pyjwt/blob/7665aa625506a11bae50b56d3e04413a3dc6fdf8/jwt/api_jwk.py#L10) or it's field [Algorithm](https://github.com/jpadilla/pyjwt/blob/7665aa625506a11bae50b56d3e04413a3dc6fdf8/jwt/api_jwk.py#L50)
- and: that key is required when calling [jwt.decode](https://github.com/jpadilla/pyjwt/blob/7665aa625506a11bae50b56d3e04413a3dc6fdf8/jwt/api_jwt.py#L147)

coyotte508: alright, it's not a blocker

severo:
Thanks to all for the reviews. Merging. Next step: modify the Hub to send a JWT.

PR #904, opened 2023-03-07
Adding deploy for missing JobRunners in chart

AndreaFrancis:
We have two new JobRunners name:
- /split-names-from-streaming -> It was previously /split-names, but it was renamed 
- /split-names-from-dataset-info -> It is a new JobRunner which is supposed to have a similar workload as /dataset-info or /parquet

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -213,9 +213,9 @@ workers:
         memory: "8Gi"
     tolerations: []
   -
-    deployName: "split-names"
+    deployName: "split-names-from-streaming"
     maxJobsPerNamespace: 20
-    workerOnlyJobTypes: "/split-names"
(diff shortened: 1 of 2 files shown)


rtrompier (approved):
LGTM

lhoestq (approved):

chart/env/prod.yaml
```diff
@@ -213,9 +213,9 @@ workers:
         memory: "8Gi"
     tolerations: []
   -
-    deployName: "split-names"
+    deployName: "split-names-from-streaming"
     maxJobsPerNamespace: 20
-    workerOnlyJobTypes: "/split-names"
+    workerOnlyJobTypes: "/split-names-from-streaming"
```
severo, line 218: wow, we missed this. The worker was not doing anything since the job did not exist

PR #905, opened 2023-03-07
fix: chart dependencies

rtrompier:


AndreaFrancis (approved):
Thanks!

severo:
Why are we versioning the dependencies? Shouldn't they be obtained from their repositories instead?

rtrompier:
We choose to push charts into the repository, because Bitnami regularly remove old versions. If we don't want be blocked by a technical migration, we have to commit these files.

severo:
OK, thanks

PR #911, opened 2023-03-08
Instrument api service

severo:


AndreaFrancis (approved):

e2e/tests/test_21_api_metrics.py
```diff
@@ -20,7 +20,14 @@ def test_metrics() -> None:
...
     content = response.text
     lines = content.split("\n")
-    metrics = {line.split(" ")[0]: float(line.split(" ")[1]) for line in lines if line and line[0] != "#"}
+    # examples:
+    # starlette_requests_total{method="GET",path_template="/metrics"} 1.0
+    # method_steps_processing_time_seconds_sum{method="healthcheck_endpoint",step="all"} 1.6772013623267412e-05
+    metrics = {
+        parts[0]: float(parts[1])
+        for line in lines
+        if line and line[0] != "#" and len(parts := line.rsplit(" ", 1)) == 2
```
albertvillanova, line 29: Nice refactoring, using the assignment expression. This was introduced in Python 3.8, but as we work with Python 3.9.15, it is safe to use it.

One question, why do you replace `split` with `rsplit`? Do you expect that `parts[0]` might contain any blanks?

Maybe we can make the refactoring simpler with something like this?
```suggestion
        if line and line[0] != "#" and (parts := line.rsplit(" ", 1))
```

services/api/tests/test_prometheus.py
```diff
@@ -13,7 +13,15 @@ def test_prometheus() -> None:
...
     print("content:", content)
     lines = content.split("\n")
-    metrics = {line.split(" ")[0]: float(line.split(" ")[1]) for line in lines if line and line[0] != "#"}
+    # examples:
+    # starlette_requests_total{method="GET",path_template="/metrics"} 1.0
+    # method_steps_processing_time_seconds_sum{method="healthcheck_endpoint",step="all"} 1.6772013623267412e-05
+    metrics = {
+        parts[0]: float(parts[1])
+        for line in lines
+        if line and line[0] != "#" and len(parts := line.rsplit(" ", 1)) == 2
```
albertvillanova, line 22: Same as above.

albertvillanova (approved):
Thanks!

I guess you are implementing an API service to measure the time it takes the different steps when creating an endpoint. This measure is always active, even in production (no need to deactivate it).

e2e/tests/test_21_api_metrics.py
severo: > One question, why do you replace split with rsplit? Do you expect that parts[0] might contain any blanks?

Yes, exactly, we have spaces in the keys

severo:
> I guess you are implementing an API service to measure the time it takes the different steps when creating an endpoint. This measure is always active, even in production (no need to deactivate it).

I'm sending metrics to the prometheus server, so that we can follow the duration of specific parts of the code in Grafana. Note that we are not restricted to the endpoint functions, we can instrument the rest of the code too. But for now, I was interested in understanding better which parts of the code are lagging and making the requests processing time increase.

services/api/tests/test_prometheus.py
severo: Done, thanks

PR #912, opened 2023-03-08
Refresh when cache is outdated because of job_runner_version

AndreaFrancis:
Given that we had a new version for `/split-names-from-streaming,` we need to refresh the outdated cache records because some records have a specific format and other with higher version have another one.

Previous context was: https://github.com/huggingface/datasets-server/pull/880
Based on comments, there should be a `K8s Job` that appends a new `dataset server Job` by each cache outdated based on job_runner_version field.


libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -356,6 +356,26 @@ def get_cache_reports(kind: str, cursor: Optional[str], limit: int) -> CacheRepo
     }
 
 
+class CacheInfo(TypedDict):
```
severo, line 359: It already exists as SplitFullName: https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/simple_cache.py#L58

Let's reuse it

libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -17,6 +17,7 @@ class _ProcessingStepSpecification(TypedDict):
 class ProcessingStepSpecification(_ProcessingStepSpecification, total=False):
     requires: Optional[str]
     required_by_dataset_viewer: Literal[True]
+    version: str
```
severo, line 20: should we move to int? see below

jobs/cache_refresh/pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[tool.poetry]
+authors = ["Andrea Soria <andrea@huggingface.co>"]
+description = "Refresh outdated cache job based on worker version"
```
severo, line 3:
```suggestion
description = "Refresh outdated cache entries based on worker version"
```

severo (changes requested):
Thanks, this is much needed, and it will serve as an example for other admin jobs.

I put some comments, I let you look at them.

Also, the CI in .github is missing (code quality, tests)

libs/libcommon/tests/test_simple_cache.py
```diff
@@ -613,3 +614,30 @@ def test_stress_get_cache_reports(num_entries: int) -> None:
         response = get_cache_reports(kind=kind, cursor=next_cursor, limit=100)
         next_cursor = response["next_cursor"]
         assert process_time() - start < MAX_SECONDS
+
+
+def test_get_cache_with_minor_version_for_kind() -> None:
+    kind = "kind"
+    current_version = "2.0.0"
+    minor_version = "1.0.0"
```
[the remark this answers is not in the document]
AndreaFrancis: Yes, it worked with "10.0.0" but it doesn't with "1.0", maybe it is doing some kind of int casting on mongo side, anyway I think we will have to change to a int (Will probably need a migration) or otherwise look for all the cache for a certain `kind` and then, validate the version as in Job_Runner.py but that will imply reading almost all the cache db. What do you think?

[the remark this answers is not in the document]
severo: I think we should move to an integer, because it's useful to be able to use it in mongo queries. We don't really need the details of semver, as we currently just use the major version (x in x.y.z)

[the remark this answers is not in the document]
AndreaFrancis: job_runner_version has been created in https://github.com/huggingface/datasets-server/pull/916 as int, this will fix the str version  issue

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: Thanks!

jobs/cache_refresh/src/cache_refresh/main.py
```diff
@@ -0,0 +1,52 @@
...
+        if not queue_resource.is_available():
+            logging.warning(
+                "The connection to the queue database could not be established. The cache refresh job is skipped."
+            )
+            return
+
+        processing_graph_config = ProcessingGraphConfig.from_env()
+        processing_graph = ProcessingGraph(processing_graph_config.specification)
+        processing_steps = list(processing_graph.steps.values())
+        refresh_cache(processing_steps, processing_graph_config)
```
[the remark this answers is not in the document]
AndreaFrancis: Thanks! good idea, I applied it

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -356,6 +356,26 @@ def get_cache_reports(kind: str, cursor: Optional[str], limit: int) -> CacheRepo
...
+    dataset: str
+    config: Optional[str]
+    split: Optional[str]
+
+
+def get_cache_with_minor_version_for_kind(kind: str, current_version: str) -> List[CacheInfo]:
+    responses = CachedResponse.objects(kind=kind, worker_version__lt=current_version).only(
```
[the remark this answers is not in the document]
AndreaFrancis: job_runner_version has been created in https://github.com/huggingface/datasets-server/pull/916 as int, this will fix the str version  issue

libs/libcommon/src/libcommon/processing_graph.py
AndreaFrancis: I renamed it to job_runner_version to be consistent with the new field

jobs/cache_refresh/src/cache_refresh/outdated_cache.py
```diff
@@ -0,0 +1,39 @@
...
+from libcommon.processing_graph import ProcessingStep
+from libcommon.queue import Priority, Queue
+from libcommon.simple_cache import get_cache_with_minor_version_for_kind
+
+
+def refresh_cache(processing_steps: List[ProcessingStep], processing_graph_config: ProcessingGraphConfig) -> None:
+    queue = Queue()
+    for processing_step in processing_steps:
+        current_version = processing_graph_config.specification[processing_step.cache_kind]["version"]
+        cache_records = get_cache_with_minor_version_for_kind(processing_step.cache_kind, current_version)
```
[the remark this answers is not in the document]
AndreaFrancis: Thanks!, I added version to ProcessingStep and change the function name

lhoestq:
Sure

> on the same model as this PR

what does that mean in practice ? just increase the version number ?

jobs/cache_refresh/tests/test_outdated_cache.py
```diff
@@ -0,0 +1,36 @@
...
+
+@pytest.mark.parametrize(
+    "worker_version, upserts_job",
+    [
+        ("0.0.0", True),
+        ("1.0.0", False),
+        ("2.0.0", False),
+    ],
+)
+def test_refresh_cache(
```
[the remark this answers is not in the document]
AndreaFrancis: Thanks, I added a dummy processing step instead of getting it from the config.

jobs/cache_refresh/pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[tool.poetry]
+authors = ["Andrea Soria <andrea@huggingface.co>"]
+description = "Refresh outdated cache entries based on worker version"
```
severo, line 3:
```suggestion
description = "Refresh outdated cache entries based on job runner version"
```

severo (approved):
cool!

severo:
> what does that mean in practice ? just increase the version number ?

Oh yes, you're right, it will be that simple!

AndreaFrancis:
> No... In that case (new endpoint) there are no existing cache entries to update. I'm not sure how we should proceed. @AndreaFrancis what do you think?

Maybe we could create a new Job that will "complete" the missing steps for datasets. Maybe the opposite as "valid" but only for the ones who as missing (not failed)

albertvillanova (approved):
Thanks.

severo:
We can ignore the issue with [chart PR](https://github.com/huggingface/datasets-server/actions/runs/4439030674/jobs/7805759136?pr=912) in the CI. It seems only to occur here; see #947, for example.

jobs/cache_refresh/pyproject.toml
severo: I think you didn't apply this suggestion. "job runner" would be more accurate than "worker"

AndreaFrancis: Yeah sorry I forgot to apply it

chart/templates/jobs/cache-refresh/job.yaml
```diff
@@ -0,0 +1,28 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels: {{ include "labels.cacheRefresh" . | nindent 4 }}
+  name: "{{ include "name" . }}-job-cache-refresh"
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-1"
+    "helm.sh/hook-delete-policy": before-hook-creation
```
rtrompier, line 14: I didn’t check the code, but it’s very important for this job to not be blocking if the DB is not yet created or ready, to avoid blocking all the deployment process during the first install 😉

rtrompier (approved):
LGTM for the chart part.

chart/templates/jobs/cache-refresh/job.yaml
severo: Ah, good point. The job works as the migration one: in case the connection to the database cannot be established, the script exits with code 1. See:
- the new job: https://github.com/huggingface/datasets-server/pull/912/files#diff-b1c58a99fef3df220ff883909890e39deb67ad586c6e0dfb833290e4a6d21d40
- the other job (migration) that we fixed to avoid blocking: https://github.com/huggingface/datasets-server/blob/main/jobs/mongodb_migration/src/mongodb_migration/main.py

AndreaFrancis:
I will merge this PR. Task `chart PR` does not work with forks so, will skip it this last time.
I will start working on branches on datasets-server repository from now on to avoid those problems.

PR #914, opened 2023-03-09
Add "context" label to filter the metrics on the endpoint

severo:


services/api/src/api/prometheus.py
```diff
@@ -24,7 +24,7 @@
 METHOD_STEPS_PROCESSING_TIME = Histogram(
     "method_steps_processing_time_seconds",
     "Histogram of the processing time of specific steps in methods (in seconds)",
```
AndreaFrancis, line 26:
```suggestion
    "Histogram of the processing time of specific steps in methods for a given context (in seconds)",
```

AndreaFrancis (approved):

PR #915, opened 2023-03-09
Add some profiling

severo:


severo:
The bad point with adding profiling is it breaks the indentation... But I didn't find a solution to this. If we just add steps between the code lines, we will not be able to measure the time of code that raises.

services/api/src/api/authentication.py
```diff
@@ -61,34 +63,47 @@ def auth_check(
...
-        )
-    elif response.status_code in [403, 404]:
-        raise ExternalAuthenticatedError(
-            "The dataset does not exist, or is not accessible with the current credentials (private or gated). Please"
-            " check the spelling of the dataset name or retry with other authentication credentials."
-        )
-    else:
-        raise ValueError(f"Unexpected status code {response.status_code}")
+    with StepProfiler(method="auth_check", step="all"):
+        with StepProfiler(method="auth_check", step="prepare parameters"):
```
AndreaFrancis, line 67: Not sure if we should profile each section, maybe only the ones that are critical or we know can take some time. I wonder if this could increase the whole method response time.

AndreaFrancis (approved):

services/api/src/api/authentication.py
severo: Yes, I was wondering the same. It seems evident that this part will not take time, but I decided to be exhaustive and always cover the whole method if I start, just in case I missed something so that the `all` step should be the sum of the other steps.
I'll merge it as is, and we can still change it afterward if needed.

PR #916, opened 2023-03-09
Change worker_version to int job_runner_version

AndreaFrancis:
According to conversation on https://github.com/huggingface/datasets-server/pull/912, worker_version needs to be an int value.
I renamed worker_version to job_runner_version as suggested in a comment in https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runner.py#L315 to keep consistency between class name and field.

### DIFF
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -26,6 +26,9 @@ from mongodb_migration.migrations._20230309123100_cache_add_progress import (
 from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
     MigrationAddJobRunnerVersionToCacheResponse,
 )
+from mongodb_migration.migrations._20230309123100_cache_add_progress import (
+    MigrationAddProgressToCacheResponse,
+)
 
 
 # TODO: add a way to automatically collect migrations from the migrations/ folder
@@ -58,7 +61,7 @@ class MigrationsCollector:
                 version="20230309123100",
                 description="add the 'progress' field with the default value (1.0) to the cached results",
             ),
-            MigrationAddJobRunnerVerionToCacheResponse(
+            MigrationAddJobRunnerVersionToCacheResponse(
                 version="20230309141600", description="add 'job_runner_version' field based on 'worker_version' value"
-            ),
+            )
         ]
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -23,6 +23,9 @@ from mongodb_migration.migrations._20230216141000_queue_split_names_from_streami
 from mongodb_migration.migrations._20230309123100_cache_add_progress import (
     MigrationAddProgressToCacheResponse,
 )
+from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
+    MigrationAddJobRunnerVerionToCacheResponse,
+)
 
 
 # TODO: add a way to automatically collect migrations from the migrations/ folder
@@ -54,5 +57,8 @@ class MigrationsCollector:
             MigrationAddProgressToCacheResponse(
                 version="20230309123100",
                 description="add the 'progress' field with the default value (1.0) to the cached results",
+            ),
+            MigrationAddJobRunnerVerionToCacheResponse(
+                version="20230309141600", description="add 'job_runner_version' field based on 'worker_version' value"
             ),
         ]
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -29,6 +29,9 @@ from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version i
 from mongodb_migration.migrations._20230309123100_cache_add_progress import (
     MigrationAddProgressToCacheResponse,
 )
+from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
+    MigrationAddJobRunnerVerionToCacheResponse,
+)
 
 
 # TODO: add a way to automatically collect migrations from the migrations/ folder
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -30,7 +30,7 @@ from mongodb_migration.migrations._20230309123100_cache_add_progress import (
     MigrationAddProgressToCacheResponse,
 )
 from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
-    MigrationAddJobRunnerVerionToCacheResponse,
+    MigrationAddJobRunnerVersionToCacheResponse,
 )
 
 
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -24,7 +24,7 @@ from mongodb_migration.migrations._20230309123100_cache_add_progress import (
     MigrationAddProgressToCacheResponse,
 )
 from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
-    MigrationAddJobRunnerVerionToCacheResponse,
+    MigrationAddJobRunnerVersionToCacheResponse,
 )
 
 
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -26,12 +26,6 @@ from mongodb_migration.migrations._20230309123100_cache_add_progress import (
 from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
     MigrationAddJobRunnerVersionToCacheResponse,
 )
-from mongodb_migration.migrations._20230309123100_cache_add_progress import (
-    MigrationAddProgressToCacheResponse,
-)
-from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
-    MigrationAddJobRunnerVersionToCacheResponse,
-)
 
 
 # TODO: add a way to automatically collect migrations from the migrations/ folder
@@ -66,5 +60,5 @@ class MigrationsCollector:
             ),
             MigrationAddJobRunnerVersionToCacheResponse(
                 version="20230309141600", description="add 'job_runner_version' field based on 'worker_version' value"
-            )
+            ),
         ]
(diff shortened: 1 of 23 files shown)


jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
```diff
@@ -0,0 +1,35 @@
...
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationAddJobRunnerVerionToCacheResponse(Migration):
+    def up(self) -> None:
+        # See https://docs.mongoengine.org/guide/migration.html#example-1-addition-of-a-field
+        logging.info("If missing, add 'job_runner_version' field based on 'worker_version' value")
+        db = get_db("cache")
+        db["cachedResponsesBlue"].update_many(
+            {"job_runner_version": {"$exists": False}},
+            [
+                {"$set": {"job_runner_version": {"$toInt": {"$first": {"$split": ["$worker_version", "."]}}}}},
```
severo, line 22: super clever!

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -48,4 +51,7 @@ def get_migrations(self) -> List[Migration]:
                     "update 'type' and 'unicity_id' fields in job from /split-names to /split-names-from-streaming"
                 ),
             ),
+            MigrationAddJobRunnerVerionToCacheResponse(
+                version="20230309141600", description="add 'job_runner_version' field based on 'worker_version' value"
```
severo, line 55: and delete 'worker_version'

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
```diff
@@ -0,0 +1,35 @@
...
+
+from libcommon.simple_cache import CachedResponse
+from mongoengine.connection import get_db
+
+from mongodb_migration.check import check_documents
+from mongodb_migration.migration import Migration
+
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationAddJobRunnerVerionToCacheResponse(Migration):
```
severo, line 14: It could be the opportunity to start using tests. This migration changes all the cache entries. It has the potential to break things!

WDYT? The idea would be to extract a small dump (ideally shuffled) of the database, version it in the code, and then, in the test, import it to the test database, launch the test, and check if the migration worked as expected.

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
```diff
@@ -0,0 +1,35 @@
...
+    def up(self) -> None:
+        # See https://docs.mongoengine.org/guide/migration.html#example-1-addition-of-a-field
+        logging.info("If missing, add 'job_runner_version' field based on 'worker_version' value")
+        db = get_db("cache")
+        db["cachedResponsesBlue"].update_many(
+            {"job_runner_version": {"$exists": False}},
```
severo, line 20: Implicitly we expect that $worker_version exists. It should, but what if it does not exist? Maybe it will break.

I worker_version does not exist, we should set a default value instead.

Also: we expect a specific format for worker_version (it should), but what if we have weird values there?

To be extra safe, maybe we could do two migrations here:
- first create a job_runner everywhere, using worker_version when possible, but also handling the case where worker_version does not exist or has not the correct format (not sure how... we could just check what is the current state of the database -> if there are no such cases, maybe just skip this part)
- in the same PR: only use job_runner_version in the code of datasets-server
- then, in a following PR, delete worker_version.

WDYT?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
```diff
@@ -0,0 +1,35 @@
...
+            [
+                {"$set": {"job_runner_version": {"$toInt": {"$first": {"$split": ["$worker_version", "."]}}}}},
+                {"$unset": "worker_version"},
+            ],  # type: ignore
+        )
+
+    def down(self) -> None:
+        logging.info("Remove 'job_runner_version' field from all the cached results")
+        db = get_db("cache")
+        db["cachedResponsesBlue"].update_many({}, {"$unset": {"job_runner_version": ""}})
```
severo, line 30: Here, worker_version is lost. I think we should really split this migration in two:
- a Migration that creates and populates the job_runner_version
- the, an IrreversibleMigration that deletes worker_version (see https://github.com/huggingface/datasets-server/blob/1e5064e2ef711741bdf7ec579898ce63495a5c43/jobs/mongodb_migration/tests/test_plan.py#L71)

services/worker/src/worker/job_runner.py
```diff
@@ -252,24 +251,8 @@ def run(self) -> Literal[Status.SUCCESS, Status.ERROR, Status.SKIPPED]:
...
-            0 if they are equal. Negative if job runner's major version is lower than other_version, positive
-              otherwise.
-        Raises:
-            :obj:`ValueError`: if job runner's version or other_version is not a valid semantic version.
-        """
-        try:
-            return version.parse(self.get_version()).major - version.parse(other_version).major
-        except Exception as err:
-            raise RuntimeError(f"Could not get major versions: {err}") from err
+    def is_major_version_than(self, other_version: Optional[int]) -> bool:
```
severo, line 254: I think we don't need a dedicated method anymore

services/worker/tests/test_job_runner.py
```diff
@@ -47,20 +47,15 @@ def get_new_splits(self, content: Mapping[str, Any]) -> set[SplitFullName]:
...
+        (None, True),
     ],
 )
-def test_compare_major_version(
-    test_processing_step: ProcessingStep,
-    other_version: str,
-    expected: int,
-    should_raise: bool,
-) -> None:
+def test_is_major_version_than(test_processing_step: ProcessingStep, other_version: int, expected: bool) -> None:
```
severo, line 58: No need to test that, I think, now that we just compare two ints

severo (changes requested):
It will much better with ints, I should have used this from start.

Main required change: I would feel more confident with tests of the migration jobs, because this complex migration could break the cache. Also: I think we should break the migration in two PRs (1. create job_runner_version and let worker_version unused, 2. delete worker_version)

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
AndreaFrancis: Yes, I think I can work on it

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
AndreaFrancis: Makes sense, it could be much better

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: I will send 'worker_version' removal in another PR

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
AndreaFrancis: Done but only with one record, simulating worker_version values and also missing value

services/worker/tests/test_job_runner.py
AndreaFrancis: I removed tests

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -20,6 +20,9 @@
 from mongodb_migration.migrations._20230216141000_queue_split_names_from_streaming import (
     MigrationQueueUpdateSplitNames,
 )
+from mongodb_migration.migrations._20230309141600_cache_add_job_runner_version import (
+    MigrationAddJobRunnerVerionToCacheResponse,
```
severo, line 24:
```suggestion
    MigrationAddJobRunnerVersionToCacheResponse,
```

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
```diff
@@ -0,0 +1,45 @@
...
+        logging.info("If missing, add 'job_runner_version' field based on 'worker_version' value")
+        db = get_db("cache")
+        db["cachedResponsesBlue"].update_many(
+            {"job_runner_version": {"$exists": False}},
+            [
+                {
+                    "$set": {
+                        "job_runner_version": {
+                            "$convert": {
+                                "input": {"$first": {"$split": ["$worker_version", "."]}},
```
severo, line 26: possibly cache entries don't have a worker_version, or it's `null` (we should check). How will it be managed? Should we do two updates? one for non-null worker_version, and another one for non-existent or null worker_version -> default value: null?

severo: Oh... it should be handled by onError and onNull, right?

You even test it :) OK!

jobs/mongodb_migration/tests/migrations/test_20230309141600_cache_add_job_runner_version.py
```diff
@@ -0,0 +1,52 @@
...
+@pytest.mark.parametrize(
+    "worker_version,expected",
+    [
+        ("2.0.0", 2),
+        ("1.5.0", 1),
+        ("WrongFormat", None),
+        (None, None),
+    ],
+)
+def test_cache_add_job_runner_version(mongo_host: str, worker_version: str, expected: int) -> None:
```
severo, line 37:
```suggestion
def test_cache_add_job_runner_version(mongo_host: str, worker_version: str, expected: Optional[int]) -> None:
```

jobs/mongodb_migration/tests/migrations/test_20230309141600_cache_add_job_runner_version.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+        migration = MigrationAddJobRunnerVerionToCacheResponse(
+            version="20230309141600", description="add 'job_runner_version' field based on 'worker_version' value"
+        )
+        migration.up()
+        result = db["cachedResponsesBlue"].find_one({"dataset": "dataset"})
+        assert result
+        if expected:
+            assert result["job_runner_version"]
+            assert result["job_runner_version"] == expected
```
severo, line 52:
```suggestion
        assert result["job_runner_version"] == expected
```

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230309141600_cache_add_job_runner_version.py
severo: (ignore the comments, the tests show that it's perfectly handled by your migration code!)

severo (approved):
Thanks! Just two small details.

The migration job's tests help a lot!

PR #917, opened 2023-03-10
feat: 🎸 change log level to DEBUG

severo:
also set the log level for requests to DEBUG. To investigate https://github.com/huggingface/datasets-server/issues/877. This commit should then be reverted.

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -68,7 +68,7 @@ common:
   # URL of the HuggingFace Hub
   hfEndpoint: "https://huggingface.co"
   # Log level
-  logLevel: "INFO"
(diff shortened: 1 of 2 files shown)


severo:
CI failing is unrelated to the PR

severo:
Note that dev already has the DEBUG log level, but I could not manually replicate the issue, and possibly there are missing logs (that's why I changed the value of the requests logger).
I want to try in production, with real traffic, and correlate the logs with the periods where the response time increases.

I plan to let this level of logs for some hours, then revert.

severo:
I'll merge and deploy without review, as it's a small and temporary change.

PR #919, opened 2023-03-10
Fix gated (gated=True -> gated="auto")

lhoestq:
this should fix the CI


lhoestq:
crap I included bad commits

PR #920, opened 2023-03-10
Fix gated (gated=True -> gated="auto")

lhoestq:
this should fix the CI

### DIFF
--- a/libs/libcommon/src/libcommon/dataset.py
+++ b/libs/libcommon/src/libcommon/dataset.py
@@ -145,6 +145,8 @@ def ask_access(
         r.raise_for_status()
     except requests.exceptions.HTTPError as err:
         if r.status_code == 400:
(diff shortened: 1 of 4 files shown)


AndreaFrancis (approved):
Thanks!

lhoestq:
There is an issue on moon landing: /ask-access returns 400 on non gated private datasets instead of 401 or 404 cc @SBrandeis

libs/libcommon/src/libcommon/dataset.py
```diff
@@ -145,6 +145,8 @@ def ask_access(
         r.raise_for_status()
     except requests.exceptions.HTTPError as err:
         if r.status_code == 400:
+            if r.headers and r.headers.get("X-Error-Code") == "RepoNotGated":
```
severo, line 148: oh, you already put it in the code! Thanks!

PR #823, opened 2023-02-15
Update dependencies

severo:
This PR upgrades Starlette (vulnerability) (already done by #821 for front/ - this PR also fixes services/admin and services/api) 

It also upgrades all the dependencies to the next minor version. I checked the important ones: starlette, uvicorn.

And nearly all the changes come from the upgrade of mypy to v1 -> I fixed a lot of types, and it exposed some (not critical) bugs.

PR #680, opened 2023-01-19
chore: 🤖 add --no-cache (poetry) and --no-cache-dir (pip)

severo:
to reduce the size of the docker images

Thanks @XciD

PR #923, opened 2023-03-13
Don't rebuild dev images

lhoestq:
To improve the local dev environment and let us iterate quicker:
1. I created `dev.Dockerfile` for the admin and api services
2. I improved the dev-docker-compose to use volumes for libcommon, admin and api
3. I removed a step in the `Dockerfile`s
 that was unecessarilly reinstalling python (and possibly at another version than 3.9.15)
4. I updated poetry (1.3.2 can't install `torch` in a linux container on mac m2) and fixed the `tensorflow` dependency to have it working locally on a mac with amd chip (this impacts the dev images but also the prod images, but I think it's ok, it was introduced in https://github.com/huggingface/datasets-server/pull/823)

With those changes I don't need to rebuild images anymore when changing code in libcommon/admin/api/worker etc.

The image is rebuilt only when the dependencies change.


services/worker/dev.Dockerfile
```diff
@@ -10,13 +10,13 @@ ENV PYTHONFAULTHANDLER=1 \
     PIP_DEFAULT_TIMEOUT=100 \
     POETRY_NO_INTERACTION=1 \
     # Versions:
-    POETRY_VERSION=1.3.2 \
+    POETRY_VERSION=1.4.0 \
```
severo, line 13: would it make sense to also upgrade in the other docker images?

services/worker/pyproject.toml
```diff
@@ -31,7 +31,7 @@ pypdf2 = "^3.0.1"
 python = "3.9.15"
 rarfile = "^4.0"
 scikit-learn = "^1.2.1"
-tensorflow-cpu = {version = "^2.11.0", markers = "sys_platform != 'darwin' or platform_machine != 'arm64'"}
+tensorflow = {version = "^2.11.0", markers = "sys_platform != 'darwin' or platform_machine != 'arm64'"}
```
severo, line 34: The docker images will include 2GB of useless binaries (for GPU) if we change back to tensorflow. See https://github.com/huggingface/datasets-server/pull/680/files.

I'm not sure what would be the best compromise to be able to work on MacOS locally and to have the smallest docker images possible.

severo (approved):
Please review the size of the worker image after changing tensorflow-cpu to tensorflow... It should add binaries for GPU which are not used in prod.

services/worker/pyproject.toml
lhoestq: hmm let me try alternatives then

lhoestq:
I updated poetry in the other places, and added tensorflow-aarch64 for linux containers running on amd macbooks. This way the prod images keep using `tensorflow-cpu`:)

severo:
Thanks!!

Issue #867, opened 2023-02-27
Change nomenclature of processing steps


severo:
Until now, every public endpoint (ie /splits) was associated with one job and one type of cache entry. But that’s no more the case:

- the response can be computed using two concurrent methods (streaming vs normal, for example), and the first successful response is returned
- the response can be computed at config level (/split-names) then aggregated at dataset level

It’s the occasion to rethink a bit the name of the steps, job types and cache kinds.

For now, every processing step has only one associated job type and cache kind, so we use the same for these three. But it’s no more related to the public API endpoint.

- First change: remove the initial slash (`/step` → `step`)
- Second change: to make it easier to think about, we can include the “level” at which the response is computed in the name of the step (`step` → `dataset-step`)

With these rules, after renaming and implementing some pending issues (https://github.com/huggingface/datasets-server/issues/735) we would end with the following set of processing steps and public endpoints:

Public endpoint | Dataset-level steps | Config-level steps | Split-level steps
-- | -- | -- | --
  | `dataset-config-names` |   |  
`/splits` | `dataset-split-names` | `config-split-names-from-streaming`, `config-split-names-from-dataset-info` |  |  
`/first-rows` |   |   | `split-first-rows-from-streaming`, `split-first-rows-from-parquet` |   |   |  
  |   | `config-parquet-and-dataset-info` |  
`/parquet` | `dataset-parquet` | `config-parquet` |  
`/dataset-info` | `dataset-dataset-info` | `config-dataset-info` |  
`/sizes` | `dataset-sizes` | `config-sizes` |  

<p>Also: every cached response would be accessible through: <code>/admin/cache/[step name]</code>, e.g. <code>/admin/cache/config-parquet-and-dataset-info?dataset=x&amp;config=y</code></p>

lhoestq:
As discussed in #926 we didn't use the -- convention

polinaeterna:
While implementing config-level dataset-info we named the steps `config-info` and `dataset-info` (instead of `config-dataset-info` and `dataset-dataset-info`). As @severo suggested [here](https://github.com/huggingface/datasets-server/pull/962#issuecomment-1481078199) we should also rename `parquet-and-dataset-info` to `parquet-and-info`. Should we do the same for `split-names-from-dataset-info`? @severo @AndreaFrancis @lhoestq  

Also for future reference - runners steps are now places under directories of their levels, like:
```
config/
   info.py
   parquet.py
   size.py
   ...
dataset/
   info.py
[...]

PR #926, opened 2023-03-13
Config-level size

lhoestq:
Needs https://github.com/huggingface/datasets-server/pull/923 to be merged first

I added `config-size` job and using #909 I implemented the fan-in `dataset-size` job.

I remember there was a discussion to name it `dataset--size` at one point. Not the biggest fan but lmk

Note for the review: it shows some changes on main while the other PR is not merged

### DIFF
--- a/libs/libcommon/src/libcommon/config.py
+++ b/libs/libcommon/src/libcommon/config.py
@@ -108,7 +108,8 @@ class ProcessingGraphConfig:
             "/parquet": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
             "/dataset-info": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
             "/split-names-from-dataset-info": {"input_type": "config", "requires": "/dataset-info"},
-            "/sizes": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
+            "/config-size": {"input_type": "config", "requires": "/parquet-and-dataset-info"},
+            "/dataset-size": {"input_type": "dataset", "requires": "/config-size"},
         }
     )
 
--- a/services/api/src/api/config.py
+++ b/services/api/src/api/config.py
@@ -111,7 +111,10 @@ class EndpointConfig:
             "/parquet-and-dataset-info": {"dataset": ["/parquet-and-dataset-info"]},
             "/parquet": {"dataset": ["/parquet"]},
             "/dataset-info": {"dataset": ["/dataset-info"]},
-            "/sizes": {"dataset": ["/sizes"]},
+            "/sizes": {
+                "dataset": ["/dataset-size"],
+                "config": ["/config-size"],
+            },
         }
     )
 
--- a/services/worker/src/worker/job_runners/config_size.py
+++ b/services/worker/src/worker/job_runners/config_size.py
@@ -8,7 +8,8 @@ from typing import Any, Literal, Mapping, Optional, TypedDict
 from libcommon.dataset import DatasetNotFoundError
 from libcommon.simple_cache import DoesNotExist, SplitFullName, get_response
 
-from worker.job_runner import JobRunner, JobRunnerError
+from worker.job_runner import CompleteJobResult, JobRunner, JobRunnerError
+from worker.job_runners.parquet_and_dataset_info import ParquetAndDatasetInfoResponse
 
 ConfigSizeJobRunnerErrorCode = Literal[
     "PreviousStepStatusError",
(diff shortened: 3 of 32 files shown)


services/worker/src/worker/job_runners/config_size.py
```diff
@@ -0,0 +1,211 @@
...
+            If the content of the previous step has not the expected format
+    </Tip>
+    """
+    logging.info(f"get size for dataset={dataset}, config={config}")
+
+    try:
+        response = get_response(kind="/parquet-and-dataset-info", dataset=dataset)
+    except DoesNotExist as e:
+        raise DatasetNotFoundError(
+            "No response found in previous step for this dataset: '/parquet-and-dataset-info' endpoint.", e
```
AndreaFrancis, line 120:
```suggestion
            "No response found in previous step for this dataset: '/parquet-and-dataset-info'.", e
```

lhoestq: this is also present in dataset_info.py and parquet.py, I guess I can do the changes in those as well while I'm at it

lhoestq:
This is ready for review :)

Not sure why the Chart CI is failing though

severo:
> Not sure why the Chart CI is failing though

It's an issue in the action itself (cc @XciD). I think I'll disable it meanwhile.

XciD:
Fixed :)

severo:
> I remember there was a discussion to name it dataset--size at one point. Not the biggest fan but lmk

Yes, the discussion is here: https://github.com/huggingface/datasets-server/issues/867

I don't really care if we use `-` or `--` as a special delimiter (which is uglier, I agree). But we must choose it now, as migrations are painful to code.

@AndreaFrancis : you already used `--` as a delimiter in a PR, right? Is it already in `main`, or still as a PR?

AndreaFrancis:
> @AndreaFrancis : you already used -- as a delimiter in a PR, right? Is it already in main, or still as a PR?


@severo, the worker for splits is still in progress https://github.com/huggingface/datasets-server/pull/900

libs/libcommon/src/libcommon/config.py
```diff
@@ -108,7 +108,8 @@ class ProcessingGraphConfig:
             "/parquet": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
             "/dataset-info": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
             "/split-names-from-dataset-info": {"input_type": "config", "requires": "/dataset-info"},
-            "/sizes": {"input_type": "dataset", "requires": "/parquet-and-dataset-info"},
```
AndreaFrancis, line 111: Could we keep for a while the "/sizes" processing step? so then, once the data is processed in prod, we will switch to it when we validate everything is correct. In a new PR we could: delete the processing step and make the "/sizes" endpoint use "/dataset-size" and remove the cached data for "/sizes" and start using the new one.

services/api/src/api/config.py
```diff
@@ -111,7 +111,10 @@ class EndpointConfig:
             "/parquet-and-dataset-info": {"dataset": ["/parquet-and-dataset-info"]},
             "/parquet": {"dataset": ["/parquet"]},
             "/dataset-info": {"dataset": ["/dataset-info"]},
-            "/sizes": {"dataset": ["/sizes"]},
+            "/sizes": {
+                "dataset": ["/dataset-size"],
```
AndreaFrancis, line 115: Here the same, we could at first use the old "/splits" and then if not available or with error point to "/dataset-size":
```suggestion
                "dataset": ["/splits", "/dataset-size"],
```
Once we validate that it works point only to                 "dataset": ["/dataset-size"],

services/worker/src/worker/job_runners/dataset_size.py
```diff
@@ -24,37 +25,26 @@ class DatasetSize(TypedDict):
...
-    config: str
-    split: str
-    num_bytes_parquet_files: int
-    num_bytes_memory: int
-    num_rows: int
-    num_columns: int
-
-
-class SizesContent(TypedDict):
+class DatasetSizeContent(TypedDict):
```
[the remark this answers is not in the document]
severo: We might even expand these later to: `/dataset-sizes` or `/dataset-sizes?filter=audio` for example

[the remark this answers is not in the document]
lhoestq: Do you mean we should be able to query multiple datasets at the same time ?

`/size` should be for one dataset or one config IMO, and modifying the API at that level seems out of scope of this PR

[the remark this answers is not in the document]
severo: > /size should be for one dataset or one config IMO, and modifying the API at that level seems out of scope of this PR

Sure, that's why I mentioned "later" (for listing the size of multiple datasets)

severo:
> I remember there was a discussion to name it dataset--size at one point. Not the biggest fan but lmk

I remember why I proposed `--`. It was because `dataset-dataset-info` looked like a typo, and `dataset--dataset-info` was a bit clearer.

But, I'm OK with both ways, let's decide on one of them.

services/worker/src/worker/job_runners/config_size.py
```diff
@@ -0,0 +1,203 @@
...
+            {
+                "dataset": dataset,
+                "config": config,
+                "num_bytes_original_files": config_dataset_info["download_size"],
+                "num_bytes_parquet_files": sum(split_size["num_bytes_parquet_files"] for split_size in split_sizes),
+                "num_bytes_memory": sum(
+                    split_size["num_bytes_memory"] for split_size in split_sizes
+                ),  # or "num_bytes_memory": config_dataset_info["dataset_size"],
+                "num_rows": sum(split_size["num_rows"] for split_size in split_sizes),
+                "num_columns": len(config_dataset_info["features"]),
```
severo, line 168:
```suggestion
                "num_columns": num_columns,
```

services/worker/src/worker/job_runners/dataset_size.py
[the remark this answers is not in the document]
severo: OK. Let's get back to this discussion in another PR or issue

services/worker/src/worker/job_runners/dataset_size.py
```diff
@@ -107,7 +97,7 @@ def compute_sizes_response(dataset: str) -> SizesResponse:
         response = get_response(kind="/parquet-and-dataset-info", dataset=dataset)
```
severo, line 97: this step depends on `/parquet-and-dataset-info` but we don't express it directly in the processing graph:
- in the current model, a step only has one parent
- the parent `config-size` is the child of `/parquet-and-dataset-info`

To take into account if we change the model one day.

severo (approved):
Nice!

Issue #865, opened 2023-02-27
Turn /parquet into a config-level job


severo:
We should:
- compute the values for every config independently: `config--parquet`
- compute the dataset-level response `dataset--parquet` each time a `config--parquet` is computed (allowing partial responses: some config responses can be missing or erroneous)
- return the appropriate response on /parquet, depending on the inputs (see https://github.com/huggingface/datasets-server/pull/834)

See #735.

severo:
> A specific challenge is how to store the parquet files. Currently, we have a dedicated "branch" (refs/convert/parquet), with an arborescence with one subdirectory per config, and parquet files inside for all the splits of the config. I think this schema would still work well, but as it's versioned with git, maybe some details could arise (concurrent access? it should be handled with the [parent_commit](https://github.com/huggingface/datasets-server/blob/main/workers/datasets_based/src/datasets_based/workers/parquet_and_dataset_info.py#L625) parameter, at least an error would be thrown in case of concurrent write attempts).

severo:
> A specific challenge is how to store the parquet files. Currently, we have a dedicated "branch" (refs/convert/parquet), with an arborescence with one subdirectory per config, and parquet files inside for all the splits of the config. I think this schema would still work well, but as it's versioned with git, maybe some details could arise (concurrent access? it should be handled with the [parent_commit](https://github.com/huggingface/datasets-server/blob/main/workers/datasets_based/src/datasets_based/workers/parquet_and_dataset_info.py#L625) parameter, at least an error would be thrown in case of concurrent write attempts).

this comment is for /parquet-and-dataset-info, not for /parquet

PR #931, opened 2023-03-14
Config-level parquet

lhoestq:
Created `config-parquet` and `dataset-parquet` jobs, and removing the old `/parquet` job.

The `/parquet` endpoints uses the two new jobs.

Close https://github.com/huggingface/datasets-server/issues/865


severo (approved):
👌

PR #932, opened 2023-03-15
refactor: 💡 fix type of the public key: str, not Any

severo:


severo:
This PR only changes types, and mypy is happy, so... merging without a review

PR #933, opened 2023-03-15
fix: 🐛 take the algorithm into account in config

severo:


severo:
Merging, as it's minor

Issue #751, opened 2023-01-31
Refactor to have only one app accessing a mongodb database 


severo:
As assessed by @rtrompier, it's generally a bad idea to have multiple apps accessing the same database concurrently. Ideally, one app should have access to the database, and expose a REST API for the other apps.

severo:
Not directly related, but while working on #763, I'm creating a new MongoConnection class.

severo:
Keeping it open for now. It means creating a microservice for the jobs, and a microservice for the cache.

PR #936, opened 2023-03-15
dataset-level split names from streaming

AndreaFrancis:


services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class ResponseNotReadyError(DatasetSplitNamesFromStreamingJobRunnerError):
+    """Raised when the response has not been processed yet from any source."""
+
+    def __init__(self, message: str):
+        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "ResponseNotReady")
+
+
+class PendingJob(TypedDict):
```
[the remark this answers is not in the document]
lhoestq: maybe no need to factorize, but yea that'b be great to use the same format

services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py
```diff
@@ -0,0 +1,176 @@
...
+            }
+        ),
+        progress,
+    )
+
+
+class DatasetSplitNamesFromStreamingJobRunner(JobRunner):
+    @staticmethod
+    def get_job_type() -> str:
+        return "dataset-split-names"
```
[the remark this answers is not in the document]
AndreaFrancis: Thanks! I fixed it

services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py
```diff
@@ -0,0 +1,176 @@
...
+    def __init__(self, message: str):
+        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "ResponseNotReady")
+
+
+class PendingJob(TypedDict):
+    dataset: str
+    config: Optional[str]
+
+
+class SuccessJob(PendingJob):
```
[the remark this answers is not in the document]
AndreaFrancis: Good idea! I moved common responses to utils.py file

services/api/src/api/config.py
```diff
@@ -110,7 +110,7 @@ class EndpointConfig:
         default_factory=lambda: {
             "/config-names": {"dataset": ["/config-names"]},
             "/splits": {
-                "dataset": ["/splits"],
+                "dataset": ["/splits", "dataset-split-names-from-streaming"],
```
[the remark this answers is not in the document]
severo: You prefer not to change the order for now? (I'm OK with both, it's not important)

services/worker/src/worker/job_runners/config/split_names_from_streaming.py
```diff
@@ -109,7 +100,7 @@ def compute_split_names_from_streaming_response(
             f"Cannot get the split names for the config '{config}' of the dataset.",
             cause=err,
         ) from err
-    return {"splits": split_name_items}
+    return SplitsList({"splits": split_name_items})
```
severo, line 103: I didn't know we could instantiate a "TypedDict" class. I was using them only as type hints. It works well, interesting!

services/worker/src/worker/job_runners/config/split_names_from_streaming.py
```diff
@@ -98,7 +89,7 @@ def compute_split_names_from_streaming_response(
     use_auth_token: Union[bool, str, None] = hf_token if hf_token is not None else False
 
     try:
-        split_name_items: List[SplitNameItem] = [
+        split_name_items: List[SplitItem] = [
             {"dataset": dataset, "config": config, "split": str(split)}
```
severo, line 93: Should we also instantiate here?

```suggestion
            SplitItem({"dataset": dataset, "config": config, "split": str(split)})
```

services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py
```diff
@@ -0,0 +1,168 @@
...
+        splits: List[SplitItem] = []
+        pending: List[ConfigItem] = []
+        failed: List[FailedConfigItem] = []
+        total = 0
+        for config_item in config_content:
+            config = config_item["config"]
+            total += 1
+            try:
+                response = get_response(kind="/split-names-from-streaming", dataset=dataset, config=config)
+            except DoesNotExist:
```
severo, line 112: in that case, should we look if a pending job exists (WAITING or STARTED), and create it if not?

severo (approved):

services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py
AndreaFrancis: Yes, I think that is a good idea, we should apply it for all the aggregates job runners, is it ok if I create a new issue that will cover all the existing runners that manages the same scenario? (Parquet, size, splits)

services/worker/src/worker/job_runners/config/split_names_from_streaming.py
AndreaFrancis: Thanks

services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py
severo: yes, better in a new issue.

By the way, it should not be done by the method we use to get the cache: it should return the content, and if missing, check if the cache should have existed. It's somewhat related to https://github.com/huggingface/datasets-server/issues/751 (having the cache in a microservice - it would be in charge of doing the check)

services/api/src/api/config.py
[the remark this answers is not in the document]
severo: https://datasets-server.huggingface.co/splits?dataset=google/xtreme_s

services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py
AndreaFrancis: I created the issue https://github.com/huggingface/datasets-server/issues/943

PR #947, opened 2023-03-17
TEST: DO NOT MERGE

rtrompier:
only for test


severo:
OK, thanks for the test. It's weird that tailscale authentication works here, and not in https://github.com/huggingface/datasets-server/actions/runs/4439030674/jobs/7805759136.
Let's just ignore it in #912

PR #951, opened 2023-03-17
feat: 🎸 add logs in case of error in /webhook

severo:
these errors should not occur, and we should be alerted when it's the case.


severo:
Merging

Issue #950, opened 2023-03-17
More than half of the received webhooks are ignored!


severo:
An exception is raised (400 is returned) and thus the update jobs are not created, leading to many outdated dataset viewers on the Hub.

severo:
It has started yesterday.

<img width="398" alt="Capture d’écran 2023-03-17 à 11 03 08" src="https://user-images.githubusercontent.com/1676121/225874016-f6b943bc-0dd9-4f5f-adee-85bff4056e10.png">

albertvillanova:
Just right now I had this issue again for: https://huggingface.co/datasets/cmu_hinglish_dog

severo:
I fixed the issue by increasing the timeout value for requests to the Hub (200ms -> 1.5s) - #952.

We have not experienced new errors in the last 8 minutes. Closing.

PR #952, opened 2023-03-17
feat: 🎸 increase the timeout to Hub from 200ms to 1.5s

severo:


severo:
I merge, there is nothing complex here and it's urgent.

severo:
Fixes #950

PR #954, opened 2023-03-17
feat: 🎸 reduce log level to INFO in prod

severo:
Investigation on #950 has finished, we can reduce the log level


severo:
Merging, since it's simple

PR #955, opened 2023-03-17
feat: 🎸 authenticate webhooks sent by the Hub

severo:
The authentication is done with a shared secret passed in X-Webhook-Secret header. If not trusted, a log is recorded and only add/update are taken into account. I.e. move and delete required the sender to be trusted. Also: before, requests were sent to the Hub to check if the dataset was supported by the datasets-server. Now, we always create the job, and it's its responsibility to check the dataset is supported. This way: a valid webhook should always be taken into account. See #950 for antecedents.


severo:
Should we require a trusted sender for all the operations? (in this PR it's required for move and delete only)

severo:
Don't merge before https://github.com/huggingface/infra/pull/536 (internal)

services/api/src/api/routes/webhook.py
```diff
@@ -125,6 +128,15 @@ async def webhook_endpoint(request: Request) -> Response:
                     logging.warning(f"/webhook: unexpected error while parsing the JSON body is invalid. Error: {e}")
                     return get_response(content, 500)
 
+                HEADER = "x-webhook-secret"
+                trust_sender = (
+                    hf_webhook_secret is not None
+                    and (secret := request.headers.get(HEADER)) is not None
+                    and secret == hf_webhook_secret
+                )
+                if not trust_sender:
```
AndreaFrancis, line 137: Maybe log `trust_sender` value anyway in before https://github.com/huggingface/datasets-server/pull/955/files#diff-75de3bd8cb5b5606e98fbf00f4e95bb2108ec49aab5fd00f7c97fb96734c7c90R66 for troubleshooting purposes.

AndreaFrancis (approved):

services/api/src/api/routes/webhook.py
severo: OK. As discussed directly, let's keep it as is for now. The idea is to log the anomalous behaviors, to avoid longing too much.

severo:
Related issues: https://github.com/huggingface/datasets-server/issues/958, https://github.com/huggingface/datasets-server/issues/957, https://github.com/huggingface/datasets-server/issues/959, https://github.com/huggingface/datasets-server/issues/956

Issue #959, opened 2023-03-19
Dataset Viewer issue for shainaraza/clinical_bias


shainaraza:
### Link

https://huggingface.co/datasets/shainaraza/clinical_bias

### Description

The dataset viewer is not working for dataset shainaraza/clinical_bias.

Error details:

```
Error code:   ResponseNotReady

```

severo:
I cannot access https://huggingface.co/datasets/shainaraza/clinical_bias anymore. Did you turn it private or delete it?

albertvillanova:
The dataset is private and the dataset viewer does not support private datasets.

albertvillanova:
The dataset is public now and the viewer works as expected.

![Screenshot from 2023-04-19 09-55-11](https://user-images.githubusercontent.com/8515462/233007595-e582ac02-c16e-4071-a532-6da0641a10cc.png)

I'm closing this issue.

Issue #958, opened 2023-03-19
Dataset Viewer issue for Zexuan/soc_data


lzxlll:
### Link

https://huggingface.co/datasets/Zexuan/soc_data

### Description

The dataset viewer is not working for dataset Zexuan/soc_data.

Error details:

```
Error code:   ResponseNotReady

```

severo:
Fixed, thanks for reporting!

<img width="1017" alt="Capture d’écran 2023-03-20 à 10 33 10" src="https://user-images.githubusercontent.com/1676121/226300115-87f8d2f1-4538-40ee-aaed-69ecfa31d03d.png">

Issue #957, opened 2023-03-18
Dataset Viewer issue for Crapp/sadQuotes


crapptrapp:
### Link

https://huggingface.co/datasets/Crapp/sadQuotes

### Description

The dataset viewer is not working for dataset Crapp/sadQuotes.

Error details:

```
Error code:   ResponseNotReady

```

severo:
Fixed, thanks for reporting!

<img width="1032" alt="Capture d’écran 2023-03-20 à 10 40 21" src="https://user-images.githubusercontent.com/1676121/226301825-81f73067-41a1-44bf-82d0-a52f621a1e12.png">

Issue #956, opened 2023-03-17
Dataset Viewer issue for vhug/xorder_dish_dataset


v-prgmr:
### Link

https://huggingface.co/datasets/vhug/xorder_dish_dataset

### Description

The dataset viewer is not working for dataset vhug/xorder_dish_dataset.

Error details:

```
Error code:   ResponseNotReady

```

albertvillanova:
@severo you mean: https://huggingface.co/datasets/vhug/xorder_dish_dataset

And the dataset has been deleted.

severo:
Yes, sorry, I edited the message

severo:
Closing

PR #961, opened 2023-03-20
feat: 🎸 reduce the number of allowed jobs for one namespace

severo:
Currently, 20 jobs are only for one namespace, blocking all the other users/datasets (220 waiting jobs). Also: moving config-names to the list of "light" jobs (even if it's not as light as the other ones, it's quick because it only creates the dataset builder.

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -191,7 +191,7 @@ api:
 workers:
   -
     deployName: "all"
-    maxJobsPerNamespace: 20
+    maxJobsPerNamespace: 5
     workerOnlyJobTypes: ""
     nodeSelector:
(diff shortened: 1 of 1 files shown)


severo:
Merging, because it's simple (and urgent to solve the prod)

albertvillanova:
I was going to accept it but it is already merged... :sweat_smile: 

Thanks for the improvement.

severo:
Related issues: https://github.com/huggingface/datasets-server/issues/958, https://github.com/huggingface/datasets-server/issues/957, https://github.com/huggingface/datasets-server/issues/959, https://github.com/huggingface/datasets-server/issues/956

Issue #864, opened 2023-02-27
Turn /dataset-info into a config-level job


severo:
We should:
- compute the values for every config independently: `config--dataset-info`
- compute the dataset-level response `dataset--dataset-info` each time a `config--dataset-info` is computed (allowing partial responses: some config responses can be missing or erroneous)
- return the appropriate response on /dataset-info, depending on the inputs (see https://github.com/huggingface/datasets-server/pull/834)

See #735.

AndreaFrancis:
For the new worker /split-names-from-dataset-info, we will probable need to change the source, currently it is /dataset-info [config] but after this is implemented, it will have to be config--dataset-info.

Issue #735, opened 2023-01-31
Change /parquet-and-dataset-info processing step


severo:
For now, /parquet-and-dataset-info is run for a dataset. But as it's a loop on all the configs, we could instead decide to turn it into a "config" job, ie. make it depend on the result of /config-names, and launch a job for each of the configs.

This means doing the same for the dependent processing steps: /dataset-info, /parquet, /sizes.

This means also being able to answer the endpoints to both:
- ?dataset=dataset&config=config: directly from the database
- ?dataset=dataset: an aggregation of the config responses. For this point, the discussion is the same as https://github.com/huggingface/datasets-server/issues/734: should we have a dedicate worker, or compute on the fly?

The benefit would be to decouple how different configs from the same dataset are treated, in particular, let a config work even if another one is broken.

A specific challenge is how to store the parquet files. Currently, we have a dedicated "branch" (refs/convert/parquet), with an arborescence with one subdirectory per config, and parquet files inside for all the splits of the config. I think this schema would still work well, but as it's versioned with git, maybe some details could arise (concurrent access? it should be handled with the [`parent_commit`](https://github.com/huggingface/datasets-server/blob/main/workers/datasets_based/src/datasets_based/workers/parquet_and_dataset_info.py#L625) parameter, at least an error would be thrown in case of concurrent write attempts).

---

Subtasks (should be implemented in that order):

- [x] https://github.com/huggingface/datasets-server/issues/863
- [x] https://github.com/huggingface/datasets-server/issues/864
- [x] https://github.com/huggingface/datasets-server/issues/865
- [x] https://github.com/huggingface/datasets-server/issues/866

albertvillanova:
I would suggest to split this issue into smaller ones, so that we implement it step by step...

polinaeterna:
- [ ] also as `config-info` is implemented, `split-names-from-dataset-info` should be adjusted to be dependent on `config-info` instead of `dataset-info` as suggested [here](https://github.com/huggingface/datasets-server/issues/864#issuecomment-1446865476)
mentioning here not to forget

severo:
We can close, right @polinaeterna?

PR #962, opened 2023-03-20
Config-level dataset info

polinaeterna:
Will solve https://github.com/huggingface/datasets-server/issues/864

Part of https://github.com/huggingface/datasets-server/issues/735

I preserved the old name (without leading "/") for dataset-level step - `"dataset-info"` and named config-level step `"config-info"` because  `"dataset-dataset-info"` and  `"config-dataset-info"` look a bit silly to me but I don't have a strong opinion here, let me know what you think.

`config-info` result format is dict with info, `dataset-info` result format remained the same (dict[config_name, infos]), like this:
* `config-info`
```
dataset_info: {
    "description": "_DESCRIPTION",
    "citation": "_CITATION",
    "homepage": "_HOMEPAGE",
    ...
}
```
* `dataset-info`
```
dataset_info: {
"config_1": {
    "description": "_DESCRIPTION",
    "citation": "_CITATION",
    "homepage": "_HOMEPAGE",
    ...
  },
"config_2": {
    "description": "_DESCRIPTION",
    "citation": "_CITATION",
    "homepage": "_HOMEPAGE",
    ...
  },
}
```

I'm not sure why CI is broken.


services/worker/src/worker/job_runners/config_info.py
```diff
@@ -0,0 +1,143 @@
...
+    @staticmethod
+    def get_job_type() -> str:
+        return "config-info"
+
+    @staticmethod
+    def get_job_runner_version() -> int:
+        return 1
+
+    def compute(self) -> CompleteJobResult:
+        if self.config is None:
```
AndreaFrancis, line 133: Maybe you could also validate dataset
        if self.dataset is None:
            raise ValueError("dataset is required")

services/worker/src/worker/job_runners/config_info.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class ConfigInfoJobRunner(JobRunner):
+    @staticmethod
+    def get_job_type() -> str:
+        return "config-info"
+
+    @staticmethod
+    def get_job_runner_version() -> int:
+        return 1
```
AndreaFrancis, line 130:
```suggestion
        return PROCESSING_STEP_CONFIG_INFO_VERSION
```
And import it from libcommon constants

AndreaFrancis:
We may need to have a migration for the change of `/dataset-info` to `dataset-info`, otherwise api will not be able to find the existing cached records for the old runner.
For both: cache and queue. You could see https://github.com/huggingface/datasets-server/blob/main/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230320163700_cache_first_rows_from_streaming.py and https://github.com/huggingface/datasets-server/blob/main/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230320165700_queue_first_rows_from_streaming.py

severo:
> I preserved the old name (without leading "/") for dataset-level step - "dataset-info" and named config-level step "config-info" because "dataset-dataset-info" and "config-dataset-info" look a bit silly to me but I don't have a strong opinion here, let me know what you think.

👍

services/worker/src/worker/job_runners/dataset_info.py
```diff
@@ -51,24 +61,31 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
...
-            If the content of the previous step has not the expected format
+            If the content of the previous step doesn't have the expected format.
+        - [`~libcommon.dataset.DatasetNotFoundError`]
+            If the dataset does not exist, or if the
+            token does not give the sufficient access to the dataset, or if the dataset is private
+            (private datasets are not supported by the datasets server)
     </Tip>
     """
-    logging.info(f"get dataset_info for dataset={dataset}")
+    logging.info(f"get dataset_info for {dataset=}")
```
severo, line 88:
```suggestion
    logging.info(f"get dataset_info for {dataset}")
```

services/worker/src/worker/job_runners/dataset_info.py
```diff
@@ -51,24 +61,31 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
...
 
     try:
         response = get_response(kind="/parquet-and-dataset-info", dataset=dataset)
```
severo, line 91: I think that we should rely on `config-names` instead to get the list of configs

services/worker/src/worker/job_runners/dataset_info.py
```diff
@@ -78,32 +95,69 @@ def compute_dataset_info_response(dataset: str) -> DatasetInfoResponse:
...
+        pending, failed = [], []
+        for config in content["dataset_info"]:
+            total += 1
+            try:
+                config_response = get_response(kind="config-info", dataset=dataset, config=config)
+            except DoesNotExist:
+                logging.debug(f"No response found in previous step for {dataset=} {config=}: 'config-info'.")
+                pending.append(
+                    PreviousJob(
+                        kind="config-size",
```
severo, line 116:
```suggestion
                        kind="config-info",
```

services/worker/src/worker/job_runners/dataset_info.py
```diff
@@ -78,32 +95,69 @@ def compute_dataset_info_response(dataset: str) -> DatasetInfoResponse:
...
+                        config=config,
+                        split=None,
+                    )
+                )
+                continue
+            if config_response["http_status"] != HTTPStatus.OK:
+                logging.debug(f"Previous step gave an error: {response['http_status']}")
+                failed.append(
+                    PreviousJob(
+                        kind="config-size",
```
severo, line 127:
```suggestion
                        kind="config-info",
```

services/worker/src/worker/job_runners/dataset_info.py
polinaeterna: should we also do this for `dataset-size`?

services/worker/src/worker/job_runners/dataset_info.py
polinaeterna: don't you like this syntax? i think it's convenient to use `f"{dataset=}"` instead of `f"dataset={dataset}"`) not to duplicate words.

services/worker/src/worker/job_runners/dataset_info.py
severo: yes! you're right, it's the same issue.

We can do it in another PR if you prefer

polinaeterna:
@AndreaFrancis @severo I addressed your comments but I still have CI errors:
`ERROR tests/test_queue.py::test__add_job - RuntimeError: Mongo resource is not available` in `libcommon` and chart diff doesn't seem working. what am I missing?

severo:
> chart diff doesn't seem working

this is because the PR is from an external branch (in your personal fork). It works only on branches in this repo.

severo:
> CI errors: ERROR tests/test_queue.py::test__add_job - RuntimeError: Mongo resource is not available in libcommon

I'm not sure why. I re-run and it works now (it failed again on the first re-run)

services/worker/src/worker/job_runners/dataset_info.py
severo: Oh, I had no idea that it was a valid syntax, I thought it was a typo. Sorry!!! And thanks for teaching me that.

severo: Doc is here: https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

services/worker/src/worker/job_runners/config_info.py
severo, line 1: I think the file should now live under job_runners/config/info.py, right @AndreaFrancis ?

services/worker/src/worker/job_runners/dataset_info.py
severo, line 1: and this file should live in job_runners/dataset/info.py

severo (approved):
Good work, thanks!

Just a detail with the name of two files

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230323160000_queue_dataset_info.py
```diff
@@ -0,0 +1,71 @@
...
+
+from libcommon.queue import Job
+from mongoengine.connection import get_db
+
+from mongodb_migration.check import check_documents
+from mongodb_migration.migration import Migration
+
+dataset_info = "/dataset-info"
+dataset_info_updated = "dataset-info"
+db_name = "cache"
```
AndreaFrancis, line 11:
```suggestion
```
It looks like this is not used

AndreaFrancis (approved):
Nice!

PR #967, opened 2023-03-20
/first-rows to first-rows-from-streaming

AndreaFrancis:
Part of https://github.com/huggingface/datasets-server/issues/755

>  Implement first-rows-from-streaming job runner at config level (It already exist as /first-rows)

`/first-rows processing` step will be renamed to `first-rows-from-streaming` and new source of split validation will be `/split-names-from-streaming` result instead of using library.


libs/libcommon/src/libcommon/constants.py
```diff
@@ -14,7 +14,7 @@
 PROCESSING_STEP_DATASET_SIZE_VERSION = 1
 PROCESSING_STEP_DATASET_SPLIT_NAMES_FROM_DATASET_INFO_VERSION = 1
 PROCESSING_STEP_DATASET_SPLIT_NAMES_FROM_STREAMING_VERSION = 1
-PROCESSING_STEP_FIRST_ROWS_VERSION = 2
+PROCESSING_STEP_FIRST_ROWS_FROM_STREAMING_VERSION = 2
```
AndreaFrancis, line 17: I didn't change the version here since it is just a `kind` renaming change and a fix for split validation.
If I upgrade this version, `cache-refresh` job could start appending new jobs for the outdated records, but that is not the intention here at least we want to re compute all the first rows again.

severo:
Should we name that step `split-first-rows-from-streaming` instead of `first-rows-from-streaming` to give the input level?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230320163700_cache_first_rows_from_streaming.py
```diff
@@ -0,0 +1,38 @@
...
+db_name = "cache"
+
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationCacheUpdateFirstRows(Migration):
+    def up(self) -> None:
+        logging.info(f"Rename cache_kind field from {first_rows} to {first_rows_from_streaming}")
+        db = get_db(db_name)
+        # update existing documents with the new kind (if any) to avoid duplicates (will be deleted later)
+        db["cachedResponsesBlue"].update_many({"kind": first_rows_from_streaming}, {"$set": {"kind": first_rows_tmp}})
```
severo, line 24: I understand, but I'm afraid: what would occur if, for some reason, this job migration is run again: we would loose all the cache entries. Is there a way to do the same without deleting the existing entries (that must be correct, by the way). Maybe there is a way to try to update, and ignore the duplicate errors?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230320163700_cache_first_rows_from_streaming.py
```diff
@@ -0,0 +1,38 @@
...
+        # update existing documents with the old kind
+        db["cachedResponsesBlue"].update_many({"kind": first_rows}, {"$set": {"kind": first_rows_from_streaming}})
+        # delete the duplicates
+        db["cachedResponsesBlue"].delete_many({"kind": first_rows_tmp})
+
+    def down(self) -> None:
+        logging.info(f"Rollback cache_kind field from {first_rows_from_streaming} to {first_rows}")
+        db = get_db(db_name)
+        db["cachedResponsesBlue"].update_many({"kind": first_rows_from_streaming}, {"$set": {"kind": first_rows}})
```
severo, line 33: in that case we also could have duplicates.

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230320165700_queue_first_rows_from_streaming.py
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    "$set": {
+                        "unicity_id": {
+                            "$replaceOne": {
+                                "input": "$unicity_id",
+                                "find": f"Job[{first_rows}]",
+                                "replacement": f"Job[{first_rows_from_streaming}]",
+                            }
+                        },
+                        "type": first_rows_from_streaming,
```
severo, line 39: in that case you could also have duplicates. I think we could do the same in all places (ie. try to update, and ignore the duplicate errors - not sure if it's possible though :) )

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230320163700_cache_first_rows_from_streaming.py
AndreaFrancis: I did the rollback of those lines. Sadly I could find a way to skip in update error. I think the problem with duplicated records might be caused in splits because of granularity level kind-dataset-config-split, since split-names-from-streaming  always has split as null/None. But that is just a suspect.
I would like to try the simple update_many approach with a sample prod data. If needed we could try to add a transaction before the update but not sure, would like to reproduce the issue locally first. wdyt?
Same for queue but since unicity_id is not a unique constraint, it shouldn't fail.

severo: By the way, I don't see how we could have existing entries with `split-first-rows-from-streaming` kind. So applying

https://github.com/huggingface/datasets-server/pull/967/files#diff-17d3c24000f1960066c66c311c9c598dde0c10fc79b2e6493830a1cafd10e2adR24-R26

should work.

severo (approved):

PR #970, opened 2023-03-21
Adding again custom workers for critical jobs

AndreaFrancis:


severo:
I think it's enough to add resources to the `all` worker.

PR #972, opened 2023-03-21
Reduce prod resources and remove cache refresh job (temporal change)

AndreaFrancis:


severo:
I'm not sure what is the best way to disable a job. I imagine that we should let the docker image details, and instead put an "enable" variable around the job template

severo:
> I'm not sure what is the best way to disable a job. I imagine that we should let the docker image details, and instead put an "enable" variable (feature switch) around the job template

Sorry, my bad, the existence of the cacheRefresh image is the feature switch (https://github.com/huggingface/datasets-server/blob/5c0e59f79d4a1f2f09b0b2ddef02b0b30a85f226/chart/templates/jobs/cache-refresh/job.yaml#L4). No need to delete job.yaml I think. Maybe just comment the cacheRefresh key in prod.yaml, with a comment saying that we disable it?

rtrompier (approved):

chart/templates/jobs/cache-refresh/job.yaml
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
 
-{{- if .Values.images.jobs.cacheRefresh }}
+{{- if and .Values.cacheRefresh.enabled .Values.images.jobs.cacheRefresh }}
```
severo, line 4: is this the syntax? `and` first, and then two values?

severo (approved):

PR #974, opened 2023-03-21
Disable cache refresh job on values and dev yaml files

AndreaFrancis:


AndreaFrancis:
Need to comment the value also in values.yaml since prod is still considering it as part of the deploy

severo (approved):

AndreaFrancis:
I will merge this temporal change, later, will implement a better mechanism to enable/disable jobs

PR #981, opened 2023-03-23
Access docker host

severo:
- give the docker services access to the host network. It is required to be able to access a local port on the same machine,
for example.
- run 4 workers in parallel, it helps processing the jobs quicker


lhoestq (approved):
nice :)

severo:
Unfortunately, it does not fully work. The idea was to be able to access a local instance of the [Hub](https://github.com/huggingface/moon-landing) (internal), ie that runs on http://localhost:some-port. Using host.docker.internal from the worker container gives access to the host network, e.g. http://host.docker.internal:some-port gives access to the local instance of the Hub. But this application does redirections to "localhost:some-port" -> this cannot be accessed anymore from the worker container.

I'm trying to get another local setup where both applications (Hub and datasets-server) can work together.

severo:
I think it should work now

severo:
I merge, since it does not affect the prod.

PR #982, opened 2023-03-23
Get config names from /config-names instead of /parquet-and-dataset-info for dataset-level size and parquet

polinaeterna:
Get config names from `/config-names` cache instead of `/parquet-and-dataset-info` in `dataset-size` and `dataset-parquet` (aligned with `dataset-info` https://github.com/huggingface/datasets-server/pull/962, discussed [here](https://github.com/huggingface/datasets-server/pull/962#discussion_r1146102010))

\+ move files to new directories structure (config-level to `config/`, dataset-level to `dataset/`


severo (approved):
Perfect. The logic is simpler that way, and the tests are a bit smaller.

PR #984, opened 2023-03-24
Make config-level `/split-names-from-dataset-info` dependent on `config-info` instead of `dataset-info`

polinaeterna:
See https://github.com/huggingface/datasets-server/issues/864#issuecomment-1446865476


AndreaFrancis (approved):
Thanks!

severo (approved):

Issue #846, opened 2023-02-24
[once in hfh] improve /parquet-and-dataset-info job runner


severo:
Once https://github.com/huggingface/huggingface_hub/pull/1331 is released, upgrade huggingface_hub and rework how we create the refs/convert/parquet branch in https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/parquet_and_dataset_info.py#L805 to have an empty history.

severo:
Released: https://github.com/huggingface/huggingface_hub/releases/tag/v0.13.0.rc0. We have to upgrade huggingface_hub to 0.13 to have it

PR #985, opened 2023-03-24
Config-level parquet-and-dataset-info

polinaeterna:
Will solve https://github.com/huggingface/datasets-server/issues/866

Part of https://github.com/huggingface/datasets-server/issues/735

# Questions:
* Didn't I miss anything? 
* Should there be any migration jobs? Technically it's a new job runner, it's not equal to the previous dataset-level one (results are not identical), and the old one doesn't exist anymore. 
* Can I rename relevant environmental variables (like `PARQUET_AND_DATASET_INFO_BLOCKED_DATASETS` -> to `PARQUET_AND_INFO_BLOCKED_DATASETS`) or might it break anything?

### DIFF
--- a/.vscode/monorepo.code-workspace
+++ b/.vscode/monorepo.code-workspace
@@ -35,6 +35,10 @@
     {
       "name": "services/reverse-proxy",
       "path": "../services/reverse-proxy"
+    },
+    {
+      "name": "front/admin_ui",
+      "path": "../front/admin_ui"
     }
   ],
   "settings": {
--- a/front/admin_ui/app.py
+++ b/front/admin_ui/app.py
@@ -71,6 +71,14 @@ with gr.Blocks() as demo:
                 gr.Markdown("*you can select multiple values by separating them with commas, e.g. split='train, test'*")
                 refresh_dataset_button = gr.Button("Force resfresh dataset")
                 refresh_dataset_output = gr.Markdown("")
+            with gr.Tab("Dataset status"):
+                dataset_name = gr.Textbox(label="dataset", placeholder="c4")
+                dataset_status_button = gr.Button("Get dataset status")
+                gr.Markdown("### Cached responses")
+                cached_responses_table = gr.DataFrame()
+                gr.Markdown("### Pending jobs")
+                jobs_table = gr.DataFrame()
+
 
     def auth(token):
         if not token:
@@ -121,7 +129,53 @@ with gr.Blocks() as demo:
                 pending_jobs_summary_table: gr.update(visible=True, value=pd.DataFrame({"Error": [f"❌ Failed to view pending jobs to {DSS_ENDPOINT} (error {response.status_code})"]})),
                 recent_pending_jobs_table: gr.update(value=None)
             }
-    
+
+    def get_dataset_status(token, dataset):
+        headers = {"Authorization": f"Bearer {token}"}
+        response = requests.get(f"{DSS_ENDPOINT}/admin/dataset-status?dataset={dataset}", headers=headers, timeout=60)
+        if response.status_code == 200:
+            dataset_status = response.json()
+            cached_responses_df = pd.DataFrame([{
+                    "type": job_type,
+                    "dataset": cached_response["dataset"],
+                    "config": cached_response["config"],
+                    "split": cached_response["split"],
+                    "http_status": cached_response["http_status"],
+                    "error_code": cached_response["error_code"],
+                    "job_runner_version": cached_response["job_runner_version"],
+                    "dataset_git_revision": cached_response["dataset_git_revision"],
+                    "progress": cached_response["progress"]
+                }
+                for job_type, content in dataset_status.items()
+                for cached_response in content["cached_responses"]
+            ])
+            jobs_df = pd.DataFrame([{
+                    "type": job_type,
+                    "dataset": job["dataset"],
+                    "config": job["config"],
+                    "split": job["split"],
+                    "namespace": job["namespace"],
+                    "force": job["force"],
+                    "priority": job["priority"],
+                    "status": job["status"],
+                    "created_at": job["created_at"],
+                    "started_at": job["started_at"],
+                    "finished_at": job["finished_at"],
+                    "last_heartbeat": job["last_heartbeat"]
+                }
+                for job_type, content in dataset_status.items()
+                for job in content["jobs"]
+            ])
+            return {
+                cached_responses_table: gr.update(value=cached_responses_df),
+                jobs_table: gr.update(value=jobs_df)
+            }
+        else:
+            return {
+                cached_responses_table: gr.update(value=None),
+                jobs_table: gr.update(value=None)
+            }
+
     def query_jobs(pending_jobs_query):
         global pending_jobs_df
         try:
@@ -169,6 +223,7 @@ with gr.Blocks() as demo:
     query_pending_jobs_button.click(query_jobs, inputs=pending_jobs_query, outputs=[pending_jobs_query_result_df])
     
     refresh_dataset_button.click(refresh_dataset, inputs=[token_box, refresh_type, refresh_dataset_name, refresh_config_name, refresh_split_name], outputs=refresh_dataset_output)
+    dataset_status_button.click(get_dataset_status, inputs=[token_box, dataset_name], outputs=[cached_responses_table, jobs_table])
 
 
 if __name__ == "__main__":
--- a/libs/libcommon/src/libcommon/simple_cache.py
+++ b/libs/libcommon/src/libcommon/simple_cache.py
@@ -187,22 +187,6 @@ def get_response_without_content(
     }
 
 
-def get_dataset_responses_without_content_for_kind(kind: str, dataset: str) -> List[CacheEntryWithoutContent]:
-    responses = CachedResponse.objects(kind=kind, dataset=dataset).only(
-        "http_status", "error_code", "job_runner_version", "dataset_git_revision", "progress"
-    )
-    return [
-        {
-            "http_status": response.http_status,
-            "error_code": response.error_code,
-            "job_runner_version": response.job_runner_version,
-            "dataset_git_revision": response.dataset_git_revision,
-            "progress": response.progress,
-        }
-        for response in responses
-    ]
-
-
 class CacheEntry(CacheEntryWithoutContent):
     content: Mapping[str, Any]
 
@@ -351,6 +335,7 @@ class CacheReport(TypedDict):
     error_code: Optional[str]
     job_runner_version: Optional[int]
     dataset_git_revision: Optional[str]
+    progress: Optional[float]
 
 
 class CacheReportsPage(TypedDict):
@@ -413,6 +398,7 @@ def get_cache_reports(kind: str, cursor: Optional[str], limit: int) -> CacheRepo
             "error_code",
             "job_runner_version",
             "dataset_git_revision",
+            "progress",
         )
         .limit(limit)
     )
@@ -427,6 +413,7 @@ def get_cache_reports(kind: str, cursor: Optional[str], limit: int) -> CacheRepo
                 "error_code": object.error_code,
                 "job_runner_version": object.job_runner_version,
                 "dataset_git_revision": object.dataset_git_revision,
+                "progress": object.progress,
             }
             for object in objects
         ],
@@ -440,6 +427,34 @@ def get_outdated_split_full_names_for_step(kind: str, current_version: int) -> L
     )
     return [
         SplitFullName(dataset=response.dataset, config=response.config, split=response.split) for response in responses
+    ]
+
+
+def get_dataset_responses_without_content_for_kind(kind: str, dataset: str) -> List[CacheReport]:
+    responses = CachedResponse.objects(kind=kind, dataset=dataset).only(
+        "kind",
+        "dataset",
+        "config",
+        "split",
+        "http_status",
+        "error_code",
+        "job_runner_version",
+        "dataset_git_revision",
+        "progress",
+    )
+    return [
+        {
+            "kind": response.kind,
+            "dataset": response.dataset,
+            "config": response.config,
+            "split": response.split,
+            "http_status": response.http_status,
+            "error_code": response.error_code,
+            "job_runner_version": response.job_runner_version,
+            "dataset_git_revision": response.dataset_git_revision,
+            "progress": response.progress,
+        }
+        for response in responses
     ]
 
 
@@ -504,6 +519,7 @@ def get_cache_reports_with_content(kind: str, cursor: Optional[str], limit: int)
             "dataset_git_revision",
             "details",
             "updated_at",
+            "progress",
         )
         .limit(limit)
     )
@@ -521,6 +537,7 @@ def get_cache_reports_with_content(kind: str, cursor: Optional[str], limit: int)
                 "dataset_git_revision": object.dataset_git_revision,
                 "details": object.details,
                 "updated_at": object.updated_at,
+                "progress": object.progress,
             }
             for object in objects
         ],
--- a/libs/libcommon/tests/test_simple_cache.py
+++ b/libs/libcommon/tests/test_simple_cache.py
@@ -488,6 +488,7 @@ def test_get_cache_reports() -> None:
             "error_code": None,
             "job_runner_version": None,
             "dataset_git_revision": None,
+            "progress": None,
         },
         {
             "kind": kind,
@@ -498,6 +499,7 @@ def test_get_cache_reports() -> None:
             "error_code": error_code_b,
             "job_runner_version": job_runner_version_b,
             "dataset_git_revision": dataset_git_revision_b,
+            "progress": None,
         },
     ]
     assert response["next_cursor"] != ""
@@ -514,6 +516,7 @@ def test_get_cache_reports() -> None:
                 "error_code": error_code_c,
                 "job_runner_version": None,
                 "dataset_git_revision": None,
+                "progress": None,
             },
         ],
         "next_cursor": "",
@@ -537,6 +540,7 @@ def test_get_cache_reports() -> None:
             "dataset_git_revision": None,
             "details": {},
             "updated_at": REDACTED_DATE,
+            "progress": None,
         },
         {
             "kind": kind,
@@ -550,6 +554,7 @@ def test_get_cache_reports() -> None:
             "dataset_git_revision": dataset_git_revision_b,
             "details": details_b,
             "updated_at": REDACTED_DATE,
+            "progress": None,
         },
     ]
     assert response_with_content["next_cursor"] != ""
@@ -571,6 +576,7 @@ def test_get_cache_reports() -> None:
                 "dataset_git_revision": None,
                 "details": details_c,
                 "updated_at": REDACTED_DATE,
+                "progress": None,
             },
         ],
         "next_cursor": "",


severo:
Wow, big PR!

> Should there be any migration jobs? Technically it's a new job runner, it's not equal to the previous dataset-level one (results are not identical), and the old one doesn't exist anymore.

I think the best strategy in this PR would have been to:
- keep the old /parquet-and-dataset-info
- add the new config-parquet-and-info step,
- make all the other steps depend on config-parquet-and-info instead of /parquet-and-dataset-info

Then in a new PR, once the cache for the new step has been filled, deprecate the old one (remove the step and delete the entries in the cache).

Otherwise, the /parquet, /info and /size endpoints will be broken for every dataset until the job has been run again.

> Can I rename relevant environmental variables (like PARQUET_AND_DATASET_INFO_BLOCKED_DATASETS -> to PARQUET_AND_INFO_BLOCKED_DATASETS) or might it break anything?

No, good idea. We just have to take care to replace every occurrence.

polinaeterna:
@severo I've got back /parquet-and-dataset-info's code and tests for it (don't be scared by the number of additions in this PR it's actually mostly because of these duplicated tests) and changed env variables (hopefully everywhere), is there anything else that must be done? It just exists but no steps are dependent on it. I don't fully understand if it would work correctly. And I'm not sure if made all the required changes to [chart/templates/_envWorker.tpl](https://github.com/huggingface/datasets-server/pull/985/files#diff-f077348863c0aa5067040ac2e812f0d34f83b6049e8a9f1f53f23c982713ef7f).

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,910 @@
...
+    try:
+        refs = hf_api.list_repo_refs(repo_id=dataset, repo_type=DATASET_TYPE)
+        if all(ref.ref != target_revision for ref in refs.converts):
+            committer_hf_api.create_branch(
+                repo_id=dataset, branch=target_revision, repo_type=DATASET_TYPE, revision=source_revision
+            )
+    except RepositoryNotFoundError as err:
+        raise DatasetNotFoundError("The dataset does not exist on the Hub.") from err
+
+    # delete:
```
severo, line 804: Here, working at config level, instead of dataset level, means that we need to work in the config directory, not in the dataset repository (every config stores the parquet files in its own directory).
If we delete all the files here, we will never have more than one config at the same time. We have to let all the other config directories untouched, and only clear the config's own directory.

Which also leads to an issue I hadn't thought about before: which step should take care of deleting obsolete config directories? It's surely a rare use case, but if a commit in the repository deletes a config (for example, the dataset had 4 configs, and now it has only 3): which step should delete the directory of the removed config?

severo (changes requested):
Two comments:

1. we still need to add some logic to let multiple configs live in parallel in the same repository. With the current implementation, one config will delete the other ones.
2. I think I was mistaken (sorry! 🙃) when I said that we must not deprecated /parquet-and-dataset-info here. I forgot that the current entries for its dependent steps will not be deleted, so... There would be no damage in removing it here. Anyway, the PR is big, so it's OK too to leave the /parquet-and-dataset-info deprecation for another PR. I let you handle it as you prefer.

services/api/src/api/config.py
```diff
@@ -120,7 +120,10 @@ class EndpointConfig:
                 "config": ["/split-names-from-streaming", "/split-names-from-dataset-info"],
             },
             "/first-rows": {"split": ["split-first-rows-from-streaming", "split-first-rows-from-parquet"]},
-            "/parquet-and-dataset-info": {"dataset": ["/parquet-and-dataset-info"]},
+            "/parquet-and-dataset-info": {
+                "dataset": ["/parquet-and-dataset-info"],
```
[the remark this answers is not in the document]
polinaeterna: yes, if I understand well too :D

services/worker/src/worker/job_runners/config/parquet_and_info.py
severo: > hf api should care for this, no?

yes: see https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/parquet_and_dataset_info.py#L840 (https://huggingface.co/docs/huggingface_hub/package_reference/hf_api#huggingface_hub.HfApi.create_commit.parent_commit)

> Specifying parent_commit ensures the repo has not changed before committing the changes, and can be especially useful if the repo is updated / committed to concurrently.

severo: > alternatively this can be performed in /config-names, because when the set of config is changed, the processing graph is triggered, right? and in this case there shouldn't be any problems with parallel deletion I guess. so maybe this option is better.

I think it's better to let the "parquet-and-info" step in charge of the parquet files. BTW: storing them on the Hub is an implementation detail, we could have stored them elsewhere (on S3).

libs/libcommon/src/libcommon/constants.py
```diff
@@ -7,18 +7,19 @@
...
+PROCESSING_STEP_DATASET_INFO_VERSION = 2
+PROCESSING_STEP_DATASET_PARQUET_VERSION = 2
+PROCESSING_STEP_DATASET_SIZE_VERSION = 2
+PROCESSING_STEP_DATASET_SPLIT_NAMES_FROM_DATASET_INFO_VERSION = 2
 PROCESSING_STEP_DATASET_SPLIT_NAMES_FROM_STREAMING_VERSION = 1
 PROCESSING_STEP_PARQUET_AND_DATASET_INFO_VERSION = 1
-PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_PARQUET_VERSION = 1
+PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_PARQUET_VERSION = 3
+PROCESSING_STEP_CONFIG_PARQUET_AND_INFO_VERSION = 1
 PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_STREAMING_VERSION = 2
```
polinaeterna, line 21: I'm not sure if I should increment `PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_STREAMING_VERSION`. It's now (https://github.com/huggingface/datasets-server/pull/1021) dependent on `/split-names-from-dataset-info`  which in turn is dependent on `config-info` which in turn is dependent on `config-parquet-and-info` :) But it's also dependent on `/split-names-from-streaming` which is not connected to `config-parquet-and-info`.

services/worker/tests/job_runners/config/test_parquet_and_info.py
```diff
@@ -0,0 +1,552 @@
...
+    assert cached_response["http_status"] == HTTPStatus.OK
+    assert cached_response["error_code"] is None
+    assert cached_response["job_runner_version"] == job_runner.get_job_runner_version()
+    assert cached_response["dataset_git_revision"] is not None
+    content = cached_response["content"]
+    assert len(content["parquet_files"]) == 1
+    assert_content_is_equal(content, hub_datasets["public"]["parquet_and_info_response"])
+
+
+def test_compute_legacy_configs(
```
polinaeterna, line 145: this checks that configs are not deleted when other configs of a dataset are pushed and that outdated configs are deleted.

polinaeterna (commented):
@severo took your comments into account. I added a test that checks that compute for one config do not delete other configs but outdated configs are deleted.

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,955 @@
...
+        if all(ref.ref != target_revision for ref in refs.converts):
+            committer_hf_api.create_branch(
+                repo_id=dataset, branch=target_revision, repo_type=DATASET_TYPE, revision=source_revision
+            )
+    except RepositoryNotFoundError as err:
+        raise DatasetNotFoundError("The dataset does not exist on the Hub.") from err
+
+    target_dataset_info = hf_api.dataset_info(repo_id=dataset, revision=target_revision, files_metadata=False)
+    # - check if there are files for configs that do not exist anymore and delete them
+    # - get dataset's config names with `datasets` lib directly
```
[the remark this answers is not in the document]
polinaeterna: I first did [like this](https://github.com/huggingface/datasets-server/pull/985/commits/e8adfff58d5a1944fc3bd95edd532162f2490d1d) - try from cache first, then with the `datasets` lib if cache response is not successful. Then I thought that the cache might be outdated(???)... But it's actually not possible because we run `config-parquet-and-info` only after `/config-names`, right? So it should be taken only from cache?

libs/libcommon/src/libcommon/constants.py
polinaeterna: so I can restore all the previous versions?

severo: YES 😬

polinaeterna: [cef5ad2](https://github.com/huggingface/datasets-server/pull/985/commits/cef5ad2652f2daa1a263e9b81a69053b5f81aad8)

services/worker/src/worker/job_runners/config/parquet_and_info.py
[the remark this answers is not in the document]
polinaeterna: [af9df65](https://github.com/huggingface/datasets-server/pull/985/commits/af9df651fe4a2335c7afe54d613afefa479c48a9)

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,955 @@
...
+            file
+            for legacy_config in legacy_configs
+            for file in repo_parquet_files
+            if file.startswith(f"{legacy_config}/")
+        }
+        legacy_configs_delete_operations: List[CommitOperation] = [
+            CommitOperationDelete(path_in_repo=file) for file in legacy_configs_files_to_delete
+        ]
+        logging.debug(f"{legacy_configs_delete_operations=}")
+        committer_hf_api.create_commit(
```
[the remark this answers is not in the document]
polinaeterna: [af9df65](https://github.com/huggingface/datasets-server/pull/985/commits/af9df651fe4a2335c7afe54d613afefa479c48a9)

polinaeterna (commented):
i hope it's fine now :D 
the tests are a bit of a mess now but maybe it's okay to refactor them in another PR

services/worker/src/worker/job_runners/splits.py
severo, line 1: this file should not exist anymore, I think

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,970 @@
...
+        commit_message=commit_message,
+        parent_commit=target_dataset_info.sha,
+    )
+
+    # call the API again to get the list of parquet files
+    target_dataset_info = hf_api.dataset_info(repo_id=dataset, revision=target_revision, files_metadata=True)
+    repo_files = [
+        repo_file
+        for repo_file in target_dataset_info.siblings
+        if repo_file.rfilename.startswith(f"{config}/") and repo_file.rfilename.endswith(".parquet")
```
severo, line 893: Are we sure the directory is always `{config}/`?

If not, we should apply the same string normalization as done in the datasets library (help from other @huggingface/datasets is welcome)

It can be in another PR too, because the bug was already present in /parquet-and-dataset-info (but slightly different: here we will just ignore configs with weird names, while in /parquet-and-dataset-info: we include them, but with the `config` field in parquet file JSON being the config directory, not the config name)

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,970 @@
...
+    except RepositoryNotFoundError as err:
+        raise DatasetNotFoundError("The dataset does not exist on the Hub.") from err
+
+    target_dataset_info = hf_api.dataset_info(repo_id=dataset, revision=target_revision, files_metadata=False)
+    # - check if there are files for configs that do not exist anymore and delete them
+    # - get configs that exist in repo
+    repo_parquet_files = {f.rfilename for f in target_dataset_info.siblings if f.rfilename.endswith(".parquet")}
+    repo_config_names = {f.split("/")[0] for f in repo_parquet_files}
+    delete_operations: List[CommitOperation] = []
+    legacy_configs = repo_config_names - config_names
```
severo, line 852: here we rely on the fact that the directory names are exactly the config names. Is is always the case? Or can they be weird characters in a config name that needs to be normalized (and is normalized by the datasets library)? In which case we would have to compare the normalized configs (ie. the directory names)

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,970 @@
...
+        for local_file in glob.glob(f"{builder.cache_dir}**/*.parquet")
+    )
+
+    # create the target revision if it does not exist yet
+    try:
+        refs = hf_api.list_repo_refs(repo_id=dataset, repo_type=DATASET_TYPE)
+        if all(ref.ref != target_revision for ref in refs.converts):
+            committer_hf_api.create_branch(
+                repo_id=dataset, branch=target_revision, repo_type=DATASET_TYPE, revision=source_revision
+            )
```
severo, line 842: Here:
- we should delete all the files obtained by forking `main`
- or, better: replace `revision=source_revision` with `revision=initial_commit` (with `initial_commit` being created as in https://huggingface.co/docs/huggingface_hub/v0.13.4/en/package_reference/hf_api#huggingface_hub.HfApi.list_repo_commits.example) to start the branch with an empty directory.

Note that we didn't have the issue with /parquet-and-dataset-info because the commit deleted all the existing files everytime.

severo: Another option is to always clean the directory in the commit below, everytime:
- get the list of configs
- create the parquet files for the current config
- ignore the directories for the other configs
- ignore the .gitattributes files
- delete the rest

We might even have both:
- create an empty branch, instead of forking main
- clean the directory everytime (it would help handle the repositories that are currently in bad state)

severo (commented):
Still some comments. Having to update a git repository only partially is not very easy.

services/worker/src/worker/job_runners/config/parquet_and_info.py
polinaeterna: Hm from what I see in `ParquetFile.repo_files()` above which is copied from `/parquet-and-dataset-info`, it creates path which always starts from `"{config"/"`

`datasets.get_config_names` doesn't normalize configs and returns them just as they are specified by users or created automatically (for packaged builders like `csv`, `jsonl` etc.) so they can contain spaces for example and it might be a problem. Config names are only checked for [these symbols](https://github.com/huggingface/datasets/blob/1c4ec00511868bd881e84a6f7e0333648d833b8e/src/datasets/naming.py#L30): 
```python
INVALID_WINDOWS_CHARACTERS_IN_PATH = r"<>:/\|?*"
```
It would raise at dataset creation so we won't get configs with these symbols for sure. But they are not normalized after.
Am I right? cc @albertvillanova @lhoestq 

> here we will just ignore configs with weird names, while in /parquet-and-dataset-info: we include them, but with the config field in parquet file JSON being the config directory, not the config name

I think I don't understand what this mean... Where these differences are reflected in the code? I think I'm missing something

polinaeterna: > they can contain spaces for example and it might be a problem

 but it's normalized by HF API.

severo: > > here we will just ignore configs with weird names, while in /parquet-and-dataset-info: we include them, but with the config field in parquet file JSON being the config directory, not the config name
> I think I don't understand what this mean... Where these differences are reflected in the code? I think I'm missing something

In config-parquet-and-info, here: if a config has a weird name, it will not be included in the list of parquet files in the response

In /parquet-and-dataset-info, if a config has a weird name, it is included in the list of parquet files in the response, but the "config" field is the name of the directory, not necessarily the name of the config, if they are different due to weird characters

severo: OK, thanks for the research. So, we surely want to delegate the comparison and/or conversion to specialized functions (so that we can unit test, and have all the logic in one place).

services/worker/src/worker/job_runners/config/parquet_and_info.py
polinaeterna: awesome idea with initial commit! is it okay to update `huggingfcae_hub` dependency to use this method? it exists since version 0.13, we have `huggingface_hub = ">=0.12.0"` in `poetry.lock`

severo: > is it okay to update huggingfcae_hub

Sure!

polinaeterna:
@severo another round :D I've implemented file deletion strategy and cloning from initial commit. I haven't implemented any normalization of config names because `datasets` lib doesn't have any normalization and `huggingface_hub` only uses normpath to filenames which leads to conversion of tabs and new lines to spaces but we anyway don't treat config with spaces correctly as far as I remember and I think we don't have a lot of datasets with config names containing whitespaces. there is also [this validation function](https://github.com/huggingface/huggingface_hub/blob/e9dbca65b636e2c5047a6174c8fd2896bfef6b0e/src/huggingface_hub/_commit_api.py#L193-L206) in `huggingface_hub` but it basically fixes/raises for files with slashes which are not allowed by `datasets` anyway. So I'd suggest to address this in another issue, wdyt?

severo:
> So I'd suggest to address this in another issue, wdyt?

Yes, totally!

libs/libcommon/tests/test_processing_steps.py
```diff
@@ -65,11 +65,11 @@ def graph() -> ProcessingGraph:
 @pytest.mark.parametrize(
     "step_name,children,ancestors",
     [
-        ("/config-names", ["/split-names-from-streaming"], []),
+        ("/config-names", ["/split-names-from-streaming", "config-parquet-and-info"], []),
```
severo, line 68: nit: we miss an entry for `"config-parquet-and-info"`

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,965 @@
...
+        for local_file in glob.glob(f"{builder.cache_dir}**/*.parquet")
+    )
+
+    # create the target revision if it does not exist yet (clone from initial commit to avoid cloning all repo's files)
+    try:
+        refs = hf_api.list_repo_refs(repo_id=dataset, repo_type=DATASET_TYPE)
+        if all(ref.ref != target_revision for ref in refs.converts):
+            initial_commit = hf_api.list_repo_commits(repo_id=dataset, repo_type=DATASET_TYPE)[-1].commit_id
+            committer_hf_api.create_branch(
+                repo_id=dataset, branch=target_revision, repo_type=DATASET_TYPE, revision=initial_commit
```
severo, line 842: Note that this closes https://github.com/huggingface/datasets-server/issues/846!

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -0,0 +1,965 @@
...
+    # - get repo parquet files
+    all_repo_files: Set[str] = {f.rfilename for f in target_dataset_info.siblings}
+    repo_parquet_files: Set[str] = {file for file in all_repo_files if file.endswith(".parquet")}
+    # - get parquet files for current config
+    config_files_to_add: Dict[str, str] = {
+        parquet_file.repo_file(): parquet_file.local_file for parquet_file in local_parquet_files
+    }
+    # - get files that will be preserved in repo: files belonging to other configs and .gitattributes
+    files_to_ignore: Set[str] = {
+        file for config in config_names for file in repo_parquet_files if file.startswith(f"{config}/")
```
severo, line 857: I think that we should exclude the current config, because if a previous version of parquet files exists in the current config directory, but with a different set of splits, we could have dangling old files among the new files we will upload.

severo: so, something like:

```suggestion
        file for other_config in config_names for file in repo_parquet_files if other_config != config and file.startswith(f"{config}/")
```

(not tested, and the style will surely make the CI fail)

severo (commented):
Sorry, still some comments. I think it's the last iteration!

severo (approved):
Thanks a lot for this PR!

PR #986, opened 2023-03-24
feat: 🎸 remove bigcode/the-stack from the blocked datasets

severo:
because the blocked datasets are taken into account before the supported datasets, it was still ignored.


severo:
Nothing difficult, so I merge

PR #987, opened 2023-03-24
[docs] Process Parquet files

stevhliu:
This PR adds a guide for how to process Parquet files:

- Eager dataframes with `pd/pl.read_parquet`
- Lazy dataframes with `pl.scan_parquet`
- Read and query with DuckDB


lhoestq (commented):
Nice ! This is super useful :)

Just two comments on the section title:

1. Do you think we can make it clearer in the title that you can query any dataset ? IMO the current title "Process Parquet files" may suggest that it can only be useful if you have parquet files already. Maybe mention that the parquet files are "published", or use a verb that suggest that the data is remote e.g. "access" in addition to "process".
2. (nit) Both polars and duckdb examples run "queries" on parquet files. Maybe say "query" instead of "process" ? I think "process" can be mentioned for data manipulation that involve user defined functions (UDF) - with Spark for example

mariosasko (commented):
Nice!

IMO it makes sense to remove the `read_polars`/`read_pandas` examples from the `List Parquet files` and instead provide a link to this guide now that we have it.

And maybe we can mention at the end of the guide that Parquet files can also be queried in the browser with DuckDB WASM.

PS: doc-builder generates (empty) CURL version of the examples, would be nice if this can be removed

stevhliu:
> PS: doc-builder generates (empty) CURL version of the examples, would be nice if this can be removed

Oh! I didn't think the doc-builder would generate `cURL` examples even though I didn't specify them in the `<inferencesnippet>`. @mishig25 is there a way to remove this?

docs/source/parquet_process.mdx
```diff
@@ -0,0 +1,235 @@
+# Query datasets from Datasets Server
+
+Datasets Server automatically converts and publishes datasets on the Hub as Parquet files.[Parquet](https://parquet.apache.org/docs/) files are column-based and they really shine when you're working with big data. There are several ways you can work with Parquet files, and this guide will show you how to:
```
mariosasko, line 3:
```suggestion
Datasets Server automatically converts and publishes datasets on the Hub as Parquet files. [Parquet](https://parquet.apache.org/docs/) files are column-based and they really shine when you're working with big data. There are several ways you can work with Parquet files, and this guide will show you how to:
```

docs/source/parquet_process.mdx
```diff
@@ -0,0 +1,235 @@
...
+- read and query Parquet files with Pandas and Polars
+- access, read and query Parquet files with DuckDB and DuckDB-Wasm
+
+## Polars 
+
+[Polars](https://pola-rs.github.io/polars-book/user-guide/introduction.html) is a fast DataFrame library written in Rust with Arrow as its foundation.
+
+<Tip>
+
+💡 Learn more about how to get the dataset URLs in the [List Parquet files](parquet) guide.
```
mariosasko, line 14: Nit: Would rename the "List Parquet files" guide from `parquet` to `parquet_list`

stevhliu:
@mariosasko, empty cURL example fixed in https://github.com/huggingface/doc-builder/pull/365!

PR #875, opened 2023-03-01
Random rows

severo:
Add a new endpoint: `/rows`

Parameters: `dataset`, `config`, `split`, `offset` and `limit`. 

It returns a list of `limit` rows of the split, from row idx=`offset`, by fetching them from the parquet files published on the Hub.

Replaces #687

PR #988, opened 2023-03-24
Split first rows from parquet new Job Runner

AndreaFrancis:
Final part of https://github.com/huggingface/datasets-server/issues/755
Based on https://github.com/huggingface/datasets-server/pull/875 for parquet reading logic


services/worker/pyproject.toml
```diff
@@ -9,12 +9,12 @@ license = "Apache-2.0"
 Pillow = "^9.4.0"
 PyICU = "^2.10.2"
 aiohttp = "^3.8.4"
-apache-beam = "2.41.0" # ^2 gives a InvalidWheelName error because it tries to install 2.42 that has not been released...
```
severo, line 12: We don't need apache-beam for the datasets-server, but some dataset scripts require it. It would be a regression for these... I understand that we remove it here to allow upgrading pyarrow to a recent version.

It's hard to have an idea of how many datasets require apache-beam. Do we have some script that analyzes the content of all the dataset scripts on the Hub @huggingface/datasets?

Another solution would be to have apache-beam work with a recent version of pyarrow. But, as https://github.com/huggingface/datasets/issues/5430 is not closed yet, I imagine that it's not currently possible.

services/worker/src/worker/asset.py
```diff
@@ -58,6 +60,34 @@ class AudioSource(TypedDict):
...
+    assets_directory: StrPath,
+) -> List[AudioSource]:
+    wav_filename = f"{filename_base}.wav"
+    mp3_filename = f"{filename_base}.mp3"
+    dir_path, url_dir_path = create_asset_dir(
+        dataset=dataset, config=config, split=split, row_idx=row_idx, column=column, assets_directory=assets_directory
+    )
+    wav_file_path = dir_path / wav_filename
+    mp3_file_path = dir_path / mp3_filename
+    array, sampling_rate = soundfile.read(io.BytesIO(array))
```
severo, line 81: this is the only difference between this new method and create_audio_files. I think we should have only one method supporting various types, or at least this method should just extract the array and sampling rate, and call the other method.

services/worker/src/worker/utils.py
```diff
@@ -35,3 +38,169 @@ class PreviousJob(TypedDict):
     dataset: str
     config: Optional[str]
     split: Optional[str]
+
```
severo, line 41: Good. At one point, all this code could go to libcommon, so that we can also use it in service/api for /rows. I think @lhoestq was working on it, right?

lhoestq: Yea I'll open a PR for image+audio support in /rows, and this requires the api to have `datasets`, `Pillow` and these utilities to be in libcommon ^^

services/worker/pyproject.toml
lhoestq: You can ignore apache beam, only wikipedia and natural_questions need it. It would be easier to switch them to non-beam datasets that supporting beam datasets x)

severo: OK. So: 
- https://huggingface.co/datasets/wikipedia
- https://huggingface.co/datasets/natural_questions

Anyway: none of them is currently working. I will open a discussion on them to remove dependency to apache_beam.

severo: - https://huggingface.co/datasets/wikipedia/discussions/9
- https://huggingface.co/datasets/natural_questions/discussions/4

AndreaFrancis:
Not sure why pip audit fails with worker project, it is now different from other folders that use `pymongo` and `mongoengine`
```
ERROR:pip_audit._virtual_env:internal pip failure: ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    pymongo<5.0,>=3.4 from https://files.pythonhosted.org/packages/38/68/928d7ce22719cfa255fb973b34aed6f04ac3ea89049ce69e3b092c30a60f/pymongo-4.3.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (from mongoengine==0.24.2->-r /tmp/tmpmnbps2mm (line 1032))

```

severo:
> Not sure why pip audit fails with worker project, it is now different from other folders that use pymongo and mongoengine

```
ERROR:pip_audit._virtual_env:internal pip failure: ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    pymongo<5.0,>=3.4 from https://files.pythonhosted.org/packages/38/68/928d7ce22719cfa255fb973b34aed6f04ac3ea89049ce69e3b092c30a60f/pymongo-4.3.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (from mongoengine==0.24.2->-r /tmp/tmpmnbps2mm (line 1032))

```

In https://github.com/huggingface/datasets-server/pull/988/commits/12289d800fe5d991e2ffec3310c03eb68f1b7635, you updated all the dependencies of the project, including some that were not related to the matter of this PR. It's better to do it in a separate PR, where we update the dependencies and ensure nothing has broken.

In this case, you updated the dev dependency `pip-audit` to https://github.com/pypa/pip-audit/releases/tag/v2.5.3, which introduced the issue (I tried with 2.5.2, and it works).

So, my recommendation for this PR would be to:
- copy/paste the version of `main` of files pypoetry.toml and poetry.lock
- run `make install`
- then `poetry remove apache-beam`
- then restore the new version (https://github.com/huggingface/datasets-server/pull/988/files#diff-5def9e9ad886d364c3edccbe42f21d5aafd1b936d6756a71fa9be4f86ecf573e) of pyproject.toml and call "update" for all the updated dependencies, one by one:
  - `poetry update datasets`
  - `poetry update hffs`
  - `poetry update pyarrow`

and I think that your poetry.lock should be good (in particular, you should be using pip-audit v2.4.14)

And then, in another PR, we could upgrade the dependencies, one by one, and checking that we don't break anything.

services/worker/src/worker/asset.py
AndreaFrancis: Agree, I fixed it on https://github.com/huggingface/datasets-server/pull/988/commits/b2f428b520bf4799497543ca3692920493529cd1  thanks!.

services/worker/tests/job_runners/split/test_first_rows_from_parquet.py
```diff
@@ -0,0 +1,162 @@
...
+            with pytest.raises(CustomError) as error_info:
+                job_runner.compute()
+            assert error_info.value.code == error_code
+        else:
+            response = job_runner.compute().content
+            assert get_json_size(response) <= rows_max_bytes
+            assert response
+            assert response["rows"]
+            assert response["features"]
+            assert len(response["rows"]) == 3  # testing file has 3 rows see config/dataset-split.parquet file
```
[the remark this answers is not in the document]
AndreaFrancis: I fixed it on https://github.com/huggingface/datasets-server/pull/988/commits/b2f428b520bf4799497543ca3692920493529cd1

AndreaFrancis:
> And then, in another PR, we could upgrade the dependencies, one by one, and checking that we don't break anything.

I introduced dependencies for this PR on https://github.com/huggingface/datasets-server/pull/996

services/worker/pyproject.toml
albertvillanova: I have created a discussion in "wiki40b" as well:
- https://huggingface.co/datasets/wiki40b/discussions/2

severo (approved):
Very nice. This, along with https://github.com/huggingface/datasets-server/pull/995, should help reduce a lot the load on the workers!

AndreaFrancis:
I will introduce same logic as https://github.com/huggingface/datasets-server/pull/995 for first rows job runners

severo:
You're right, #995 was for split-names. Let's do it for first-rows now!

PR #990, opened 2023-03-27
Support splits with dots

lhoestq:
The split regex is defined at https://github.com/huggingface/datasets/blob/4db8e33eb9cf6cd4453cdfa246c065e0eedf170c/src/datasets/naming.py#L28, I updated the one used to parque parquet file paths to use it


severo (approved):
perfect

PR #1000, opened 2023-03-30
Update job_runner.py

### LINKED ISSUE #854: Turn `get_new_splits` into an abstract method
See https://github.com/huggingface/datasets-server/pull/839#issuecomment-1443264927. cc @AndreaFrancis

Aniket1299:
This issue resolves #854 by turning get_new_splits into an abstract method.

### DIFF
--- a/services/worker/src/worker/job_runner.py
+++ b/services/worker/src/worker/job_runner.py
@@ -411,6 +411,7 @@ class JobRunner(ABC):
         pass
 
     # should be overridden if the job has children jobs of type "split"
+    @abstractmethod
     def get_new_splits(self, content: Mapping[str, Any]) -> set[SplitFullName]:
         """Get the set of new splits, from the content created by the compute.
 


services/worker/src/worker/job_runner.py
```diff
@@ -411,6 +411,7 @@ def compute(self) -> JobResult:
         pass
 
     # should be overridden if the job has children jobs of type "split"
```
AndreaFrancis, line 413:
```suggestion

```
I think the comment is redundant here, since it is going to be an abstract method.

AndreaFrancis:
Thank you very much for the contribution @Aniket1299, here some details you could add fix some of the failed checks:

- We have here a Dummy job runner, https://github.com/huggingface/datasets-server/blob/main/services/worker/tests/job_runners/test__datasets_based_worker.py#L25 you could here add a dummy result for the new abstract method 
- Same for this one https://github.com/huggingface/datasets-server/blob/main/services/worker/tests/test_loop.py#L15

services/worker/src/worker/job_runner.py
AndreaFrancis: Also, you could pass the result in line 425 instead of returning a set()

PR #1005, opened 2023-03-31
Raise error when the viewer is disabled

### LINKED ISSUE #1004: Fail quickly when the dataset card contains `viewer: false`
The Hub shows a specific message, instead of the dataset viewer, when the dataset card contains the metadata `viewer: false`.

I think that we should take it into account when checking if a dataset is supported, and fail fast (adding this case in https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/dataset.py#L165, along to checking if the dataset is private)

albertvillanova:
This PR raises error when the viewer is disabled, thus making the server fail quickly instead of relying on moon-landing.

Fix #1004.

### DIFF
--- a/docs/source/valid.mdx
+++ b/docs/source/valid.mdx
@@ -128,6 +128,12 @@ If a dataset is not valid, then the response looks like:
 {"valid": false}
 ```
 
+Some cases where a dataset is not valid are:
+- the dataset viewer is disabled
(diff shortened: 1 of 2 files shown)


AndreaFrancis (approved):

severo (approved):
Excellent, thanks!

Issue #994, opened 2023-03-28
Job Migration should not be dependent of hf-token secret


rtrompier:
### Description

Job migration should not require hf token secret :

https://github.com/huggingface/datasets-server/blob/main/chart/templates/jobs/mongodb-migration/job.yaml#L23
https://github.com/huggingface/datasets-server/blob/main/chart/templates/_envCommon.tpl#L9

severo:
fixed with https://github.com/huggingface/datasets-server/pull/1006

rtrompier:
Many thanks

PR #1006, opened 2023-03-31
feat: 🎸 move COMMON_LOG_LEVEL to LOG_LEVEL

severo:
This allows to define LOG_LEVEL without having to define other "common" variables like HF_ENDPOINT and HF_TOKEN

Required by https://github.com/huggingface/datasets-server/issues/994

### DIFF
--- a/libs/libcommon/tests/test_config.py
+++ b/libs/libcommon/tests/test_config.py
@@ -1,9 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
 
-from libcommon.config import CommonConfig
+from libcommon.config import LogConfig
 
 
 def test_common_config() -> None:
-    common_config = CommonConfig()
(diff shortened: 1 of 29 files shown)


severo:
After that PR, the "common" parameters only include hf_endpoint and hf_token. Maybe we should rename them too ("common" is like "others", "utils", etc: it's better if they don't exist and if we use precise names)

libs/libcommon/tests/test_config.py
```diff
@@ -1,9 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
 
-from libcommon.config import CommonConfig
+from libcommon.config import LogConfig
 
 
 def test_common_config() -> None:
```
AndreaFrancis, line 7: Maybe
```suggestion
def test_log_config() -> None:
```

AndreaFrancis (approved):
Great!

Issue #1014, opened 2023-04-04
Allow a step to depend on (be triggered by) several other steps


severo:
This is needed for https://github.com/huggingface/datasets-server/issues/891.

And by the way, various steps already implicitly depend on various steps (they read the value of the cache for various steps). But their refresh is only triggered by one of them.

severo:
Proposal

- instead of depending on one (or zero) step, all steps now have a list of upstream steps (length: 0, 1 or more)
- an upstream step means a required step which provides some data (from the cache)
- the update of an upstream step will trigger an update of the step, directly or indirectly:
  - directly in most cases
  - indirectly: if step B and step C are upstream steps for step A, and step C is an upstream step for step B, therefore a job will be created for step A only when step B is updated, not when step C is updated

severo:
TODO:

- [ ] depend on [networkx](https://networkx.org/documentation) to manage the graph
- [ ] add the ability to depend on multiple steps

severo:
Follow-up:
- [ ] add `/split-names-from-dataset-info` to the list of required steps of `split-first-rows-from-streaming` (and try to read its value in the job runner)
- [ ] replace `dataset-split-names-from-dataset-info` and `dataset-split-names-from-streaming` with a single `dataset-split-names` that requires both `/split-names-from-streaming` and `/split-names-from-dataset-info`

severo:
nearly done, when https://github.com/huggingface/datasets-server/pull/1141 will be merged

severo:
Done!

PR #1016, opened 2023-04-04
Use networkx to compute the processing graph

severo:
We now rely on a specialized library to compute the graph. It will help when we allow multiple parents per step.

See https://github.com/huggingface/datasets-server/issues/1014


AndreaFrancis (approved):
Great!

PR #1017, opened 2023-04-04
Moving /backfill to a k8s job

AndreaFrancis:
First step of https://github.com/huggingface/datasets-server/issues/740
New k8s job under /jobs/cache_maintenance folder


AndreaFrancis:
I did a test running against an prod backup (from two weeks ago) and it took about 7-8 minutes running.

severo:
> I did a test running against an prod backup (from two weeks ago) and it took about 7-8 minutes running.

And did you note how many new jobs were created? It must be tens of thousands (with priority=LOW)

severo:
>  Not sure if name cache_monitor is the better thought, maybe cache_fixer, cache_doctor, I am open to new ideas

I understand that you preferred these names more than cache_backfill because you plan to extend the capabilities of the job to other actions on the cache, right?
To me: it's simpler mentally to have one job per action, and we could let the job with the name cache_backfill. That way, we will be able, later, to launch the different jobs (if there are multiple) with cron at different intervals.

chart/templates/jobs/cache-monitor/job.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+{{- if .Values.images.jobs.cacheMonitor }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels: {{ include "labels.cacheMonitor" . | nindent 4 }}
+  name: "{{ include "name" . }}-job-cache-monitor"
+  namespace: {{ .Release.Namespace }}
+spec:
+  ttlSecondsAfterFinished: 300
```
severo, line 12: we can let it live longer (one day?) after it finished, in order to be able to look at the logs

severo (approved):
Just a detail about the HF token

AndreaFrancis:
> And did you note how many new jobs were created? It must be tens of thousands (with priority=LOW)

It appended 3 records for each dataset (27537 in the backup) and a total of 82611 new Jobs with low priority.

chart/templates/jobs/cache-monitor/job.yaml
AndreaFrancis: Sure

chart/templates/jobs/cache-monitor/_container.tpl
```diff
@@ -0,0 +1,15 @@
...
+# Copyright 2022 The HuggingFace Authors.
+
+{{- define "containerCacheMonitor" -}}
+- name: "{{ include "name" . }}-cache-monitor"
+  image: {{ include "jobs.cacheMonitor.image" . }}
+  imagePullPolicy: {{ .Values.images.pullPolicy }}
+  env:
+  {{ include "envLog" . | nindent 2 }}
+  {{ include "envQueue" . | nindent 2 }}
+  {{ include "envCommon" . | nindent 2 }}
```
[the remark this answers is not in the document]
rtrompier: IMO `post-install` and `post-upgrade` should work.
HF Token secrets a created / updated during the `install` and `upgrade` process.

severo:
> total of 82611 new Jobs with low priority

which in turn created other jobs... OK

AndreaFrancis:
> I understand that you preferred these names more than cache_backfill because you plan to extend the capabilities of the job to other actions on the cache, right?

Yes, I think I will rename to cache_maintenance 
> To me: it's simpler mentally to have one job per action, and we could let the job with the name cache_backfill. That way, we will be able, later, to launch the different jobs (if there are multiple) with cron at different intervals.

In case of cron, maybe we could add another Job def to the helm chart but using same code with different time intervals and action parameters. 
> (on the other hand: creating a new Job is a lot of duplicated code... So maybe having one Job, with a parameter/var env to chose the action, is not a bad idea after all stuck_out_tongue )
Yes, it is a lot of code,  I will go first with the approach of creating a single job with only  backfill action supported. I would also like to move to this forlder the cache_refresh code (since it is also simple and is related to cache monitoring/fixing tasks).

My idea is:
- To have an ACTION env variable that will indicate the action to perform, initially only backfill is supported, but it could be None (In this case no action will be executed).
- By default ACTION will be None for dev and prod env, in case we want to turn on backfill, we will have to send a PR to change its value. Initially, I dont think backfill is something to run periodically since it would append a lot of jobs but after we support only append to  "complete" missing cache, it will have more sense to run in a time basis.
- In case we want a cron in the future, we could add a new K8s job definition but using the same folder cache_maintenance but with different ACTION value and specifying the interval.

chart/templates/jobs/cache-maintenance/job.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+{{- if .Values.images.jobs.cacheMaintenance }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels: {{ include "labels.cacheMaintenance" . | nindent 4 }}
+  name: "{{ include "name" . }}-job-cache-maintenance"
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-upgrade
```
rtrompier, line 12: not needed after an install ?

rtrompier (approved):
LGTM for helm part

chart/templates/jobs/cache-maintenance/job.yaml
severo: Yes, I think that we could run it after an install to populate the cache database. So, post-install too, maybe.

Note that a next step is to change it to a cron job, to have more control on the time it is run.

.github/workflows/cd.yml
```diff
@@ -25,6 +25,8 @@ jobs:
             project: mongodb_migration
           - directory: jobs
             project: cache_refresh
```
severo, line 27: Will cache_refresh later be merged inside cache_maintenance?

chart/env/prod.yaml
```diff
@@ -101,6 +106,16 @@ mongodbMigration:
     limits:
       cpu: 1
 
+# --- jobs (post-upgrade hooks) ---
+
+cacheMaintenance:
+  # action: "backfill" -- Set this value according to required action for maintenance
```
severo, line 112: > Set this value according to required action for maintenance

where?

Shouldn't it be set here, and if we need another action to be run, change it here too?

chart/values.yaml
```diff
@@ -195,6 +200,16 @@ cacheRefresh:
       cpu: 0
   tolerations: []
 
+cacheMaintenance:
+  # action: "backfill" -- Set this value according to required action for maintenance
```
severo, line 204: maybe set the default action here (since the job will use one action anyway, I think it's better to be explicit here about what is the default action)

severo (approved):

severo:
Also: you need to run `poetry update libcommon` inside the job/cache_maintenance directory to install the missing dependency (networkx)

chart/templates/jobs/cache-maintenance/job.yaml
AndreaFrancis: I fixed it on https://github.com/huggingface/datasets-server/pull/1017/commits/e67005f028c2117e44c80a88f7d170ab664afaef thanks!

chart/env/prod.yaml
AndreaFrancis: I fixed it on https://github.com/huggingface/datasets-server/pull/1017/commits/e67005f028c2117e44c80a88f7d170ab664afaef

chart/values.yaml
AndreaFrancis: Thanks, I fixed it on https://github.com/huggingface/datasets-server/pull/1017/commits/e67005f028c2117e44c80a88f7d170ab664afaef

PR #1019, opened 2023-04-05
Allow multiple parent steps

severo:
A step can now have multiple parents, ie multiple steps can trigger it. Note: the ProcessingStep API has changed: parent is removed (since it's not used in the code) and requires is now a list of strings.

Related to https://github.com/huggingface/datasets-server/issues/1014


AndreaFrancis (approved):
Thanks! this will help a lot to parallel job runners

PR #1025, opened 2023-04-06
Create step `dataset-split-names`

severo:
See https://github.com/huggingface/datasets-server/issues/1014#issuecomment-1497597525

It aims at replacing `dataset-split-names-from-dataset-info` and `dataset-split-names-from-streaming`, now that one step can have two parents (or more)

Note the new method:

https://github.com/huggingface/datasets-server/blob/4e776c95799e23a6874ca1ebce0a1cfd7cb29e39/libs/libcommon/src/libcommon/simple_cache.py#L241-L267

It might be used in other places, for example in `service/api` instead of:

https://github.com/huggingface/datasets-server/blob/e89a249ac3dc7548517f8c516089b9f961258565/services/api/src/api/routes/endpoint.py#L52


services/worker/src/worker/job_runners/dataset/split_names.py
```diff
@@ -0,0 +1,176 @@
...
+            ]
+        elif best_response.kind == "dataset-info":
+            config_names = list(best_response.response["content"]["dataset-info"].keys())
+        else:
+            raise PreviousStepFormatError(
+                "Previous step 'config-names' or 'dataset-info' did not return the expected content."
+            )
+    except DoesNotExist as e:
+        raise DatasetNotFoundError(
+            "No response found in previous step for this dataset: 'config-names' or 'dataset-info'.", e
```
AndreaFrancis, line 93:
```suggestion
            "No response found in previous step for this dataset: '/config-names' or 'dataset-info'.", e
```

services/worker/src/worker/job_runners/dataset/split_names.py
```diff
@@ -0,0 +1,176 @@
...
+        best_response = get_best_response(["/config-names", "dataset-info"], dataset)
+        if best_response.kind == "/config-names":
+            config_names = [
+                config_name_item["config"] for config_name_item in best_response.response["content"]["config_names"]
+            ]
+        elif best_response.kind == "dataset-info":
+            config_names = list(best_response.response["content"]["dataset-info"].keys())
+        else:
+            raise PreviousStepFormatError(
+                "Previous step 'config-names' or 'dataset-info' did not return the expected content."
```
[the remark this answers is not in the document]
AndreaFrancis: But not sure if it is worth to change since we will change the nomenclature in https://github.com/huggingface/datasets-server/issues/867

services/worker/src/worker/job_runners/dataset/split_names.py
```diff
@@ -0,0 +1,176 @@
...
+            )
+    except DoesNotExist as e:
+        raise DatasetNotFoundError(
+            "No response found in previous step for this dataset: 'config-names' or 'dataset-info'.", e
+        ) from e
+    except KeyError as e:
+        raise PreviousStepFormatError(
+            "Previous steps 'config-names' or 'dataset-info' did not return the expected content.", e
```
AndreaFrancis, line 97:
```suggestion
            "Previous steps '/config-names' or 'dataset-info' did not return the expected content.", e
```

services/worker/src/worker/job_runners/dataset/split_names.py
```diff
@@ -0,0 +1,176 @@
...
+        ) from e
+    if any(not isinstance(config_name, str) for config_name in config_names):
+        raise PreviousStepFormatError(
+            "Previous steps 'config-names' or 'dataset-info' did not return a list of config names."
```
AndreaFrancis, line 101:
```suggestion
            "Previous steps '/config-names' or 'dataset-info' did not return a list of config names."
```

AndreaFrancis (approved):
Great! 
I just put couple of minor comments more related to docs and logs but maybe some of them are not valid.

PR #1026, opened 2023-04-06
Random access: image and audio files support

lhoestq:
I extended the `/rows` API endpoint to also return image and audio files.

Similarly to the `first-rows` job, it writes the image and audio files to the NFS. But instead of using the `assets/` directory, it uses `cached-assets/`. The reason is that this directory is a cache: the image and audio files may be removed from time to time to free disk space.

I added a new config `CachedAssetsConfig` to define the cached assets URLs, storage directory and the cache parameters (e.g. the number of most recent rows to keep)

Because the API service now requires `datasets`, `Pillow` etc. I added those dependencies in libcommon and factorized the viewer assets and features code in libcommon in a `viewer_utils` submodule.

Finally I added tests for the regular `/rows` for both text and image cases.

## How it works

1. there's a request to an image dataset to `/rows`
2. (optional - in 5% of requests) the cached assets directory is cleaned to save disk space using a simple (?) heuristic:
    * it takes a big sample of rows from the cache using glob
    * it keeps the most recent ones (max 200)
    * it keeps the rows below a certain index (max 100)
    * it discards the rest
3. the arrow data are retrieved from the parquet files using the row indexer
4. the cached assets and URLs are created on disk using the same mechanism as the `first-rows` job
5. the response is returned and contains URLs to the cached assets

## Other details

- I updated `pip-audit` everywhere to fix issues with extras/duplicated requirements
- I introduced the `MockFileSystem` (originally from `datasets`) that can be used to test the parquet data indexing for random access


lhoestq:
This is ready for review ! The PR may seem big because I moved some files from worker to libcommon, but don't be scared ^^

libs/libcommon/src/libcommon/config.py
```diff
@@ -29,21 +29,58 @@
...
 ASSETS_BASE_URL = "assets"
-ASSETS_STORE_DIRECTORY = None
+ASSETS_STORAGE_DIRECTORY = None
 
 
 @dataclass(frozen=True)
 class AssetsConfig:
     base_url: str = ASSETS_BASE_URL
-    storage_directory: Optional[str] = ASSETS_STORE_DIRECTORY
+    storage_directory: Optional[str] = ASSETS_STORAGE_DIRECTORY
```
severo, line 38: the changes in environment variables (renaming, addition) must be reflected in the docker compose files, and in the Helm chart

libs/libcommon/src/libcommon/config.py
```diff
@@ -29,21 +29,58 @@
...
+                storage_directory=env.str(name="STORAGE_DIRECTORY", default=ASSETS_STORAGE_DIRECTORY),
+            )
+
+
+CACHED_ASSETS_BASE_URL = "cached-assets"
+CACHED_ASSETS_STORAGE_DIRECTORY = None
+CACHED_ASSETS_CLEAN_CACHE_PROBA = 0.05
+CACHED_ASSETS_KEEP_ROWS_BELOW_INDEX = 100
+CACHED_ASSETS_KEEP_N_MOST_RECENT_ROWS = 200
+CACHED_ASSETS_MAX_CLEAN_SAMPLE_SIZE = 10_000
```
severo, line 55: nit: when we have a size, it's a best practice to end with the unit. bytes?

libs/libcommon/src/libcommon/config.py
```diff
@@ -29,21 +29,58 @@
...
                 base_url=env.str(name="BASE_URL", default=ASSETS_BASE_URL),
-                storage_directory=env.str(name="STORAGE_DIRECTORY", default=ASSETS_STORE_DIRECTORY),
+                storage_directory=env.str(name="STORAGE_DIRECTORY", default=ASSETS_STORAGE_DIRECTORY),
+            )
+
+
+CACHED_ASSETS_BASE_URL = "cached-assets"
+CACHED_ASSETS_STORAGE_DIRECTORY = None
+CACHED_ASSETS_CLEAN_CACHE_PROBA = 0.05
+CACHED_ASSETS_KEEP_ROWS_BELOW_INDEX = 100
```
severo, line 53: nit: ideally the unit (rows) would be at the end

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -24,6 +27,30 @@ def create_asset_dir(
...
+    length: int,
+    assets_directory: StrPath,
+) -> None:
+    row_dirs_path = Path(assets_directory).resolve() / dataset / DATASET_SEPARATOR / config / split
+    for row_idx in range(offset, offset + length):
+        if (row_dirs_path / str(row_idx)).is_dir():
+            # update the directory's last modified date
+            if (row_dirs_path / str(row_idx) / DATASETS_SERVER_MDATE_FILENAME).is_file():
+                (row_dirs_path / str(row_idx) / DATASETS_SERVER_MDATE_FILENAME).unlink()
+            (row_dirs_path / str(row_idx) / DATASETS_SERVER_MDATE_FILENAME).touch()
```
severo, line 51: is it necessary to delete the file if it exists? afaiu, touch is meant to update the modification date

https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

> If the file already exists, the function succeeds if exist_ok is true (and its modification time is updated to the current time), otherwise [FileExistsError](https://docs.python.org/3/library/exceptions.html#FileExistsError) is raised.

libs/libcommon/src/libcommon/viewer_utils/features.py
```diff
@@ -76,8 +77,10 @@ def image(
                 image=value,
                 assets_base_url=assets_base_url,
                 assets_directory=assets_directory,
+                overwrite=overwrite,
             )
         except OSError:
+            raise
```
severo, line 83: why are we changing the logic here? the idea was to try the formats in order.

services/api/Dockerfile
```diff
@@ -17,6 +17,7 @@ ENV PYTHONFAULTHANDLER=1 \
 # System deps:
 RUN apt-get update \
     && apt-get install -y build-essential unzip wget \
+    libicu-dev ffmpeg libavcodec-extra libsndfile1 llvm pkg-config \
```
severo, line 20: now that every service and job depends on libcommon, which in turn requires datasets with the audio and image extras, do we need to install these packages in all the docker images? Is the poetry install working as expected without the packages?

severo: Also: do we need all these packages?

severo (commented):
It requires a rebase, due to the huggingface_hub's upgrade

libs/libcommon/src/libcommon/viewer_utils/features.py
lhoestq: oops it was for debugging

services/api/Dockerfile
lhoestq: Yes we need those,I had errors if they were not all installed

libs/libcommon/src/libcommon/viewer_utils/asset.py
lhoestq: `touch` is not enough to modify the last modified time of **the parent directory** when the file already exists.

libs/libcommon/src/libcommon/config.py
lhoestq: Done in the docker compose and in `_envAssets.tpl`

libs/libcommon/src/libcommon/viewer_utils/asset.py
severo: hmmm ok

services/api/Dockerfile
severo: OK. And thus: do we need to add those lines in the other dockerfiles?

services/api/src/api/routes/rows.py
```diff
@@ -256,35 +278,151 @@ class RowItem(TypedDict):
...
+        return on_error
+
+
+def clean_cached_assets(
+    dataset: str,
+    cached_assets_directory: StrPath,
+    keep_rows_below_index: int,
+    keep_n_most_recent_rows: int,
+    max_clean_sample_size: int,
+) -> None:
```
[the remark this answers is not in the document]
severo: Nice explanation, thanks

libs/libcommon/src/libcommon/config.py
severo: it seems like you didn't push your commit

lhoestq: To be more clear: I only renamed the python variable. The env variable was already `ASSETS_STORAGE_DIRECTORY`

services/api/Dockerfile
severo: OK

libs/libcommon/src/libcommon/config.py
severo: OK!

Still: the new environment variables are not in the chart

chart/templates/_initContainerCachedAssets.tpl
```diff
@@ -0,0 +1,21 @@
...
+
+{{- define "initContainerCachedAssets" -}}
+- name: prepare-assets
+  image: ubuntu:focal
+  imagePullPolicy: {{ .Values.images.pullPolicy }}
+  command: ["/bin/sh", "-c"]
+  args:
+  - chown {{ .Values.uid }}:{{ .Values.gid }} /mounted-path;
+  volumeMounts:
+  - mountPath: /mounted-path
```
lhoestq, line 12: Why is `/mounted-path` hardcoded here ? It may cause issues to have the same as the one used with `assets.subpath` ?

tools/docker-compose-dev-datasets-server.yml
```diff
@@ -43,20 +45,29 @@ services:
     depends_on:
       - mongodb
     restart: always
-    network_mode: host
+    ports:
+      - ${ADMIN_UVICORN_PORT-8081}:${ADMIN_UVICORN_PORT-8081}
```
severo, line 49: are you reverting these changes?

chart/values.yaml
```diff
@@ -167,6 +167,12 @@ assets:
   # Directory on the shared storage (audio files and images)
   storageDirectory: "/assets"
 
+cachedAssets:
+  # base URL for the cached assets files. It should be set accordingly to the datasets-server domain, eg https://datasets-server.huggingface.co/cached-assets
+  # baseUrl: "not used for now"
+  # Directory on the cached shared storage (audio files and images)
+  storageDirectory: "/cached-assets"
+
```
severo, line 175: Even if it's not as important as the other variables, we should have all the env vars in this file:

```
CACHED_ASSETS_CLEAN_CACHE_PROBA = 0.05
CACHED_ASSETS_KEEP_FIRST_ROWS_NUMBER = 100
CACHED_ASSETS_KEEP_MOST_RECENT_ROWS_NUMBER = 200
CACHED_ASSETS_MAX_CLEANED_ROWS_NUMBER = 10_000
```

chart/templates/services/api/_container.tpl
```diff
@@ -43,6 +44,8 @@
     value: {{ .Values.api.uvicornNumWorkers | quote }}
   - name: API_UVICORN_PORT
     value: {{ .Values.api.uvicornPort | quote }}
+  volumeMounts:
+  {{ include "volumeMountCachedAssetsRO" . | nindent 2 }}
```
severo, line 48: RO in volumeMountCachedAssetsRO was for read-only. Here, we need to have access in writing mode -> RW

severo:
```suggestion
  {{ include "volumeMountCachedAssetsRW" . | nindent 2 }}
```

chart/templates/_initContainerCachedAssets.tpl
severo: no, I don't think so, but we have to change the name above

chart/templates/_initContainerCachedAssets.tpl
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 The HuggingFace Authors.
+
+{{- define "initContainerCachedAssets" -}}
+- name: prepare-assets
```
severo, line 5:
```suggestion
- name: prepare-cached-assets
```

severo (commented):
Still some small details

tools/docker-compose-dev-datasets-server.yml
lhoestq: Yes, `network_mode: host` doesn't work on mac :/
But keeping the ports is fine

lhoestq:
added the missing variables :)

severo (approved):
gogogo!

PR #1028, opened 2023-04-06
Disable full backfill action

AndreaFrancis:
Disabling backfill Job action since it was intended to run only one time.
Later, it will be a cron job (But should backfill only missing cache records in order to avoid appending too many jobs).

### DIFF
--- a/chart/env/dev.yaml
+++ b/chart/env/dev.yaml
@@ -103,7 +103,7 @@ mongodbMigration:
 # --- jobs (post-upgrade hooks) ---
 
 cacheMaintenance:
-  action: "backfill"
+  action: "skip"
   resources:
     requests:
       cpu: 100m
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -109,7 +109,7 @@ mongodbMigration:
 # --- jobs (post-upgrade hooks) ---
 
 cacheMaintenance:
-  action: "backfill"
+  action: "skip"
   resources:
     requests:
       cpu: 100m


chart/env/dev.yaml
```diff
@@ -103,7 +103,7 @@ mongodbMigration:
 # --- jobs (post-upgrade hooks) ---
 
 cacheMaintenance:
-  action: "backfill"
+  action: "skip"
```
severo, line 106: This action does not exist, right? It's OK, because it's explicit that the job will not do anything, but I'm afraid that it will exist with error code 1, ie. it will make the whole "sync" being reported as failed in ArgoCD

Capture d’écran 2023-04-06 à 20 08 12

AndreaFrancis: No, it will print
`        logging.warning(f"Wrong action mode selected, supported actions are {supported_actions}.")`

severo: Oh, right.

severo (approved):

severo:
I'm not sure how the jobs work, but I imagine that we have to wait for the previous upgrade to finish (ie for the post-upgrade job to finish) before deploying again, no?

rtrompier:
> I'm not sure how the jobs work, but I imagine that we have to wait for the previous upgrade to finish (ie for the post-upgrade job to finish) before deploying again, no?
> 
> cc @rtrompier

Yes, I confirm.

severo:
Merging because @AndreaFrancis is off

Issue #891, opened 2023-03-03
Store /valid and /is-valid in the cache


severo:
Pre-computing /is-valid (for each dataset) and /valid (for all the datasets) and storing the responses in that cache would help reduce the latency of these endpoints and the compute load on the API service. They are the only two endpoints that are calculated on the fly.

Another benefit: it would remove the particular field `required_by_dataset_viewer` in https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/processing_graph.py because the same property could be defined as dependencies between processing steps:
- /is-valid would depend on /splits and /first-rows
- /valid would depend on /is-valid

Note that we would have to add a value to [InputType](https://github.com/huggingface/datasets-server/blob/0539fbd1950bfbe89605ddd40db1aaebfc757438/libs/libcommon/src/libcommon/processing_graph.py#L10) for /valid. None? Optional?

Thanks, @AndreaFrancis, for the discussion that allowed rethinking this part.

---

- [x] create `dataset-is-valid` https://github.com/huggingface/datasets-server/pull/1032
- [ ] create `valid`
  - [ ] add `none` in InputType, and make the `dataset` parameter optional where it makes sense (it is currently mandatory nearly everywhere in the code)
  - [ ] add the `valid` step
- [ ] deploy and fill the cache. Done for `is-valid`
- [ ] switch the /valid and /is-valid endpoints. PR for `is-valid`: https://github.com/huggingface/datasets-server/pull/1160

severo:
We have to work on this to reduce the latency

severo:
I'll work on adding the step `dataset-is-valid which will only depend on `split-first-rows-from-streaming` and `split-first-rows-from-parquet` (if we have at least one of them working, it means that we have the list of configs and splits).

severo:
- [ ] create `dataset-is-valid` https://github.com/huggingface/datasets-server/pull/1032
- [ ] create `valid` https://github.com/huggingface/datasets-server/pull/1032
- [ ] deploy and fill the cache
- [ ] switch the /valid and /is-valid endpoints

severo:
For `valid`: maybe include the list of feature types, to enable filtering by feature type?

See https://github.com/huggingface/datasets-server/issues/561 and https://github.com/huggingface/datasets-server/pull/833#issuecomment-1517689866. cc @lhoestq

lhoestq:
Oh that could be great !

severo:
keep open

[...]

PR #1032, opened 2023-04-06
Create step dataset-is-valid

severo:
See https://github.com/huggingface/datasets-server/issues/891


services/worker/src/worker/job_runners/dataset/is_valid.py
```diff
@@ -0,0 +1,73 @@
...
+]
+FIRST_ROWS_KINDS = ["split-first-rows-from-streaming", "split-first-rows-from-parquet"]
+
+
+def compute_is_valid_response(dataset: str) -> Tuple[DatasetIsValidResponse, float]:
+    """
+    Get the response of /is-valid for one specific dataset on huggingface.co.
+
+    A dataset is valid if:
+    - /splits is valid
```
albertvillanova, line 35: One question:
```suggestion
    - /splits is valid for at least one of the splits 
```

albertvillanova (approved):
Thanks.

services/worker/src/worker/job_runners/dataset/is_valid.py
severo: You're right. (nit: configs, not splits)

severo: done

PR #1034, opened 2023-04-07
fix: 🐛 enable the two last migrations

severo:
It was missing in
https://github.com/huggingface/datasets-server/pull/1033

FYI @albertvillanova


severo:
Merging, no need for a review

albertvillanova (commented):
Thanks.

Issue #1035, opened 2023-04-07
Harmonize the TTL of k8s jobs after being deleted


severo:
See https://github.com/huggingface/datasets-server/pull/788: it added a TTL of 5 minutes, in order not to block an uninstall.

But we recently added a new job (https://github.com/huggingface/datasets-server/pull/1017) with a TTL of 24h.

We put 24h because we wanted to be able to inspect the logs if necessary.

I understand it's not the way to do: we should instead access the logs in elasticsearch. We should reduce the TTL to 300s

PR #1036, opened 2023-04-07
fix: 🐛 reduce the k8s job TTL to 5 minutes

severo:
It is required to allow helm uninstall quickly if necessary.

✅ Closes: https://github.com/huggingface/datasets-server/issues/1035


severo:
Merging, as it's easy

Issue #822, opened 2023-02-15
Add /dataset-status to the admin panel


severo:
See https://github.com/huggingface/datasets-server/pull/815

severo:
Let's keep it open

Issue #948, opened 2023-03-17
Add dataset,config,split to /admin/dataset-status


severo:
The report on https://huggingface.co/datasets/mweiss/fashion_mnist_corrupted is missing information about the configs and splits (in first rows in particular):

```json
{
  "/config-names": {
    "cached_responses": [
      {
        "http_status": 200,
        "error_code": null,
        "job_runner_version": 1,
        "dataset_git_revision": "705573226dc92bd56883d3882d62b1b9bb6b2d73",
        "progress": 1.0
      }
    ],
    "jobs": []
  },
  "/split-names-from-streaming": {
    "cached_responses": [
      {
        "http_status": 500,
        "error_code": "SplitNamesFromStreamingError",
        "job_runner_version": 2,
        "dataset_git_revision": "705573226dc92bd56883d3882d62b1b9bb6b2d73",
        "progress": null
      }
    ],
    "jobs": []
  },
  "/splits": {
    "cached_responses": [
      {
        "http_status": 500,
        "error_code": "SplitsNamesError",
        "job_runner_version": 2,
        "dataset_git_revision": "705573226dc92bd56883d3882d62b1b9bb6b2d73",
        "progress": null
      }
    ],
    "jobs": []
  },
  "/first-rows": {
    "cached_responses": [
      {
        "http_status": 501,
        "error_code": "JobRunnerCrashedError",
        "job_runner_version": 2,
        "dataset_git_revision": "6a4780f015ba3159709d35c7777ca8b83e30b3cd",
        "progress": null
      },
      {
        "http_status": 500,
        "error_code": "StreamingRowsError",
        "job_runner_version": 2,
        "dataset_git_revision": "6a4780f015ba3159709d35c7777ca8b83e30b3cd",
        "progress": null
      }
    ],
    "jobs": []
  },
  "/parquet-and-dataset-info": {
    "cached_responses": [
      {
        "http_status": 500,
        "error_code": "UnexpectedError",
        "job_runner_version": 1,
        "dataset_git_revision": "705573226dc92bd56883d3882d62b1b9bb6b2d73",
        "progress": null
      }
    ],
    "jobs": []
  },
  "config-parquet": { "cached_responses": [], "jobs": [] },
  "dataset-parquet": { "cached_responses": [], "jobs": [] },
  "/dataset-info": {
    "cached_responses": [
      {
        "http_status": 200,
        "error_code": null,
        "job_runner_version": 1,
[...]

PR #1040, opened 2023-04-10
Dataset status in admin ui

AndreaFrancis:
Adding new tab in admin-ui to get dataset status
Closes https://github.com/huggingface/datasets-server/issues/822
Depends on https://github.com/huggingface/datasets-server/pull/1041

Screenshot from 2023-04-10 14-20-51
Screenshot from 2023-04-10 14-20-23

### DIFF
--- a/.vscode/monorepo.code-workspace
+++ b/.vscode/monorepo.code-workspace
@@ -35,6 +35,10 @@
     {
       "name": "services/reverse-proxy",
       "path": "../services/reverse-proxy"
+    },
+    {
+      "name": "front/admin_ui",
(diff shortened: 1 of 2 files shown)


AndreaFrancis:
Once https://github.com/huggingface/datasets-server/issues/948 is done, it will be necessary to add those columns in the ui as well

severo (approved):
Nice! Very useful

AndreaFrancis:
> Though I'm getting errors (we still need to deploy before it can work, I guess ?):

Yes, it was fixed with prod deploy but anyway I will make the table show the response as it is instead of hardcoded columns.

Issue #1055, opened 2023-04-18
Dataset Viewer issue for togethercomputer/RedPajama-Data-1T


carlosseda:
### Link

https://huggingface.co/datasets/togethercomputer/RedPajama-Data-1T

### Description

The dataset viewer is not working for dataset togethercomputer/RedPajama-Data-1T.

Error details:

```
Error code:   PreviousStepFormatError

```

severo:
Working on it

albertvillanova:
The viewer is working fine now:

![Screenshot from 2023-04-19 12-36-26](https://user-images.githubusercontent.com/8515462/233049503-252e3fb1-83da-45ee-85c9-c7043d855986.png)

CC: @severo 

Also note that this issue page is getting filled with many cross-references, because of the commit messages...

PR #1056, opened 2023-04-18
Fix dataset level aggregators

severo:
copy the error if the previous step is erroneous, instead of errors that cannot be understood by the user, like https://github.com/huggingface/datasets-server/issues/1055. That way, an error (like 'empty dataset' for example) is propagated to the descendants.

---

Before, on an empty dataset repository, the dataset-split-names, for example, returned:

Capture d’écran 2023-04-18 à 16 47 01

Now:

Capture d’écran 2023-04-18 à 16 45 05


severo:
I agree, that's the logic behind the versions. In that case, the goal is precisely to update (part of) the responses, to replace the PreviousStepStatusError with the original error (see the screenshots in the description of the PR).
Maybe the debate is whether we should increment the version when only part of the responses should be updated, or if in that case we should better identify the affected responses, and refresh them specifically...
In that case, I would say that as the steps are very quick, so I'll merge as is. But feel free to give your opinion.

PR #1061, opened 2023-04-19
feat: 🎸 add backfill tasks, and button to launch backfill

severo:
See it in action. Here we see than a repo that does not exists shows an error (we could improve it), and then when the repo is created (but empty), the cache entries are created. Note that currently, the setup is a bit weird:
- the local hub does not send webhooks to the local datasets-server
- but: as the local hub has tried to get the list of first rows from the datasets-server, it has triggered an update of the dataset (see how service/api works when a cache entry is missing)
- thus: some cache entries have been created
- BUT: it was done using the "job runner" logic, where every step creates the following steps after the job has been finished. This logic has some flaws, which is why the backfill plan shows some missing tasks: some other jobs should have been created

https://user-images.githubusercontent.com/1676121/233086733-133866c0-bcaf-4f62-903b-9bb7a5c29f40.mov

Here we show that, once we add files to the empty repository, the backfill plan is to refresh a lot of cache entries. We execute the backfill, and then the dataset viewer works.

https://user-images.githubusercontent.com/1676121/233087075-d8aeeb11-817f-4bae-b412-ea792e32796a.mov

### DIFF
--- a/front/admin_ui/app.py
+++ b/front/admin_ui/app.py
@@ -78,6 +78,10 @@ with gr.Blocks() as demo:
                 cached_responses_table = gr.DataFrame()
                 gr.Markdown("### Pending jobs")
                 jobs_table = gr.DataFrame()
+                backfill_message = gr.Markdown("", visible=False)
+                backfill_plan_table = gr.DataFrame(visible=False)
+                backfill_execute_button = gr.Button("Execute backfill plan", visible=False)
+                backfill_execute_error = gr.Markdown("", visible=False)
 
 
     def auth(token):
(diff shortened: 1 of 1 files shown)


front/admin_ui/app.py
```diff
@@ -175,6 +179,46 @@ def get_dataset_status(token, dataset):
...
+        dataset_state = response.json()
+        tasks_df = pd.DataFrame(dataset_state["plan"])
+        has_tasks = len(tasks_df) > 0
+        return {
+            backfill_message: gr.update(
+                value="""### Backfill plan
+
+The cache is outdated or in an incoherent state. Here is the plan to backfill the cache."""
+            ,visible=has_tasks),
+            backfill_plan_table: gr.update(value=tasks_df,visible=has_tasks),
```
AndreaFrancis, line 202:
```suggestion
            backfill_plan_table: gr.update(value=tasks_df, visible=has_tasks),
```

front/admin_ui/app.py
```diff
@@ -175,6 +179,46 @@ def get_dataset_status(token, dataset):
...
                 jobs_table: gr.update(value=None)
             }
+    
+    def get_backfill_plan(token, dataset):
+        headers = {"Authorization": f"Bearer {token}"}
+        response = requests.get(f"{DSS_ENDPOINT}/admin/dataset-state?dataset={dataset}", headers=headers, timeout=60)
+        if response.status_code != 200:
+            return {
+                backfill_message: gr.update(value=f"❌ Failed to get backfill plan for {dataset} (error {response.status_code})", visible=True),
+                backfill_plan_table: gr.update(value=None,visible=False),
```
AndreaFrancis, line 189:
```suggestion
                backfill_plan_table: gr.update(value=None, visible=False),
```

front/admin_ui/app.py
```diff
@@ -175,6 +179,46 @@ def get_dataset_status(token, dataset):
...
+                backfill_execute_button: gr.update( visible=False),
```
AndreaFrancis, line 190:
```suggestion
                backfill_execute_button: gr.update(visible=False),
```

front/admin_ui/app.py
```diff
@@ -175,6 +179,46 @@ def get_dataset_status(token, dataset):
...
+                backfill_execute_error: gr.update( visible=False)
```
AndreaFrancis, line 191:
```suggestion
                backfill_execute_error: gr.update(visible=False)
```

AndreaFrancis (approved):
Great! Interface looks really friendly and useful. I just put minor details but not blocking at all

front/admin_ui/app.py
severo: If someone wants to create an issue and assign themself, please do!

AndreaFrancis: Sure no problem, we can handle it in another PR/issue

PR #1066, opened 2023-04-20
Propagate previous step error

severo:
- ensure that all the job runners will propagate the error from a previous step, if it's the case. This will help the users of the dataset viewer: they will see the original error (e.g.: the dataset is empty), instead of weird applications about steps

Also, refactor: remove special cases by making PreviousJobError a CustomError, as the other exceptions


libs/libcommon/src/libcommon/constants.py
```diff
@@ -9,21 +9,21 @@
...
+PROCESSING_STEP_DATASET_IS_VALID_VERSION = 2
+PROCESSING_STEP_DATASET_PARQUET_VERSION = 2
+PROCESSING_STEP_DATASET_SIZE_VERSION = 2
+PROCESSING_STEP_DATASET_SPLIT_NAMES_FROM_DATASET_INFO_VERSION = 2
+PROCESSING_STEP_DATASET_SPLIT_NAMES_FROM_STREAMING_VERSION = 2
 PROCESSING_STEP_PARQUET_AND_DATASET_INFO_VERSION = 1
-PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_PARQUET_VERSION = 1
-PROCESSING_STEP_CONFIG_PARQUET_AND_INFO_VERSION = 1
+PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_PARQUET_VERSION = 2
+PROCESSING_STEP_CONFIG_PARQUET_AND_INFO_VERSION = 2
```
lhoestq, line 23: maybe let's update the image row group size at the same time, to not recompute the parquet files twice

severo: hmmm, anyway the version upgrade k8s job is not enabled for now, so, it's not critical, I think. Right @AndreaFrancis ?

lhoestq: oh ok !

AndreaFrancis: Yes, It is disable with action=skip

AndreaFrancis (approved):

PR #1069, opened 2023-04-20
feat: 🎸 create children jobs even in case of error

### LINKED ISSUE #949: An error in a step should be propagated in the dependent steps
See https://github.com/huggingface/datasets-server/issues/948:
- the entries of "/first-rows" should have been deleted because the upstream jobs are errors and we cannot get the list of splits.
- the entry of "/dataset-info" should be an error, since the result of "/parquet-and-dataset-info" is an error, and both are at dataset level.
- etc.

I think we have to review the logic of what occurs in case of error, to have a coherent status for the cache of a dataset.

severo:
fixes #949, along with #1066


severo:
Note also that two interdependant steps mean... a cycle. Which should not exist in a DAG.

AndreaFrancis (approved):

severo:
review is welcome 👋 🤗

services/worker/src/worker/job_runner.py
```diff
@@ -544,16 +542,25 @@ def create_children_jobs(self) -> None:
...
+                SplitFullName(dataset=self.dataset, config=self.config, split=self.split)
+            }
+            new_split_full_names_for_config = {SplitFullName(dataset=self.dataset, config=self.config, split=None)}
+        elif self.processing_step.input_type == "config":
+            new_split_full_names_for_split = set()
+            new_split_full_names_for_config = {SplitFullName(dataset=self.dataset, config=self.config, split=None)}
+
+        else:
+            new_split_full_names_for_split = set()
+            new_split_full_names_for_config = set()
```
lhoestq, line 561: does this mean that there won't be any new job added if the child job is at split or config level ?

lhoestq (approved):
Sounds good !

PR #1075, opened 2023-04-21
refactor: 💡 create the configs at startup

severo:
instead of creating a FirstRowsConfig every time


AndreaFrancis (approved):
Looks good, thanks!

PR #1078, opened 2023-04-21
fix: 🐛 remove the wrong concept of blocked_by_parent

severo:
the relation created by "requires" means that the step will be refreshed when the required step will be refreshed. (should we change the name?).


AndreaFrancis (approved):

AndreaFrancis:
> the relation created by "requires" means that the step will be refreshed when the required step will be refreshed. (should we change the name?).

Yes, I think it could be misunderstood that it has to be ran only when all their parents are finished but that is not true. Maybe `triggered_by` or something like that could be more explicit.

Issue #1074, opened 2023-04-21
Move dataset,config and split validation to job_runner instead of compute method


AndreaFrancis:
We validate the parameters `dataset, config and split` at each job runner level here:

- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config_names.py#L109
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/parquet.py#L93
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/parquet_and_info.py#L934
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/size.py#L162
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/split_names_from_dataset_info.py#L108
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/split_names_from_streaming.py#L128
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/dataset/parquet.py#L141
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/dataset/is_valid.py#L60
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/dataset/size.py#L169
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/dataset/split_names.py#L152
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/dataset/split_names_from_dataset_info.py#L137
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/dataset/split_names_from_streaming.py#L153
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py#L315
- https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/first_rows_from_streaming.py#L490

But we could have it once before calling `compute` method https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runner.py#L466
Based on `self.processing_step.input,` we could perform the corresponding validation and throw the error before compute.
The change should be applied also in the job runners not listed here (Probable new ones after this issue was created)

PR #1079, opened 2023-04-21
Params validation to job_runner

AndreaFrancis:
Closes https://github.com/huggingface/datasets-server/issues/1074

### DIFF
--- a/services/worker/src/worker/job_runner.py
+++ b/services/worker/src/worker/job_runner.py
@@ -463,10 +463,20 @@ class JobRunner(ABC):
                 raise NoGitRevisionError(f"Could not get git revision for dataset {self.dataset}")
             try:
                 self.pre_compute()
+
+                # validate input parameters based on processing_step
+                input_type = self.processing_step.input_type
+                if self.dataset is None:
+                    raise ParameterMissingError("'dataset' parameter is required")
+                if input_type in {"split", "config"} and self.config is None:
+                    raise ParameterMissingError("'config' parameter is required")
+                if input_type == "split" and self.split is None:
+                    raise ParameterMissingError("'split' parameter is required")
+
                 job_result = self.compute()
                 content = job_result.content
 
-                # Validate content size
+                # validate content size
                 if len(orjson_dumps(content)) > self.worker_config.content_max_bytes:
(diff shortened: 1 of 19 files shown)


services/worker/src/worker/job_runner.py
```diff
@@ -463,10 +463,20 @@ def process(
                 raise NoGitRevisionError(f"Could not get git revision for dataset {self.dataset}")
             try:
                 self.pre_compute()
+
+                # validate input parameters based on processing_step
```
severo, line 467: maybe extract in `self.check_inputs()`?

services/worker/src/worker/job_runners/config/info.py
```diff
@@ -92,11 +91,9 @@ def get_job_runner_version() -> int:
         return PROCESSING_STEP_CONFIG_INFO_VERSION
 
     def compute(self) -> CompleteJobResult:
-        if self.dataset is None:
-            raise ParameterMissingError("'dataset' parameter is required")
-        if self.config is None:
-            raise ParameterMissingError("'config' parameter is required")
-        return CompleteJobResult(compute_config_info_response(dataset=self.dataset, config=self.config))
+        return CompleteJobResult(
+            compute_config_info_response(dataset=self.dataset, config=self.config)  # type: ignore
```
severo, line 95: we have to find a better solution than adding `# type: ignore`, otherwise it's better to keep the code untouched.

Not sure how. Should we have DatasetJobRunner, ConfigJobRunner and SplitJobRunner, inherited from JobRunner?

services/worker/tests/job_runners/config/test_parquet_and_info.py
```diff
@@ -529,7 +529,7 @@ def test_compute_splits_response_simple_csv_ok(
 @pytest.mark.parametrize(
     "name,error_code,cause",
     [
-        ("does_not_exist", "ParameterMissingError", None),
+        ("does_not_exist", "DatasetNotFoundError", None),
```
severo, line 532: why is this changing? we didn't have an CI error before?

AndreaFrancis:
Will close this PR and send a new one with another approach

Issue #1087, opened 2023-04-25
Full Spawning urls scan 


AndreaFrancis:
Currently we have urls scan implemented for first 100K rows [here ](https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py)
We need to do it for full dataset but there could be some concerns:
- Full scan could take too much time for big datasets (Like [laion](https://huggingface.co/datasets/laion/laion2B-en)) and we have a task configured [here](https://github.com/huggingface/datasets-server/blob/bdda4ed42c3f884995917ae6e7575ef67e338f48/services/worker/src/worker/executor.py#L121) that kills long running jobs that can prevent Full scan to do not finish
- We have a response size validation [here](https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runner.py#L470) which can lead to a database error in case of having big content (Too much opt in/out urls) to store

Approach 1:
Run full dataset scan in one single job
Will need the following actions:
- Implement custom timeout configurations to kill "long running jobs" i.e We could have a WORKER_MAX_JOB_DURATION_SECONDS value for split-opt-in-out-urls-scan Job set up to 10 hrs and for other jobs 20 min (as  it is currently)
- Store the content response (or maybe only opt in/out urls list) in an external storage. It was suggested by @severo  to store it on a parquet file (see thread https://github.com/huggingface/datasets-server/pull/1044#discussion_r1173625223). I think it could be a good idea but will need to think where to store the file (Maybe https://datasets-server.huggingface.co/assets? Another branch in each repository like we actually do for parquets https://huggingface.co/datasets/<dataset>/tree/refs%2Fconvert%2Fparquet? )
- In case we dont need yet to show the list of opt in/out urls, we can skip previous step and only store number of urls see sample: https://datasets-server.huggingface.co/opt-in-out-urls-scan?dataset=laion/laion2B-en&config=laion--laion2B-en&split=train

<div class="toolbar"><div class="devtools-separator"></div><div class="devtools-searchbox"><input class="searchBox devtools-filterinput" placeholder="Filter JSON" value=""></div></div><div class="panelContent" id="json-scrolling-panel" tabindex="0">

[...]

PR #1088, opened 2023-04-25
Spawning full scan

AndreaFrancis:
Closes https://github.com/huggingface/datasets-server/issues/1087
Included changes:
- Separate worker definition in chart for prod and dev in a dedicated one for spawning with max long running time set to 10 hrs
- Removing max rows scanning limitation
- Store in assets a csv file with list of opted in/out urls (Headers: url, column, row_idx, opt_in, opt_out)


libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -27,6 +28,13 @@ def create_asset_dir(
     return dir_path, url_dir_path
 
 
+def create_file_dir(dataset: str, config: str, split: str, assets_directory: StrPath) -> Tuple[Path, str]:
```
severo, line 31: maybe rename this one into `create_split_asset_dir`, and the previous one (`create_asset_dir`) into `create_cell_asset_dir`.

severo: maybe both could use a common function with optional parameters, to avoid code duplication

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -93,6 +101,41 @@ class AudioSource(TypedDict):
...
+
+    with open(file_path, "w") as output_file:
+        fc = csv.DictWriter(
+            output_file,
+            fieldnames=headers,
+        )
+        fc.writeheader()
+        fc.writerows(data)
+
+    return {"src": f"{assets_base_url}/{url_dir_path}/{file_name}", "type": "csv"}
```
severo, line 136:
```suggestion
    return {"src": f"{assets_base_url}/{url_dir_path}/{file_name}", "type": "text/csv"}
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -93,6 +101,41 @@ class AudioSource(TypedDict):
...
+    file_name: str,
+    assets_directory: StrPath,
+) -> FileSource:
+    dir_path, url_dir_path = create_file_dir(
+        dataset=dataset,
+        config=config,
+        split=split,
+        assets_directory=assets_directory,
+    )
+    makedirs(dir_path, ASSET_DIR_MODE, exist_ok=True)
```
severo, line 125: isn't it redundant?

(it's the same for the two other functions)

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -93,6 +101,41 @@ class AudioSource(TypedDict):
...
+    file_path = dir_path / file_name
+
+    with open(file_path, "w") as output_file:
+        fc = csv.DictWriter(
+            output_file,
+            fieldnames=headers,
+        )
+        fc.writeheader()
+        fc.writerows(data)
```
severo, line 134: we rely on the fact that the caller of this function passes coherent values for `headers` and `data`:
- same set of values
- in the same order

Is it OK? maybe add some docstring for that.

severo: Also: we never created unit tests for the assets functions. Should we use this PR to add them?

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -125,26 +128,25 @@ async def check_spawning(
 
 async def opt_in_out_task(
     image_urls: List[str], session: ClientSession, semaphore: Semaphore, limiter: AsyncLimiter, spawning_url: str
-) -> Tuple[List[Any], List[Any]]:
+) -> Any:
     try:
         spawning_response = await check_spawning(image_urls, session, semaphore, limiter, spawning_url)
     except Exception:
         raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}")
```
severo, line 135:
```suggestion
    except Exception as e:
        raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}", cause=e) from e
```

(not from this PR)

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -125,26 +128,25 @@ async def check_spawning(
 
 async def opt_in_out_task(
```
severo, line 129: maybe this function is no more useful, and the try/except could go to the check_spawning function?

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -125,26 +128,25 @@ async def check_spawning(
...
 async def opt_in_out_scan_urls(
     urls: List[str],
+    urls_columns: List[str],
     urls_number_per_batch: int,
     spawning_token: str,
     max_concurrent_requests_number: int,
     max_requests_per_second: int,
     spawning_url: str,
-) -> Tuple[List[int], List[int]]:
+) -> Tuple[List[OptUrl], int, int]:
```
severo, line 149: Personal taste (feel free to ignore): I prefer using named tuples, TypedDict, or dataclasses (in the same way I prefer keyword arguments over positional arguments in the call to functions) to be more explicit

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
severo: Or move the checks below, to the loop where you use the contents of spawning_response

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -160,30 +162,43 @@ async def opt_in_out_scan_urls(
...
-        for batch_opt_in_urls_idx in batch_opt_in_urls_indices:
-            opt_in_urls_indices.append(offset + batch_opt_in_urls_idx)
-        for batch_opt_out_urls_idx in batch_opt_out_urls_indices:
-            opt_out_urls_indices.append(offset + batch_opt_out_urls_idx)
-
-    return opt_in_urls_indices, opt_out_urls_indices
+        spawning_response = task.result()
+        spawning_response_urls = spawning_response["urls"]
+        for i in range(len(spawning_response_urls)):
+            if spawning_response_urls[i]["optIn"] or spawning_response_urls[i]["optOut"]:
```
severo, line 172: should we specifically check that spawning_response_urls[i] is a dict and that it contains the fields "optIn" and "optOut", or should we just let an exception be raised?

severo, line 172: each one is called three times, maybe create variables:

```suggestion
            is_opt_in = spawning_response_urls[i]["optIn"]
            is_opt_out = spawning_response_urls[i]["optOut"]
            if is_opt_in or is_opt_out:
```
(or `opt_in` / `opt_out`). Also: should you check the type of is_opt_in and is_opt_out (boolean)?

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -240,8 +255,7 @@ def compute_opt_in_out_urls_scan_response(
     if not urls_columns:
         return OptInOutUrlsScanResponse(
             urls_columns=[],
-            opt_in_urls=[],
-            opt_out_urls=[],
+            opt_in_out_urls_source=None,
```
severo, line 258: maybe something clearer for the users? `urls_file`? `urls_csv_file`? (we already know that we are in the context of opt in and opt out)

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -282,30 +296,24 @@ def compute_opt_in_out_urls_scan_response(
...
-    ]
-    opt_out_urls = [
-        OptUrl(
-            url=urls[url_idx],
-            row_idx=url_idx // len(urls_columns),
-            column_name=urls_columns[url_idx % len(urls_columns)],
-        )
-        for url_idx in opt_out_urls_indices
-    ]
+    headers = ["url", "row_idx", "column_name", "opt_in", "opt_out"]
```
severo, line 299: `column_name` is clear, but we use the term `feature` in /row and /first-rows. Should we stick to `feature` / `feature_name`, or is it better to use `column` here?

severo: Also: is it OK to use `opt_in` and `opt_out` (which is what sampling.ai returns in its API (optIn/optOut)), or should we prefix with `is_` (or `opted_in`/`opted_out`) to hint that it's a boolean?

services/worker/tests/job_runners/split/test_opt_in_out_urls_scan_from_streaming.py
```diff
@@ -150,12 +152,7 @@ def _get_job_runner(
             {
                 "has_urls_columns": True,
                 "num_scanned_rows": 4,
-                "opt_in_urls": [
-                    {"url": "http://testurl.test/test_image3-optIn.jpg", "row_idx": 3, "column_name": "col"}
-                ],
-                "opt_out_urls": [
-                    {"url": "http://testurl.test/test_image-optOut.jpg", "row_idx": 0, "column_name": "col"}
-                ],
+                "opt_in_out_urls_source": "assets/",
```
severo, line 155: why is this the value? there is no dataset, config, split and filename ?!

services/worker/tests/job_runners/split/test_opt_in_out_urls_scan_from_streaming.py
```diff
@@ -191,10 +188,22 @@ def test_compute(
...
+    assert content["num_scanned_rows"] == expected_content["num_scanned_rows"]
+    assert content["urls_columns"] == expected_content["urls_columns"]
+    assert content["num_opt_out_urls"] == expected_content["num_opt_out_urls"]
+    assert content["num_opt_in_urls"] == expected_content["num_opt_in_urls"]
+    assert content["num_urls"] == expected_content["num_urls"]
+    if expected_content["has_urls_columns"]:
+        assert content["opt_in_out_urls_source"] is not None and isinstance(content["opt_in_out_urls_source"], dict)
+    else:
+        assert content["opt_in_out_urls_source"] is None
+
```
severo, line 206: We might want to also test that the contents of the file are what we expect (ie. read the CSV and check the format and values)

chart/env/dev.yaml
```diff
@@ -224,3 +224,20 @@ workers:
         cpu: 1
         memory: "4Gi"
     tolerations: []
+  -
+    deployName: "spawning"
+    maxJobsPerNamespace: 5
+    maxJobDurationSeconds: 36_000
+    # 10 hours
```
lhoestq, line 231: IIRC LAION took more than 12h to be processed. Maybe increase to 20h ?

services/worker/src/worker/utils.py
```diff
@@ -284,6 +284,8 @@ def get_rows(
         raise TypeError("load_dataset should return a Dataset in normal mode")
     if column_names:
         ds = ds.select_columns(column_names)
+    if not rows_max_number:
+        return list(ds)
```
lhoestq, line 288: You'll OOM if you do that, some datasets are hundreds of gigabytes.

You can use it as a generator instead of a list

(same for `rows_plus_one ` a few lines later - we can't use a list with the full dataset)

lhoestq (commented):
Cool ! 

I think you can add a retry mechanism for such long jobs in `check_spawning`. Once every 10k request you may gate an error 5XX or a rate limiting error or a timeout.

I used this personally:

```python
RETRY_TIMES = ([1] * 5) + ([10] * 5) + ([10 * 60] * 10) + ([60 * 60] * 3)
TIMEOUT = 60

async def check_spawning(image_urls: List[str], session: aiohttp.ClientSession, semaphore, limiter) -> dict:
    url = f"https://opts-api.spawningaiapi.com/api/v2/query/urls"
    if not image_urls:
        return {"urls": []}
    elif len(image_urls) == 1:
        image_urls = image_urls + [""]  # the API requires >1 urls
    async with semaphore:
        async with limiter:
            resp_body = None
            for retry_time in RETRY_TIMES:
                try:
                    async with session.post(
                        url=url,
                        data="\n".join(image_urls),
                        timeout=TIMEOUT,
                    ) as resp:
                        resp_body = await resp.text()
                        spawning_response = await resp.json()
                        assert "urls" in spawning_response, str(resp_body)
                        return spawning_response
                except Exception as e:
                    logger.warning(str(e))
                    pass
                logger.warning(f"Retrying in {retry_time} seconds")
                time.sleep(retry_time)
            raise RuntimeError(str(resp_body))
```

chart/env/dev.yaml
```diff
@@ -199,7 +199,7 @@ workers:
   -
     deployName: "all"
     maxJobsPerNamespace: 1
-    workerOnlyJobTypes: ""
+    workerOnlyJobTypes: "/split-names-from-streaming,config-parquet-and-info,split-first-rows-from-parquet,dataset-is-valid"
```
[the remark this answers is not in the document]
severo: Here is the PR: https://github.com/huggingface/datasets-server/pull/1092

With that, you will be able to setup: `workerJobTypesBlocked: "split-opt-in-out-urls-scan"`

severo:
> I think you can add a retry mechanism for such long jobs in check_spawning. Once every 10k request you may gate an error 5XX or a rate limiting error or a timeout.

Note that we have a `retry` method that might be used, or refactored to manage this case.

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: Sure, I will add them

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
AndreaFrancis: I think we can leave as feature to keep consistency with firs_rows logic and also rename opt_x to is_opt_x (in/out)

AndreaFrancis:
Closing PR for now, I will think in another approach.

PR #1090, opened 2023-04-26
feat: 🎸 use a common function to forge unique ids

### LINKED ISSUE #861: Avoid collisions in unicity_id field in the jobs collection
See https://github.com/huggingface/datasets-server/pull/837#pullrequestreview-1315579718

severo:
Used for the Job unicity_id field, and for the artifact ids.

BREAKING CHANGE: 🧨 the job unicity_id has changed a bit

The only issue with the breaking change is that it could temporarily start twice the same job. Not a big issue, I think.

fixes #861


severo:
The only place where the unicity_id field is used is to select the next job to start: it allows checking if another job is already started for the same "artifact" (step + inputs). The only conflict we could have is to start twice the same job at the same time: it could only occur if the current started job is lasting for a long time, and if another job for the same artifact is created before it ends, but after the deploy has been done. At worst, it might occur for a handful of jobs, and the effect would be to overwrite on the same cache entry. So: not a problem, I would say

PR #1092, opened 2023-04-27
Add WORKER_JOB_TYPES_BLOCKED

severo:
Now, we can list:
- the jobs types that a worker cannot process: `WORKER_JOB_TYPES_BLOCKED`
- the jobs types that a worker can process: `WORKER_JOB_TYPES_ONLY`

### DIFF
--- a/services/worker/src/worker/loop.py
+++ b/services/worker/src/worker/loop.py
@@ -125,7 +125,16 @@ class Loop:
         logging.debug("try to process a job")
 
         try:
-            job_info = self.queue.start_job(job_types_only=self.worker_config.job_types_only)
+            job_info = self.queue.start_job(
+                job_types_blocked=self.worker_config.job_types_blocked,
+                job_types_only=self.worker_config.job_types_only,
+            )
+            if self.worker_config.job_types_blocked and job_info["type"] in self.worker_config.job_types_blocked:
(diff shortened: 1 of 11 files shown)


severo:
CI is failing because the Hub is currently down

services/worker/src/worker/loop.py
```diff
@@ -125,11 +125,20 @@ def process_next_job(self) -> bool:
         logging.debug("try to process a job")
 
         try:
-            job_info = self.queue.start_job(only_job_types=self.worker_config.only_job_types)
-            if self.worker_config.only_job_types and job_info["type"] not in self.worker_config.only_job_types:
+            job_info = self.queue.start_job(
+                job_types_blocked=self.worker_config.job_types_blocked,
+                job_types_only=self.worker_config.job_types_only,
+            )
+            if self.worker_config.job_types_blocked and job_info["type"] in self.worker_config.job_types_blocked:
```
AndreaFrancis, line 132: Do we really need this validation twice? One here and the other in queue start which is called in process_next_job -> run

AndreaFrancis (approved):

services/worker/src/worker/loop.py
severo: I think we can remove one, you're right

PR #1096, opened 2023-04-28
fix: 🐛 add missing environment variables for migration job

severo:


severo:
Merging, because it's simple.

PR #1098, opened 2023-04-28
Add logs and move backfill job to one shot job

severo:


severo:
Merging since it's simple.

PR #1103, opened 2023-04-28
fix: 🐛 revert to the previous value

severo:
When mongoengine viewed the new value (delete finished entries after 2 days instead of 7 days), it tried to create the index, but raised the following error:

```
pymongo.errors.OperationFailure: An equivalent index already exists with the same name but different options. Requested index: { v: 2, key: { finished_at: 1 }, name: "finished_at_1", background: false, expireAfterSeconds: 172800 }, existing index: { v: 2, key: { finished_at: 1 }, name: "finished_at_1", background: false, expireAfterSeconds: 604800 }, full error: {'ok': 0.0, 'errmsg': 'An equivalent index already exists with the same name but different options. Requested index: { v: 2, key: { finished_at: 1 }, name: "finished_at_1", background: false, expireAfterSeconds: 172800 }, existing index: { v: 2, key: { finished_at: 1 }, name: "finished_at_1", background: false, expireAfterSeconds: 604800 }', 'code': 85, 'codeName': 'IndexOptionsConflict'
```

Reverting for now.


severo:
cc @AndreaFrancis fyi

severo:
Merging

Issue #1104, opened 2023-04-28
Delete finished jobs immediately?


severo:
Currently, finished jobs are deleted after 7 days by an index. See https://github.com/huggingface/datasets-server/blob/259fd092c12d240d9b8d733c965c4b9362e90684/libs/libcommon/src/libcommon/queue.py#L144

But we never use the finished jobs, so:
- we could delete them immediately after finishing
- we could reduce the duration from 7 days to 1 hour (can be complementary to the previous action, to clean uncaught jobs)

For point 2, see https://github.com/huggingface/datasets-server/pull/1103

Stats:
- 9.805.591 jobs
- 13.345 are not finished! (0.1% of the jobs)

severo:
First change, in https://github.com/huggingface/datasets-server/pull/1105: reduce from 7 days to 1 day.

AndreaFrancis:
> we could delete them immediately after finishing

Maybe we could remove immediately if it is a success job (status=200) and for failed ones, leave as it is for now.

severo:
Yes, good idea

severo:
In https://github.com/huggingface/datasets-server/commit/69ca3c4bae5d581dde1e1d40c74b20e9d5e129bb we reduced the duration from 1 day to 10 minutes. It reduces a lot the number of entries in the Jobs collection. At the same time, it still allow to get the following metrics (they are computed every 2 minutes), which I find very useful for monitoring.

<img width="389" alt="Capture d’écran 2023-05-31 à 14 19 51" src="https://github.com/huggingface/datasets-server/assets/1676121/fdb05451-acf3-476e-8365-ad1261c074fd">

I close for now.

PR #1105, opened 2023-04-28
feat: 🎸 change TTL for finished jobs from 7 days to 1 day

severo:
We need to delete the index before mongoengine creates it again with a new value

See https://github.com/huggingface/datasets-server/issues/1104


AndreaFrancis (approved):

severo:
Thanks for the review. Merging because the CI is still failing, due to LFS outage on Hub CI

PR #1108, opened 2023-04-28
Aggregated config and dataset level for opt-in-out-urls-scan

AndreaFrancis:
Currently we have spawning urls scan numbers but at split level. This PR add config and dataset aggregation level (Only to summarize numbers).


libs/libcommon/src/libcommon/config.py
```diff
@@ -270,6 +272,16 @@ class ProcessingGraphConfig:
                 "requires": ["split-first-rows-from-streaming"],
                 "job_runner_version": PROCESSING_STEP_SPLIT_OPT_IN_OUT_URLS_SCAN_VERSION,
             },
+            "config-opt-in-out-urls-scan": {
+                "input_type": "config",
+                "requires": ["split-opt-in-out-urls-scan"],
```
severo, line 277: I think it also requires the list of split names, right?

libs/libcommon/src/libcommon/config.py
```diff
@@ -270,6 +272,16 @@ class ProcessingGraphConfig:
...
+                "job_runner_version": PROCESSING_STEP_CONFIG_OPT_IN_OUT_URLS_SCAN_VERSION,
+            },
+            "dataset-opt-in-out-urls-scan": {
+                "input_type": "dataset",
+                "requires": ["config-opt-in-out-urls-scan"],
```
severo, line 282: I think that it should also depend on /config-names

services/worker/src/worker/config.py
```diff
@@ -19,9 +19,9 @@
 WORKER_HEARTBEAT_INTERVAL_SECONDS = 60
 WORKER_KILL_LONG_JOB_INTERVAL_SECONDS = 60
 WORKER_KILL_ZOMBIES_INTERVAL_SECONDS = 10 * 60
-WORKER_MAX_DISK_USAGE_PCT = 90
+WORKER_MAX_DISK_USAGE_PCT = 70
```
severo, line 22: I think we should change this in another PR

services/worker/src/worker/job_runners/config/opt_in_out_urls_scan_from_streaming.py
severo, line 1: The tests are missing for the two new job runners

services/worker/src/worker/job_runners/dataset/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -0,0 +1,126 @@
...
+    """Raised when the content of the previous step has not the expected format."""
+
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "PreviousStepFormatError", cause, False)
+
+
+def compute_opt_in_out_urls_scan_response(dataset: str) -> Tuple[DatasetOptInOutUrlsScanResponse, float]:
+    logging.info(f"get opt-in-out-urls-scan for dataset={dataset}")
+
+    config_names_response = get_previous_step_or_raise(kinds=["/config-names"], dataset=dataset)
```
severo, line 58: we depend on /config-names, it should be in the list of requires

services/worker/src/worker/job_runners/config/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(message, HTTPStatus.INTERNAL_SERVER_ERROR, "PreviousStepFormatError", cause, False)
+
+
+def compute_opt_in_out_urls_scan_response(dataset: str, config: str) -> Tuple[OptInOutUrlsScanResponse, float]:
+    logging.info(f"get config-opt-in-out-urls-scan for dataset={dataset} config={config}")
+
+    split_names_response = get_previous_step_or_raise(
+        kinds=["/split-names-from-streaming"], dataset=dataset, config=config
```
severo, line 50: we depend on /split-names-from-streaming, we should add it to the requires of the step

severo (commented):
Some details, but it's good

libs/libcommon/src/libcommon/config.py
AndreaFrancis: Yes I added it here https://github.com/huggingface/datasets-server/blob/config-dataset-spawning-level/libs/libcommon/src/libcommon/config.py#L283

libs/libcommon/src/libcommon/config.py
AndreaFrancis: Yes, I added it here https://github.com/huggingface/datasets-server/blob/config-dataset-spawning-level/libs/libcommon/src/libcommon/config.py#L288 thanks!

services/worker/src/worker/job_runners/config/opt_in_out_urls_scan_from_streaming.py
AndreaFrancis: I added them https://github.com/huggingface/datasets-server/blob/config-dataset-spawning-level/services/worker/tests/job_runners/dataset/test_opt_in_out_urls_count.py and https://github.com/huggingface/datasets-server/blob/config-dataset-spawning-level/services/worker/tests/job_runners/config/test_opt_in_out_urls_count.py

services/worker/src/worker/config.py
AndreaFrancis: Sorry, it was a mistake, I removed it from the PR

services/worker/src/worker/job_runners/config/opt_in_out_urls_scan_from_streaming.py
AndreaFrancis: Fixed in https://github.com/huggingface/datasets-server/blob/config-dataset-spawning-level/libs/libcommon/src/libcommon/config.py#L283

libs/libcommon/src/libcommon/config.py
```diff
@@ -276,6 +278,16 @@ class ProcessingGraphConfig:
                 "requires": ["split-opt-in-out-urls-scan"],
                 "job_runner_version": PROCESSING_STEP_SPLIT_OPT_IN_OUT_URLS_COUNT_VERSION,
             },
+            "config-opt-in-out-urls-count": {
+                "input_type": "config",
+                "requires": ["/split-names-from-streaming", "split-opt-in-out-urls-scan"],
```
severo, line 283: maybe we can depend on `split-opt-in-out-urls-count` now?

services/worker/src/worker/job_runners/config/opt_in_out_urls_count.py
```diff
@@ -0,0 +1,121 @@
...
+    num_urls = 0
+    num_scanned_rows = 0
+    try:
+        total = 0
+        pending = 0
+        for split_item in content["splits"]:
+            split = split_item["split"]
+            total += 1
+            try:
+                response = get_response(kind="split-opt-in-out-urls-scan", dataset=dataset, config=config, split=split)
```
severo, line 68: maybe use `split-opt-in-out-urls-count` instead?

severo (approved):
Two details:
- maybe depend on `split-opt-in-out-urls-count` instead of `split-opt-in-out-urls-scan` to reduce the size of the data fetched from the database.
- fix the flake8 error.

services/worker/src/worker/job_runners/config/opt_in_out_urls_count.py
AndreaFrancis: Yes sure, I forgot to change this one but it was fixed on https://github.com/huggingface/datasets-server/pull/1108/commits/f142efba4f815c1af9b654eb3bdb346dd501b9a7

PR #1113, opened 2023-05-02
Add error detail in admin UI

severo:


lhoestq (approved):
thanks ! I think later we can also simply exclude "contents" in the UI and show all the other fields rather than hardcoding them

PR #1116, opened 2023-05-02
fix: 🐛 extend takes one argument (a list), not one arg per elem

severo:


severo:
After merging, we should relaunch the jobs with `UnexpectedError` error code.

Note the increase in this Error code after the big backfill we did last week:

Capture d’écran 2023-05-02 à 11 48 34

Not sure if all of them are due to this bug (we would have to look at the `details` field)

lhoestq (approved):
good catch !

severo:
That's what unit tests are helpful for... and I hadn't written one.

Issue #1121, opened 2023-05-02
Remove urls content in Spawning response


AndreaFrancis:
Currently SplitOptInOutJobRunner job runner  displays following fields:
```
    urls_columns: List[str]
    num_opt_in_urls: int
    num_opt_out_urls: int
    num_urls: int
    num_scanned_rows: int
    has_urls_columns: bool
    opt_in_urls: List[OptUrl]
    opt_out_urls: List[OptUrl]
```

For first phase, `opt_in_urls` and  `opt_out_urls` shouldn't be shown meanwhile we found a better way to store it.
cc. @julien-c @severo

PR #1124, opened 2023-05-02
Remove spawning url content from API /opt-in-out-urls

AndreaFrancis:
Closes https://github.com/huggingface/datasets-server/issues/1121
A new job runner is created that will copy only the num values from previous response split-opt-in-out-urls-scan.
This new cache kind will be the new response for /opt-in-out-urls endpoint


services/api/tests/routes/test_endpoint.py
```diff
@@ -75,6 +75,12 @@ def test_endpoints_definition() -> None:
     assert len(size["dataset"]) == 1  # Only has one processing step
     assert len(size["config"]) == 1  # Only has one processing step
 
+    opt_in_out_ursl = definition["/opt-in-out-urls"]
+    assert opt_in_out_ursl is not None
+    assert sorted(list(opt_in_out_ursl)) == ["split"]
+    assert opt_in_out_ursl["split"] is not None
+    assert len(opt_in_out_ursl["split"]) == 1  # Only has one processing step
```
severo, line 82:
```suggestion
    opt_in_out_urls = definition["/opt-in-out-urls"]
    assert opt_in_out_urls is not None
    assert sorted(list(opt_in_out_urls)) == ["split"]
    assert opt_in_out_urls["split"] is not None
    assert len(opt_in_out_urls["split"]) == 1  # Only has one processing step
```

severo (approved):
Perfect

PR #1131, opened 2023-05-03
fix: 🐛 fix the URL for /admin/force-refresh

severo:
The URL is weird, but while some steps still start with `/`, we are a bit stuck with that (otherwise, the double slash would be converted to one slash, and would break the route).


lhoestq (approved):
Ah lol

PR #1132, opened 2023-05-03
Plot processing graph

polinaeterna:
I updated admin-ui with the new `Processing graph` tab with `Plot processing graph` button. I wasn't sure that the plot would have appropriate size by default for all screens so I added sliders for width and height:
image

I doesn't look very nice but okay for the first version? 
I also updated some requirements, specifically `gradio`, because the old version of `gradio` wasn't rendering plots well.

I think it was a bad idea to update `poetry.lock` because now it has all the `libcommon` (from local path) dependencies. Maybe remove it completely from this directory? In the ui `libcommon` is installed from git, main branch.


front/admin_ui/requirements.txt
```diff
@@ -1,5 +1,7 @@
-gradio~=3.18.0
+gradio==3.28.1
+libcommon @ git+https://github.com/huggingface/datasets-server@main#subdirectory=libs/libcommon
```
lhoestq, line 2: I think it will raise an error when deploying to Spaces. On my laptop it I get
```
ERROR: Package 'libcommon' requires a different Python: 3.10.11 not in '==3.9.15'
```
maybe we can have an endpoint that returns the processing graph as a JSON object, and the UI can query it with a GET request ? And this way it shows the actual processing graph of the server at the state at which is it deployed, not the one on 'main'

lhoestq (commented):
Cool ! just one comment:

front/admin_ui/requirements.txt
severo: > maybe we can have an endpoint that returns the processing graph as a JSON object, and the UI can query it with a GET request ? And this way it shows the actual processing graph of the server at the state at which is it deployed, not the one on 'main'

Another way would be to update the space when we deploy datasets-server to prod.
Maybe we could split libcommon into two libs, if it helps reducing the number of dependencies in the space.

front/admin_ui/app.py
```diff
@@ -35,6 +39,23 @@ def healthcheck():
         return f"❌ Failed to connect to {DSS_ENDPOINT} (error {response.status_code})"
 
 
+def draw_graph(width, height):
+    config = ProcessingGraphConfig()
+    libcommon_graph = ProcessingGraph(config.specification)
+    steps = libcommon_graph.steps
+
+    graph = nx.DiGraph()
+    for name, step in steps.items():
```
severo, line 48: We can also give access to the nx graph as an attribute of libcommon_graph, so that you don't have to recreate the graph (I'm doing this in https://github.com/huggingface/datasets-server/pull/1094, along others things)

severo (approved):
It's excellent! I was dreaming of this for so long!

Next step: showing the status of a dataset in this graph :)

front/admin_ui/requirements.txt
polinaeterna: > Another way would be to update the space when we deploy datasets-server to prod.

How to do this correctly? Currently `libcommon` dependency is provided as `libcommon @ git+https://github.com/huggingface/datasets-server@main#subdirectory=libs/libcommon`, as far as I understand this is not equivalent to what is deployed to prod? 

> Maybe we could split libcommon into two libs, if it helps reducing the number of dependencies in the space.

On my side the container is being build pretty fast anyway, so maybe the number of dependencies for the space is not that crucial

severo: > How to do this correctly? According to this change libcommon dependency in the space will be provided as libcommon @ git+https://github.com/huggingface/datasets-server@main#subdirectory=libs/libcommon, as far as I understand this is not equivalent to what is deployed to prod?

OK, I had not understood the issue. Not sure how we can manage this.

polinaeterna: Ah but if we update the space only when we deploy, by installing from `main` we will get exactly the version on the prod, right? At least most likely :D if nothing was merged to main meanwhile. But that means we must not forget about it each time.

severo: Yes, I think it's ok for now to do so

Issue #1127, opened 2023-05-03
Too large row group size for parquet exports of image datasets


lhoestq:
See https://huggingface.co/datasets/sasha/dog-food/blob/refs%2Fconvert%2Fparquet/sasha--dog-food/parquet-train.parquet which still has a row group size of 1K even though it's supposed to be 100 after #833

severo:
I force refreshed sasha/dog-food, but the parquet files have not been updated: https://huggingface.co/datasets/sasha/dog-food/tree/refs%2Fconvert%2Fparquet/sasha--dog-food

lhoestq:
it did an empty commit: https://huggingface.co/datasets/sasha/dog-food/commit/a81197249e873ae4c9341eb1b3751a05705b9142

severo:
Ah ok, thanks, I was looking at the list of commits in https://huggingface.co/datasets/sasha/dog-food/tree/refs%2Fconvert%2Fparquet/sasha--dog-food (it showed only 2), while at the root (https://huggingface.co/datasets/sasha/dog-food/tree/refs%2Fconvert%2Fparquet), we see the empty ones

lhoestq:
found the issue: the feature somehow disappeared from `main`- will re-make a PR

PR #1134, opened 2023-05-03
Re-lower row group

lhoestq:
Re-add the changes from https://github.com/huggingface/datasets-server/pull/833 that were inadvertently removed by https://github.com/huggingface/datasets-server/pull/985

close https://github.com/huggingface/datasets-server/issues/1127


severo (approved):
Thanks!

One question: should we include bytes too? I have the feeling that it's another case where the calls to /rows would be slow if we have big rowgroups

lhoestq:
I got a bunch of BadRequestError in the CI - is it OK to merge anyway ? Or is there anything I need to do to fix them ?

severo:
Seems like an issue with the LFS on Hub CI

severo:
> We have the bytes from the preview computation no ?

I didn't understand

lhoestq:
Sorry I misunderstood. I thought you were mentioning the actual number of bytes per row.

But you meant the "binary" type right ? I agree we should also include it - let me add it real quick

severo (approved):

PR #1136, opened 2023-05-03
Fix parent class for split-names-from-dataset-info and first-rows-from-parquet

AndreaFrancis:
split-names-from-dataset-info and first-rows-from-parquet don't need to inherit from DatasetsBasedJobRunner, changing to JobRunner


severo (approved):
good catch.

PR #1142, opened 2023-05-04
Moving input (dataset,config,split) params validation to parent job runners

AndreaFrancis:
Replaces https://github.com/huggingface/datasets-server/pull/1079
Closes https://github.com/huggingface/datasets-server/issues/1074

PR #1146, opened 2023-05-08
WIP - Separate job runner logic from operations

AndreaFrancis:
Based on https://github.com/huggingface/datasets-server/pull/1142, this is a proposal to start separating operation logic like compute, pre-compute and post-compute from Job Runner so that, it will be easier to switch to an orchestrator some day.

I introduce "operators" concept which can be defined as a class that only compute (pre and post) a result and then it is managed in job runner.
An operator do not need to know about other actions like store cache, validate if there already exist a parallel cache, validate if it has to skip the compute or other actions different than "compute" the response.
I also introduce parent classes dataset_job_operator, config_job_operator and split_job_operator (They are different in the attributes needed by each operator, it will close https://github.com/huggingface/datasets-server/issues/1074).
I think this way we could extend in the future more levels of granularity (Like partitions for example for https://github.com/huggingface/datasets-server/issues/1087

This is just a POC and might need more teaks.
So far, it is pending:

- [ ] Fix unittests
- [ ] Refactor in some classes


services/worker/src/worker/job_runners/_datasets_based_job_runner.py
```diff
@@ -34,24 +34,29 @@ def __init__(
...
             processing_step=processing_step,
-            processing_graph=processing_graph,
         )
         self.datasets_based_config = app_config.datasets_based
         self.base_datasets_cache = hf_datasets_cache
 
     def get_cache_subdirectory(self, date: datetime) -> str:
         date_str = date.strftime("%Y-%m-%d-%H-%M-%S")
-        payload = (date_str, self.get_job_type(), self.dataset, self.config, self.split, self.force)
+        # TODO: Refactor, need a way to generate payload based only on provided params
```
severo, line 49: what is the issue here?

services/worker/src/worker/loop.py
```diff
@@ -133,10 +136,14 @@ def process_next_job(self) -> bool:
             return False
 
         job_runner = self.job_runner_factory.create_job_runner(job_info)
-        finished_status = job_runner.run()
-        self.queue.finish_job(job_id=job_runner.job_id, finished_status=finished_status)
+        processing_graph = ProcessingGraph(self.app_config.processing_graph.specification)
```
severo, line 139: we should create it before, instead of creating it again in every loop

services/worker/src/worker/job_runner.py
```diff
@@ -1,322 +1,20 @@
...
-        job_info (:obj:`JobInfo`):
-            The job to process. It contains the job_id, the job type, the dataset, the config, the split
-            the force flag, and the priority level.
-        common_config (:obj:`CommonConfig`):
-            The common config.
-        processing_step (:obj:`ProcessingStep`):
-            The processing step to process.
-    """
-
-    job_id: str
```
severo, line 309: why to you remove the fields that are initialized in __init__?

services/worker/src/worker/job_manager.py
```diff
@@ -0,0 +1,308 @@
...
+            the force flag, and the priority level.
+        common_config (:obj:`CommonConfig`):
+            The common config.
+        processing_step (:obj:`ProcessingStep`):
+            The processing step to process.
+    """
+
+    job_id: str
+    dataset: str
+    job_params: JobParams
```
severo, line 51: why both dataset and job_params, that contains "dataset"?

services/worker/src/worker/job_manager.py
```diff
@@ -0,0 +1,308 @@
...
+        self.job_type = job_info["type"]
+        self.job_id = job_info["job_id"]
+        self.force = job_info["force"]
+        self.priority = job_info["priority"]
+        self.job_params = job_info["params"]
+        self.dataset = self.job_params["dataset"]
+        self.common_config = app_config.common
+        self.worker_config = app_config.worker
+        self.job_runner = job_runner
+        self.processing_graph = processing_graph
```
severo, line 77: could we add it to the declared fields above? (and the other missing ones, if there are some?)

services/worker/src/worker/job_runners/dataset/dataset_job_runner.py
```diff
@@ -0,0 +1,50 @@
...
+        hf_datasets_cache: Path,
+    ) -> None:
+        DatasetsBasedJobRunner.__init__(
+            self=self,
+            job_info=job_info,
+            app_config=app_config,
+            processing_step=processing_step,
+            hf_datasets_cache=hf_datasets_cache,
+        )
+        DatasetJobRunner.__init__(
```
severo, line 45: What is the difference between calling super().__init__() and, like here, to call explicitly the parent class names? Is one of them super() too? 
I'm not familiar with double inheritance.

severo (approved):
Some questions, and some small refactoring, but it's very useful! I love it. Let's merge as soon as possible.

Some comments about the jobmanager being specific to a dataset:
- we currently need to get the dataset revision. It could be a job param at one point (not now)
- we need to know the dataset to run DatasetState.backfill()
The current state is OK, but it could be good to extract these both things in some way, so that the jobmanager is not specifically a dataset job manager

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -232,6 +267,10 @@ def get_response(kind: str, dataset: str, config: Optional[str] = None, split: O
     }
 
 
+def get_response_params(kind: str, dataset: str, job_params: JobParams) -> CacheEntry:
```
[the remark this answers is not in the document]
AndreaFrancis: Makes sense, I will remove dataset, we dont need it

services/worker/src/worker/job_runners/_datasets_based_job_runner.py
AndreaFrancis: I don't like the way we use params in lines 53 - 55, because for dataset job runners we don't need config and split for example.

services/worker/src/worker/job_manager.py
AndreaFrancis: I removed it on https://github.com/huggingface/datasets-server/pull/1146/commits/9b6bf8e0f8b93176cd3f8dd7e36025d62edb3305

services/worker/src/worker/job_manager.py
AndreaFrancis: Thanks, I added it on https://github.com/huggingface/datasets-server/pull/1146/commits/9b6bf8e0f8b93176cd3f8dd7e36025d62edb3305

services/worker/src/worker/job_runner.py
AndreaFrancis: Most of them were moved to job_manager

services/worker/src/worker/job_runners/dataset/dataset_job_runner.py
AndreaFrancis: For this specific case the problem was having different parameters/types in the parent constructors, and calling to super().__init__() would expect to have the same parameters sign. One solution was to have *args and **kwargs in ancestors but I didn't want to complicate them. But maybe there is a better way to handle it.

services/worker/src/worker/job_runners/_datasets_based_job_runner.py
severo: Oh. It's not important, we will just have `None, None,`. The payload is only used for the hash. I think it's OK. We could even put `self.job_info["params"]` to be simpler.

services/worker/src/worker/job_runners/dataset/dataset_job_runner.py
severo: OK, no idea. Thanks

services/worker/src/worker/job_runner.py
severo: but you're still initializing them in `__init__`: https://github.com/huggingface/datasets-server/pull/1146/files/f96451eed128bc531ea383b98d4e62e550413d0c#diff-7ccf0bcf2a0d7e66a9eae13d7168a08a4b883b8792b418a0a60cd86d7f0e8d7fR37

severo: they should not be initialized in job_runner I think.

AndreaFrancis: Totally agree, sorry I misunderstood.
I will fix it on https://github.com/huggingface/datasets-server/pull/1170

Issue #1002, opened 2023-03-31
Update datasets to 2.11.0


severo:
See https://github.com/huggingface/datasets/releases/tag/2.11.0

TODO: See discussions below
- [x] #1009
- [x] #1280
- [x] #1281
- [x] Use writer_batch_size for ArrowBasedBuilder
- [ ] Use direct cast from binary to Audio/Image
- [ ] Refresh datasets that use numpy.load

Useful changes for the datasets server (please complete if there are more, @huggingface/datasets)

> Use soundfile for mp3 decoding instead of torchaudio by @polinaeterna in https://github.com/huggingface/datasets/pull/5573
> 
>  - this allows to not have dependencies on pytorch to decode audio files
>  - this was possible with soundfile 0.12 which bundles libsndfile binaries at a recent version with MP3 support

should we remove the dependency to torch and torchaudio? cc @polinaeterna

> Add writer_batch_size for ArrowBasedBuilder by @lhoestq in https://github.com/huggingface/datasets/pull/5565
> - allow to specofy the row group / record batch size when you download_and_prepare() a dataset

Needed for https://github.com/huggingface/datasets-server/pull/833 I think; cc @lhoestq 

> Allow direct cast from binary to Audio/Image by @mariosasko in https://github.com/huggingface/datasets/pull/5644

Should we adapt the code in https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/features.py due to that?

> Support streaming datasets with numpy.load by @albertvillanova in https://github.com/huggingface/datasets/pull/5626

should we refresh some datasets after that?

albertvillanova:
I would suggest splitting this issue into several minor ones. I'm editing the description above...

severo:
Keep open. Note also #1099 to upgrade to 2.12.0

albertvillanova:
Keep open

albertvillanova:
I think "Use writer_batch_size for ArrowBasedBuilder" is already done by:
- #833

CC: @lhoestq @severo

PR #1147, opened 2023-05-09
Update datasets dependency to 2.12.0 version

### LINKED ISSUE #1099: Upgrade datasets to 2.12.0
https://github.com/huggingface/datasets/releases/tag/2.12.0

albertvillanova:
After 2.12.0 datasets release, update dependencies on it.

Fix #1099.


severo:
We have to upgrade in libcommon too (https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/pyproject.toml#L10). BTW, it might be the occasion to remove the explicit dependency from worker, since it depends on datasets through libcommon

albertvillanova:
I didn't know about this change in the way we explicit dependencies... Before it was only set in worker. I am updating this PR...

severo:
Yes, we need datasets for the `/rows` endpoint in services/api. Note that possibly we will move /rows to the workers, just with a highest priority, to get quick answers. So... possibly it will change again soon. Meanwhile, let's define it only in one place.

albertvillanova:
The CI error on Build PR Documentation is unrelated: `504 Server Error: Gateway Time-out`.

The error in services/worker / unit-tests is weird: it should be an error because too long but gave a success. I also think it is unrelated: https://github.com/huggingface/datasets-server/actions/runs/4926280974/jobs/8801621665?pr=1147
```
FAILED tests/test_executor.py::test_executor_stops_on_long_job - AssertionError: must be an error because too long
assert <Status.SUCCESS: 'success'> == <Status.ERROR: 'error'>
  - error
  + success
```

severo:
> The error in services/worker / unit-tests is weird: it should be an error because too long but gave a success. I also think it is unrelated: https://github.com/huggingface/datasets-server/actions/runs/4926280974/jobs/8801621665?pr=1147

Yes, it happens a lot of times. Relaunching generally fixes the test, but unit test that work 50% of the time is not very useful, we should find the issue and fix the test (in another issue).

severo:
I created https://github.com/huggingface/datasets-server/issues/1156

albertvillanova:
All CI tests are green now and the CI issues were unrelated to this PR. I think this can be approved.

albertvillanova:
> Apart from the pure dependency upgrade, I think we still have to finish the tasks for the previous version upgrade to 2.11.0 (#1002). It can be in other PRs.
> 
> Also: are there similar tasks related to 2.12.0 release (https://github.com/huggingface/datasets/releases/tag/2.12.0). Maybe the dependency to beam?

Yes, indeed I am working on the list of tasks of the previous datasets release. I can work on the new one for this release as well.

PR #1149, opened 2023-05-09
Use state for job creation

severo:
We centralize the creation of queue jobs using DatasetState.backfill(), for:
-> webhook
-> endpoint in services/api (when the response is not found, we analyze if the cache should have existed, or is in progress, and respond adequately + launch the job if needed)

Note that jobs are also created by the job runners when they finish. We will change that in another PR (https://github.com/huggingface/datasets-server/pull/1157).


severo:
issues on github: https://www.githubstatus.com/

Capture d’écran 2023-05-10 à 15 21 07

PR #1157, opened 2023-05-10
feat: 🎸 do full backfill instead of creating jobs for children

severo:
it will reduce the incoherencies in the cache. In particular, parallel steps lead to deleting valid cache entries.


AndreaFrancis (approved):
Nice! It will close https://github.com/huggingface/datasets-server/issues/854 as well

PR #1158, opened 2023-05-10
Replace legacy hffs dependency with huggingface-hub

### LINKED ISSUE #1150: Use of hffs dependency is legacy
The `hffs`dependency is legacy and their repo has been archived: https://github.com/huggingface/hffs

We should use `huggingface-hub` >= 0.14 instead.

albertvillanova:
This PR removes legacy `hffs` dependency and uses `huggingface-hub` instead.

Fix #1150.


services/api/src/api/routes/rows.py
```diff
@@ -67,19 +67,27 @@ class ParquetDataProcessingError(Exception):
 
 # TODO: how to invalidate the cache when the parquet branch is created or deleted?
 @lru_cache(maxsize=128)
```
albertvillanova, line 69: I didn't change the use of `lru_cache` here (in rows.py), but I noticed it is not used in first_rows_from_parquet.py.

Is there any reason for this? Also note that now the returned fs will always be the same as it no longer depends on the dataset name...

services/api/src/api/routes/rows.py
```diff
@@ -147,12 +155,14 @@ def __post_init__(
                 if not sources:
                     raise ParquetResponseEmptyError("No parquet files found.")
             with StepProfiler(method="rows.index", step="get the Hub's dataset filesystem"):
-                fs = get_parquet_fs(dataset=self.dataset, hf_token=hf_token)
+                fs = get_hf_fs(hf_token=hf_token)
+            with StepProfiler(method="rows.index", step="get the source URIs"):
```
albertvillanova, line 159: Note I added a profiler step here.

services/api/src/api/routes/rows.py
severo: you're right, we could use it too in the worker. Possibly it was more crucial here: the worker works asynchronously, while the code here is called synchronously when doing a request on /rows.

services/api/src/api/routes/rows.py
severo: (even if it should always be ultra fast, since it's a loop that creates strings). But it's better to be coherent with the rest of the code.

services/worker/poetry.lock
```diff
@@ -1735,9 +1709,9 @@ develop = true
 
 [package.dependencies]
 appdirs = "^1.4.4"
-datasets = {version = "^2.11.0", extras = ["audio", "vision"]}
+datasets = {version = "^2.12.0", extras = ["audio", "vision"]}
```
severo, line 1712: why is that? was it missing from https://github.com/huggingface/datasets-server/pull/1147/files?

(it's not very important, I just wonder what occurred)

severo (approved):
Very good!

Just some type annotations missing + huggingface_hub should be ignored by mypy in services/admin/pyproject.toml

services/worker/poetry.lock
albertvillanova: This is the commit that should have updated that file (after having updated libcommon before): https://github.com/huggingface/datasets-server/pull/1147/commits/2d907bffc099000ff1365c2fdc1042df072459ca

albertvillanova: I ran `poetry update datasets` and maybe I should have run differently, because the dependency on datasets is not direct but indirect through libcommon.

Maybe should I have run instead `poetry update libcommon`?

albertvillanova: See my comment below, related to this.

albertvillanova:
The CI gave error for services/api: https://github.com/huggingface/datasets-server/actions/runs/4935960556/jobs/8822912855?pr=1158
```
__________________ ERROR collecting tests/routes/test_rows.py __________________
ImportError while importing test module '/home/runner/work/datasets-server/datasets-server/services/api/tests/routes/test_rows.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.9.15/x64/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/routes/test_rows.py:20: in <module>
    from api.routes.rows import Indexer, RowsIndex, clean_cached_assets, create_response
src/api/routes/rows.py:17: in <module>
    from huggingface_hub import HfFileSystem
E   ImportError: cannot import name 'HfFileSystem' from 'huggingface_hub' (/home/runner/.cache/pypoetry/virtualenvs/api-fuJHx4a0-py3.9/lib/python3.9/site-packages/huggingface_hub/__init__.py)
```
I looked at the installed dependencies and found:
```
• Installing huggingface-hub (0.12.1)
```

I had run: `poetry update huggingface-hub`, but this did not update `huggingface-hub` because it is not a direct dependency, but an indirect one, through libcommon.

Now I have run: `poetry update libcommon` and it worked:
```
Updating dependencies
Resolving dependencies... (2.4s)

Writing lock file

Package operations: 0 installs, 2 updates, 0 removals

  - Updating huggingface-hub (0.12.1 -> 0.14.1)
  - Updating datasets (2.11.0 -> 2.12.0)
```

albertvillanova:
I have to fix the tests, now that we have URIs...

albertvillanova:
@severo any hint why the CI error on e2e? https://github.com/huggingface/datasets-server/actions/runs/4936661148/jobs/8824461485?pr=1158
```
FAILED tests/test_11_api.py::test_rows_endpoint - AssertionError: 500 - {'Server': 'nginx/1.20.2', 'Date': 'Wed, 10 May 2023 12:02:54 GMT', 'Content-Type': 'application/json', 'Transfer-Encoding': 'chunked', 'Connection': 'keep-alive', 'cache-control': 'max-age=10', 'x-error-code': 'UnexpectedError', 'content-encoding': 'gzip', 'vary': 'Accept-Encoding'} - {"error":"Unexpected error."} - http://localhost:8000
```

albertvillanova:
Thanks for the hints. I'm investigating it.

albertvillanova:
It's all green now...

services/api/src/api/routes/rows.py
```diff
@@ -67,19 +68,27 @@ class ParquetDataProcessingError(Exception):
...
+
+    Args:
+        paths (List[str]): List of paths.
+        dataset (str): The dataset name.
     Returns:
-        HfFileSystem: The parquet filesystem.
+        List[str]: List of Parquet URIs.
     """
-    return HfFileSystem(dataset, repo_type="dataset", revision=PARQUET_REVISION, token=hf_token)
+    return [f"hf://datasets/{dataset}@{quote_plus(PARQUET_REVISION)}/{path}" for path in paths]
```
severo, line 91: OK, it's hard to forge an URL (quote_plus). Shouldn't we have it in huggingface_hub instead? cc @Wauplin

severo (approved):

services/api/src/api/routes/rows.py
albertvillanova: We have `save_quote` in `huggingface_hub.hf_file_system`.

albertvillanova: Otherwise, we can use `HfFileSystemResolvedPath`.unresolve()`. Let me do the change.

severo: OK, but I meant: huggingface_hub could provide a way to forge the URLs

severo: OK :)  Note that I didn't ask you to change the PR in any way, it's perfect for now. It was a feature request for hfh

PR #1162, opened 2023-05-10
feat: 🎸 upgrade libcommon in all the code

severo:
following #1158


AndreaFrancis (approved):

albertvillanova (approved):
Thanks for the fix. 

Maybe we should write in our dev docs how to properly update "indirect" dependencies, so that we do not repeat the same mistake in the future.

severo:
Yes. Basically, if the poetry.lock has been changed in libcommon, we have to run `poetry update libcommon` in all the dependent apps, which currently means: jobs/ and services/.

Issue #1086, opened 2023-04-25
What's left in refactoring of processing steps and endpoints


polinaeterna:
Here I want to summarize once again the final changes that we need to make datasets-server consistent in naming and logic, see https://github.com/huggingface/datasets-server/issues/867 (because it's all scattered in comments to different issues)

## Rename steps

- [x] `/split-names-from-streaming` -> `config-split-names-from-streaming` https://github.com/huggingface/datasets-server/pull/1168
- [x] `/split-names-from-dataset-info` -> `config-split-names-from-info` https://github.com/huggingface/datasets-server/pull/1226
- [x] `/config-names` -> `dataset-config-names` https://github.com/huggingface/datasets-server/pull/1246

## Delete steps

- [x] `parquet-and-dataset-info` https://github.com/huggingface/datasets-server/pull/1043
- [x] `dataset-split-names-from-streaming` https://github.com/huggingface/datasets-server/pull/1106
- [x] `dataset-split-names-from-dataset-info` https://github.com/huggingface/datasets-server/pull/1141

## Endpoints
- [x] `/dataset-info` -> `/info` https://github.com/huggingface/datasets-server/pull/1468
- [x] delete `/parquet-and-dataset-info` https://github.com/huggingface/datasets-server/pull/1488
- [x] delete `/config-names`

## Misc
- [x] update the docs 
- [x] update the OpenAPI spec.

severo:
> /split-names-from-streaming -> split-names-from-streaming
> /split-names-from-dataset-info -> /split-names-from-info
> /config-names -> config-names

I think it's:

/split-names-from-streaming -> config-split-names-from-streaming
/split-names-from-dataset-info -> config-split-names-from-info
/config-names -> config-names -> dataset-config-names

severo:
> Rename endpoints - should we?
>
> /dataset-info -> /info
> /parquet-and-dataset-info -> /parquet-and-info - do we need this endpoint actually?

The endpoints are defined here: https://github.com/huggingface/datasets-server/blob/main/services/api/src/api/config.py#L118

OK to remove `/parquet-and-dataset-info`
OK to rename `/dataset-info` to `/info`

Maybe we could have a better name than "/opt-in-out-urls-scan"
And also: it's inconsistent to have "/config-names" but "/splits". Maybe rename "/config-names" to "/configs", or remove it altogether?

severo:
Finally: to finish this "cycle", we also have to update the docs and the OpenAPI spec.

[...]

PR #1168, opened 2023-05-11
Rename `/split-names-from-streaming` job runner

polinaeterna:
Part of https://github.com/huggingface/datasets-server/issues/1086 and https://github.com/huggingface/datasets-server/issues/867


jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
```diff
@@ -1,44 +1,23 @@
...
-split_names_tmp = "/split-names-TMP"
+from mongodb_migration.renaming_migrations import CacheRenamingMigration
 
 
 # connection already occurred in the main.py (caveat: we use globals)
-class MigrationCacheUpdateSplitNames(Migration):
+class MigrationCacheUpdateSplitNames(CacheRenamingMigration):
     def up(self) -> None:
-        logging.info(f"Rename cache_kind field from {split_names} to {split_names_from_streaming}")
+        tmp_kind = f"{self.cache_kind}-TMP"
```
polinaeterna, line 15: is the intermediate step actually required now? or this migration can also be refactored as simple `CacheRenamingMigration`?

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -193,4 +194,22 @@ def get_migrations(self) -> List[Migration]:
             MigrationQueueDeleteIndexWithoutRevision(
```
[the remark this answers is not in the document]
severo: on main, or in this branch?

jobs/mongodb_migration/src/mongodb_migration/renaming_migrations.py
```diff
@@ -0,0 +1,97 @@
...
+        db = get_db(self.MONGOENGINE_ALIAS)
+        db[self.COLLECTION_JOBS].update_many(
+            {"type": self.job_type},
+            [
+                {
+                    "$set": {
+                        "unicity_id": {
+                            "$replaceOne": {
+                                "input": "$unicity_id",
+                                "find": f"{self.job_type}",
```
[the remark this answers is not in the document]
severo: good question. I would say: it's not very important to bother about this, I think the only database to be migrated is our production one, and it has already migrated these steps.

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
severo: I don't know. The idea was to ensure that, in the case some cache entries already existed with the new kind, they would not be overwritten with older entries. But since the migration should be blocking the upgrade, thus: the new cache kind should not had been used already, I don't remember why we added this. @AndreaFrancis do you remind?

libs/libcommon/src/libcommon/constants.py
```diff
@@ -24,12 +24,11 @@
 PROCESSING_STEP_DATASET_OPT_IN_OUT_URLS_COUNT_VERSION = 2
 PROCESSING_STEP_DATASET_PARQUET_VERSION = 2
 PROCESSING_STEP_DATASET_SIZE_VERSION = 2
-PROCESSING_STEP_PARQUET_AND_DATASET_INFO_VERSION = 2
 PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_PARQUET_VERSION = 2
 PROCESSING_STEP_CONFIG_PARQUET_AND_INFO_VERSION = 3
 PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_STREAMING_VERSION = 3
 PROCESSING_STEP_SPLIT_NAMES_FROM_DATASET_INFO_VERSION = 3
-PROCESSING_STEP_SPLIT_NAMES_FROM_STREAMING_VERSION = 3
+PROCESSING_STEP_CONFIG_SPLIT_NAMES_FROM_STREAMING_VERSION = 4
```
severo, line 31: why do we upgrade the version? If I understand correctly, the job runner in itself has not changed, right? ie. what would we gain by running it again on all the splits?

libs/libcommon/tests/state/utils.py
```diff
@@ -69,9 +69,13 @@ def assert_dataset_state(
...
-        assert_equality(computed_cache_status[key], value, key)
-    assert_equality(dataset_state.queue_status.as_response(), queue_status, context="queue_status")
-    assert_equality(dataset_state.plan.as_response(), tasks, context="tasks")
+        assert_equality(computed_cache_status[key], sorted(value), key)
+    assert_equality(
+        dataset_state.queue_status.as_response(),
+        {key: sorted(value) for key, value in queue_status.items()},
+        context="queue_status",
+    )
+    assert_equality(dataset_state.plan.as_response(), sorted(tasks), context="tasks")
```
severo, line 78: ok. alternative was to move the rows to keep the alphabetic order in the test (I did it that way because I found it easier to read the tests if ordered). But it's ok that way, maybe it's too much maintenance burden for not much impact

severo (approved):
Thanks for bearing with these ungrateful issues of renaming steps, and for improving the migration jobs at the same time!

Just one thing before merging: I think we must not increment the job runner version. The other comments are not blocking

libs/libcommon/tests/state/utils.py
polinaeterna: i decided that it's inconvenient to always remember where to put a new name, or to move the same runner to a different line just because the name changed (like in this PR), and the order doesn't reflect the order of steps in the graph that way anyway. but i don't mind rolling back!

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
polinaeterna: I can just add this step to the common `CacheRenamingMigration` class just in case, but if @AndreaFrancis could explain what the case is exactly :D and if it's possible to have collisions like this, would be good for me to understand.
And actually, aren't new kind entries deleted anyway by this code and changed by renamed old ones if any?

libs/libcommon/src/libcommon/constants.py
polinaeterna: ah true, thanks for noticing! changed it

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230216112500_cache_split_names_from_streaming.py
severo: Code comes from https://github.com/huggingface/datasets-server/pull/873. I think (from reviewing the preceding commits) that it was a way to fix a migration we had done wrong: job runners were already creating entries with the new name, and we applied the migration later, which lead to conflicts. Now, it should not occur, so, please feel free to remove this extra step!

AndreaFrancis: So sorry, I missed this thread. Sylvain added it on https://github.com/huggingface/datasets-server/pull/873, I think a similar case of duplicated key happened when renaming /split-names to /split-names-from-streaming.
But for some reason, it does not happened when renaming /first-rows to /split-first-rows-from-streaming

PR #1177, opened 2023-05-12
Adding full_scan field in opt-in-out cache

AndreaFrancis:
According to PR https://github.com/huggingface/moon-landing/pull/6289/files, we will need full_scan flag for UI


lhoestq (approved):
Awesome ! lgtm :)

libs/libcommon/src/libcommon/constants.py
```diff
@@ -21,7 +21,7 @@
 PROCESSING_STEP_CONFIG_OPT_IN_OUT_URLS_COUNT_VERSION = 1
```
severo, line 21:
```suggestion
PROCESSING_STEP_CONFIG_OPT_IN_OUT_URLS_COUNT_VERSION = 2
```

severo: Otherwise the content will not be updated at config level, thus neither at dataset level, thus in the endpoint we use to check the number of opted-in/opted-out urls.

severo (approved):
OK. Just a detail about a missing version increase

libs/libcommon/src/libcommon/constants.py
AndreaFrancis: My mistake, thanks!

PR #1187, opened 2023-05-15
Set git revision at job creation

severo:
The proposal in the PR is to add a field `revision` to the jobs, at creation, and it must be non-null (it should be the commit hash).

This way, the job runners don't have to reach the hub to check for the current revision, and we're preparing for (one day) handle multiple revisions in the cache for the same dataset.

Meanwhile, the model is the following:
- we cannot have two jobs for the same dataset but with different revisions: if we create a new job with a new revision, the existing ones are canceled.
- the cache contains the revision in the field "dataset_git_revision", and it's still optional for backward compatibility reasons
- there is no semantics about the time or ordering relation between commits: if we create a job for a given revision, it will be used blindly and replace the cache entries even if this is an older commit.


jobs/mongodb_migration/tests/migrations/test_20230309141600_cache_add_job_runner_version copy.py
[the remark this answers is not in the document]
severo: no, it's an error!

jobs/mongodb_migration/tests/migrations/test_20230516101600_queue_delete_index_without_revision.py
```diff
@@ -0,0 +1,35 @@
...
+from mongodb_migration.migration import IrreversibleMigrationError
+from mongodb_migration.migrations._20230516101600_queue_delete_index_without_revision import (
+    INDEX_DEFINITION,
+    MigrationQueueDeleteIndexWithoutRevision,
+    get_index_names,
+)
+
+
+def test_queue_delete_index_without_revision(mongo_host: str) -> None:
+    with MongoResource(database="test_queue_delete_indexes_with_force", host=mongo_host, mongoengine_alias="queue"):
```
AndreaFrancis, line 18:
```suggestion
    with MongoResource(database="test_queue_delete_index_without_revision", host=mongo_host, mongoengine_alias="queue"):
```

services/worker/tests/test_job_manager.py
```diff
@@ -300,6 +299,7 @@ def test_doesnotexist(app_config: AppConfig) -> None:
         job_info=job_info, app_config=app_config, job_runner=job_runner, processing_graph=processing_graph
     )
 
-    assert not job_manager.process()
-    with pytest.raises(DoesNotExist):
```
AndreaFrancis, line 304: I think test name should change now that it does not throw an exception anymore

AndreaFrancis (approved):

jobs/mongodb_migration/tests/migrations/test_20230309141600_cache_add_job_runner_version copy.py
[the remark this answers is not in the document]
severo: fixed

services/worker/tests/test_job_manager.py
severo: Oh, I understand it's confusing, but "doesnotexist" referred to the fact that the dataset does not exist, not to the fact that the cache would not be created (thus raising DoesNotExist).
I'll let the name, even if now that we don't reach the Hub, the job manager does now that the dataset does not exist.

PR #1188, opened 2023-05-16
feat: 🎸 ignore result of job runner if job has been canceled

severo:
also: refactor to remove two queue methods (kill_zombies, kill_long_jobs): job_manager is now in charge of finishing the jobs, and updating the cache (if needed).


albertvillanova (approved):
Thank you.

services/worker/tests/test_job_manager.py
```diff
@@ -171,24 +192,24 @@ def test_job_runner_set_crashed(
     test_processing_step: ProcessingStep,
```
AndreaFrancis, line 192: Not part of this PR, but I should have changed the name of this test, maybe only test_set_crashed since the logic is now handled by job_manager and not job_runner.

AndreaFrancis (approved):
Thanks! Nice refactor.

services/worker/tests/test_job_manager.py
severo: OK... Let's change this when we see it again in a PR

lhoestq (approved):
nice ! :)

severo:
follow-up: https://github.com/huggingface/datasets-server/pull/1227

PR #1165, opened 2023-05-11
Remove "force" field from the queue Jobs

severo:
Now that all the jobs are created by the DatasetState.backfill() method, we don't want to skip jobs.

Another PR will remove the "skip job" mechanism

PR #1197, opened 2023-05-18
Set datetime types in admin ui

lhoestq:
to fix errors when duckdb tries to cast the columns like "started_at"

(already deployed on HF - I ran my tests there ^^')

### DIFF
--- a/front/admin_ui/app.py
+++ b/front/admin_ui/app.py
@@ -189,7 +189,6 @@ with gr.Blocks() as demo:
                     "config": job["config"],
                     "split": job["split"],
                     "namespace": job["namespace"],
-                    "force": job["force"],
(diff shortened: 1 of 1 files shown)


front/admin_ui/app.py
```diff
@@ -183,6 +189,7 @@ def get_dataset_status(token, dataset):
                     "config": job["config"],
                     "split": job["split"],
                     "namespace": job["namespace"],
+                    "force": job["force"],
```
AndreaFrancis, line 192: I think `force` field was removed on https://github.com/huggingface/datasets-server/pull/1165

severo (approved):
OK! Just think to remove "force" before merging

Issue #655, opened 2022-11-30
Add a new endpoint: /features


severo:
The features are already provided under /first-rows, but some clients, like Autotrain (cc @SBrandeis), are interested in having access to the features, without the first rows.

SBrandeis:
low priority though, as we already have the information with `/rows` where we need that ingo

severo:
`/first-rows`

SBrandeis:
Would it make sense to expose a "simplified" version of the `/first-rows` endpoint, that is easier to consume (especially from JS frontends)?

I understand the advantages of being transparent with the actual objects from `datasets`, but parsing the features can be challenging, especially when dealing with nested features (see https://github.com/huggingface/moon-landing/tree/main/server/lib/datasets).

We could imagine a new endpoint that does all the heavy lifting and outputs data in an easy-to-use format, wdyt? It would also be a way to centralize the processing logic for all dataset-viewer likes in the company.

cc @beurkinger & @lhoestq

severo:
As you mention, /first-rows returns rows and features, where features are an exact dump of the Features returned by the datasets library.

Previously, when /first-rows was called /rows, we were returning instead "columns", with a column type among:
```
"JSON", "BOOL", "INT", "FLOAT", "STRING", "IMAGE_URL", "RELATIVE_IMAGE_URL", "AUDIO_RELATIVE_SOURCES"
```

See https://github.com/huggingface/datasets-server/blob/b36e416cae5a1fb4a34c91d5723032c5f13a5bb3/libs/libutils/src/libutils/types.py#L4 or https://github.com/huggingface/moon-landing/pull/3650/files#diff-64a67deff17881efb7c2974c3d4cb47d9558c70a4d1e24ec7ed6f0c712d1f29aL9 for reference.

I decided to remove this intermediate layer and expose the original `datasets` features. But if needed we can add this kind of simplification of the types if needed.

severo:
Another option would be to offer Python and JavaScript code to parse the responses. Not sure what is the best.

beurkinger:
@severo @SBrandeis 

Not directly related to this PR I think (I'm actually curious to know about the "only the first row" stuff), but something that would be VERY nice to have is to have the `_type` of each cell direclty accessible, so we don't have to do stuff like 
```
features.find((f) => f.name === key)?.type._type === 'ClassLabel'
```
when parsing. It would also makes Cells typing/checking sooooo much easier.

[...]

PR #1199, opened 2023-05-18
Descriptive statistics

polinaeterna:
This PR introduces the following measurements/statistics:
### numerical columns (float and int):
- nan values count
- nan values percentage
- min
- max
- mean
- median
- std
- histogram: 
  - for float: fixed number of bins (which is a global config parameter - tell me if it's an overkill :D)
  - for integers: bin size is always an integer value. if difference between max and min is less then fixed number of bins parameter, bin size is equal to 1, otherwise it's `round((max-min)/num_bins)` - which means that bins might be not equal (like the last one being smaller) but otherwise there could be float bin edges which is nonsense.

### categorical columns
These are **only** `ClassLabel` columns which is not ideal because some integers also might be categories, as well as strings, but we can't be sure in advance.
- nan values count
- nan values percentage
- number of unique values
- counts for each value

<details><summary>Here's the result for my toy dataset with 1000 rows </summary>
<p>

[polinaeterna/delays_nans](https://huggingface.co/datasets/polinaeterna/delays_nans)

```python
{
    "num_examples": 100000,
    "statistics": [
        {
            "column_name": "class_col",
            "column_type": "class_label",
            "column_statistics": {
                "nan_count": 0,
                "nan_proportion": 0.0,
                "n_unique": 5,
                "frequencies": {
                    "this": 19834,
                    "are": 20159,
                    "random": 20109,
                    "words": 20172,
                    "test": 19726
                }
            }
        },
        {
            "column_name": "class_col_nans",
            "column_type": "class_label",
            "column_statistics": {
                "nan_count": 49972,
                "nan_proportion": 0.49972,
                "n_unique": 5,
                "frequencies": {
                    "this": 9904,
                    "are": 10021,
                    "random": 10126,
                    "words": 10061,
                    "test": 9916
                }
            }
        },
        {
            "column_name": "delay",
            "column_type": "float",
            "column_statistics": {
                "nan_count": 0,
                "nan_proportion": 0.0,
                "min": -10.206,
                "max": 8.48053,
                "mean": 2.10174,
                "median": 3.4012,
                "std": 3.12487,
                "histogram": {
                    "hist": [
                        2,
                        34,
                        256,
                        15198,
                        9037,
                        2342,
                        12743,
                        45114,
                        14904,
                        370
                    ],
                    "bin_edges": [
                        -10.206,
                        -8.33734,
                        -6.46869,
                        -4.60004,
                        -2.73139,
                        -0.86273,
                        1.00592,
                        2.87457,
                        4.74322,
                        6.61188,
                        8.48053
                    ]
                }
            }
        },
        {
            "column_name": "delay_nans",
            "column_type": "float",
            "column_statistics": {
                "nan_count": 49892,
                "nan_proportion": 0.49892,
                "min": -10.206,
                "max": 8.48053,
                "mean": 2.11288,
                "median": 3.4012,
                "std": 3.11722,
                "histogram": {
                    "hist": [
                        1,
                        17,
                        137,
                        7515,
                        4522,
                        1197,
                        6481,
                        22593,
                        7473,
                        172
                    ],
                    "bin_edges": [
                        -10.206,
                        -8.33734,
                        -6.46869,
                        -4.60004,
                        -2.73139,
                        -0.86273,
                        1.00592,
                        2.87457,
                        4.74322,
                        6.61188,
                        8.48053
                    ]
                }
            }
        },
        {
            "column_name": "temp",
            "column_type": "float",
            "column_statistics": {
                "nan_count": 0,
                "nan_proportion": 0.0,
                "min": 0.6,
                "max": 14.9,
                "mean": 7.28953,
                "median": 7.5,
                "std": 3.05441,
                "histogram": {
                    "hist": [
                        4711,
                        9781,
                        10349,
                        9781,
                        20166,
                        17046,
                        13965,
                        8794,
                        4055,
                        1352
                    ],
                    "bin_edges": [
                        0.6,
                        2.03,
                        3.46,
                        4.89,
                        6.32,
                        7.75,
                        9.18,
                        10.61,
                        12.04,
                        13.47,
                        14.9
                    ]
                }
            }
        },
        {
            "column_name": "temp_nans",
            "column_type": "float",
            "column_statistics": {
                "nan_count": 49959,
                "nan_proportion": 0.49959,
                "min": 0.6,
                "max": 14.9,
                "mean": 7.29404,
                "median": 7.5,
                "std": 3.05107,
                "histogram": {
                    "hist": [
                        2392,
                        4855,
                        5143,
                        4879,
                        10040,
                        8571,
                        7058,
                        4452,
                        2014,
                        637
                    ],
                    "bin_edges": [
                        0.6,
                        2.03,
                        3.46,
                        4.89,
                        6.32,
                        7.75,
                        9.18,
                        10.61,
                        12.04,
                        13.47,
                        14.9
                    ]
                }
            }
        },
        {
            "column_name": "vehicle_type",
            "column_type": "int",
            "column_statistics": {
                "nan_count": 0,
                "nan_proportion": 0.0,
                "min": 0,
                "max": 2,
                "mean": 0.82646,
                "median": 1.0,
                "std": 0.72333,
                "histogram": {
                    "hist": [
                        36343,
                        44668,
                        18989
                    ],
                    "bin_edges": [
                        0,
                        1,
                        2,
                        2
                    ]
                }
            }
        },
        {
            "column_name": "vehicle_type_nans",
            "column_type": "int",
            "column_statistics": {
                "nan_count": 50247,
                "nan_proportion": 0.50247,
                "min": 0,
                "max": 2,
                "mean": 0.82542,
                "median": 1.0,
                "std": 0.72384,
                "histogram": {
                    "hist": [
                        18135,
                        22169,
                        9449
                    ],
                    "bin_edges": [
                        0,
                        1,
                        2,
                        2
                    ]
                }
            }
        },
        {
            "column_name": "weekday",
            "column_type": "int",
            "column_statistics": {
                "nan_count": 0,
                "nan_proportion": 0.0,
                "min": 0,
                "max": 6,
                "mean": 3.08063,
                "median": 3.0,
                "std": 1.90347,
                "histogram": {
                    "hist": [
                        10282,
                        15416,
                        15291,
                        15201,
                        15586,
                        15226,
                        12998
                    ],
                    "bin_edges": [
                        0,
                        1,
                        2,
                        3,
                        4,
                        5,
                        6,
                        6
                    ]
                }
            }
        },
        {
            "column_name": "weekday_nans",
            "column_type": "int",
            "column_statistics": {
                "nan_count": 50065,
                "nan_proportion": 0.50065,
                "min": 0,
                "max": 6,
                "mean": 3.07762,
                "median": 3.0,
                "std": 1.90272,
                "histogram": {
                    "hist": [
                        5136,
                        7695,
                        7711,
                        7550,
                        7749,
                        7646,
                        6448
                    ],
                    "bin_edges": [
                        0,
                        1,
                        2,
                        3,
                        4,
                        5,
                        6,
                        6
                    ]
                }
            }
        }
    ]
}
```

</p>
</details>


lhoestq:
we could also use duckdb (should be faster and scale well) but afaik it doesn't implement numeric_histogram

polinaeterna:
@severo @lhoestq @AndreaFrancis can you please take a look at the current implementation? It's not as fast as I was hoping to for bigger datasets unfortunately but I believe it's because I didn't cook `duckdb` well :D and it might be improved.

The main problem here is that statistics for all features are computed in a single job. For bigger datasets that might take a long time but if it breaks in the middle for some feature, all the results for previously computed features will be lost (basically the same problem as with dataset-level `parquet-and-info` and [Spawning urls scan](https://github.com/huggingface/datasets-server/issues/1087)). Maybe not a big deal for now when we set the limit? But we should take into account for future. (An obvious solution is to add another - `feature` - level to workers but not sure it's a good idea to add more and more granularity levels.)  

Not marking the PR as ready yet but want to know your opinion.

lhoestq (commented):
Cool ! Do you know which steps may take too much time ?

polinaeterna:
@lhoestq for numerical stuff, but there are also more steps (three queries vs one query for categories now). I just checked for 
[amazon_us_reviews](https://huggingface.co/datasets/amazon_us_reviews), `Apparel_v1_00` config - it's 6M rows, about 1Gb parquet files size:

```
2023-06-16 13:42:17,790 INFO root Compute descriptive statistics for dataset='polinaeterna/amazon_us_reviews', config='Apparel_v1_00', split='train'
2023-06-16 13:42:17,791 INFO root Compute statistics for categorical features
2023-06-16 13:42:17,797 DEBUG root Compute statistics for ClassLabel feature **vine**
2023-06-16 13:43:25,743 DEBUG root ClassLabel **vine** computed
2023-06-16 13:43:25,743 DEBUG root Compute statistics for ClassLabel feature **verified_purchase**
2023-06-16 13:45:07,228 DEBUG root ClassLabel **verified_purchase** computed
100%|██████████| 2/2 [02:49<00:00, 84.72s/it]
2023-06-16 13:45:07,228 INFO root Compute min, max, mean, median, std, histogram for numerical features. 
2023-06-16 13:45:07,228 DEBUG root Compute min, max ... for **star_rating**
2023-06-16 13:46:36,710 DEBUG root min, max ... for **star_rating** computed
2023-06-16 13:46:36,710 DEBUG root Compute nans for **star_rating**
2023-06-16 13:47:43,033 DEBUG root Nans for **star_rating** computed
2023-06-16 13:47:43,033 DEBUG root Compute histogram for **star_rating**
2023-06-16 13:48:49,270 DEBUG root histogram for **star_rating** computed
```
and then I broke because of HTTP error :( and I encounter this pretty frequently actually and wasn't able to debug it for a long time until the last bugfix release and [this fix](https://github.com/duckdb/duckdb/pull/7683) - it wasn't throwing an underlying error

services/worker/src/worker/job_runners/split/descriptive_stats.py
```diff
@@ -0,0 +1,288 @@
...
+
+def compute_histogram(
+    con: duckdb.DuckDBPyConnection,
+    column_name: str,
+    urls: List[str],
+    bin_size: int,
+) -> Histogram:
+    hist_query = f"""
+    SELECT FLOOR("{column_name}"/{bin_size})*{bin_size}, COUNT(*) FILTER (WHERE {column_name} IS NOT NULL)
+     FROM read_parquet({urls}) GROUP BY 1 ORDER BY 1;
```
AndreaFrancis, line 75: Not sure if it would help, but I am seeing that everywhere it is reading the parquets from HTTP, would it be helpful if first we copy the info in a local db? In the future, this step and `split-duckdb-index` could depend on another one where first: all data is ingested in a duckdb db file and then, everything is run against duckdb engine directly. 
Just a random idea (feel free to ignore it)

services/worker/src/worker/job_runners/split/descriptive_stats.py
```diff
@@ -0,0 +1,288 @@
...
+        minimum, maximum, mean, median, std = np.round([minimum, maximum, mean, median, std], DECIMALS).tolist()
+    elif dtype in INTEGER_DTYPES:
+        if maximum - minimum < n_bins:
+            bin_size = 1
+        else:
+            bin_size = int(np.round((maximum - minimum) / n_bins))
+        mean, median, std = np.round([mean, median, std], DECIMALS).tolist()
+    else:
+        raise ValueError("Incorrect dtype, only integers and float are allowed. ")
+    nan_query = f"SELECT COUNT(*) FILTER (WHERE {column_name} IS NULL) FROM read_parquet({urls});"
```
AndreaFrancis, line 113: Not sure if this will apply to all datasets but I did a small test of different performances for:
`EXPLAIN ANALYZE SELECT COUNT(*) FROM read_parquet("https://huggingface.co/datasets/glue/resolve/refs%2Fconvert%2Fparquet/ax/glue-test.parquet") WHERE premise IS NULL
`and
`EXPLAIN ANALYZE SELECT COUNT(*) FROM read_parquet("https://huggingface.co/datasets/glue/resolve/refs%2Fconvert%2Fparquet/ax/glue-test.parquet") WHERE premise IS NULL;
`
 And looks like the second one is faster:
First result: Total Time: 0.382s
Second result: Total Time: 0.0001s

Maybe EXPLAIN ANALYZE command can help you identify critical steps https://duckdb.org/dev/profiling.html

AndreaFrancis (commented):
I see it really great! 
I left a couple of minor comments/suggestions, feel free to ignore them for this step.

services/worker/src/worker/job_runners/split/descriptive_stats.py
```diff
@@ -0,0 +1,352 @@
...
+def compute_histogram(
+    con: duckdb.DuckDBPyConnection,
+    column_name: str,
+    parquet_filename: Path,
+    bin_size: int,
+    min_value: Union[int, float],
+    n_bins: int,
+    n_samples: Optional[int] = None,
+) -> Histogram:
+    hist_query = f"""
```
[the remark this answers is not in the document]
polinaeterna: also can't figure out how to avoid possible sql injection here and in all queries below. I need to pass `column_name` as a parameter, but when doing something like 
```
con.execute("select $column_name" from df ...", {"column_name": column_name})
```
`column_name` works like a string value, not like a column reference (like it's `select 'column_name' from df`, not `select column_name from df`).

polinaeterna:
@lhoestq @AndreaFrancis @severo can you please take a look? I've left comments where I'm not sure what to do. Basically I'm struggling with avoiding sql injections, hope you can advice what to do

severo:
Sure! Looking at it.

libs/libcommon/src/libcommon/config.py
```diff
@@ -318,6 +351,15 @@ class ProcessingGraphConfig:
                 ],
                 "job_runner_version": PROCESSING_STEP_DATASET_SPLIT_NAMES_VERSION,
             },
+            "split-descriptive-stats": {
+                "input_type": "split",
+                "triggered_by": [
+                    "config-split-names-from-info",
+                    "config-split-names-from-streaming",
+                    "config-parquet-and-info",
```
lhoestq, line 359: why not just by "config-parquet-and-info" ?

services/worker/src/worker/job_runners/split/descriptive_stats.py
```diff
@@ -0,0 +1,352 @@
...
+        stats.append(
+            StatsPerColumnItem(
+                column_name=feature_name,
+                column_type="float" if feature_dtype in FLOAT_DTYPES else "int",
+                column_dtype=feature_dtype,
+                column_stats=num_column_stats,
+            )
+        )
+
+    return SplitDescriptiveStatsResponse(
```
AndreaFrancis, line 312: Don't forget to `con.close()` at the end. It might not be a problem in this case but as a best practice, I think it is better to prevent it.

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -65,18 +69,7 @@ def compute_index_rows(
...
-        kinds=["config-split-names-from-streaming", "config-split-names-from-info"], dataset=dataset, config=config
-    )
-    try:
-        splits_content = split_names_best_response.response["content"]["splits"]
-    except Exception as e:
-        raise PreviousStepFormatError("Previous step did not return the expected content.", e) from e
-
-    if split not in [split_item["split"] for split_item in splits_content]:
-        raise SplitNotFoundError(f"The split '{split}' does not exist for the config '{config}' of the dataset.")
+    check_split_exists(dataset=dataset, config=config, split=split)
```
AndreaFrancis, line 72: Thanks! Nice refactor!

AndreaFrancis:
Great @polinaeterna, really nice PR!  I put a couple of minor/non critical comments

severo:
re

> for float: fixed number of bins (which is a global config parameter - tell me if it's an overkill :D)

Perfect for now. There is plenty of room for configuration if we want, see https://observablehq.com/plot/transforms/bin for example:

Capture d’écran 2023-06-29 à 18 54 33

severo:
> These are only ClassLabel columns which is not ideal because some integers also might be categories, as well as strings, but we can't be sure in advance.

also taking example from Observable: https://observablehq.com/@observablehq/data-table-cell. Strings are considered as categories.

services/worker/src/worker/job_runners/split/descriptive_stats.py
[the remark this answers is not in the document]
polinaeterna: > I think it's not really possible to pass anything bad in column names of a dataset or a parquet filename

no, actually it is, I thought we don't allow spaces and other symbols in column names but apparently we do :D 

image

severo:
> An obvious solution is to add another - feature - level to workers but not sure it's a good idea to add more and more granularity levels

what is your opinion after two weeks? I think it would make sense... a column, in columnar data, is a well-defined concept.

Another comment, related to the columns too: it would be good to have docs about what we support (column type, eg: we have issues with datetime) and which statistics we run (description of the stats, format)

chart/templates/_envWorker.tpl
```diff
@@ -110,4 +110,13 @@
...
 - name: DUCKDB_INDEX_EXTENSIONS_DIRECTORY
   value: "/tmp/duckdb-extensions"
+# specific to 'split-descriptive-stats'
+- name: DESCRIPTIVE_STATS_HISTOGRAM_NUM_BINS
+  value: {{ .Values.descriptiveStats.histogramNumBins | quote }}
+- name: DESCRIPTIVE_STATS_MAX_PARQUET_SIZE_BYTES
+  value: {{ .Values.descriptiveStats.maxParquetSizeBytes | quote }}
+- name: DESCRIPTIVE_STATS_STORAGE_DIRECTORY
+  value: {{ .Values.descriptiveStats.storageDirectory | quote }}
+- name: DESCRIPTIVE_STATS_EXTENSIONS_DIRECTORY
```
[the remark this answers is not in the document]
severo: what is stored inside? could there be conflicts? will the data inside be persisted after a job finishes?

libs/libcommon/src/libcommon/config.py
severo: as for https://github.com/huggingface/datasets-server/pull/1441#discussion_r1244897544, maybe we can remove "config-parquet-and-info" to save resources (ie run the step a bit less often)

chart/templates/_envWorker.tpl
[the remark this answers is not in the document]
AndreaFrancis: Maybe there is no conflict since duckdb tries to download the extensions only if they don't exist yet.

services/worker/src/worker/job_runners/split/descriptive_stats.py
```diff
@@ -0,0 +1,355 @@
...
+    con.sql("INSTALL httpfs")
+    con.sql("LOAD httpfs")
+    con.sql("SET enable_progress_bar=true;")
+
+    # store data as local parquet file for fast querying
+    local_parquet_path = (
+        Path(local_parquet_directory) / PARQUET_FILENAME if local_parquet_directory else Path(PARQUET_FILENAME)
+    )
+    logging.info(f"Copying remote data to a local parquet file {local_parquet_path}. ")
+    con.sql(f"COPY (SELECT * FROM read_parquet({parquet_files_urls})) TO '{local_parquet_path}' (FORMAT PARQUET);")
```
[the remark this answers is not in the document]
severo: BTW: if at one point we only do the stats per column (new step level), we could only download the data for that column

services/worker/src/worker/job_runners/split/descriptive_stats.py
```diff
@@ -0,0 +1,355 @@
...
+            feature_name,
+            parquet_filename=local_parquet_path,
+            n_bins=histogram_num_bins,
+            n_samples=num_examples,
+            dtype=feature_dtype,
+        )
+        stats.append(
+            StatsPerColumnItem(
+                column_name=feature_name,
+                column_type="float" if feature_dtype in FLOAT_DTYPES else "int",
```
severo, line 308: are "float" and "int" part of a new vocabulary that we're providing? If so, it should be well documented. Maybe related to the discussion in https://github.com/huggingface/datasets-server/issues/655 (having a "simpler" set of column types, because features, with type and dtype, is a bit complex)

services/worker/src/worker/job_runners/split/descriptive_stats.py
```diff
@@ -0,0 +1,355 @@
...
+    frequencies, nan_count = {}, 0
+    for cat_id, freq in categories:
+        if cat_id is not None:
+            frequencies[class_label_names[cat_id]] = freq
+        else:
+            nan_count = freq
+    nan_prop = np.round(nan_count / n_samples, DECIMALS).item() if nan_count != 0 else 0.0
+    return CategoricalStatsItem(
+        nan_count=nan_count,
+        nan_prop=nan_prop,
```
severo, line 182: could we rename to `nan_proportion`?

services/worker/tests/job_runners/split/test_descriptive_stats.py
```diff
@@ -0,0 +1,263 @@
...
+    ) -> SplitDescriptiveStatsJobRunner:
+        processing_step_name = SplitDescriptiveStatsJobRunner.get_job_type()
+        processing_graph = ProcessingGraph(
+            {
+                "dataset-config-names": {"input_type": "dataset"},
+                "config-split-names-from-info": {
+                    "input_type": "config",
+                    "triggered_by": "dataset-config-names",
+                },
+                # "config-split-names-from-streaming": {
```
severo, line 46: remove dead code

services/worker/tests/job_runners/split/test_descriptive_stats.py
```diff
@@ -0,0 +1,263 @@
...
+                "nan_count": 8,
+                "nan_prop": 0.4,
+                "min": 0.2,
+                "max": 9.9,
+                "mean": 5.09167,
+                "median": 4.9,
+                "std": 3.87497,
+                "histogram": {
+                    "hist": [3, 0, 2, 0, 1, 1, 0, 0, 2, 3],
+                    "bin_edges": [0.2, 1.17, 2.14, 3.11, 4.08, 5.05, 6.02, 6.99, 7.96, 8.93],
```
severo, line 197: could we have more tests on the histogram:
- negative values (min=-10, max=-5)
- crossing zero (min=-5, max=5)
- far from zero (min=50,max=60)

services/worker/src/worker/job_runners/split/descriptive_stats.py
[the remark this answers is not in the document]
lhoestq: > no, actually it is, I thought we don't allow spaces and other symbols in column names but apparently we do :D

Ah nice xD But yea anyway we can see later how to fix the sql, it's not that important imo (just add a TODO maybe ?)

chart/templates/_helpers.tpl
```diff
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# Copyright 2022 The HuggingFace Authors.
+# Copyright 2023 The HuggingFace Authors.
```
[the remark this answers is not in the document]
lhoestq: people usually keep the date of the file creation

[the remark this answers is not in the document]
severo: Yes, it seems simpler. Let's keep the existing dates as is, and use the current year for new files

chart/env/prod.yaml
```diff
@@ -106,6 +106,8 @@ optInOutUrlsScan:
   rowsMaxNumber: 100_000
   urlsNumberPerBatch: 1000
 
+descriptiveStats:
+  maxParquetSizeBytes: "1_000_000_000"
```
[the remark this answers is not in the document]
polinaeterna: I actually just aligned the value with duckdb-index, for some reason I thought that it also uses 1Gb. But yes, I think here it can be increased to 5G.

services/api/src/api/config.py
```diff
@@ -148,6 +148,7 @@ class EndpointConfig:
             "/is-valid": {
                 "dataset": ["dataset-is-valid"],
             },
+            "/stats": {"split": ["split-descriptive-stats"]},
```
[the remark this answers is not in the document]
polinaeterna: and change step name to `SplitDescriptiveStatistics` too?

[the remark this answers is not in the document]
severo: yes; Or `SplitStatistics` if it's too long?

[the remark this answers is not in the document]
polinaeterna: I like the word "descriptive" :D and at first I was keeping in mind that we would be adding other more complex computations in the future, maybe as separate steps (but this is not clear for now if it's gonna be separate steps actually).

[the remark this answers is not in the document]
severo: OK for me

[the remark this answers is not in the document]
severo: or just ignore this for the code, and only use full words for the public API, if it's really too ugly

libs/libcommon/src/libcommon/config.py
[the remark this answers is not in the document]
AndreaFrancis: I sent  a PR to move DuckdbIndexConfig https://github.com/huggingface/datasets-server/pull/1474

libs/libcommon/src/libcommon/config.py
```diff
@@ -299,6 +300,14 @@ class ProcessingGraphConfig:
                 "job_runner_version": PROCESSING_STEP_DATASET_SPLIT_NAMES_VERSION,
                 "difficulty": 20,
             },
+            "split-descriptive-stats": {
+                "input_type": "split",
+                "triggered_by": [
+                    "config-split-names-from-info",
+                    "config-split-names-from-streaming",
+                ],
+                "job_runner_version": PROCESSING_STEP_SPLIT_DESCRIPTIVE_STATISTICS_VERSION,
```
lhoestq, line 309: you can add the difficulty here :)

I'd say 70 ?
```suggestion
                "job_runner_version": PROCESSING_STEP_SPLIT_DESCRIPTIVE_STATISTICS_VERSION,
                "difficulty": 70,
```

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -0,0 +1,412 @@
...
+        n_unique=len(categories) - 1 if nan_count else len(categories),
+        frequencies=frequencies,
+    )
+
+
+def compute_descriptive_statistics_response(
+    dataset: str,
+    config: str,
+    split: str,
+    local_parquet_directory: Optional[Path],
```
lhoestq, line 247: If I understand correctly it doesn't need to be Optional, since the SplitDescriptiveStatisticsJobRunner always provides the local_parquet_directory

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -0,0 +1,412 @@
...
+        return PROCESSING_STEP_SPLIT_DESCRIPTIVE_STATISTICS_VERSION
+
+    def compute(self) -> CompleteJobResult:
+        return CompleteJobResult(
+            compute_descriptive_statistics_response(
+                dataset=self.dataset,
+                config=self.config,
+                split=self.split,
+                local_parquet_directory=self.cache_subdirectory,
+                extensions_directory=self.descriptive_statistics_config.extensions_directory,
```
lhoestq, line 408: Maybe extensions_directory can be a path inside local_parquet_directory ?
This way no need to define a second directory, and avoid having the optional parameter `extensions_directory `

services/worker/tests/job_runners/split/test_descriptive_statistics.py
```diff
@@ -0,0 +1,283 @@
...
+            }
+    return expected_statistics
+
+
+@pytest.mark.parametrize(
+    "hub_dataset_name,expected_error_code",
+    [
+        ("descriptive_statistics", None),
+        # ("audio", "NoSupportedFeaturesError"),
+        # ("big", "SplitWithTooBigParquetError"),
```
lhoestq, line 221: I think you can re-add them ?

lhoestq (commented):
Awesome ! It looks all good to me. I just added a minor comments about `local_parquet_directory` that doesn't need to be Optional IMO

chart/templates/_helpers.tpl
```diff
@@ -187,6 +187,15 @@ The duckdb-index/ subpath in the NFS
 {{- printf "%s/%s/%s/" .Chart.Name .Release.Name "duckdb-index" }}
 {{- end }}
 
+{{/*
+The stats-cache/ subpath in the NFS
```
AndreaFrancis, line 191: @severo  well we need to have  this subdirectory to EFS as well as in https://github.com/huggingface/datasets-server/pull/1495?

AndreaFrancis (approved):
Awesome!

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -0,0 +1,412 @@
...
+    con.sql("LOAD httpfs")
+    con.sql("SET enable_progress_bar=true;")
+
+    # store data as local parquet file for fast querying
+    local_parquet_path = (
+        Path(local_parquet_directory) / PARQUET_FILENAME if local_parquet_directory else Path(PARQUET_FILENAME)
+    )
+    logging.info(f"Downloading remote data to a local parquet file {local_parquet_path}. ")
+    con.sql(
+        COPY_PARQUET_DATA_COMMAND.format(parquet_files_urls=parquet_files_urls, local_parquet_path=local_parquet_path)
```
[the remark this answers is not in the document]
polinaeterna: oops i think i'm not. how/where should i do this?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: sounds good. separate `extensions_directory` came from `split-duckdb-index` but maybe it's really not needed, @AndreaFrancis ?

chart/templates/_helpers.tpl
polinaeterna: could it be in the same volume as `duckdb-index` (`duckdb-data` if I understand it right...)?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
AndreaFrancis: I think in your case it could be local_parquet_directory. In `split-duckdb-index` I didn't have another alternative folder (besides duckdb_index_file_directory root).

services/worker/src/worker/job_runners/split/descriptive_statistics.py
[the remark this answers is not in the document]
AndreaFrancis: Downloaded parquet files should be deleted after processing stats. Maybe at the end of job runner execution? or maybe in post_compute method?

services/worker/src/worker/job_runners/split/descriptive_stats.py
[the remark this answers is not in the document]
polinaeterna: > is this faster than downloading the files using huggingface_hub?

I don't know how to do this in a straightforward way tbh... But current query should be fast

services/worker/src/worker/job_runners/split/descriptive_statistics.py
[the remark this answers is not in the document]
polinaeterna: as discussed on slack, `JobRunnerWithCache` automatically deletes the cache dir so this should be fine

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: this is because `cache_subdirectory` attribute of `JobRunnerWithCache` is [typed as `Optional[Path]`](https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/_job_runner_with_cache.py#L23), not sure what to do, here it really doesn't look logical I agree.

polinaeterna: i mean, `mypy` gives error if make it just `Path`, I can put `type: ignore` ofc

chart/templates/_helpers.tpl
polinaeterna: I've created (hopefully correctly... ?) a separate volume for now, though don't know if it's worth it cc @severo

severo: OK. I think it's a good idea to have created a separate volume. Let's monitor, and if it proves to be overkill, we can merge afterward.

chart/env/prod.yaml
```diff
@@ -93,6 +93,9 @@ parquetPersistence:
 duckdbPersistence:
   existingClaim: "datasets-server-duckdb-pvc"
 
+statisticsPersistence:
+  existingClaim: "datasets-server-statistics-pvc"
```
severo, line 97: Is there something needed on infra to create this pvc?

severo: cc @AndreaFrancis

libs/libcommon/src/libcommon/constants.py
```diff
@@ -6,6 +6,7 @@
 CACHE_MONGOENGINE_ALIAS = "cache"
 CACHED_ASSETS_CACHE_APPNAME = "datasets_server_cached_assets"
 PARQUET_METADATA_CACHE_APPNAME = "datasets_server_parquet_metadata"
+DESCRIPTIVE_STATISTICS_CACHE_APPNAME = "datasets_server_statistics"
```
severo, line 9: ultra-nit

```suggestion
DESCRIPTIVE_STATISTICS_CACHE_APPNAME = "datasets_server_descriptive_statistics"
```

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: Yes, it should not be optional. See https://github.com/huggingface/datasets-server/blob/0691cb7af3cd3a2293ad0120c684bb686f7aa804/services/worker/src/worker/job_runners/split/duckdb_index.py#L281-L282

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -0,0 +1,409 @@
...
+            "No columns for statistics computation found. Currently supported feature types are: "
+            f"{NUMERICAL_DTYPES} and ClassLabel. "
+        )
+
+    con = duckdb.connect(":memory:")  # we don't load data in local db file, use local parquet file instead
+    # configure duckdb extensions
+    con.sql(f"SET extension_directory='{local_parquet_directory}';")
+    con.sql("INSTALL httpfs")
+    con.sql("LOAD httpfs")
+    con.sql("SET enable_progress_bar=true;")
```
severo, line 315: does it generate logs? with which level? not important now, but we might have to monitor to avoid flooding the production logs

severo (approved):
I think the dockerfiles should also be updated (as well as the github actions?)

Also: the e2e tests are missing.

Alternatively, we can do all of this in a following PR if you want to finish this today, and somebody from the team can do it.

chart/env/prod.yaml
AndreaFrancis: Yes, I think we need these assets to be created for the new claim like https://github.com/huggingface/infra/blob/main/projects/hub/kubernetes_app_datasets_server_duckdb_efs.tf. Maybe @rtrompier can help us with it?

rtrompier: New volume is created and deployed on production (`hub-prod-us-east-1` cluster).

polinaeterna:
@lhoestq i can't figure this out either, first I thought maybe because statistics takes too much time but for the same dataset with token it works :thinking:

lhoestq:
Thanks docker

```
#0 30.32   • Installing types-pillow (9.5.0.4)
#0 30.80   • Installing types-psutil (5.9.5.13)
#0 31.03   • Installing types-requests (2.31.0.1)
#0 31.15   • Installing wget (3.2)
#0 38.14 
#0 38.14   AssertionError
#0 38.14 
#0 38.14   
#0 38.14 
#0 38.14   at /usr/local/lib/python3.9/site-packages/poetry/installation/executor.py:742 in _download_link
#0 38.16       738│         # to the original archive.
#0 38.16       739│         archive = self._chef.get_cached_archive_for_link(link, strict=False)
#0 38.16       740│         # 'archive' can at this point never be None. Since we previously downloaded
#0 38.16       741│         # an archive, we now should have something cached that we can use here
#0 38.16     → 742│         assert archive is not None
#0 38.16       743│ 
#0 38.16       744│         if archive.suffix != ".whl":
#0 38.16       745│             message = (
#0 38.16       746│                 f"  • {self.get_operation_message(operation)}:"
#0 38.16 
#0 65.25   • Installing libcommon (0.6.8 /src/libs/libcommon)
------
failed to solve: executor failed running [/bin/sh -c poetry install --no-cache]: exit code: 1
make[1]: *** [up] Error 17
make: *** [test] Error 2`
```

lhoestq:
I have the same error on `main` so probably just an environment error on my side - can someone else try to run the e2e tests locally ?

polinaeterna:
@severo @lhoestq @AndreaFrancis is the tests are passed now, can we merge?

lhoestq (approved):
Looks all good !! :D

severo:
Yes, let's merge!  (we can add the e2e tests later, I'll take care of it)

severo:
🎉🎉🎉 GOOD JOB!

PR #1203, opened 2023-05-19
Again: ignore result of job runner if job has been canceled

severo:
First PR: #1188
Reverted by #1196

New try. First, I get the code again, and then will commit the fix, once I find the issue.

To investigate, the commit has been deployed in production between Wed May 17 2023 18:11:12 GMT+0200 and Thu May 18 2023 15:49:11 GMT+0200.

### DIFF
--- a/libs/libcommon/src/libcommon/queue.py
+++ b/libs/libcommon/src/libcommon/queue.py
@@ -8,7 +8,7 @@ from collections import Counter
 from datetime import datetime, timedelta
 from itertools import groupby
 from operator import itemgetter
-from typing import Generic, List, Literal, Optional, Type, TypedDict, TypeVar
+from typing import Generic, List, Optional, Type, TypedDict, TypeVar
 
 import pytz
 from mongoengine import Document, DoesNotExist
@@ -357,7 +357,7 @@ class Queue:
                 status=Status.WAITING, namespace__nin=set(started_job_namespaces), priority=priority, **filters
             )
             .order_by("+created_at")
-            .only("type", "dataset", "revision", "config", "split")
+            .only("type", "dataset", "revision", "config", "split", "priority")
             .no_cache()
             .first()
         )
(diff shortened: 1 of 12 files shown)


severo:
To investigate, the commit has been deployed in production between Wed May 17 2023 18:11:12 GMT+0200 and Thu May 18 2023 15:49:11 GMT+0200.

Indeed, the volume of worker's logs speak for itself: the workers have not done much during that period:

Capture d’écran 2023-05-19 à 14 05 46

Extract of the logs:

```
INFO: 2023-05-18 00:38:22,171 - root - get config-opt-in-out-urls-count for dataset=allenai/nllb config=bho_Deva-kas_Arab
INFO: 2023-05-18 00:38:22,171 - root - [config-opt-in-out-urls-count] compute JobManager(job_id=64649094cc2847d3c6cda88f dataset=allenai/nllb job_info={'job_id': '64649094cc2847d3c6cda88f', 'type': 'config-opt-in-out-urls-count', 'params': {'dataset': 'allenai/nllb', 'revision': 'main', 'config': 'bho_Deva-kas_Arab', 'split': None}, 'priority': <Priority.NORMAL: 'normal'>}
```

In normal mode, the logs don't help much, we cannot see what occurs at queue level

---

What is clear is that we don't have any `WARNING: 2023-05-19 12:13:25,521 - root - job /config-names,boligmappa/home-standard-results has a non-empty finished_at field. Force finishing anyway.` message during that period, meanwhile we had a lot the rest of the time (it seems to be nearly the "normal" way to work).

Capture d’écran 2023-05-19 à 14 13 51

So, the underlying issue could be that when a job finished, in general, it has already been cancelled! As upsert_job currently works by cancelling all the similar jobs, then creating a new one, it means that another process has asked to create that job again.

---

In reality, we have 7 occurrences of this message, I'm not sure what this means, because we have a lot of started jobs during that time. Are the job just not existing anymore, so: queue.finish_job() returns immediately?

Capture d’écran 2023-05-19 à 15 03 14

Even with the new code, the message "WARNING ... has a non-empty finished_at field. Force finishing anyway.` should still appear.

severo:
I cannot reproduce on dev env

severo:
Closing. I'll do smaller PRs

PR #1213, opened 2023-05-19
feat: Part #3: Adding "partition" granularity level logic

AndreaFrancis:
First part of approach # 2 of https://github.com/huggingface/datasets-server/issues/1087
Adding a new granularity level - "Partition", will imply to also have a new field in Job and Cache.
Depends on https://github.com/huggingface/datasets-server/pull/1263, https://github.com/huggingface/datasets-server/pull/1259
and https://github.com/huggingface/datasets-server/pull/1260


severo:
I think it's a good idea, but I would like to avoid merging it for now, because we already struggle a lot with the jobs. Adding a new level might make the number of jobs explode.

About the concept of partitions: it would be chunks of 100_000 rows, right? If so, I think we should hardcode it, instead of using an environment variable, because changing the value after some time would break everything, right? If we really need multiple values, we might hardcode a scale, ie: 1_000, 100_000, 10_000_000... and let set an environment variable to chose the granularity among the fixed values.

AndreaFrancis:
> I think it's a good idea, but I would like to avoid merging it for now because we already struggle a lot with the jobs. Adding a new level might make the number of jobs explode.

Ok no problem, full scan will be on hold for now. 

> About the concept of partitions: it would be chunks of 100_000 rows, right? 

Yes
> If so, I think we should hardcode it, instead of using an environment variable, because changing the value after some time would break everything, right? 

Currently, there is no logic that depends on the chunk size so I think there is no problem to have it as an env variable. I put it in case we would like to increase the chunk size in the future when we can handle bigger datasets but even with that change, I think nothing could be broken since partition parameters will be stored in the job.

> If we really need multiple values, we might hardcode a scale, ie: 1_000, 100_000, 10_000_000... and let set an environment variable to choose the granularity among the fixed values.

Not sure if it is worth giving much logic to chunks for now.

AndreaFrancis:
Making this PR draft again since I need to merge with last changes on https://github.com/huggingface/datasets-server/pull/1260

AndreaFrancis:
Closing, for now, will focus on viewer features and keep this PR on stand by meanwhile.

Issue #1186, opened 2023-05-15
Re-enable image and audio in the viewer


lhoestq:
The current caching mechanism from #1026  is almost never used:
a. the parquet index is stored in memory per worker and there are too many of them
b. the image/audio files are always recreated

Because of that the viewer was too slow for image and audio datasets and we disabled it in #1144 

To fix this issue we could
- [x] 1. store the parquet index on disk using a worker
- [x] 2. use the local parquet index in /rows
- [ ] 3. (optional) avoid reloading images/audio files that are already on disk

lhoestq:
I managed to fix 1. by loading the parquet metadata from a local file.
The metadata file needs to be prepared using a dedicated job and simply stores the parquet metadata (usually a few KB).

It takes <1ms to load a parquet index instead of around 1s.

I'll open a PR to implement this logic

<details>

<summary> The code I used </summary>

```python
from datasets import Features
import pyarrow as pa
import pyarrow.parquet as pq

import asyncio
import logging
from fsspec.implementations.http import HTTPFileSystem, HTTPFile

import time
import numpy as np
from functools import partial

from typing import List, Callable


# Init

# logging.basicConfig(level=logging.DEBUG)
httpfs = HTTPFileSystem()
session = asyncio.run(httpfs.set_session())

# DB cache

cached_urls = ["https://huggingface.co/datasets/sasha/dog-food/resolve/refs%2Fconvert%2Fparquet/sasha--dog-food/parquet-train.parquet"]
cached_metadata_paths = ["playground/parquet-train.metadata.parquet"]
cached_num_bytes = [199271534]
cached_num_rows = [2100]

# Query params

query_offset, query_length = 0, 100

# Get index

_start = time.time()

parquet_file_offsets = np.cumsum(cached_num_rows)

last_row_in_parquet = parquet_file_offsets[-1] - 1
first_row = min(query_offset, last_row_in_parquet)
last_row = min(query_offset + query_length - 1, last_row_in_parquet)
first_parquet_file_id, last_parquet_file_id = np.searchsorted(
    parquet_file_offsets, [first_row, last_row], side="right"
)
offset = query_offset - parquet_file_offsets[first_parquet_file_id - 1] if first_parquet_file_id > 0 else query_offset
urls = cached_urls[first_parquet_file_id:last_parquet_file_id + 1]
metadata_paths = cached_metadata_paths[first_parquet_file_id:last_parquet_file_id + 1]
[...]

PR #1214, opened 2023-05-20
Use parquet metadata in /rows

lhoestq:
Step 2 of https://github.com/huggingface/datasets-server/issues/1186

## Implementation details

I implemented ParquetIndexWithMetadata (new) and ParquetIndexWithoutMetadata (from the existing code).:

- ParquetIndexWithMetadata is used when `config-parquet-metadata` is cached and is fast
- ParquetIndexWithoutMetadata is the old code, that runs when `config-parquet-metadata` is not available yet

I think in the long run we should remove ParquetIndexWithoutMetadata. I just kept it so that the UI doesn't show an error when `config-parquet-metadata` is not available.

I re-added support for image and audio for the ParquetIndexWithMetadata


severo (approved):

severo:
Excellent. As talked by chat: please enable the images and audio files only on a hardcoded list of datasets for now, in order to be able to test well on the real infrastructure.

PR #1221, opened 2023-05-22
fix: 🐛 if a step depend on parallel steps, both must be used

severo:
otherwise, the "error" "Response has already been computed and stored in cache kind: split-first-rows-from-parquet. Compute will be skipped" is propagated, instead of using the other cache entry as it was meant to.

Unfortunately, we will have to relaunch a lot of jobs


severo:
merging

Issue #1209, opened 2023-05-19
A lot of jobs finish with Warning: ... has a non-empty finished_at field. Force finishing anyway


severo:
See https://github.com/huggingface/datasets-server/pull/1203#issuecomment-1554544553

Started jobs should have an empty finished_at field.

severo:
Looking at the database, it seems to be the case: no `finished_at` field in the entries. Maybe an issue with how we use mongoengine, maybe it puts a default value?

severo:
still working on it

severo:
keep open

PR #1222, opened 2023-05-22
fix: 🐛 backfill the dataset after finishing the job

severo:


severo:
I think that it will fix https://github.com/huggingface/datasets-server/issues/1209

PR #1226, opened 2023-05-22
Rename /split-names-from-dataset-info

polinaeterna:
part of https://github.com/huggingface/datasets-server/issues/1086


jobs/mongodb_migration/src/mongodb_migration/deletion_migrations.py
```diff
@@ -16,6 +16,11 @@
 
 
 class MetricsDeletionMigration(MetricsMigration):
+    def __init__(self, job_type: str, cache_kind: str, version: str, description: Optional[str] = None):
```
polinaeterna, line 19: I got tired from copy-pasting descriptions too :smile:  is this change okay? doesn't it affect old migrations (i'm not sure how description is used internally).

severo: yes, much better! thanks for all these improvements

PR #1235, opened 2023-05-23
Instrument backfill

severo:
- add StepProfiler to libcommon.state, to be able to profile the code duration when doing a backfill
- refactor code to manage prometheus from libcommon
- detail: don't put empty (0) values in cache and queue metrics if the metrics database is empty. it's ok not to have values until the background metrics job has run


lhoestq:
Cool ! I hope it helps to understand what to optimize 🤞

PR #1241, opened 2023-05-23
Adding temporal hardcoded data for opt in/out: laion/laion2B-en and kakaobrain/coyo-700m

AndreaFrancis:
As per discussion in https://github.com/huggingface/moon-landing/pull/6332#discussion_r1202989143
Fake data will be hardcoded on the server side so that it is consistent with what API returns and what we show in the UI.
NOTE.- This is a temporal solution, once https://github.com/huggingface/datasets-server/issues/1087 is implemented, this code can be removed.


severo (approved):
thanks, it's better here than in moonlanding (it's our debt)

PR #1246, opened 2023-05-24
Rename `/config-names` processing step

polinaeterna:
the last change of graph steps names in https://github.com/huggingface/datasets-server/issues/1086
next: endpoints and docs :)

**reminder: don't forget to stop the workers before mirgrations!**


services/admin/tests/test_app_real.py
```diff
@@ -45,7 +45,7 @@ def test_force_refresh(
 ) -> None:
     dataset = "glue"
     processing_graph = ProcessingGraph(real_app_config.processing_graph.specification)
-    first_step = processing_graph.get_processing_steps()[0]
+    first_step = processing_graph.get_processing_steps(order="topological")[0]
```
polinaeterna, line 48: i added `"topological"` here because otherwise the first step was not `dataset-config-names` but first alphabetically

AndreaFrancis (approved):
Thanks for making the steps' nomenclature better!

services/admin/tests/test_app_real.py
severo: Well seen

severo (approved):
Perfect

PR #1247, opened 2023-05-24
feat: 🎸 create all jobs in backfill in one operation

severo:
instead of one (or multiple) operations for every job creation


severo:
let's try!

PR #1256, opened 2023-05-25
Separate opt in out urls scan

AndreaFrancis:
Part of the second approach for spawning full scan https://github.com/huggingface/datasets-server/issues/1087
"Run full scan in separated Jobs and store results in separated cache entries".

I am moving logic to inspect if a split has image URL columns to another step `"split-image-url-columns"`.
Now, `"split-opt-in-out-urls-scan"` will depend on `"split-image-url-columns"`.
I am also adding validation on the new step to consider only those image URLs as discussed in https://huggingface.slack.com/archives/C0311GZ7R6K/p1684962431285069.


services/worker/src/worker/job_runners/split/image_url_columns.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+        features = upstream_response_content["features"]
+        first_rows = upstream_response_content["rows"]
+    except KeyError as e:
+        raise PreviousStepFormatError("Previous step did not return the expected content.", e) from e
+
+    # look for URLs columns using the first rows
+    string_type_dict = {"dtype": "string", "_type": "Value"}
+    string_columns = [feature["name"] for feature in features if feature["type"] == string_type_dict]
```
severo, line 50: Possibly `feature["type"]` can contain extra fields, so I think it would be safer to check the existence and value of `dtype` and `_type`, than checking the equality between objects

services/worker/src/worker/job_runners/split/image_url_columns.py
```diff
@@ -0,0 +1,82 @@
...
+    urls_columns = []
+    for string_column in string_columns:
+        urls_count = sum(
+            1
+            for row in first_rows
+            if isinstance(row["row"].get(string_column), str) and is_image_url(text=row["row"][string_column])
+        )
+        if urls_count and urls_count / len(first_rows) > 0.3:
```
severo, line 58: nit: can you set the value in a constant at the start of the file? I find it's a best practice to always set a constant for the hardcoded numerical values, and to group them (at the start of the file, or in a dedicated constants file) -> easier to search for instances in the code, easier to maintain... (it might be a personal preference)

severo: Also, just in case, we should handle the case len(first_rows) == 0

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
severo, line 1: side note: now, only the image URLs will be scanned, which is not reflected in the name of the step. I think it's better to avoid renaming again, but maybe we should ensure that it's well documented.

severo (approved):
Excellent!

services/worker/src/worker/job_runners/split/image_url_columns.py
AndreaFrancis: Done

services/worker/src/worker/job_runners/split/image_url_columns.py
AndreaFrancis: Sure, I think is a good practice and added it.
Also added len validation.

PR #1260, opened 2023-05-26
feat: 🎸 create orchestrator

severo:
tests are missing


libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -0,0 +1,757 @@
...
+                            "revision": self.revision,
+                            "config": artifact_state.config,
+                            "split": artifact_state.split,
+                        },
+                        "priority": self.priority,
+                    }
+                )
+            else:
+                pending_jobs_to_delete_df.drop(valid_pending_jobs_df.index, inplace=True)
+        # Better keep this order: delete, then create
```
AndreaFrancis, line 557: There is precisely similar logic in AfterJobPlan post_init, could it be factored in a method?

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -0,0 +1,757 @@
...
+        if len(first_processing_steps) < 1:
+            raise ValueError("Processing graph has no first step")
+        if any(first_processing_step.input_type != "dataset" for first_processing_step in first_processing_steps):
+            raise ValueError("One of the first processing steps is not a dataset step")
+        with StepProfiler(
+            method="DatasetOrchestrator.set_revision",
+            step="all",
+            context=f"dataset={self.dataset}",
+        ):
+            logging.info(f"Analyzing {self.dataset}")
```
AndreaFrancis, line 599: After this line it looks similar logic for the backfill method in line 720, maybe you could call it with different parameters if needed?

libs/libcommon/src/libcommon/state.py
```diff
@@ -345,204 +245,67 @@ def __post_init__(self) -> None:
...
     artifact_state_by_step: Dict[str, ArtifactState] = field(init=False)
-    cache_status: CacheStatus = field(init=False)
-    plan: Plan = field(init=False)
-    should_be_backfilled: bool = field(init=False)
 
     def __post_init__(self) -> None:
         with StepProfiler(
-            method="DatasetState.__post_init__",
-            step="all",
+            method="DatasetBackfillPlan.__post_init__",
```
AndreaFrancis, line 265:
```suggestion
            method="DatasetState.__post_init__",
```

libs/libcommon/src/libcommon/state.py
```diff
@@ -345,204 +245,67 @@ def __post_init__(self) -> None:
...
-                            (self.cache_entries_df["kind"] == processing_step.cache_kind)
-                            & (self.cache_entries_df["config"].isnull())
-                            & (self.cache_entries_df["split"].isnull())
-                        ],
-                    )
-                    for processing_step in self.processing_graph.get_input_type_processing_steps(input_type="dataset")
-                }
-            with StepProfiler(
-                method="DatasetState.__post_init__",
+                method="DatasetBackfillPlan.__post_init__",
```
AndreaFrancis, line 293: I don't understand, should we use in the method the name of the current class + method? Why it changed to DatasetBackfillPlan?

AndreaFrancis (approved):
Thanks! LGTM overall but I put a couple of questions

libs/libcommon/src/libcommon/state.py
severo: It's an error!

libs/libcommon/src/libcommon/orchestrator.py
severo: Hmm it's only four lines, I think it's not a problem to keep them duplicated.

PR #1264, opened 2023-05-30
Fix missing slash in admin endpoints

albertvillanova:
As discussed with @severo, this PR fixes the missing slash in some admin endpoints, e.g.: 
- https://datasets-server.huggingface.co/admin/force-refreshdataset-config-names

After this PR, it will be:
- https://datasets-server.huggingface.co/admin/force-refresh/dataset-config-names

Related to:
- #1246


AndreaFrancis (approved):
Good catch! Thanks!

Issue #1281, opened 2023-06-01
Remove dependency on torch and torchaudio


albertvillanova:
As suggested by @severo:
> should we remove the dependency to torch and torchaudio? cc @polinaeterna

- [ ] #1282
- [ ] Remove torch

albertvillanova:
I have already removed the dependency on `torchaudio`, once all the Audio feature functionalities are backed by `soundfile`.

In relation with `torch`, are we sure we want to stop supporting dataset loading scripts that use `torch`. If so, what about the ones using `tensorflow`?

Maybe @lhoestq can give better insight.

severo:
I was asking because, if I remember correctly, we didn't depend on torch (because we had no dataset depending on torch) before working with audio cells.

But I wouldn't want to rely on my poor memory. At one point it would be good to store statistics about the dependencies found in the scripts to be able to take decisions. Not prioritary.

albertvillanova:
I agree with you that we should track that information.

Honestly, I don't know how many loading scripts on Hub dataset depend on `torch` (or `tensorflow`)...

lhoestq:
we might want to keep torchaudio after all, e.g. for https://huggingface.co/datasets/cdminix/libritts-aligned/blob/main/libritts-aligned.py

severo:
should we revert https://github.com/huggingface/datasets-server/pull/1282?

albertvillanova:
I think the relevant question is how many loading scripts use `torchaudio`. I don't think we should support all packages for all scripts, but we should do sensible choices based on impact.

Also please note that even if reverting @1282, the viewer will not work for https://huggingface.co/datasets/cdminix/libritts-aligned
- It has also 2 other third-party dependencies which are not installed on the datasets-server:
```
Error code:   DatasetModuleNotInstalledError
Exception:    ImportError
Message:      To be able to use cdminix/libritts-aligned, you need to install the following dependencies: alignments, phones.
```

lhoestq:
Ah I see, maybe not worth it for now then

albertvillanova:
There are still 2 questions to be answered: https://github.com/huggingface/datasets-server/issues/1281#issuecomment-1571875857
> In relation with `torch`, are we sure we want to stop supporting dataset loading scripts that use `torch`?

> If so, what about the ones using `tensorflow`?

severo:
OK. Let's keep the current situation and do statistics about `DatasetModuleNotInstalledError` regularly.

For example, for now:

```
[...]

PR #1282, opened 2023-06-01
Remove torchaudio dependency

albertvillanova:
Remove `torchaudio` dependency.

Fix partially #1281.


severo:
do we still require torch?

severo (approved):

Issue #629, opened 2022-11-07
Index the (text) datasets contents to enable full-text search


severo:
See an example by @ola13 on the ROOTS corpus: https://huggingface.co/spaces/bigscience-data/roots-search.

Implementation details are here: https://huggingface.co/spaces/bigscience-data/scisearch/resolve/main/roots_search_tool_specs.pdf

Internal discussion: https://huggingface.slack.com/archives/C0311GZ7R6K/p1667586339997379

AndreaFrancis:
I would love to work in this feature! 
Also I saw the hub already implemented it on https://github.com/huggingface/moon-landing/pull/4971 using `meilisearch`, I will continue with the research and put my finding here.

ola13:
More of a research-oriented solution but check out this. as well: https://arxiv.org/abs/2302.14534 :)

AndreaFrancis:
The API phase can start now that the first phase of FTS has been implemented on https://github.com/huggingface/datasets-server/pull/1296. 
Note that duckdb index files (Initially for those small datasets with parquet files under 100Mb) are being generated and stored in revision "refs/convert/parquet" in the file "index.duckdb". 
So, my proposal is:

1. We need a preparation for FTS (This includes downloading the index file) so, there could be a `POST /fts-prepare` triggered by the hub that will do a couple of actions in datasets-server side:
		- Store an entry in a new mongo collection `"temporal_file"` with 
			- dataset
			- config
			- split
			- location (path of the downloaded file)
			- created_at
			- last_used_at (last date that API used the file to query fts)
		- Download the index file	
2. When requesting a search `GET /fts?q=<query>`
	       - Look for the record in "temporal_file" collection and get the downloaded index location 
		- If exists: connect to db -> with retry mechanism in case another process is connected to the same locked file
		- Return results with pagination
		- Update last_used_at file in "temporal_file" collection
3. To have a k8s Job that runs every X minutes
		- Delete all files whose last_used_at field is more than Y minutes
		- Delete related records in "temporal_file" collection
	

WDYT @huggingface/datasets-server

severo:
I think we can do even simpler:
- on GET /fts?q=<query>, if no local file, download it
- use the path to locate the file (dataset/config/split)
- have a k8s job that deletes the old files (based on the last access date)

lhoestq:
Nice ! 

> -> with retry mechanism in case another process is connected to the same locked file

If you open the db as read-only there shouldn't be any lock no ?

[...]

Issue #1272, opened 2023-05-31
Store the list of parquet shards on the Hub?


severo:
We could reorganize the way we store the parquet files on the Hub:

- one directory per config (already done)
- then, one directory per split
- then, one file per shard, with name `n.parquet` (starting at `0.parquet`)
- plus: a file `shards.json` which is a JSON file with an "ordered" array of the shard URLs, ie: `["https://huggingface.co/datasets/LLMs/Alpaca-ShareGPT/parquet/train/0.parquet", "https://huggingface.co/datasets/LLMs/Alpaca-ShareGPT/parquet/train/1.parquet", ..., "https://huggingface.co/datasets/LLMs/Alpaca-ShareGPT/parquet/train/13.parquet"]`

Alternative: don't change the directory structure, nor the filenames, but use redirections. Anyway, add the shard.json files (one per split)

severo:
No, we won't store the list of shards on the Hub as such. For the structure, see #1463

PR #1296, opened 2023-06-02
feat: Index the (text) datasets contents to enable full-text search - DuckDB

AndreaFrancis:
First part for https://github.com/huggingface/datasets-server/issues/629 (index only)
New job runner to index parquet files at a split level using duckdb 
- Using parquet content, it will validate if there are string columns, if so, it will index all columns at least there is any image/audio/binary feature type (won't be supported yet since it could affect the viewer whenever search API is available, need a better design for this case).
- Will create a duckdb index file that contains the parquet + FTS search
- Will create a new branch duckdb/index in the dataset repository and upload the duckdb index file


services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,147 @@
...
+from worker.job_runners.split.split_job_runner import SplitJobRunner
+from worker.utils import (
+    CompleteJobResult,
+    IndexRowsResponse,
+    get_previous_step_or_raise,
+)
+
+STRING_FEATURE_DTYPE = "string"
+VALUE_FEATURE_TYPE = "Value"
+DUCKDB_DEFAULT_DB_NAME = "index.db"
```
severo, line 29: nit: here and in other places where you mention `db_name`, maybe replace with `index_filename`, which is more meaningful (at least for me)

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,147 @@
...
+INSTALL_EXTENSION_COMMAND = "INSTALL '{extension}';"
+LOAD_EXTENSION_COMMAND = "LOAD '{extension}';"
+# TODO: What if __id field already exist?
+
+
+def compute_index_rows(dataset: str, config: str, split: str, duckdb_index_directory: StrPath) -> IndexRowsResponse:
+    logging.info(f"get index-rows for dataset={dataset} config={config} split={split}")
+
+    # get the first rows from previous job
+    upstream_response = get_previous_step_or_raise(
```
severo, line 44: use the /parquet endpoint to get the features? and thus remove the need to depend on split-first-rows-from-streaming and split-first-rows-from-parquet?

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,147 @@
...
+    con = duckdb.connect(str(db_location))
+    con.sql(CREATE_SEQUENCE_COMMAND)
+    con.sql(f"{CREATE_TABLE_COMMAND} read_parquet({parquet_urls});")
+    con.sql(DROP_INDEX_COMMAND)
+
+    # TODO: by default, 'porter' stemmer is being used, use a specific one by dataset language in the future
+    # see https://duckdb.org/docs/extensions/full_text_search.html for more deails about 'stemmer' parameter
+    con.sql(CREATE_INDEX_COMMAND)
+
+    return IndexRowsResponse(duckdb_db_name=duck_db_name)
```
severo, line 111: duck_db_name => duckdb_index_filename?

severo (commented):
two main comments:
- should it be a config level step instead, that depends only on config-parquet?
- should we store the index file on the Hub (as the parquet files themselves), instead of storing them on the NFS?

AndreaFrancis:
Just one question, I was thinking that it would be better to have this step as `"split"` level because of the indexing time. We have some datasets with many configs but small data, what if in the middle of processing, the job runner crashes because of timeout?
We will have an inconsistent state with no record in the cache but some created indices in the branch. 
Not sure if we have the same problem with parquet_and_info.

severo:
Good point, OK to let it at split level.

libs/libcommon/src/libcommon/config.py
```diff
@@ -321,6 +338,15 @@ class ProcessingGraphConfig:
                 "triggered_by": ["dataset-config-names", "config-opt-in-out-urls-count"],
                 "job_runner_version": PROCESSING_STEP_DATASET_OPT_IN_OUT_URLS_COUNT_VERSION,
             },
+            "split-duckdb-index": {
+                "input_type": "split",
+                "triggered_by": [
+                    "split-first-rows-from-streaming",
+                    "split-first-rows-from-parquet",
```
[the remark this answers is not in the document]
AndreaFrancis: I changed to use parquet instead of first-rows, but kept split level as per the comment https://github.com/huggingface/datasets-server/pull/1296#issuecomment-1578855161

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: Done

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: I changed the response structure to:
```
    dataset: str
    config: str
    split: str
    url: str
    filename: str
    size: int
```
I think it makes more sense to have similar format as we have on /parquet

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: I changed it to DUCKDB_DEFAULT_INDEX_FILENAME

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,280 @@
...
+        repo_file for repo_file in target_dataset_info.siblings if repo_file.rfilename == index_file_location
+    ]
+
+    if not repo_files:
+        raise NotAvailableIndexFileError("No index file was found")
+
+    if len(repo_files) != 1:
+        logging.warning(f"Found {len(repo_files)} index files, should be only 1")
+
+    remove_dir(dir_path)
```
severo, line 226: OK, good!
I think we should reuse the mechanism in https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/_datasets_based_job_runner.py to avoid concurrency issues.

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,280 @@
...
+    duckdb.execute(INSTALL_EXTENSION_COMMAND.format(extension="fts"))
+    duckdb.execute(LOAD_EXTENSION_COMMAND.format(extension="fts"))
+
+    # index all columns
+    con = duckdb.connect(str(db_location))
+    con.sql(CREATE_SEQUENCE_COMMAND)
+    con.sql(f"{CREATE_TABLE_COMMAND} read_parquet({parquet_urls});")
+
+    # TODO: by default, 'porter' stemmer is being used, use a specific one by dataset language in the future
+    # see https://duckdb.org/docs/extensions/full_text_search.html for more deails about 'stemmer' parameter
```
severo, line 177:
```suggestion
    # see https://duckdb.org/docs/extensions/full_text_search.html for more details about 'stemmer' parameter
```

libs/libcommon/src/libcommon/viewer_utils/index_utils.py
```diff
@@ -0,0 +1,17 @@
...
+from os import makedirs
+from pathlib import Path
+
+from libcommon.storage import StrPath
+
+DATASET_SEPARATOR = "--"
+INDEX_DIR_MODE = 0o755
+
+
+def create_index_dir_split(dataset: str, config: str, split: str, index_directory: StrPath) -> Path:
```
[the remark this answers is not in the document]
AndreaFrancis: > is it temporary? do we delete it later?

Yes, I call     remove_dir(dir_path) in line 226, but will better apply the same approach as in `datasets library cache` (create a random directory to avoid concurrency in pre-compute and delete it in post-compute)

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,280 @@
...
+
+    if not string_columns:
+        raise NoIndexableColumnsError("No string columns available to index.")
+
+    # look for image, audio and binary columns, if present, raise exeception (not supported yet)
+    if any(
+        feature
+        for feature in features.values()
+        if next(
+            (feature_type for feature_type in UNSUPPORTED_FEATURES_MAGIC_STRINGS if feature_type in str(feature)), None
```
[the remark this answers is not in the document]
AndreaFrancis: It does not break indexing but I was thinking that it will need preparation like we do in first rows when mapping image/audio to a file in assets or maybe something similar like in /rows (It surely will support algo pagination)

services/worker/src/worker/utils.py
```diff
@@ -138,6 +151,15 @@ class RowsContent(TypedDict):
     all_fetched: bool
 
 
+class ParquetFileItem(TypedDict):
```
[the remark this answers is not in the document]
AndreaFrancis: Thanks, I renamed it to SplitHubFile.

services/worker/src/worker/job_runners/split/duckdb_index.py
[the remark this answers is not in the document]
AndreaFrancis: But maybe I could leave indexing also image/binary/audio and then when implementing API layer think about it or apply any change needed, what do you think @severo ?

libs/libcommon/src/libcommon/viewer_utils/index_utils.py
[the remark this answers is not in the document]
severo: OK, let's try to use the same code for both (even if the datasets library cache has some additional code to set and unset the config variable

services/worker/src/worker/job_runners/split/duckdb_index.py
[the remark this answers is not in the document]
severo: Sorry, I don't understand. As any other non-text column, they will be ignored by this step, so... why do they receive a special treatment?

severo:
Another comment, related to the way all the datasets-server works.

On one side, it's good to have steps that do only one task. And we take advantage of the format of Parquet files to be able to query them efficiently using a remote URL.

But on the other side, we're downloading (part of) the parquet files, after creating and deleting them on the disk before (in the config-parquet-and-info step). It feels like a waste of resources. In a sense, we might want to do everything related to parquet in one step. WDYT @huggingface/datasets-server?

services/worker/src/worker/job_runners/split/duckdb_index.py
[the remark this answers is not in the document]
AndreaFrancis: Maybe it is something I am overthinking but my idea is that when using the index on the API layer, it will require something like https://github.com/huggingface/datasets-server/blob/main/services/api/src/api/routes/rows.py#L144 or https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py#L37 or https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/first_rows_from_streaming.py#L42 right?
I think when implementing /search  endpoint, it will download the index (cache for a certain time period), search the specified text and return the dataframe result of it (no need to read from parquet, just from duckdb index file), I wonder how will we return those datasets with image/audio/binary data to the viewer?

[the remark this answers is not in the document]
severo: > no need to read from parquet, just from duckdb index file

is this how the index works? I thought that it would only contain an index for the text columns, while having to access the parquet file to get the full rows (the other columns, basically). Isn't it how it works?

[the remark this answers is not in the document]
AndreaFrancis: No, like a typical database engine, it needs to have all the data in a table and then it adds an index to it.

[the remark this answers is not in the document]
severo: hmmm... OK. And how will we manage sharded parquet files? will the index be sharded too? Or is it a huge file? is it what we want?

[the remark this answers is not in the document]
AndreaFrancis: At least we create different tables/index files (one per parquet file) and, in API layer, join all the results 
what do you think @huggingface/datasets-server  @lhoestq ?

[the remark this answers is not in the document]
lhoestq: I think one big duckdb file on the first 5GB is fine for now. The size is available in the `config-parquet` response afaik

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,280 @@
...
+            partial(get_parquet_file, fs=fs, hf_token=hf_token), parquet_urls, desc=desc, unit="pq", disable=True
+        )
+    except Exception as e:
+        raise FileSystemError(f"Could not read the parquet files: {e}") from e
+
+    # get the features
+    features = Features.from_arrow_schema(parquet_files[0].schema.to_arrow_schema())
+
+    # look for string columns using the first rows
+    string_columns = [column for column, feature in features.items() if STRING_FEATURE_DTYPE in str(feature)]
```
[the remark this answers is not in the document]
polinaeterna: and in this case, if only first parquet file is taken, is it needed to get all of them (in line 101)?

for visibility: @lhoestq approved that "features" field should always exist in "dataset_info" and if it's not - it's a bug. So it should be safe to take features from there, instead of reading parquet schema from files directly.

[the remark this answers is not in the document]
AndreaFrancis: Thanks @polinaeterna ! I will do the change to read from dataset_info instead

services/worker/src/worker/job_runners/split/duckdb_index.py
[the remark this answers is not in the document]
severo: Is it easy to truncate by size?

Most of the datasets should have parquet files under 5GB, but we have:
- false positives for datasets with scripts, where we don't detect the size is above 5GB (like https://huggingface.co/datasets/Muennighoff/multi_eurlex)
- parquet-native datasets (like https://huggingface.co/datasets/bigcode/the-stack) that can be enormous

For those, we would have to cut somewhere.

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: Done in Done in https://github.com/huggingface/datasets-server/pull/1296/commits/3c9b4eef5f394e3f8dc0b60f7c740cfbdfb666b6

libs/libcommon/src/libcommon/config.py
```diff
@@ -319,6 +348,15 @@ class ProcessingGraphConfig:
                 "triggered_by": ["dataset-config-names", "config-opt-in-out-urls-count"],
                 "job_runner_version": PROCESSING_STEP_DATASET_OPT_IN_OUT_URLS_COUNT_VERSION,
             },
+            "split-duckdb-index": {
+                "input_type": "split",
+                "triggered_by": [
+                    "config-split-names-from-info",
+                    "config-split-names-from-streaming",
```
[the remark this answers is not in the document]
AndreaFrancis: I created this issue https://github.com/huggingface/datasets-server/issues/1358

services/worker/src/worker/job_runners/split/duckdb_index.py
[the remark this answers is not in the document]
AndreaFrancis: I changed the parent to "config-parquet-and-info" and now, it gets features from cache and no need to load from parquet file  schema.

libs/libcommon/src/libcommon/config.py
```diff
@@ -104,6 +105,34 @@ def from_env(cls) -> "ParquetMetadataConfig":
             )
 
 
+DUCKDB_INDEX_STORAGE_DIRECTORY = None
+DUCKDB_INDEX_COMMIT_MESSAGE = "Update duckdb index file"
+DUCKDB_INDEX_COMMITTER_HF_TOKEN = None
+DUCKDB_INDEX_TARGET_REVISION = "duckdb/index"
```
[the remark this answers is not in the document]
AndreaFrancis: I changed the target revision to point to `refs/convert/parquet` instead, and also the index file will be under `{config}/{split}/ `subdirectory but just to the new file. To standardize the already existing logic for parquet files I think we can handle it in another PR, job runner will be modified and will probably have to think about how to remove the old structure in the branch `{config}/{dataset}-{split}-{shard}.parquet` to `{config}/{split}/{shard}.parquet` maybe?
I could create another issue for it.

services/worker/src/worker/job_runners/split/duckdb_index.py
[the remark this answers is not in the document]
AndreaFrancis: For now, I limited to those splits with parquets under 5 GB (sum of all shards). I think  it can scale later to cut to the first N GB?

libs/libcommon/src/libcommon/config.py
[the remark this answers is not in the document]
AndreaFrancis: I saw it already exists for parquets file https://github.com/huggingface/datasets-server/issues/1272

services/worker/src/worker/utils.py
```diff
@@ -398,3 +402,14 @@ def get_rows_or_raise(
...
             ) from err
+
+
+# TODO: use huggingface_hub's hf_hub_url after
+# https://github.com/huggingface/huggingface_hub/issues/1082
+def hf_hub_url(repo_id: str, filename: str, hf_endpoint: str, revision: str, url_template: str) -> str:
+    return (hf_endpoint + url_template) % (repo_id, quote(revision, safe=""), filename)
+
+
+def get_parquet_file(url: str, fs: HTTPFileSystem, hf_token: Optional[str]) -> ParquetFile:
```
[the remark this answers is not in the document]
polinaeterna: sorry it's being used in other places, ignore it...

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,250 @@
...
+from worker.utils import CompleteJobResult, hf_hub_url
+
+DATASET_TYPE = "dataset"
+STRING_FEATURE_DTYPE = "string"
+VALUE_FEATURE_TYPE = "Value"
+DUCKDB_DEFAULT_INDEX_FILENAME = "duckdb_index.db"
+CREATE_SEQUENCE_COMMAND = "CREATE OR REPLACE SEQUENCE serial START 1;"
+CREATE_INDEX_COMMAND = "PRAGMA create_fts_index('data', '__id', '*', overwrite=1);"
+CREATE_TABLE_COMMAND = "CREATE OR REPLACE TABLE data AS SELECT nextval('serial') AS __id, * FROM"
+# TODO: What if __id field already exist?
```
lhoestq, line 43: maybe use something even less likely to exist, like `__hf_index_id`

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,250 @@
...
+    delete_operations: List[CommitOperation] = []
+    if index_file_location in all_repo_files:
+        delete_operations.append(CommitOperationDelete(path_in_repo=index_file_location))
+
+    # send the files to the target revision
+    add_operations: List[CommitOperation] = [
+        CommitOperationAdd(path_in_repo=index_file_location, path_or_fileobj=db_location)
+    ]
+
+    committer_hf_api.create_commit(
```
lhoestq, line 168: Here we'll need to use the new locking mechanism to avoid concurrent commits like in `https://github.com/huggingface/datasets-server/pull/1355`

Maybe let's agree on using something like
```python
lock_key = json.dumps({"dataset": dataset, "resource": "parquet conversion branch"})
with lock(key=lock_key, job_id=job_id):
    committer_hf_api.create_commit(...)
```

I can do the change in my PR to use this

services/worker/tests/fixtures/hub.py
```diff
@@ -714,4 +722,13 @@ def hub_datasets(
...
         },
+        "duckdb_index": {
+            "name": hub_public_duckdb_index,
+            "config_names_response": create_config_names_response(hub_public_duckdb_index),
+            "splits_response": create_splits_response(hub_public_duckdb_index),
+            "first_rows_response": create_first_rows_response(hub_public_duckdb_index, TEXT_cols, TEXT_rows),
+            "parquet_and_info_response": create_parquet_and_info_response(
+                dataset=hub_public_duckdb_index, data_type="csv"
+            ),
+        },
```
lhoestq, line 733: Just FYI I refactored this fixture at https://github.com/huggingface/datasets-server/pull/1373 - so depending on which one is merged first we'll have to fix a small conflict

lhoestq (commented):
Love it ! This will be sooo useful

just added some minor comments:

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: Good idea! I renamed it to __hf_index_id

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: Thanks, I applied it as in parquet_and_info.

services/worker/tests/fixtures/hub.py
AndreaFrancis: Done, I created a new fixture for duckdb_index

services/worker/src/worker/job_runners/split/duckdb_index.py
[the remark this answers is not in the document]
AndreaFrancis: It is going to be limited to datasets with parquets under 100 MB initially (small), then we can think of a better way to scale for bigger datasets.

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,259 @@
...
+        refs = hf_api.list_repo_refs(repo_id=dataset, repo_type=DATASET_TYPE)
+        if all(ref.ref != target_revision for ref in refs.converts):
+            initial_commit = hf_api.list_repo_commits(repo_id=dataset, repo_type=DATASET_TYPE)[-1].commit_id
+            committer_hf_api.create_branch(
+                repo_id=dataset, branch=target_revision, repo_type=DATASET_TYPE, revision=initial_commit, exist_ok=True
+            )
+    except RepositoryNotFoundError as err:
+        raise DatasetNotFoundError("The dataset does not exist on the Hub.") from err
+
+    target_dataset_info = hf_api.dataset_info(repo_id=dataset, revision=target_revision, files_metadata=False)
```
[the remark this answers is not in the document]
severo: I think so, because we use `parent_commit=target_dataset_info.sha` inside the lock, so we don't want the commits to have changed or it will raise

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,259 @@
...
+                revision=target_revision,
+                operations=delete_operations + add_operations,
+                commit_message=commit_message,
+                parent_commit=target_dataset_info.sha,
+            )
+    except TimeoutError as err:
+        raise LockedDatasetTimeoutError("the dataset is currently locked, please try again later.") from err
+
+    # call the API again to get the index file
+    target_dataset_info = hf_api.dataset_info(repo_id=dataset, revision=target_revision, files_metadata=True)
```
severo, line 186: this could be inside the lock too

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,259 @@
...
+    repo_files = [
+        repo_file for repo_file in target_dataset_info.siblings if repo_file.rfilename == index_file_location
+    ]
+
+    if not repo_files:
+        raise NotAvailableIndexFileError("No index file was found")
+
+    if len(repo_files) != 1:
+        logging.warning(f"Found {len(repo_files)} index files, should be only 1")
```
severo, line 195: I think this could be merged with the previous test and raise one Exception for all the cases (not 1 index file)

severo (commented):
Amazing work!

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,250 @@
...
+        job_info: JobInfo,
+        app_config: AppConfig,
+        processing_step: ProcessingStep,
+        duckdb_index_directory: StrPath,
+    ) -> None:
+        super().__init__(
+            job_info=job_info,
+            app_config=app_config,
+            processing_step=processing_step,
+            hf_datasets_cache=Path(duckdb_index_directory).resolve(),
```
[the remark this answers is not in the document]
AndreaFrancis: PR ready https://github.com/huggingface/datasets-server/pull/1388

lhoestq:
btw I noticed that duckdb files can only be read by the exact same duckdb version. For example using a differerent version can raise

```
IOException: IO Error: Trying to read a database file with version number 43, but we can only read version 51.
The database file was created with DuckDB version v0.7.0 or v0.7.1.

The storage of DuckDB is not yet stable; newer versions of DuckDB cannot read old database files and vice versa.
The storage will be stabilized when version 1.0 releases.

For now, we recommend that you load the database file in a supported version of DuckDB, and use the EXPORT DATABASE command followed by IMPORT DATABASE on the current version of DuckDB.

See the storage page for more information: https://duckdb.org/internals/storage
```

I don't think it implies anything for this PR in particular, but let's keep this in mind if we ever want to upgrade

AndreaFrancis:
> I don't think it implies anything for this PR in particular, but let's keep this in mind if we ever want to upgrade

Thanks @lhoestq, hmmm maybe in the future if we would like to increase the dependency version it will require also an upgrade in the job runner version so that everything is computed again :-(. On the other hand, if the duckdb file will be available in the branch and somehow the usage depends on the version, I think maybe at least we can put the version on the file name. Currently, it will be `index.duckdb` but it could be -> index-0.8.1.duckdb?  Or in a subfolder `/0.8.1/index.duckdb` ?

lhoestq:
No need to put it in the file or directory name. This way no need to ask the server for the index file URL.

But if we ever need to take the duckdb version into account in the future we can add it to the cache response alongside the URL of the index file.

Having the duckdb version in the cache version could also help if we ever need to support multiple versions in the API: then we can dispatch the user query to a python env with the right duckdb version.

Though let's hope we won't have to do that :p

AndreaFrancis:
Thanks @severo @lhoestq @polinaeterna, I think this PR is ready for a new revision, I tried to address all the suggestions and also added some test scenarios.
Thanks for all your patience!

chart/templates/_envWorker.tpl
```diff
@@ -87,4 +87,23 @@
 - name: CONFIG_NAMES_MAX_NUMBER
   value: {{ .Values.configNames.maxNumber | quote }}
 
+# specific to 'split-duckdb-index' job runner
+- name: DUCKDB_INDEX_COMMIT_MESSAGE
+  value: {{ .Values.duckDBIndex.commitMessage | quote }}
+- name: DUCKDB_INDEX_COMMITTER_HF_TOKEN
+  {{- if .Values.secrets.appParquetConverterHfToken.fromSecret }}
```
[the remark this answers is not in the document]
julien-c: i would say we can use the same user (files are sent to the same refs/convert/parquet ref, right?)

chart/values.yaml
```diff
@@ -214,6 +214,17 @@ parquetMetadata:
...
+  # Directory on the shared storage (used temporarily to prepare the duckdb indexes before sending to the Hub)
+  storageDirectory: "/duckdb-index"
+  # the git commit message when the duckdb index file is uploaded to the Hub. Defaults to `Update duckdb index files`.
+  commitMessage: "Update duckdb index files"
+  # the git revision of the dataset where to store the duckdb index file. Defaults to `refs/convert/parquet`.
+  targetRevision: "refs/convert/parquet"
+  # the URL template to build the duckdb index file URL. Defaults to `/datasets/%s/resolve/%s/%s`.
+  urlTemplate: "/datasets/%s/resolve/%s/%s"
+  # the maximum size of the split parquets.
+  maxParquetSizeBytes: "100_000_000"
```
lhoestq, line 227: If we cant to increase this it we'll need to know how much time it takes to index datasets and to query them. We can see that later by adding some profiling.

services/worker/tests/job_runners/split/test_duckdb_index.py
```diff
@@ -0,0 +1,234 @@
...
+        ("duckdb_index", 1_000, "SplitWithTooBigParquetError"),  # parquet size is 2812
+        ("public", None, "NoIndexableColumnsError"),  # dataset does not have string columns to index
+    ],
+)
+def test_compute(
+    get_parquet_job_runner: GetParquetJobRunner,
+    get_job_runner: GetJobRunner,
+    app_config: AppConfig,
+    hub_responses_public: HubDatasetTest,
+    hub_responses_duckdb_index: HubDatasetTest,
```
lhoestq, line 137: maybe delete the remote index file, in case another test uses the same repository (fixture scope is session) ?

chart/values.yaml
```diff
@@ -214,6 +214,17 @@ parquetMetadata:
   # Directory on the shared storage (parquet metadata files used for random access in /rows)
   storageDirectory: "/parquet-metadata"
 
+duckDBIndex:
+  # Directory on the shared storage (used temporarily to prepare the duckdb indexes before sending to the Hub)
+  storageDirectory: "/duckdb-index"
```
lhoestq, line 219: Using a non-shared temporary directory would also have worked no ?

I'm fine with using a shared storage though

lhoestq (approved):
Looks all good to me :D

services/worker/tests/job_runners/split/test_duckdb_index.py
polinaeterna: > in case another test uses the same repository

Is it possible?

libs/libcommon/src/libcommon/utils.py
```diff
@@ -65,6 +65,15 @@ class JobResult(TypedDict):
     output: Optional[JobOutput]
 
 
+class SplitHubFile(TypedDict):
```
polinaeterna, line 68: what's the idea of having a `Split` prefix? to reflect that the item must have a non-empty split?

polinaeterna (approved):
looks awesome to me :fire:

severo:
Reviewing again, now that https://www.notion.so/discussion-URLs-and-directories-structure-for-Parquet-files-562399c61db6483c886272a7d44fbb93 is settled.

chart/values.yaml
severo: I think it's due to the size (the pods have nearly no space, while the shared storage has as much space as we need (EFS))

libs/libcommon/src/libcommon/exceptions.py
```diff
@@ -498,8 +502,36 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
         super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "UnsupportedExternalFilesError", cause, True)
 
 
+class NoIndexableColumnsError(CacheableError):
```
severo, line 505: nit: sort the class definitions alphabetically

libs/libcommon/src/libcommon/exceptions.py
```diff
@@ -498,8 +502,36 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
...
+
+
 class DatasetWithTooManyConfigsError(CacheableError):
     """Raised when the number of configs of a dataset exceeded the limit."""
 
     def __init__(self, message: str, cause: Optional[BaseException] = None):
         super().__init__(message, HTTPStatus.NOT_IMPLEMENTED, "DatasetWithTooManyConfigsError", cause, True)
+
+
+class CachedDirectoryNotInitializedError(CacheableError):
```
severo, line 533: I think that we renamed CachedDirectory in another PR, we should update here too

services/worker/pyproject.toml
```diff
@@ -45,6 +45,7 @@ transformers = "^4.30.0"
 trec-car-tools = { path = "vendors/trec-car-tools/python3" }
 typer = "^0.4.2"
 wget = "^3.2"
+duckdb = "^0.8.1"
```
severo, line 48: nit : alphabetical order

tools/docker-compose-dev-datasets-server.yml
```diff
@@ -116,6 +117,12 @@ services:
       PARQUET_AND_INFO_TARGET_REVISION: ${PARQUET_AND_INFO_TARGET_REVISION-refs/convert/parquet}
       PARQUET_AND_INFO_URL_TEMPLATE: ${PARQUET_AND_INFO_URL_TEMPLATE-/datasets/%s/resolve/%s/%s}
       PARQUET_METADATA_STORAGE_DIRECTORY: ${PARQUET_METADATA_STORAGE_DIRECTORY-/parquet_metadata}
+      DUCKDB_INDEX_STORAGE_DIRECTORY: ${DUCKDB_INDEX_STORAGE_DIRECTORY-/duckdb-index}
+      DUCKDB_INDEX_COMMIT_MESSAGE: ${DUCKDB_INDEX_COMMIT_MESSAGE-Update duckdb index files}
+      DUCKDB_INDEX_COMMITTER_HF_TOKEN: ${DUCKDB_INDEX_COMMITTER_HF_TOKEN-}
+      DUCKDB_INDEX_TARGET_REVISION: ${DUCKDB_INDEX_TARGET_REVISION-duckdb/index}
```
severo, line 123:
```suggestion
      DUCKDB_INDEX_TARGET_REVISION: ${DUCKDB_INDEX_TARGET_REVISION-refs/convert/parquet}
```

tools/docker-compose-datasets-server.yml
```diff
@@ -112,6 +113,12 @@ services:
       PARQUET_AND_INFO_TARGET_REVISION: ${PARQUET_AND_INFO_TARGET_REVISION-refs/convert/parquet}
       PARQUET_AND_INFO_URL_TEMPLATE: ${PARQUET_AND_INFO_URL_TEMPLATE-/datasets/%s/resolve/%s/%s}
       PARQUET_METADATA_STORAGE_DIRECTORY: ${PARQUET_METADATA_STORAGE_DIRECTORY-/parquet_metadata}
+      DUCKDB_INDEX_STORAGE_DIRECTORY: ${DUCKDB_INDEX_STORAGE_DIRECTORY-/duckdb-index}
+      DUCKDB_INDEX_COMMIT_MESSAGE: ${DUCKDB_INDEX_COMMIT_MESSAGE-Update duckdb index file}
+      DUCKDB_INDEX_COMMITTER_HF_TOKEN: ${DUCKDB_INDEX_COMMITTER_HF_TOKEN-}
+      DUCKDB_INDEX_TARGET_REVISION: ${DUCKDB_INDEX_TARGET_REVISION-duckdb/index}
```
severo, line 119:
```suggestion
      DUCKDB_INDEX_TARGET_REVISION: ${DUCKDB_INDEX_TARGET_REVISION-refs/convert/parquet}
```

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,263 @@
...
+        ) from e
+
+    # configure duckdb extensions
+    duckdb.execute(INSTALL_EXTENSION_COMMAND.format(extension="httpfs"))
+    duckdb.execute(LOAD_EXTENSION_COMMAND.format(extension="httpfs"))
+    duckdb.execute(INSTALL_EXTENSION_COMMAND.format(extension="fts"))
+    duckdb.execute(LOAD_EXTENSION_COMMAND.format(extension="fts"))
+
+    # index all columns
+    if duckdb_index_file_directory is None:
```
severo, line 134: we could instead force `duckdb_index_file_directory` to be str, not Optional[str], no?

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,263 @@
...
+        raise CachedDirectoryNotInitializedError("Cache directory has not been initialized.")
+    db_path = Path(duckdb_index_file_directory.resolve() / DUCKDB_DEFAULT_INDEX_FILENAME)
```
severo, line 136: nit: is Path() necessary? to me the `/` operator returns a Path

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,263 @@
...
+    logging.debug(create_command_sql)
+    con.sql(create_command_sql)
+
+    # TODO: by default, 'porter' stemmer is being used, use a specific one by dataset language in the future
+    # see https://duckdb.org/docs/extensions/full_text_search.html for more details about 'stemmer' parameter
+    logging.debug(CREATE_INDEX_COMMAND)
+    con.sql(CREATE_INDEX_COMMAND)
+    con.close()
+
+    # create the target revision if it does not exist yet (clone from initial commit to avoid cloning all repo's files)
```
severo, line 152: please move this part inside the lock, as done in config-parquet-and-info

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -0,0 +1,263 @@
...
+        refs = hf_api.list_repo_refs(repo_id=dataset, repo_type=DATASET_TYPE)
+    except RepositoryNotFoundError as err:
+        raise DatasetNotFoundError("The dataset does not exist on the Hub.") from err
+
+    create_branch(
+        dataset=dataset, target_revision=target_revision, refs=refs, hf_api=hf_api, committer_hf_api=committer_hf_api
+    )
+
+    try:
+        sleeps = [1, 1, 1, 10, 10, 100, 100, 100, 300]
```
severo, line 166: please use the same sleeps as config-parquet-and-info (I think we could move this sleeps definition to utils.py)

chart/values.yaml
severo: (even if, by default, the local duckdb will be small)

severo (commented):
Still some changes are required, but very good PR!

severo:
Additional comment: the index.duckdb file will be deleted by `config-parquet-and-info`:

https://github.com/huggingface/datasets-server/blob/ad4351dc9ee2995cf716aa83afb50f8f1fb305f1/services/worker/src/worker/job_runners/config/parquet_and_info.py#L866

We have to fix this part of the code, and add an integration test (or e2e test) where we ensure that both steps don't erase the other work.

lhoestq:
I opened https://github.com/huggingface/datasets-server/pull/1429 to keep duckdb files

severo (approved):

PR #1298, opened 2023-06-02
feat: 🎸 use app token for commits to the Hub

severo:
User token is no more accepted for PARQUET_AND_INFO_COMMITTER_HF_TOKEN. It now must be an app token, associated with a user for the datasets-maintainers org. See details in the README.

BREAKING CHANGE: 🧨 the PARQUET_AND_INFO_COMMITTER_HF_TOKEN value must be changed.

Note that this PR includes https://github.com/huggingface/datasets-server/pull/1303 (upgrade dependencies, to fix a vulnerability, and adapt the code)

Also: note that we deleted the following error codes: `AskAccessHubRequestError`, `GatedDisabledError` and `GatedExtraFieldsError`: so, we should refresh all the datasets with these errors

PR #1303, opened 2023-06-05
feat: 🎸 update dependencies

severo:
main upgrades: cryptography (fixes a vulnerability that made the CI fail), pyarrow 11 -> 12. Also: mypy 1.1.1 ->1.3.0 (should not affect)


severo:
OK, the mypy upgrade seems to trigger an error when we use `# type: ignore` (rightly, because it's ignored)

severo:
The two remaining CI errors come from a change in users populated in Hub CI (see https://github.com/huggingface/moon-landing/pull/6502, internal). It will be fixed with https://github.com/huggingface/datasets-server/pull/1298.

I'm rebasing on use-app-token-for-parquet-uploading, and changing the destination branch of this PR to it

severo:
Merging

PR #1311, opened 2023-06-06
fix: Parallel job runner for split-first-rows-from-parquet

AndreaFrancis:
It must be split-first-rows-from-streaming, not config-split-names-from-info


severo (approved):
😮

well spotted!

severo:
The good point here is that it did not affect the cache: as `config-split-names-from-info` is config level, it cannot have `split != None`, which is the case for all the split level artifacts in `split-first-rows-from-parquet`. Confirmed in the database: 0 cache entry for `{kind: "split-first-rows-from-parquet", error_code:"split-first-rows-from-parquet"}`.

So: nothing to refresh after that.

AndreaFrancis:
> We have... 51,751 cache entries to refresh! Not sure why.

Should I increase the job runner version to trigger refresh for those datasets?

severo:
hmm, I'll do it manually for now

PR #1315, opened 2023-06-07
Complete raise if dataset requires manual download

### LINKED ISSUE #1307: Do not support datasets that require manual download
Currently, datasets that require manual download show a weird error message due to a TypeError. See: https://huggingface.co/datasets/timit_asr/discussions/2
```
expected str, bytes or os.PathLike object, not NoneType
```

I think we should not support theses datasets, as we already do with private datasets.

EDIT: TODO:
- [x] Raise `DatasetManualDownloadError` from `config-parquet-and-info`
  - [x] #1309
- [x] Raise `DatasetManualDownloadError` from `config-split-names-from-streaming`
  - [x] #1315
- [ ] Propagate the error codes to the aggregator steps (config -> dataset, in this case) so that we can show the appropriate message on the Hub

albertvillanova:
See: https://github.com/huggingface/datasets-server/issues/1307#issuecomment-1579288660

Fix #1307.

### DIFF
--- a/services/worker/src/worker/job_runners/config/split_names_from_streaming.py
+++ b/services/worker/src/worker/job_runners/config/split_names_from_streaming.py
@@ -5,12 +5,17 @@ import logging
 from typing import List, Optional, Union
 
 from datasets import get_dataset_split_names
+from datasets.builder import ManualDownloadError
 from datasets.data_files import EmptyDatasetError as _EmptyDatasetError
 from libcommon.constants import (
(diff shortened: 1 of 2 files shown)


severo:
> @severo any hint for the propagation of the error code? Thanks.

Good question. We should manage it in another PR/issue because it might affect the response format for all the "aggregation steps". Maybe it's "just" adding the `error_code` field (if not none) to the `failed` entries.

severo (approved):
Thanks, perfect

Issue #1323, opened 2023-06-07
The number of started jobs is a lot bigger than the number of workers


severo:
We currently have nearly 2000 started jobs, and "only" 200 workers

<img width="389" alt="Capture d’écran 2023-06-07 à 18 27 10" src="https://github.com/huggingface/datasets-server/assets/1676121/df48b775-b423-4d99-af59-7117355685c8">

My guess: the jobs raise an uncaught exception and are never finished properly in the database. They will be deleted afterwards anyway (zombie killer) but possibly we must fix something here.

severo:
Note that the three affected steps are `dataset-opt-in-out-urls-count`, `config-opt-in-out-urls-count`, `split-opt-in-out-urls-count`, which at least seems coherent in some way.

maybe @AndreaFrancis you want to check this?

AndreaFrancis:
Sure, I will check

AndreaFrancis:
Interestingly, there are jobs with "started" status but have a  "finished_at" value and a cache record in the DB. I think only the status is not being set correctly, will put here my findings.

AndreaFrancis:
I found some other records in the queue for other steps (Different than dataset-opt-in-out-urls-count, config-opt-in-out-urls-count, split-opt-in-out-urls-count) using filter` {status:"started", finished_at:{$exists:true}`}. But I was not able to reproduce the issue.
I think setting the TTL index to drop only for those status success/error would help us identify the problem.

severo:
Hmmm. No need to investigate too much. If we don't find the root reason quickly, we can let it unchanged: anyway, the cron backfill will fix the remaining issues once a day

AndreaFrancis:
> Hmmm. No need to investigate too much. If we don't find the root reason quickly, we can let it unchanged: anyway, the cron backfill will fix the remaining issues once a day

Sure, anyway I think we should not delete the inconsistent records, those are not being finished by zombie killer, and looks like queue's finish method is not letting us know if there is a wrong combination of state+dates https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/queue.py#LL632C15-L632C15 it returns a boolean but nowhere is being evaluated.
I opened https://github.com/huggingface/datasets-server/pull/1325, and would like to understand what is happening and if there is a pattern for this behavior.

severo:
[...]

PR #1325, opened 2023-06-07
Adding condition to Jobs Collection - ttl index

AndreaFrancis:
Context: https://github.com/huggingface/datasets-server/issues/1323
Some records in the queue have `finished_at` field but their final status is different than success/error, we should not have these scenarios.
Adding a condition to the TTL index will let us identify potential issues without losing data for troubleshooting.

Second part of https://github.com/huggingface/datasets-server/issues/1326
Depends on https://github.com/huggingface/datasets-server/pull/1378


libs/libcommon/src/libcommon/queue.py
```diff
@@ -131,7 +131,11 @@ class Job(Document):
             ("status", "namespace", "priority", "type", "created_at"),
             ("status", "namespace", "unicity_id", "priority", "type", "created_at"),
             "-created_at",
-            {"fields": ["finished_at"], "expireAfterSeconds": QUEUE_TTL_SECONDS},
+            {
+                "fields": ["finished_at"],
+                "expireAfterSeconds": QUEUE_TTL_SECONDS,
+                "partialFilterExpression": {"status": {"$in": [Status.SUCCESS, Status.ERROR]}},
```
severo, line 137: why not Status.CANCELLED?

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -215,4 +215,11 @@ def get_migrations(self) -> List[Migration]:
                 version="20230524192300",
             ),
             MetricsDeletionMigration(job_type="/config-names", cache_kind="/config-names", version="20230524192400"),
+            MigrationQueueDeleteTTLIndex(
```
severo, line 218: Beware, this migration has never worked, by the way, because any old code that runs during the migration will recreate the old index, and we will have a conflict (same name, different parameters). I'm not sure what is the solution.

severo (approved):

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: I created a bug for it, maybe we can handle it in a separate PR https://github.com/huggingface/datasets-server/issues/1326

libs/libcommon/src/libcommon/queue.py
AndreaFrancis: I added it to the list, thanks.

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: I would like to have a try, my guess is that the issue is the usage of `mongoengine` with the desired schema is trying to re-create the index every time while doing validations https://github.com/huggingface/datasets-server/issues/1326#issuecomment-1581688644.
But since this is the last validation, I think no others will affect the index deletion.

severo: But the workers, the api service, the admin service (and even the backfill job, if running) are running during the migration, and they use the old version of queue.py, thus they recreate the index.
Also, I'm not sure if the index deletion is immediate (I think it takes some time once asked for).

severo: maybe we have to create a new index (have both in parallel)
then delete the old one in another migration, in a new deployment (so that no running app would try to recreate it)

severo:
It's blocked by https://github.com/huggingface/datasets-server/issues/1326, right?

AndreaFrancis:
> It's blocked by #1326, right?

Yes

severo:
OK, so: we have to split this into two PR:
1. change the index (new name) in queue.py -> merge, deploy
2. then, migration to delete the old index -> merge, deploy

severo (approved):

Issue #1122, opened 2023-05-02
Create a cronjob to clean the dangling cache directories


severo:
Currently, the disk used to store the `datasets` library cache for the workers is full (shortage of inodes). It's mainly due to dangling cache directories from days or weeks. We should create a cronjob that runs every day, and deletes the directories older than 2 days.

severo:
keep open

severo:
fyi, to clean directories older than 3 days:

```bash
find . -maxdepth 1 -mtime +6 -type d -exec rm -r {} +
```

severo:
keep open.

severo:
I think it has been done for all the cache directories, right @AndreaFrancis?

AndreaFrancis:
Yes, it is enabled for: duckdb-index, hf-datasets-cache and stats-cache

PR #1332, opened 2023-06-08
Generate random cache subdirectory for dataset job runner

albertvillanova:
Generate random cache subdirectory for dataset job runner.

This might fix the stale file handle error:
```
OSError: [Errno 116] Stale file handle
```

See: https://huggingface.co/datasets/tapaco/discussions/4

### DIFF
--- a/services/worker/src/worker/job_runners/_datasets_based_job_runner.py
+++ b/services/worker/src/worker/job_runners/_datasets_based_job_runner.py
@@ -3,8 +3,8 @@
 
 import json
 import logging
+import random
 import re
-from datetime import datetime
 from hashlib import sha1
 from pathlib import Path
 from typing import Optional
@@ -44,18 +44,18 @@ class DatasetsBasedJobRunner(JobRunner):
         self.datasets_based_config = app_config.datasets_based
         self.base_datasets_cache = hf_datasets_cache
 
-    def get_cache_subdirectory(self, date: datetime) -> str:
-        date_str = date.strftime("%Y-%m-%d-%H-%M-%S")
+    def get_cache_subdirectory(self, digits: int = 14) -> str:
+        random_str = f"{random.randrange(10**(digits - 1), 10**digits)}"
         # TODO: Refactor, need a way to generate payload based only on provided params
         payload = (
-            date_str,
+            random_str,
             self.get_job_type(),
             self.job_info["params"]["dataset"],
             self.job_info["params"]["config"],
             self.job_info["params"]["split"],
         )
         hash_suffix = sha1(json.dumps(payload, sort_keys=True).encode(), usedforsecurity=False).hexdigest()[:8]
-        prefix = f"{date_str}-{self.get_job_type()}-{self.job_info['params']['dataset']}"[:64]
+        prefix = f"{random_str}-{self.get_job_type()}-{self.job_info['params']['dataset']}"[:64]
         subdirectory = f"{prefix}-{hash_suffix}"
         return "".join([c if re.match(r"[\w-]", c) else "-" for c in subdirectory])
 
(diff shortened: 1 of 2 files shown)


services/worker/src/worker/job_runners/_datasets_based_job_runner.py
```diff
@@ -44,18 +44,18 @@ def __init__(
         self.datasets_based_config = app_config.datasets_based
         self.base_datasets_cache = hf_datasets_cache
 
-    def get_cache_subdirectory(self, date: datetime) -> str:
-        date_str = date.strftime("%Y-%m-%d-%H-%M-%S")
+    def get_cache_subdirectory(self, digits: int = 14) -> str:
+        random_str = f"{random.randrange(10**(digits - 1), 10**digits)}"  # nosec B311
```
severo, line 48: I would prefer to keep the date, because it allows to manually clean the old stale directories. Normally the directory is removed automatically, but if the worker crashes (which still occurs a lot) it's kept on the disk.
See #1122.

btw: I'm currently doing such a maintenance operation, removing the dangling cache directories for may:

```bash
find . -maxdepth 1 -name "2023-05*" -type d -exec rm -rf {} \;
```

With a random string only, we lose this possibility

severo: > removing the dangling cache directories for may

it finished, and made the inodes usage fall from 82% to 65%. So, definitively not a detail

severo, line 48:
```suggestion
    def get_cache_subdirectory(self, date: datetime, digits: int = 14) -> str:
        date_str = date.strftime("%Y-%m-%d-%H-%M-%S")
        random_str = f"{random.randrange(10**(digits - 1), 10**digits)}"  # nosec B311
```

services/worker/src/worker/job_runners/_datasets_based_job_runner.py
```diff
@@ -3,8 +3,8 @@
 
 import json
 import logging
+import random
 import re
-from datetime import datetime
```
severo, line 7: keep the import

services/worker/src/worker/job_runners/_datasets_based_job_runner.py
```diff
@@ -44,18 +44,18 @@ def __init__(
...
-            date_str,
+            random_str,
             self.get_job_type(),
             self.job_info["params"]["dataset"],
             self.job_info["params"]["config"],
             self.job_info["params"]["split"],
         )
         hash_suffix = sha1(json.dumps(payload, sort_keys=True).encode(), usedforsecurity=False).hexdigest()[:8]
-        prefix = f"{date_str}-{self.get_job_type()}-{self.job_info['params']['dataset']}"[:64]
+        prefix = f"{random_str}-{self.get_job_type()}-{self.job_info['params']['dataset']}"[:64]
```
severo, line 58:
```suggestion
        prefix = f"{date_str}-{random_str}-{self.get_job_type()}-{self.job_info['params']['dataset']}"[:64]
```

severo (changes requested):
Please keep the date, as it's needed for maintenance tasks

services/worker/src/worker/job_runners/_datasets_based_job_runner.py
severo: Removing my comments!

severo (approved):

albertvillanova:
OK, as you prefer... :smile:

services/worker/src/worker/job_runners/_datasets_based_job_runner.py
severo: The command to clean after 3 days, for example:

```bash
find . -maxdepth 1 -mtime +3 -type d -exec rm -r {} +
```

PR #1340, opened 2023-06-09
[documentation] grammatical fixes in first_rows.mdx

LiamSwayne:
The changes made are grammatical and do not affect the ideas communicated in the file.


severo:
Thanks for the contribution!

Nevertheless, in that case, the text is part of a dataset, and should be maintained unchanged. WDYT @stevhliu?

stevhliu:
Yeah I agree with @severo, it is an example from a dataset and not something we authored.

severo:
Thanks for the proposal. I'll close this one, but happy to get other feedback and contributions on the docs, we try to improve it regularly.

PR #1348, opened 2023-06-12
Fix admin app

severo:
- fix the columns of dataset status
- prefill the step name (in a select element) in refresh tab

### DIFF
--- a/front/admin_ui/app.py
+++ b/front/admin_ui/app.py
@@ -168,7 +168,7 @@ with gr.Blocks() as demo:
         if response.status_code == 200:
             dataset_status = response.json()
             cached_responses_df = pd.DataFrame([{
-                    "type": job_type,
+                    "kind": cached_response["kind"],
                     "dataset": cached_response["dataset"],
                     "config": cached_response["config"],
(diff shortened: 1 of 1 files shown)


front/admin_ui/app.py
```diff
@@ -78,7 +79,8 @@ def draw_graph(width, height):
                 query_pending_jobs_button = gr.Button("Run")
                 pending_jobs_query_result_df = gr.DataFrame()
             with gr.Tab("Refresh dataset"):
-                refresh_type = gr.Textbox(label="Processing step type", placeholder="split-first-rows-from-streaming")
+                job_types = [processing_step.job_type for processing_step in PROCESSING_GRAPH.get_topologically_ordered_processing_steps()]
```
lhoestq, line 82: Note that the admin app ProcessingGraph is not synced with the production one in general, but I think it's fine

lhoestq (approved):

front/admin_ui/app.py
severo: I think that the solution is to automatize the deployment of the space from the CI.

PR #1353, opened 2023-06-12
refactor: 💡 rename required_by_dataset_viewer: enables_preview

severo:
Change the response of `/valid`, from:

```json
{
  "valid": ["datasets", "that", "have", "/first-rows"]
}
```

to

```json
{
  "preview": ["datasets", "that", "only", "have", "/first-rows"],
  "viewer": ["datasets", "that", "have", "/rows", "and", "/size"],
  "valid": ["preview", "and", "viewer"]
}
```

Note that `valid` is the same as before, and will be removed in a future PR, once the Hub has been updated to use "preview" and "viewer".

### Notes

I propose to first try this. If it works, we will remove the `valid` field, update the docs (https://huggingface.co/docs/datasets-server/valid) and do the same switch for `/is-valid`.

### DIFF
--- a/chart/static-files/openapi.json
+++ b/chart/static-files/openapi.json
@@ -290,10 +290,7 @@
         "description": "A string that identifies the underlying error for 500 on /first-rows.",
         "schema": {
           "type": "string",
-          "enum": [
-            "RowsPostProcessingError",
-            "UnexpectedError"
-          ]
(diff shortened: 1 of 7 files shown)


AndreaFrancis (approved):

severo:
Thanks for the review @AndreaFrancis. Note that it's still in draft mode, and I will ask you to review it when it's finished!

Issue #1356, opened 2023-06-13
Locking mechanism


lhoestq:
Certain jobs like config-parquet-and-info write to a repository and may have concurrency issues (see e.g. https://github.com/huggingface/datasets-server/issues/1349)

For config-parquet-and-info we need to lock the job at the dataset level, so that no other job of the same type for the same dataset write at the same time.

This can be implemented using mongodb

severo:
> lock the job at the dataset level

indeed, because all the configs share the same git branch, and we don't want to interleave commits.

PR #1357, opened 2023-06-13
Add lock for queue based on mongodb

lhoestq:
Example of usage:

```python
with lock(key=f"{job.type}--{job.dataset}", job_id=job.pk):
    ...
```
or
```python
try:
    lock(key=f"{job.type}--{job.dataset}", job_id=job.pk).acquire()
except TimeoutError:
    ...
```

### Implementation details

- I use the atomicity of `.update()` in mongoengine.
- I defined a new collection "locks" in the "queue" db
- I added created_at and updated_at in case we want to define a TTL for it later

Close https://github.com/huggingface/datasets-server/issues/1356


severo:
Just a comment:

```
key=f"{job.type}--{job.dataset}"
```

To avoid collisions, I like to use this kind of serialization
```
json.dumps({"type": job.type, "dataset": job.dataset})
```

I may be a bit less efficient, but it frees your mind from thinking about collisions.

libs/libcommon/tests/test_queue.py
```diff
@@ -383,3 +386,30 @@ def test_has_ttl_index_on_finished_at_field() -> None:
...
+    with open(tmp_file, "w") as f:
+        f.write(str(current + 1))
+
+
+def locked_increment(tmp_file: Path) -> None:
+    with lock(key="test_lock", job_id=str(os.getpid())):
+        increment(tmp_file)
+
+
+def test_lock(tmp_path_factory: pytest.TempPathFactory, queue_mongo_resource: QueueMongoResource) -> None:
```
severo, line 403: good idea to have used the same database (Queue), we don't have to create another "resource".

libs/libcommon/src/libcommon/queue.py
```diff
@@ -198,6 +202,53 @@ def flat_info(self) -> FlatJobInfo:
...
+
+class lock:
+    def __init__(self, key: str, job_id: str) -> None:
+        self.key = key
+        self.job_id = job_id
+
+    def acquire(self, sleeps: Sequence[float] = (0.05, 0.05, 0.05, 1, 1, 1, 5)) -> None:
+        try:
+            Lock(key=self.key, job_id=self.job_id, created_at=get_datetime()).save()
+        except NotUniqueError:
```
severo, line 224: clever way to have only one occurrence

libs/libcommon/src/libcommon/queue.py
```diff
@@ -198,6 +202,53 @@ def flat_info(self) -> FlatJobInfo:
...
 
 
+class Lock(Document):
+    meta = {"collection": QUEUE_COLLECTION_LOCKS, "db_alias": QUEUE_MONGOENGINE_ALIAS}
+    key = StringField(primary_key=True)
+    job_id = StringField()
+
+    created_at = DateTimeField(required=True)
+    updated_at = DateTimeField()
+
```
severo, line 212: you will have to create some indexes to speed up the queries

libs/libcommon/src/libcommon/queue.py
```diff
@@ -198,6 +202,53 @@ def flat_info(self) -> FlatJobInfo:
         )
 
 
+class Lock(Document):
+    meta = {"collection": QUEUE_COLLECTION_LOCKS, "db_alias": QUEUE_MONGOENGINE_ALIAS}
+    key = StringField(primary_key=True)
```
severo, line 207: does `primary_key` imply `unique=True`? I imagine so, but could not find it in the docs of mongoengine

libs/libcommon/src/libcommon/queue.py
```diff
@@ -198,6 +202,53 @@ def flat_info(self) -> FlatJobInfo:
...
+    job_id = StringField()
+
+    created_at = DateTimeField(required=True)
+    updated_at = DateTimeField()
+
+    objects = QuerySetManager["Lock"]()
+
+
+class lock:
```
severo, line 216: the Lock objects are created and mutated, but never deleted. I think it's not an issue because the data is small, right?

severo (approved):
Good! It will be very useful, possibly in multiple places of the code.

Main comment: please add indexes before merging.

libs/libcommon/src/libcommon/queue.py
lhoestq: Yea we can figure deletion later - there won't be a ton of locks anyway.

The trivial implementation that deletes the lock on `.release()` doesn't work because another worker could try to acquire it at the same time.

libs/libcommon/src/libcommon/queue.py
lhoestq: I don't think it's needed since I use the primary key in the queries, and there's always an index for the primary key

severo: hmmm, for https://github.com/huggingface/datasets-server/pull/1357/files#diff-44fa7cb2645881e55953db64dafa198b2e007a2e531f70acaeebfc50ffa67953R228, I think we need 

```
"indexes": [
            ("key", "job_id")
]
```

libs/libcommon/src/libcommon/queue.py
```diff
@@ -214,16 +214,40 @@ class Lock(Document):
 
 
 class lock:
-    def __init__(self, key: str, job_id: str) -> None:
+    """
+    Provides a simple way of inter-worker communication using a MondoDB lock.
```
severo, line 218:
```suggestion
    Provides a simple way of inter-worker communication using a MongoDB lock.
```

lhoestq:
added the index and fixed the typo

severo:
gogogo

PR #1364, opened 2023-06-14
feat: 🎸 run jobs killing routines with some random delay

### LINKED ISSUE #1330: Use randomness in zombie+long jobs killing process?
Every worker runs the function to kill zombies every 10 minutes. But as all the workers are launch about at the same, it's pretty strange: nothing happens during 10 minutes, then 80 workers try to kill the zombies.

Same for the long jobs.

The interval between run could be random between 5 and 15 minutes, for example.

severo:
instead of running the zombie killing routine, and the long jobs killing routine on a regular basis, we run them between 50% and 150% of that delay. Fixes #1330

### DIFF
--- a/services/worker/src/worker/executor.py
+++ b/services/worker/src/worker/executor.py
@@ -27,7 +27,7 @@ START_WORKER_LOOP_PATH = start_worker_loop.__file__
 async def every(
     func: Callable[..., Optional[Any]],
     *args: Any,
-    seconds: Union[float, Tuple[float, float]],
+    seconds: Union[int, Tuple[float, float]],
     stop_on: Optional[Any] = None,
     **kwargs: Any,
 ) -> None:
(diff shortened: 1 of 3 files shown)


services/worker/src/worker/executor.py
```diff
@@ -24,13 +25,20 @@
 
 
 async def every(
-    func: Callable[..., Optional[Any]], *args: Any, seconds: int, stop_on: Optional[Any] = None, **kwargs: Any
+    func: Callable[..., Optional[Any]],
+    *args: Any,
+    seconds: Union[int, Tuple[float, float]],
```
lhoestq, line 30:
```suggestion
    seconds: Union[float, Tuple[float, float]],
```

lhoestq (approved):
lgtm !

services/worker/src/worker/executor.py
severo: it's what I had, but we are passing `int` to the function. I'll change this.

PR #1368, opened 2023-06-14
feat: 🎸 reduce the resources

severo:


severo:
it's so good not to trigger the docs and the CodeQL analysis when we change a variable in the Helm chart!

PR #1376, opened 2023-06-15
[docs] Improvements

stevhliu:
Based on @mishig25's [feedback](https://huggingface.slack.com/archives/C0311GZ7R6K/p1684484245379859), this adds:
- a response to the code snippets in the Quickstart
- an end-to-end example of using `/parquet` to get a dataset, analyze it, and plot the results
- button to open in a Colab notebook to run examples right away

### DIFF
--- a/docs/source/quick_start.mdx
+++ b/docs/source/quick_start.mdx
@@ -1,5 +1,7 @@
 # Quickstart
 
+[[open-in-colab]]
+
 In this quickstart, you'll learn how to use the Datasets Server's REST API to:
 
 - Check whether a dataset on the Hub is functional.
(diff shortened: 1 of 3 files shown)


docs/source/quick_start.mdx
```diff
@@ -87,6 +89,13 @@ curl https://datasets-server.huggingface.co/is-valid?dataset=rotten_tomatoes \
 </curl>
 </inferencesnippet>
 
+You;ll see the following error if you're trying to access a gated dataset without providing your user token:
```
severo, line 92:
```suggestion
You'll see the following error if you're trying to access a gated dataset without providing your user token:
```

severo (approved):
excellent!

I could not access the colab notebooks, is it normal?

Issue #1361, opened 2023-06-14
Increase PIL limit for images size


severo:
See https://huggingface.co/datasets/datadrivenscience/ship-detection/discussions/4#6489674ff4d5239fac762f2a

PIL has a limit for the image size (178,956,970 pixels). When the datasets library tries to get the value for the big image, it raises `DecompressionBombError`

https://github.com/huggingface/datasets-server/blob/7b4b78f04473d8f00e7c2226381982a9a84ab48c/services/worker/src/worker/utils.py#L341

<strike>I'm not sure if we could handle this better (ie: ignoring the error just for these images, replacing with a placeholder?)</strike> 
- [ ] increase the limit
- [ ] catch the exception and return a proper advice for the user

wdyt @huggingface/datasets?

lhoestq:
I think we can just increase the limit. The idea is just not to OOM

severo:
Duplicate of #1192 (but we will keep this one)

severo:
Copying references found by @albertvillanova 

https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open

> If the number of pixels is greater than twice MAX_IMAGE_PIXELS, then a DecompressionBombError will be raised instead.

> This threshold can be changed by setting MAX_IMAGE_PIXELS. It can be disabled by setting Image.MAX_IMAGE_PIXELS = None.

> The default value for MAX_IMAGE_PIXELS is 0.25GB, which allows opening images up to 0.5GB (twice the value)

PR #1387, opened 2023-06-16
fix: 🐛 support bigger images

severo:
fixes https://github.com/huggingface/datasets-server/issues/1361


services/worker/src/worker/utils.py
```diff
@@ -33,6 +34,9 @@
 from libcommon.exceptions import NormalRowsError, StreamingRowsError
 from libcommon.utils import orjson_dumps
 
+MAX_IMAGE_PIXELS = 1_000_000_000
```
albertvillanova, line 37: Do you think it is good to have this constant here instead of in a config file?

albertvillanova (approved):
Thanks!
Just wondering if we can properly handle this larger size...

services/worker/src/worker/utils.py
severo: I was wondering if we really want to configure this, or if it's OK to hardcode. For now, I would say it's OK. Maybe we could move it to a constants.py file if it's cleaner (I think it's not the only hardcoded value not in a constants.py file).

lhoestq: ok to have it here imo, but if it's used somewhere else in the future we can move it

lhoestq (approved):
cool !

PR #1388, opened 2023-06-16
New parent job runner for cached data

AndreaFrancis:
Currently, we have datasets_based_job_runner but we need a new one that only creates a chance folder without modifying the datasets library config.
Context https://github.com/huggingface/datasets-server/pull/1296#discussion_r1232512427


services/worker/tests/job_runners/test__datasets_based_worker.py
severo, line 1: the new file should also have its tests file

services/worker/src/worker/job_runners/_cached_based_job_runner.py
```diff
@@ -0,0 +1,63 @@
...
+from libcommon.processing_graph import ProcessingStep
+from libcommon.storage import init_dir, remove_dir
+from libcommon.utils import JobInfo
+
+from worker.config import AppConfig
+from worker.job_runner import JobRunner
+
+
+class CachedBasedJobRunner(JobRunner):
+    """Base class for job runners that use a cached directory."""
```
severo, line 20:
```suggestion
    """Base class for job runners that use a temporary cache directory."""
```

services/worker/src/worker/job_runners/_cached_based_job_runner.py
```diff
@@ -0,0 +1,63 @@
...
+
+from libcommon.processing_graph import ProcessingStep
+from libcommon.storage import init_dir, remove_dir
+from libcommon.utils import JobInfo
+
+from worker.config import AppConfig
+from worker.job_runner import JobRunner
+
+
+class CachedBasedJobRunner(JobRunner):
```
severo, line 19:
```suggestion
class CacheDirectoryJobRunner(JobRunner):
```

services/worker/src/worker/job_runners/_cached_based_job_runner.py
```diff
@@ -0,0 +1,63 @@
...
+        super().__init__(
+            job_info=job_info,
+            app_config=app_config,
+            processing_step=processing_step,
+        )
+        self.base_cache_directory = cache_directory
+
+    def get_cache_subdirectory(self, digits: int = 14) -> str:
+        random_str = f"{random.randrange(10**(digits - 1), 10**digits)}"  # nosec B311
+        # TODO: Refactor, need a way to generate payload based only on provided params
```
severo, line 41: do we need this comment? what is the idea?

services/worker/src/worker/job_runners/_datasets_based_job_runner.py
```diff
@@ -10,10 +10,10 @@
 from libcommon.utils import JobInfo
 
 from worker.config import AppConfig
-from worker.job_runners._cached_based_job_runner import CachedBasedJobRunner
+from worker.job_runners._cached_directory_job_runner import CacheDirectoryJobRunner
```
severo, line 13:
```suggestion
from worker.job_runners._cache_directory_job_runner import CacheDirectoryJobRunner
```

services/worker/src/worker/job_runners/_cached_based_job_runner.py
AndreaFrancis: I copied it from the previous implementation, currently, it generates the payload from job_info (dataset, config, and split) but I thought that in the future we could have a better way to generate it dynamically. 
But I can remove it does not makes sense.

services/worker/tests/job_runners/test__datasets_based_worker.py
AndreaFrancis: Done, I separated the current test file in two according to the scenario.

polinaeterna (approved):
awesome! i will also use it in https://github.com/huggingface/datasets-server/pull/1199

PR #1390, opened 2023-06-18
Add Docker internal to extra_hosts

baskrahmer:
This is required to connect to the local DB instance on Linux; it is already added to `tools/docker-compose-dev-datasets-server.yml`


severo (approved):
I think it's OK, right @lhoestq?

lhoestq (approved):
yup, thanks !

Issue #1163, opened 2023-05-11
Catch commit conflit in parquet branch


severo:
See https://huggingface.co/datasets/cais/mmlu/discussions/9.

We have to catch the exception (412) here: https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/parquet_and_info.py#L904-L911

and raise a retriable exception so that the error is only temporary.

severo:
We have to fix this

severo:
Note that the lock should make this error disappear. See https://github.com/huggingface/datasets-server/pull/1355 and https://github.com/huggingface/datasets-server/pull/1357.

severo:
But we have another one:

```
Give up after 6 attempts with <class 'huggingface_hub.utils._errors.HfHubHTTPError'>

Error code:   UnexpectedError
```

which is due to timeout after waiting for too long for the lock to be released.

I think we should 1. catch this better to make the error retryable, 2. increase the duration

severo:
cc @lhoestq

lhoestq:
hmm actually this happens when the dataset is locked and the commits are being done.

But it seems to fail because the precondition "parent_commit" fails, though it is defined as
```python
parent_commit=parent_commit if not commit_infos else commit_infos[-1].oid
```

which sounds correct to me

lhoestq:
Maybe it's an issue with consistency - we could add a write/read concern for the lock requests to try to fix that

severo:
> But it seems to fail because the precondition "parent_commit" fails, though it is defined as

Indeed, it's a 412 error code, that seem to occur when we set parent_commit, and a commit has happened since
https://github.com/huggingface/huggingface_hub/blob/92028da882e47d61703e8c9144028e1ecadab415/tests/test_hf_api.py#L732

severo:
On https://huggingface.co/datasets/gsarti/flores_101/viewer, it seems to occur for half of the configs, which is a lot!

<img width="754" alt="Capture d’écran 2023-06-20 à 13 55 12" src="https://github.com/huggingface/datasets-server/assets/1676121/134af3d5-1b03-4cef-b5a0-7861deb431e1">

severo:
> Maybe it's an issue with consistency - we could add a write/read concern for the lock requests to try to fix that

it could be indeed

severo:
> `parent_commit=parent_commit if not commit_infos else commit_infos[-1].oid`
> which sounds correct to me.

[...]

PR #1393, opened 2023-06-20
Raise retryable error on hfhubhttperror

severo:
see https://github.com/huggingface/datasets-server/issues/1163


lhoestq (approved):
nice !

severo:
Note: it fixes https://github.com/huggingface/datasets-server/issues/1163 but not the underlying issue which is: why do the commits conflict?

PR #1355, opened 2023-06-12
Multi commit Parquet copies

lhoestq:
close https://github.com/huggingface/datasets-server/issues/1349

PR #1398, opened 2023-06-20
Rename parent job runners

AndreaFrancis:
Based on the new job runner for those that need a  cached directory on https://github.com/huggingface/datasets-server/pull/1388
We will need a new split job runner that inherits from the new JobRunnerWithCache to be used as part of https://github.com/huggingface/datasets-server/pull/1199 and https://github.com/huggingface/datasets-server/pull/1296/files.
According to the discussions here, I renamed the parent job runners to:

> CacheDirectoryJobRunner -> JobRunnerWithCache
> DatasetsBasedJobRunner -> JobRunnerWithDatasetsCache

And their children with Dataset/Config/Split + JobRunnerWithCache/JobRunnerWithDatasetsCache


lhoestq:
DatasetsBasedConfigJobRunner ?

lhoestq:
DatasetsBasedJobRunner could also be renamed JobRunnerWithHFDatasetsCache.

This way we can have ConfigJobRunnerWithHFDatasetsCache

severo:
Who can propose a longer name? Maybe we have a Java dev in the team 😛

polinaeterna:
@severo I can! I have the following suggestion:

`DatasetsBasedJobRunner` -> `HfDatasetsCacheDirectoryJobRunner` (since it inherits from `CacheDirectoryJobRunner` so it has the same logic but the cache dir is not arbitrary, it's related to `hf-datasets`)
and then just add a level to the left as we did before.

So from `CacheDirectoryJobRunner` we have:
* `DatasetCacheDirectoryJobRunner`
* `ConfigCacheDirectoryJobRunner`
* `SplitCacheDirectoryJobRunner`

and from `HfDatasetsCacheDirectoryJobRunner` we have the same:
* `DatasetHfDatasetsCacheDirectoryJobRunner`
* `ConfigHfDatasetsCacheDirectoryJobRunner`
* `SplitHfDatasetsCacheDirectoryJobRunner`

as a note aside - my husband is a Java dev :laughing:

polinaeterna:
okay that's horrible i see

AndreaFrancis:
I thought only having `SplitCacheDirectoryJobRunner`, `ConfigCacheDirectoryJobRunner,` and `DatasetCacheDirectoryJobRunner` and leaving the others as they were was enough.

severo:
> upd: to save some symbols we can cut Directory to Dir 😄

as I wrote in https://github.com/huggingface/datasets-server/pull/1355#discussion_r1232141816:
> I think we should avoid abbreviations like rev, I agree with https://github.com/huggingface/moon-landing/pull/5339/files <- every variable/function should ideally be made of valid words

And: I'm OK with long explicit names. We have to try to avoid having long names if we can have shorter ones, but if we have the shortest explicit name, it's ok to me

polinaeterna:
> I thought only having SplitCacheDirectoryJobRunner, ConfigCacheDirectoryJobRunner, and DatasetCacheDirectoryJobRunner and leaving the others as they were was enough.
I think it's good boot I just thought that old names (with `Cached`) are maybe not easily interpretable (what is cached here?)

AndreaFrancis:
> I think it's good but I just thought that old names (with Cached) are maybe not easily interpretable (what is cached here and what's the difference from Cache?)

Agree, DatasetsBasedJobRunner and its children should be renamed.

I liked @polinaeterna suggestion:

>     DatasetCacheDirectoryJobRunner
>     ConfigCacheDirectoryJobRunner
>     SplitCacheDirectoryJobRunner


But for the library ones, I don't have a strong opinion about putting the explicit relation for DatasetsBasedJobRunner and CacheDirectoryJobRunner, that is why I thought something like LibraryBasedJobRunner  and then we could have:

- SplitLibraryBasedJobRunner
- ConfigLibraryBasedJobRunner
- DatasetLibraryBasedJobRunner 

But not sure, let's vote for the better suggestion!

polinaeterna:
> But for the library ones, I don't have a strong opinion about putting the explicit relation for DatasetsBasedJobRunner and CacheDirectoryJobRunner, that is why I thought something like LibraryBasedJobRunner and then we could have:

    SplitLibraryBasedJobRunner
    ConfigLibraryBasedJobRunner
    DatasetLibraryBasedJobRunner

I totally see the point in this, it's just that here it's not clear what "library" means

AndreaFrancis:
Thanks! @polinaeterna @lhoestq and @severo 
I renamed it as the final comment, this PR is ready for review again.

polinaeterna (approved):
looks all good to me, thank you!

severo (approved):
Good!

lhoestq (approved):
Looks good !

PR #1404, opened 2023-06-20
Retry get_parquet_file_and_size

lhoestq:
In prod I got an ArrowInvalid when instantiating a pq.ParquetFile for bigcode/the-stack-dedup even though all the parquet files are valid (I ran a script and checked I could get all the pq.ParquetFile objects)


severo (approved):
one comment

PR #1412, opened 2023-06-21
unblock DFKI-SLT/few-nerd

lhoestq:


severo (approved):
I imagine it's OK, because GitHub diff does not help here

Issue #1416, opened 2023-06-21
Truncated first-rows may crop image URLs


lhoestq:
see https://huggingface.co/datasets/Antreas/TALI-base

because of that the images are not show in the UI

PR #1417, opened 2023-06-21
keep image and audio untruncated

lhoestq:
close https://github.com/huggingface/datasets-server/issues/1416


lhoestq:
Maybe we can leave device=None except when the user had set somethign like `torch.set_default_tensor_type('torch.cuda.FloatTensor')`

severo (approved):
Good idea.

Should we not ensure the content is not `{bytes: ... a long series of bytes ...}`? I remember having seen this kind of cells in Image columns

lhoestq:
> Should we not ensure the content is not {bytes: ... a long series of bytes ...}? I remember having seen this kind of cells in Image columns

It was when the Image() feature type was not recognized and was fixed recently.
Though if it ever happens again it will get truncated.

PR #1418, opened 2023-06-22
Create /filter endpoint

albertvillanova:
Create /filter endpoint.

I have tried to follow roughly the same logic as the /rows endpoint.

TODO:
- [x] e2e tests
- [x] chart: only increase the number of replicas
- ~~docker-compose files~~
- [x] openapi specification
- [x] documentation pages: draft of `filter.mdx' 

Subsequent PRs:
- Index all datasets, not only those with text features. See:
  - #1854
- Complete the `filter.mdx` docs page
  - Give examples for all the supported operators
  - Enumerate all the supported column data types
- More complex validation of request parameters (to avoid SQL injection)
- Maybe rename the `search` service to `query`, with the 2 endpoints `search` and `filter`


services/api/src/api/routes/filter.py
```diff
@@ -0,0 +1,111 @@
...
+
+def get_config_parquet_metadata_from_cache(
+    dataset: str, config: str, processing_graph: ProcessingGraph
+) -> tuple[list[ParquetFileMetadataItem], Optional[str]]:
+    config_parquet_metadata_processing_steps = processing_graph.get_config_parquet_metadata_processing_steps()
+    if not config_parquet_metadata_processing_steps:
+        raise RuntimeError("No processing steps are configured to provide config-parquet-metadata response.")
+    cache_kinds = [step.cache_kind for step in config_parquet_metadata_processing_steps]
+    try:
+        result = get_previous_step_or_raise(
```
albertvillanova, line 100: The function `get_previous_step_or_raise` is not tested anywhere.

severo: could you add one, or open an issue?

severo:
@albertvillanova I'll let you reopen

lhoestq:
I think we can continue this PR now no ? This will be super useful for classification dataset (text/image/audio) and for tabular datasets with numerical values

albertvillanova:
I'm retaking this task...

albertvillanova:
I started following the logic from /rows, but /rows has substantially changed during the last months. I'm integrating the relevant changes.

albertvillanova:
The /filter endpoint needs to be moved to its own service. Is there any script for this or should I do it by hand?

CC: @severo

severo:
yes, you have to do it by hand, the same way as /rows and /search. At least, if you think none of those can share the same codebase as /filter (rows uses parquet, search uses duckdb)

albertvillanova:
Maybe we could start a round of reviews... :sweat_smile: 

My current pending tasks are factorizing `get_config_parquet_metadata_from_cache` and `get_features_from_parquet_file_metadata`, whose code is also used by /rows.

services/filter/src/filter/main.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 The HuggingFace Authors.
```
severo, line 2: nit: replace 2022 with 2023 in the new files

services/filter/src/filter/routes/filter.py
```diff
@@ -0,0 +1,240 @@
...
+from libcommon.storage import StrPath
+from libcommon.utils import PaginatedResponse
+from libcommon.viewer_utils.features import (
+    get_supported_unsupported_columns,
+    to_features_list,
+)
+from starlette.requests import Request
+from starlette.responses import Response
+
+MAX_ROWS = 100
```
severo, line 39: nit: factor this constant and other parts of the file that are shared with /rows and/or /search.

services/filter/src/filter/routes/filter.py
```diff
@@ -0,0 +1,240 @@
...
+
+# audio still has some errors when librosa is imported
+UNSUPPORTED_FEATURES = [Value("binary")]
+
+FILTER_QUERY = """\
+    SELECT {columns}
+    FROM read_parquet({parquet_file_urls})
```
severo, line 46: we use Duckdb to query parquet files, but we also have the Duckdb version of the file.

1. Would it make sense to use the Duckdb file instead? what are the pros/cons?
2. A specific aspect of parquet vs. Duckdb: some datasets are huge, with many parquet files (might be more than 1TB of parquet files), while the Duckdb index file is always under 10GB at most. How does/filter perform on big datasets?

services/filter/src/filter/routes/filter.py
```diff
@@ -0,0 +1,240 @@
...
+    WHERE {where}
+    LIMIT {limit}
+    OFFSET {offset}"""
```
severo, line 49: would it make sense to have `ORDER BY`? (not sure about the performance implications)

services/filter/src/filter/routes/filter.py
```diff
@@ -0,0 +1,240 @@
...
+                        parquet_file_metadata_item=parquet_file_metadata_items[0],
+                        parquet_metadata_directory=parquet_metadata_directory,
+                    )
+                with StepProfiler(method="filter_endpoint", step="get supported and unsupported columns"):
+                    supported_columns, unsupported_columns = get_supported_unsupported_columns(
+                        features,
+                        unsupported_features=UNSUPPORTED_FEATURES,
+                    )
+                with StepProfiler(method="filter_endpoint", step="execute filter query"):
+                    num_rows_total, pa_table = execute_filter_query(
```
severo, line 136: is pa_table copied in the memory?

severo (commented):
Very nice! Some questions in the review.

I also list what is missing:
- e2e tests
- chart
- docker-compose files
- openapi specification
- documentation pages

services/filter/src/filter/routes/filter.py
albertvillanova: I have no idea of the feasibility of using the DuckDB index instead of the Parquet files. Maybe @AndreaFrancis can give some insights?
- Is it possible to use the DuckDB index to filter the files based on a WHERE clause (like e.g. `age = 30`)?

services/filter/src/filter/routes/filter.py
albertvillanova: Naively I would say no: this is just a "filter" (WHERE clause), not a complete SQL SELECT endpoint.
- the additional LIMIT and OFFSET clauses were added because we need them for pagination

Having said that, however, it may be a nice feature to have as well...

services/filter/src/filter/routes/filter.py
albertvillanova: I aligned with the ocde in /search:
https://github.com/huggingface/datasets-server/blob/88257755b2b8ab5a16540f98b621aab826f1dfe5/services/search/src/search/routes/search.py#L268

And yes, I think the pa_table is created in memory. It only has `MAX_ROWS = 100` though... Anyway, sooner or later we need the data in memory to render it on the viewer. See code in `to_rows_list`:
https://github.com/huggingface/datasets-server/blob/88257755b2b8ab5a16540f98b621aab826f1dfe5/libs/libapi/src/libapi/utils.py#L212

services/filter/src/filter/routes/filter.py
severo: OK. Let's keep that for a next iteration of the endpoint, if needed on the Hub

services/filter/src/filter/routes/filter.py
severo: OK, you're right, it's small

services/filter/src/filter/main.py
albertvillanova: Done.

services/filter/src/filter/routes/filter.py
albertvillanova: Done.

services/filter/src/filter/routes/filter.py
AndreaFrancis: I think we can use the same file as in /search. I think the index does not affect the performance because it creates like another table in another schema.
**Duckdb**
pros:
- We can share the downloaded data for /filter and /search.  
- I think that native SQL queries on an already physical DB should be faster than SQL-parquet

**Using remote parquet**
pros:
- No need to download files 
cons:
- Connectivity issues

severo: Also:
- duckdb: limited to the first 5GB
- remote parquet: no limit (see https://huggingface.co/datasets/EleutherAI/the_pile_deduplicated/tree/main/data for example), but... the filter operation might take ages

albertvillanova: Thanks a lot for your insights, @AndreaFrancis and @severo. :hugs: 

I think I will use the index then, so that /search and /filter are aligned.

severo: Note that, if it makes sense, they can share the same "service" (ie: two endpoints: /search and /filter) in services/search. I don't know if it make sense, but it's worth evaluating. cc @AndreaFrancis

AndreaFrancis: I think it makes sense, we would have to increase the number of replicas for the existing service

albertvillanova: Thanks again for your hints. I will need your help with the replicas...

severo: re Replicas: it's just the number of pods that run the service. It's set here:

https://github.com/huggingface/datasets-server/blob/1afb2818bf88746494bf2770dfc33616eb68c661/chart/env/prod.yaml#L342

We will possibly have to multiply the number by 2

lhoestq: Note that DuckDB indexes are currently only created for datasets that have FTS-indexable (=string) columns. We would have to change this to have a DuckDB file for every datasets

albertvillanova: Ups, thanks @lhoestq. Should that also be addressed in this PR, or better in another one?

lhoestq: Better in another one IMO. For now you can raise an error if the duckdb-index step is missing or has an error

albertvillanova: OK. Thanks again.

services/search/src/search/routes/filter.py
```diff
@@ -0,0 +1,334 @@
...
+                                cache_folder=f"{duckdb_index_file_directory}/{HUB_DOWNLOAD_CACHE_FOLDER}",
+                                index_folder=index_folder,
+                                target_revision=target_revision,
+                                dataset=dataset,
+                                repo_file_location=repo_file_location,
+                                hf_token=hf_token,
+                            )
+                with StepProfiler(method="filter_endpoint", step="get features"):
+                    if "features" in content and isinstance(content["features"], dict):
+                        features = Features.from_dict(content["features"])
```
albertvillanova, line 177: @huggingface/datasets-server can we safely assume that this condition is always true?

polinaeterna: i think `content["features"]` also might be `None` if i'm not mistaken. not sure though

albertvillanova: Thanks @Polina for your answer. :hugs: 
Any idea about in which cases can it be None?

lhoestq: You can raise an error if it's None imo.

It can only happen in very rare cases (parquet dataset processed with old version of `datasets` for example)

albertvillanova: Thanks, I am removing the else branch as well and all the parquet metadata functions! :hugs:

docs/source/filter.mdx
```diff
@@ -0,0 +1,12 @@
...
+
+Datasets Server provides a `/filter` endpoint for filtering rows in a dataset.
+
+<Tip warning={true}>
+  Currently, only <a href="./parquet">datasets with Parquet exports</a>
+  are supported so Datasets Server can index the contents and run the search without
+  downloading the whole dataset.
+</Tip>
+
+This guide shows you how to use Datasets Server's `/filter` endpoint to filter rows based on a query string.
```
severo, line 11: Nice to have prepared the docs. It would be nice to explain more (here or in openapi.json) the values / syntax accepted for `query`

services/search/src/search/routes/filter.py
```diff
@@ -0,0 +1,251 @@
...
+                        dataset=dataset,
+                        config=config,
+                        split=split,
+                        hf_endpoint=hf_endpoint,
+                        hf_token=hf_token,
+                        hf_timeout_seconds=hf_timeout_seconds,
+                        cache_max_days=cache_max_days,
+                    )
+                    revision = duckdb_index_cache_entry["dataset_git_revision"]
+                    if duckdb_index_cache_entry["http_status"] != HTTPStatus.OK:
```
AndreaFrancis, line 136: Maybe you will need to add a similar logic as in https://github.com/huggingface/datasets-server/pull/1854/files#diff-b31b94865cf356d2b241fc7f38f98bd9484d61464a991d06364bdf82c2c5ba4eR245 (but it depends which one is going to be merged first).

docs/source/filter.mdx
albertvillanova: Done.

services/search/src/search/routes/filter.py
albertvillanova: Done.

albertvillanova:
Do you think this PR could be merged?

services/filter/src/filter/routes/filter.py
severo: > Note that DuckDB indexes are currently only created for datasets that have FTS-indexable (=string) columns. We would have to change this to have a DuckDB file for every datasets

done here: https://github.com/huggingface/datasets-server/pull/1854

libs/libcommon/src/libcommon/utils.py
```diff
@@ -100,6 +100,9 @@ class PaginatedResponse(TypedDict):
     num_rows_per_page: int
 
 
+MAX_NUM_ROWS_PER_PAGE = 100
```
severo, line 103: should we handle this with `FIRST_ROWS_MAX_NUMBER`? Or hardcode `FIRST_ROWS_MAX_NUMBER` and use `MAX_NUM_ROWS_PER_PAGE` instead (I think so)

services/search/src/search/routes/filter.py
```diff
@@ -0,0 +1,278 @@
...
+                    auth_check(
+                        dataset=dataset,
+                        external_auth_url=external_auth_url,
+                        request=request,
+                        hf_jwt_public_keys=hf_jwt_public_keys,
+                        hf_jwt_algorithm=hf_jwt_algorithm,
+                        hf_timeout_seconds=hf_timeout_seconds,
+                    )
+                # TODO: duplicated in /search
+                with StepProfiler(method="search_endpoint", step="validate indexing was done"):
```
lhoestq, line 129:
```suggestion
                with StepProfiler(method="filter_endpoint", step="validate indexing was done"):
```

services/search/src/search/routes/filter.py
```diff
@@ -0,0 +1,278 @@
...
+                    revision = duckdb_index_cache_entry["dataset_git_revision"]
+                    if duckdb_index_cache_entry["http_status"] != HTTPStatus.OK:
+                        return get_json_error_response(
+                            content=duckdb_index_cache_entry["content"],
+                            status_code=duckdb_index_cache_entry["http_status"],
+                            max_age=max_age_short,
+                            error_code=duckdb_index_cache_entry["error_code"],
+                            revision=revision,
+                        )
+                with StepProfiler(method="search_endpoint", step="download index file if missing"):
```
lhoestq, line 151:
```suggestion
                with StepProfiler(method="filter_endpoint", step="download index file if missing"):
```

libs/libcommon/src/libcommon/utils.py
severo: hmmm: but we set `FIRST_ROWS_MAX_NUMBER` to 4 in the tests...

libs/libcommon/src/libcommon/viewer_utils/features.py
```diff
@@ -30,6 +30,9 @@
 from libcommon.utils import FeatureItem
 from libcommon.viewer_utils.asset import create_audio_file, create_image_file
 
+# audio still has some errors when librosa is imported
```
severo, line 33: remove this comment?

```suggestion
```

docs/source/openapi.json
```diff
@@ -3317,6 +3317,370 @@
         }
       }
     },
+    "/filter": {
+      "get": {
+        "summary": "Filter rows of a split",
+        "description": "Returns the rows matching the filter query, ordered by row index. Up to 100 rows are returned. The offset and length parameters allow to navigate the results.",
+        "externalDocs": {
+          "description": "See search (Hub docs).",
```
severo, line 3325:
```suggestion
          "description": "See filter (Hub docs).",
```

docs/source/openapi.json
```diff
@@ -3317,6 +3317,370 @@
...
+            },
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/PaginatedResponse"
+                },
+                "examples": {
+                  "Filter on a text dataset (julien-c/titanic-survival)": {
+                    "summary": "The first 2 rows where the 'Age' column equals to 30 (where=Age=30&length=2).",
+                    "description": "Try with: https://datasets-server.huggingface.co/filter?dataset=julien-c/titanic-survival&config=default&split=train&where=Age=30&length=2",
```
severo, line 3423: could you add some more examples with different queries (different operators + different column types, for example)?

services/search/tests/routes/test_filter.py
```diff
@@ -0,0 +1,94 @@
...
+def test_execute_filter_query(index_file_location: str) -> None:
+    columns, where, limit, offset = ["name", "age"], "gender = 'female'", 1, 1
+    num_rows_total, pa_table = execute_filter_query(
+        index_file_location=index_file_location, columns=columns, where=where, limit=limit, offset=offset
+    )
+    assert num_rows_total == 2
+    assert pa_table == pa.Table.from_pydict({"name": ["Simone"], "age": [30]})
+
+
+# def test_create_response(ds: Dataset, app_config: AppConfig, cached_assets_directory: StrPath) -> None:
```
lhoestq, line 54: Do you plan to uncomment this test ?
If you don't need it you can remove it or explain that it's a function that was copy pasted from /rows and is already tested there

chart/env/prod.yaml
```diff
@@ -341,7 +341,7 @@ search:
   uvicornNumWorkers: "9"
   nodeSelector:
     role-datasets-server-search: "true"
-  replicas: 8
+  replicas: 16
```
lhoestq, line 344: We have only 6 machines that can have up to 2 pods each. You can set it to 12 maximum
```suggestion
  replicas: 12
```

see https://github.com/huggingface/infra/blob/69465295f1331c68dd7768aafa3a497f0ab6044e/projects/hub/00-aws/variables/hub-aws-prod-us-east-1.tfvars#L190-L199

docs/source/filter.mdx
```diff
@@ -0,0 +1,33 @@
+# Filter rows in a dataset
+
+Datasets Server provides a `/filter` endpoint for filtering rows in a dataset.
+
+<Tip warning={true}>
+  Currently, only <a href="./parquet">datasets with Parquet exports</a>
+  are supported so Datasets Server can index the contents and run the search without
```
severo, line 7: this part should be adapted a bit for "filter"

docs/source/filter.mdx
```diff
@@ -0,0 +1,33 @@
...
+  downloading the whole dataset.
+</Tip>
+
+This guide shows you how to use Datasets Server's `/filter` endpoint to filter rows based on a query string.
+Feel free to also try it out with [ReDoc](https://redocly.github.io/redoc/?url=https://datasets-server.huggingface.co/openapi.json#operation/filterRows).
+
+The `/filter` endpoint accepts five query parameters:
```
severo, line 14:
```suggestion
The `/filter` endpoint accepts six query parameters:
```

or remove the value (more robust to future edits)

docs/source/filter.mdx
```diff
@@ -0,0 +1,33 @@
...
+- `where`: the filter condition
+- `offset`: the offset of the slice, for example `150`
+- `length`: the length of the slice, for example `10` (maximum: `100`)
+
+The `where` parameter must be expressed as a comparison predicate, which can be:
+- a simple predicate composed of a column name, a comparison operator, and a value
+  - the comparison operators are: `=`, `<>`, `>`, `>=`, `<`, `<=`
+- a composite predicate composed of two or more simple predicates (optionally grouped with parentheses to indicate the order of evaluation), combined with logical operators
+  - the logical operators are: `AND`, `OR`, `NOT`
+
```
severo, line 28: what are the supported column types? Maybe give several examples below

Also: I was wondering: do we have to apply a specific encoding to the URL when passing the query?

lhoestq (approved):
Amazing ! Looking forward to play with it ❤️

services/search/src/search/routes/filter.py
```diff
@@ -0,0 +1,278 @@
...
+
+    return filter_endpoint
+
+
+def execute_filter_query(
+    index_file_location: str, columns: list[str], where: str, limit: int, offset: int
+) -> tuple[int, pa.Table]:
+    con = duckdb.connect(database=index_file_location, read_only=True)
+    # TODO: Address possible SQL injection CWE-89
+    filter_query = FILTER_QUERY.format(columns=",".join(columns), where=where, limit=limit, offset=offset)
```
severo, line 226: OK. We really don't look at what is passed in the where clause. I understand that it would be complex, but... we don't check it's against a supported column type, or if the query makes sense, or if it's dangerous...

I'm no specialist, so I would prefer other opinions on this. cc @AndreaFrancis maybe

services/search/src/search/routes/filter.py
```diff
@@ -0,0 +1,278 @@
...
+    pa_table = con.sql(filter_query).arrow()
+    filter_count_query = FILTER_COUNT_QUERY.format(where=where)
```
severo, line 228: same here

services/search/tests/routes/test_filter.py
severo, line 54: remove dead code?

services/search/tests/routes/test_filter.py
```diff
@@ -0,0 +1,94 @@
...
+    os.remove(index_file_location)
+
+
+def test_execute_filter_query(index_file_location: str) -> None:
+    columns, where, limit, offset = ["name", "age"], "gender = 'female'", 1, 1
+    num_rows_total, pa_table = execute_filter_query(
+        index_file_location=index_file_location, columns=columns, where=where, limit=limit, offset=offset
+    )
+    assert num_rows_total == 2
+    assert pa_table == pa.Table.from_pydict({"name": ["Simone"], "age": [30]})
```
severo, line 51: related to the comment above, on just passing the `where` parameter to duckdb. I think we want:
- to do a lot of tests, at least for every supported column type, and for several operators
- to pass wrong or "dangerous" queries and see what happens

severo (commented):
I'm uneasy with just passing the contents of `where` to duckdb. At least, if we do so, I would prefer having more tests to ensure we handle well the expected cases (wrong query, "dangerous" query that modifies the duckdb file, all the supported column types, several operators, etc)

chart/env/prod.yaml
severo: we can also increase the value in infra

lhoestq:
> I'm uneasy with just passing the contents of where to duckdb. At least, if we do so, I would prefer having more tests to ensure we handle well the expected cases (wrong query, "dangerous" query that modifies the duckdb file, all the supported column types, several operators, etc)

The duckdb file is opened in read-only mode. I also think we can improve error handling as a second step

lhoestq:
Maybe just ban a few keywords for now like select, copy, export, install, load

lhoestq:
I'm checking if there's a lib like sqlite that can create the full query and make sure it's safe

lhoestq:
Alternatively we could just have triplets (column, operand, value) and make sure columns and operands are valid, and escape the value using dedicated tools like psycopg2

albertvillanova:
Thanks a lot @lhoestq and @severo for your detailed reviews! :hugs: 

I am addressing your suggestions.

chart/env/prod.yaml
albertvillanova: I set it to 12 until we increase the value in infra.

services/search/tests/routes/test_filter.py
albertvillanova: Uncommented, so it is also tested with the search service app config.

albertvillanova: I uncommented it.

libs/libcommon/src/libcommon/viewer_utils/features.py
albertvillanova: Done.

docs/source/filter.mdx
albertvillanova: I replaced "search" with "filter query". Is it OK?

docs/source/filter.mdx
albertvillanova: Done.

e2e/tests/test_53_search_filter.py
```diff
@@ -0,0 +1,67 @@
...
+    filter_response = poll_until_ready_and_assert(
+        relative_url=(
+            f"/filter?dataset={dataset}&config={config}&split={split}&offset={offset}&length={length}&where={where}"
+        ),
+        expected_status_code=expected_status_code,
+        expected_error_code=expected_error_code,
+        headers=headers,
+        check_x_revision=True,
+    )
+    if not expected_error_code:
```
AndreaFrancis, line 28: I think we can remove this validation since we know expected_error_code is always None.
I saw same code in test_rows_endpoint and test_search_endpoint  :confused:

libs/libapi/src/libapi/response.py
```diff
@@ -0,0 +1,63 @@
...
+from datasets import Features
+from libcommon.s3_client import S3Client
+from libcommon.storage import StrPath
+from libcommon.storage_options import DirectoryStorageOptions, S3StorageOptions
+from libcommon.utils import MAX_NUM_ROWS_PER_PAGE, PaginatedResponse
+from libcommon.viewer_utils.features import to_features_list
+
+from libapi.utils import to_rows_list
+
+CACHED_ASSETS_S3_SUPPORTED_DATASETS: list[str] = ["asoria/image"]  # for testing
```
AndreaFrancis, line 11: After https://github.com/huggingface/datasets-server/pull/1882, I think you can remove this constant.

libs/libapi/src/libapi/response.py
```diff
@@ -0,0 +1,63 @@
...
+    offset: int,
+    features: Features,
+    unsupported_columns: list[str],
+    num_rows_total: int,
+) -> PaginatedResponse:
+    if set(pa_table.column_names).intersection(set(unsupported_columns)):
+        raise RuntimeError(
+            "The pyarrow table contains unsupported columns. They should have been ignored in the row group reader."
+        )
+    storage_options = (
```
AndreaFrancis, line 32: Also here, you can directly use a S3StorageOptions

services/search/src/search/routes/filter.py
AndreaFrancis: I think we are good for now but maybe wrap this part in a try/except block to provide better feedback to the client.
Also, consider that we load the db file in read-only mode.
Regarding dangerous code, I'm not sure if we can use the where wildcard (?) with prepared statements https://duckdb.org/docs/api/c/prepared.html (avoid SQL Injection).

services/search/src/search/routes/filter.py
AndreaFrancis: Maybe I am wrong but I saw you added 
`if duckdb_index_cache_entry["content"]["has_fts"] is not True:`
in search.py but I can't see something similar here

AndreaFrancis (approved):
Thanks for all the great work refactoring /rows and /search!

libs/libapi/src/libapi/response.py
albertvillanova: Yes, there are again conflicts with the main branch. I am going to fix them once more... :sweat_smile: 
```
This branch has conflicts that must be resolved
Use the web editor or the to resolve conflicts.

Conflicting files:
services/rows/src/rows/routes/rows.py
services/rows/tests/routes/test_rows.py
services/search/src/search/routes/search.py
```

severo:
> Regarding dangerous code, I'm not sure if we can use the where wildcard (?) with prepared statements https://duckdb.org/docs/api/c/prepared.html (avoid SQL Injection).

It seems wise to use this!

albertvillanova:
> Regarding dangerous code, I'm not sure if we can use the where wildcard (?) with prepared statements https://duckdb.org/docs/api/c/prepared.html (avoid SQL Injection).

I remember having read this some time ago and if I remember correctly, it was not possible to use it for completely free-form `where`, as we require.

That could be used only if the where had the column name and the operator already fixed by us, and the user was providing only the value.

Let me check it again.

albertvillanova:
Additionally, I have been testing the /filter with different SQL injections and I have found that:
- as it is opened in read-only mode, no modification of the data can be performed
- users can inject code to have access to all the table/indexes names and all the content of the table:
  - I think this is useless for malicious actors because they have already complete access to the `index.duckdb` file itself (it is public)
  - Anyway, we could implement some validation to avoid any potential disruption of the viewer
    - As @lhoestq suggested, we can check there are no other SQL commands (SELECT, SHOW TABLES, DESCRIBE,...) and no SQL special characters (`;`, `--`, `/*`,...)

libs/libapi/src/libapi/response.py
albertvillanova: Done.

libs/libapi/src/libapi/response.py
albertvillanova: Done.

albertvillanova:
Finally, instead of checking for the presence of SQL commands (the list of forbidden words would be quite long and may interfere with potential valid column names), I validate there are no SQL syntax symbols required to inject SQL commands, like `;` or comments `--`.

What do you think?

lhoestq:
Sounds good to me as soon as we add some additional protections. In particular I think you can run those commands when initializing the connection:

1. Disable local and remote access

```duckdb
SET disabled_filesystems='LocalFileSystem,HTTPFileSystem';
```

2. disable access external states (e.g. install/load extensions, copy to, from, read CSVs, scan pandas)

```duckdb
SET enable_external_access=false;
```

3. disable extensions autoload

```duckdb
SET autoload_known_extensions=false;
```

lhoestq:
Arf it seems that `disabled_filesystems` is only available in `duckdb`0.9 which can't read the current duckdb indexes that were created using 0.8.1

lhoestq:
I think running this should be fine instead

```duckdb
SET enable_external_access=false;
SET file_search_path='hf://';
SET lock_configuration=true;
```

setting the file_search_path to this value (`hf://`, or any value containing `://`) prevents glob from working at all (for both relative and absolute paths)

albertvillanova:
I am fixing the conflicts:
```
This branch has conflicts that must be resolved

Conflicting files
e2e/tests/test_11_api.py
e2e/tests/test_52_search.py
```

e2e/tests/test_53_search_filter.py
albertvillanova: Done.

albertvillanova:
I am fixing:
```
This branch has conflicts that must be resolved

Conflicting files
services/search/pyproject.toml
```

services/api/src/api/routes/filter.py
albertvillanova: Done:
- #1908

libs/libcommon/src/libcommon/utils.py
severo: yes, sure!

docs/source/openapi.json
albertvillanova: There are 4 examples now on:
- float equality
- string equality
- float inequality
- logical and with integer equality/inequality

docs/source/filter.mdx
albertvillanova: Finally I just wrote a single example but more complex: with AND/OR and equality/inequality operators, with float, integer and string data types.

I think that when writing more complete docs in a subsequent PR, we could add several detailed examples. Is it OK?

services/search/src/search/routes/filter.py
albertvillanova: I implemented the `try... except` around the SQL query, so that we return an `InvalidParameterError`.

services/search/src/search/routes/filter.py
albertvillanova: If the `where` is problematic, it raises an exception in the precedent SQL query.

services/search/tests/routes/test_filter.py
albertvillanova: See `test_validate_where_parameter`, `test_validate_where_parameter_raises`, `test_execute_filter_query` and `test_execute_filter_query_raises` in `test_filter.py` file.

albertvillanova:
> EDIT: setting enable_external_access to false seems to block http requests and globbing local files!
@lhoestq I am reverting those configurations settings then...

services/search/src/search/routes/filter.py
albertvillanova: At the end, I put the count query under try-except as well.

services/search/src/search/routes/filter.py
albertvillanova: In relation with the validation of `where` parameter, see `validate_where_parameter` done when getting `where` from `response` in the body of `create_filter_endpoint`.

albertvillanova:
@severo may I merge this PR?

albertvillanova:
As mentioned in the description of this PR, it will remain for subsequent PRs:
- Complete the `filter.mdx` docs page
  - Give examples for all the supported operators
  - Enumerate all the supported column data types
- More complex validation of request parameters (to avoid SQL injection)
- Maybe rename the `search` service to `query`, with the 2 endpoints `search` and `filter`

libs/libapi/src/libapi/duckdb.py
```diff
@@ -0,0 +1,119 @@
...
+        hf_token=hf_token,
+        hf_timeout_seconds=hf_timeout_seconds,
+        cache_max_days=cache_max_days,
+        blocked_datasets=blocked_datasets,
+    )
+
+
+def duckdb_connect(**kwargs: Any) -> duckdb.DuckDBPyConnection:
+    con = duckdb.connect(read_only=True, **kwargs)
+    con.sql("SET enable_external_access=false;")
```
severo, line 117: we never use duckdb to directly access the remote parquet files or duckdb index file stored in refs/convert/parquet, right?

services/search/tests/routes/test_filter.py
severo: OK. Maybe add more tests in the next PRs if we get bugs (test to reproduce the bug, and check it's fixed)

severo (approved):
gogogo!

(only one comment)

libs/libapi/src/libapi/duckdb.py
albertvillanova: https://github.com/huggingface/datasets-server/pull/1418#issuecomment-1742788076
> EDIT: setting enable_external_access to false seems to block http requests and globbing local files!

albertvillanova: @severo I would say that this duckdb connection is only used with the downloaded local index because the `duckdb_connect` is only used in the /search and /filter endpoints, both with the argument `index_file_location`, that is the path of the previously downloaded index duckdb file (which we download with `get_index_file_location_and_download_if_missing`).

Therefore I'm merging this PR; anyway we could revert this configuration setting if necessary before deploying.

PR #1420, opened 2023-06-22
Ensure only one job is started for the same unicity_id

### LINKED ISSUE #1323: The number of started jobs is a lot bigger than the number of workers

severo:
To avoid multiple job runners getting the same job at the same time, for a given unicity_id (identifies job type + parameters):
- a lock is used during the update of the selected job
- we ensure no other job is already started
- we select the newest (in date order) job from all the waiting jobs and start it (status, started_at)
- we cancel all the other waiting jobs

Should fix #1323.

### DIFF
--- a/libs/libcommon/src/libcommon/queue.py
+++ b/libs/libcommon/src/libcommon/queue.py
@@ -408,45 +408,6 @@ class Queue:
         except Exception:
             return 0
 
-    def cancel_jobs(
-        self,
-        job_type: str,
-        dataset: str,
-        config: Optional[str] = None,
-        split: Optional[str] = None,
-        statuses_to_cancel: Optional[List[Status]] = None,
-    ) -> List[JobDict]:
-        """Cancel jobs from the queue.
-
-        Note that the jobs for all the revisions are canceled.
-
-        Returns the list of canceled jobs (as JobDict, before they are canceled, to be able to know their previous
-        status)
-
-        Args:
-            job_type (`str`): The type of the job
-            dataset (`str`): The dataset on which to apply the job.
-            config (`str`, optional): The config on which to apply the job.
-            split (`str`, optional): The config on which to apply the job.
-            statuses_to_cancel (`list[Status]`, optional): The list of statuses to cancel. Defaults to
-                [Status.WAITING, Status.STARTED].
-
-        Returns:
-            `list[JobDict]`: The list of canceled jobs
-        """
-        if statuses_to_cancel is None:
-            statuses_to_cancel = [Status.WAITING, Status.STARTED]
-        existing = Job.objects(
-            type=job_type,
-            dataset=dataset,
-            config=config,
-            split=split,
-            status__in=statuses_to_cancel,
-        )
-        job_dicts = [job.to_dict() for job in existing]
-        existing.update(finished_at=get_datetime(), status=Status.CANCELLED)
-        return job_dicts
-
     def cancel_jobs_by_job_id(self, job_ids: List[str]) -> int:
         """Cancel jobs from the queue.
 
--- a/libs/libcommon/src/libcommon/queue.py
+++ b/libs/libcommon/src/libcommon/queue.py
@@ -4,6 +4,7 @@
 import contextlib
 import json
 import logging
+import random
 import time
 import types
 from collections import Counter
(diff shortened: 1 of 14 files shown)


lhoestq:
I'm fine with renaming job_id -> owner

lhoestq:
If a worker crashes, the global lock will stay acquired and it will block the full datasets-server.

Maybe instead you can do one lock per job unicity-id ?

severo:
> I'm fine with renaming job_id -> owner
done

severo:
> Maybe instead you can do one lock per job unicity-id ? And if a worker can't acquire it it must retry picking a new job in the queue.

You're right, we have to add some rules around `unicity_id` (the tuple `type,dataset,config,split`). For one `unicity_id` value, we can have only 4 states:
- `w0s0`: no waiting job, no started job
- `w1s0`: one waiting job, no started job
- `w0s1`: no waiting job, one started job
- `w1s1`: one waiting job, one started job

The transitions of the state machine are:
- `w0s0 > w1s0`: add job
- `w0s1 > w1s1`: add job
- `w1s0 > w1s0`: add job (the new job replaces the current waiting job)
- `w1s1 > w1s1`: add job (the new job replaces the current waiting job)
- `w1s0 > w0s1`: start job
- `w0s1 > w0s0`: end job
- `w1s1 > w1s0`: end job

AndreaFrancis:
I don't quite understand why if a Job is processed twice it still remains in the "`started`" state.
In case of a second job runner processing the Job, it should have modified the finished_at and also the state to error/success.

Besides that, I had another naive idea to solve reading multiple times the same "waiting" Job, but it could affect our concurrent pipeline processing speed.
Currently, our [Queue](https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/queue.py#L303) implementation is not really a global FIFO queue, because it is like we have one queue per pod. I think that is the problem.
We should have a single point of entry to get the `get_next_waiting_job` and as a FIFO queue, we could set the new state as "started" (Like doing a dequeue operation so that no one else will try to process the Job).
This global queue should be accessible and shared among all the pods/nodes/applications that want to process a Job (maybe an API service or something like that).
In case we need to scale maybe we can create like topics? (Maybe one topic by job type?).

Not sure if that is the way queue messaging applications work, but in those cases, they remove the next item in the queue. 
We can't do the same because we need to keep the history of the item (Job).

severo:
> because it is like we have one queue per pod

No, the jobs collection is the same for all the pods. And that's what creates the race conditions.

severo:
I'll try to propose the following changes (thanks @lhoestq for the ideas):
- one lock per "unicity_id": if a worker wants to lock a job, but cannot acquire it, it sleeps and loops, as occurs with EmptyJobQueue (already in the PR)
- remove the current restriction on the waiting jobs (only one waiting job, the others are canceled when creating a new one)
- on job start (ie. when changing the state of a waiting job to started), cancel all the other waiting jobs. As we're inside the lock, we should not have concurrency issues. Note that other waiting jobs might be created meanwhile, but it's the expected behavior.

severo:
> Yes, I meant, we have many different instances running get_next_waiting_job (from code perspective, not database).

OK. Yes, we also spoke about moving to a single component that would start the jobs. But: possibly we want it to scale at one point, and it would require locks anyway. So... let's first try the locks?

libs/libcommon/src/libcommon/queue.py
```diff
@@ -609,17 +547,70 @@ def get_next_waiting_job(
...
+                ).order_by("-created_at")
+                datetime = get_datetime()
+                # raise if any job has already been started for unicity_id
+                num_started_jobs = waiting_jobs(status=Status.STARTED).count()
+                if num_started_jobs > 0:
+                    if num_started_jobs > 1:
+                        logging.critical(f"job {job.unicity_id} has been started {num_started_jobs} times. Max is 1.")
+                    raise AlreadyStartedError(f"job {job.unicity_id} has been started by another worker")
+                # get the most recent one
+                first_job = waiting_jobs.first()
```
lhoestq, line 582: What if the first job is a job that is based on a parent job that has not been run yet ? It might run on outdated data no ?

libs/libcommon/src/libcommon/queue.py
```diff
@@ -96,6 +96,18 @@ class JobDoesNotExistError(DoesNotExist):
     pass
 
 
+class AlreadyStartedError(Exception):
```
AndreaFrancis, line 99:
```suggestion
class JobAlreadyStartedError(Exception):
```
Or something like that?

AndreaFrancis (approved):
Thanks!

libs/libcommon/src/libcommon/queue.py
severo: Yes, it happens a lot, and it's not an issue. It's better to have an explicit error or a partial response than an empty cache. And it will be re-run when the parent job is finished.

Anyway, this PR does not change anything in that aspect.

lhoestq: ah yes indeed !

PR #1423, opened 2023-06-23
Remove too restrictive __all__ definitions

albertvillanova:
Remove too restrictive `__all__` from `libcommon/simple_cache`:
- it contained only one attribute: `DoesNotExist`
- Python considers as private all the module attributes which are not defined in `__all__`

```
from libcommon.simple_cache import upsert_response

Warning: Accessing a protected member of a class or a module:
'upsert_response' is not declared in __all__
```

### DIFF
--- a/jobs/mongodb_migration/src/mongodb_migration/database_migrations.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/database_migrations.py
@@ -5,7 +5,6 @@ import types
 from typing import Generic, Type, TypeVar
 
 from mongoengine import Document
-from mongoengine.errors import DoesNotExist
 from mongoengine.fields import StringField
 from mongoengine.queryset.queryset import QuerySet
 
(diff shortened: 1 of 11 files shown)


severo (approved):
OK.

I don't like having an import from mongoengine in many places of the code. To me, only the modules that operate on the database (queue, cache, etc) should explicitly import from mongoengine, and in the rest of the code, we should consume an error raised by these modules. I think I shouldn't have re-exported, but instead, I should have caught the DoesNotExist exception and raised another exception from it.

Nevermind, it's a nit, I think, more a conceptual issue for now

albertvillanova:
Yes, that is the right approach if you don't want to have an explicit dependence on `mongoengine` in other parts of the code: re-raise a custom error.

severo (commented):
thanks a lot for the refactoring!

severo (approved):

severo:
It seems like the CI is failing, because some DoesNotExist exceptions have not been caught

albertvillanova:
Yes I saw that on Friday, but needed to go out... I'm fixing those now.

severo:
> What do you think?

Sure. I don't think there are many cases where we call "Document" (or inherited) objects from external code, at least I always took care of not doing it, but if there are some we should change.

Issue #1341, opened 2023-06-09
Create missing Jobs when /rows cache does not exists yet


AndreaFrancis:
In https://github.com/huggingface/datasets-server/pull/1287, /rows were modified to depend on `parquet_utils` which calls `get_previous_step_or_raise` instead of `get_cache_entry_from_steps`. This implies that if the cache entry does not exist yet, it will throw a `CachedArtifactError` error instead of retrying later + create missing Jobs if needed.  
See discussion https://github.com/huggingface/datasets-server/pull/1287#discussion_r1224576658
We should keep consistency in the responses of API level and have the same behavior on /rows as in other endpoints.

PR #1425, opened 2023-06-23
 Create missing Jobs when /rows cache does not exists yet

AndreaFrancis:
Should fix https://github.com/huggingface/datasets-server/issues/1341


severo (approved):
With a nice refactoring!

PR #1426, opened 2023-06-23
Move dtos to its own file

AndreaFrancis:
Currently, we have data transfer objects in the `utils.py` file, but this one is growing and growing and sometime we could have duplicated code if don't verify the existing models.
This PR just moves all the abstractions of responses to a `dtos.py` file.
Most of the files are changes on the `imports` definition.


severo (approved):
Good idea!

Issue #1433, opened 2023-06-27
Can't copy parquet files with path that can be URL encoded


lhoestq:
we get this error for bigcode/the-stack in config-parquet-and-info

```
huggingface_hub.utils._errors.EntryNotFoundError: Cannot copy data/c%2B%2B/train-00000-of-00214.parquet at revision 349a71353fd5868fb90b593ef09e311379da498a: file is missing on repo.
```

PR #1434, opened 2023-06-27
Unquote path and revision when copying parquet files

lhoestq:
close https://github.com/huggingface/datasets-server/issues/1433


severo:
nice! Do you think we could have the same bug in other parts of the code? In particular when we write on the disk (assets, etc.)

lhoestq:
> Do you think we could have the same bug in other parts of the code? In particular when we write on the disk (assets, etc.)

No it happens only when we get the file name out of the URL of the file and we only do it here afaik

Issue #1257, opened 2023-05-25
Generate 5GB parquet files for big datasets


AndreaFrancis:
For datasets over 5GB, let's generate 5GB parquet files (with shards) instead of ignoring them. The fact that the dataset was truncated should be stored somewhere.

---

Currently, datasets server gets stores parquet files if the dataset size is less than PARQUET_AND_INFO_MAX_DATASET_SIZE config.
- `PARQUET_AND_INFO_MAX_DATASET_SIZE`: the maximum size in bytes of the dataset to pre-compute the parquet files. Bigger datasets, or datasets without that information, are ignored. Defaults to `100_000_000`.
For future implementations/features we would probably need a full dataset reading, big datasets won't be available for it.

There are a couple of suggestions to mitigate excluding big datasets from full reading:
- Copy at least first x GB of data
- If a dataset is already in parquet files copy them directly instead of processing
- To have a data lake?

severo:
Yes, I think the first two points are what we should do.

For `Copy at least first x GB of data`: we will have to transfer that information in all the processing steps that depend on it. eg. the opt-in/out box on the Hub should say that the analysis was made only on the first 5GB, or the /rows last page should mention that the following rows are not available (only 5GB can be accessed)...

For "If a dataset is already in parquet files copy them directly instead of processing", see https://github.com/huggingface/datasets-server/issues/1273

severo:
See also https://github.com/huggingface/datasets-server/issues/908

lhoestq:
Currently we download everything and generate everything, and we want at least partial generation.
But I think we also need partial download (or it may take ages to download the data and fill the disk).

Therefore ideally we should only download what's needed and only generate what's needed.
Shall we use streaming mode on big datasets in this case ? :)

For big non-streamable datasets it's less easy but doable (maybe with a LazyDownloadManager). IMO we can see later if we need to support non-streamable datasets ?

severo:
I have no idea of what is possible with the datasets library. OK to do with streaming first.

severo:
Once done, try on mozilla-foundation/common_voice_10_0 and answer https://twitter.com/cosmos_218/status/1667801640710328320

PR #1448, opened 2023-06-28
Stream convert to parquet

lhoestq:
Allow to have datasets partially converted to parquet, like c4, refinedweb, oscar, etc.

Datasets above 5GB are streamed to generate 5GB (uncompressed) of parquet files.

## Implementation details

I implemented a context manager `limite_parquet_writes` that does some monkeypatching  in the `datasets` lib to stop the dataset generation at the right time. This is not the kind of features that would be implemented in the `datasets` library natively, so I'm convinced this is the easiest way to do it. I added a test for this helper to make sure it works as expected and keeps working in the future.

I'm also using the new path scheme `config/split/partial/SSSS.parquet`. I didn't change the scheme for regular parquet conversion though - it can be done later.

close https://github.com/huggingface/datasets-server/issues/1257


jobs/mongodb_migration/src/mongodb_migration/migrations/_20230628180100_cache_add_partial_field_in_config_parquet_and_info.py
severo, line 1: can you add a test?

severo: Also: you have to list it in collector.py, or it will be ignored

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -53,10 +68,6 @@
     DatasetInBlockListError,
     DatasetManualDownloadError,
     DatasetNotFoundError,
-    DatasetTooBigFromDatasetsError,
-    DatasetTooBigFromHubError,
-    DatasetWithTooBigExternalFilesError,
-    DatasetWithTooManyExternalFilesError,
```
severo, line 59: Excellent

- can you also remove them from the exceptions file (I think they are not used anymore)?
- in this PR we should include a backfill that retries these error codes: https://github.com/huggingface/datasets-server/blob/6ae555fb078e2e47ca3794174f8006bb57ccefe2/chart/env/prod.yaml#L124

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -83,29 +94,52 @@
...
+        self.shard_idx = shard_idx
+        self.partial = partial
 
     @property
     def path_in_repo(self) -> str:
-        return f'{self.config}/{self.local_file.removeprefix(f"{self.local_dir}/")}'
+        if self.partial:
+            return f"{self.config}/partial/{self.split}/{self.shard_idx:04d}.parquet"
+        else:
+            return f'{self.config}/{self.local_file.removeprefix(f"{self.local_dir}/")}'
```
severo, line 118: you didn't try to move the other files to the same directory structure? I think it's better to manage this in another PR indeed

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -83,29 +94,52 @@
...
         self.config = config
+        self.split = split
+        self.shard_idx = shard_idx
+        self.partial = partial
 
     @property
     def path_in_repo(self) -> str:
-        return f'{self.config}/{self.local_file.removeprefix(f"{self.local_dir}/")}'
+        if self.partial:
+            return f"{self.config}/partial/{self.split}/{self.shard_idx:04d}.parquet"
```
severo, line 116: for safety, I would add a test that raises if shard_idx >= 10000

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -1064,6 +1214,7 @@ def compute_config_parquet_and_info_response(
         for repo_file in target_dataset_info.siblings
         if repo_file.rfilename.startswith(f"{config}/") and repo_file.rfilename.endswith(".parquet")
     ]
+    repo_files.sort(key=lambda repo_file: repo_file.rfilename)  # type: ignore
```
severo, line 1217: indeed, lambda functions and mypy are enemies. https://stackoverflow.com/questions/33833881/is-it-possible-to-type-hint-a-lambda-function

alternative: create a function and call it by its name (ie: no lambda)

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -311,20 +322,8 @@ def raise_if_not_supported(
         `ParquetResponseResult`: An object with the parquet_response
           (dataset and list of parquet files) and the dataset_git_revision (sha) if any.
     Raises the following errors:
-        - [`libcommon.exceptions.DatasetManualDownloadError`]:
-          If the dataset requires manual download.
         - [`libcommon.exceptions.DatasetRevisionNotFoundError`]
```
severo, line 325: can this function still raise all the remaining listed exceptions?

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -639,7 +629,7 @@ def fill_builder_info(builder: DatasetBuilder, hf_token: Optional[str]) -> None:
             split = str(split)  # in case it's a NamedSplit
             try:
                 parquet_files_and_sizes: List[Tuple[pq.ParquetFile, int]] = thread_map(
-                    partial(retry_get_parquet_file_and_size, fs=fs, hf_token=hf_token),
+                    functools.partial(retry_get_parquet_file_and_size, fs=fs, hf_token=hf_token),
```
severo, line 632: ah ah, name collision

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -671,10 +817,7 @@ def convert_to_parquet(builder: DatasetBuilder) -> List[CommitOperationAdd]:
     builder.download_and_prepare(
         file_format="parquet"
     )  # the parquet files are stored in the cache dir and it fills the info
```
severo, line 819: does it mean that anyway: the full dataset must be downloaded by the datasets library? It might break things

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -660,8 +650,164 @@ def fill_builder_info(builder: DatasetBuilder, hf_token: Optional[str]) -> None:
...
+        num_shards = len(split_info.shard_lengths) if isinstance(split_info.shard_lengths, list) else 1
+        fname = f"{builder.name}-{split}.parquet"
+        local_parquet_files.extend(
+            [
+                ParquetFile(
+                    local_file=os.path.join(
+                        builder.cache_dir,
+                        fname.replace(
+                            ".parquet",
+                            f"-{shard_idx:05d}-of-{num_shards:05d}.parquet" if num_shards > 1 else ".parquet",
```
severo, line 758: isn't this implicit dependency on the datasets library implementation?

services/worker/src/worker/job_runners/config/parquet_and_info.py
severo: or maybe I didn't understand well: at this point, are the files already downloaded locally?

If we stream the dataset later, it wouldn't make sense to download them, so... I imagine not.

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -660,8 +650,164 @@ def fill_builder_info(builder: DatasetBuilder, hf_token: Optional[str]) -> None:
                 builder.info.dataset_size += approx_num_bytes
 
 
+class limit_parquet_writes:
```
severo, line 653: isn't it worth going to the datasets library?

severo: (not blocking for this PR)

severo (approved):
Incredible!

This is a game changer!

services/worker/src/worker/job_runners/config/parquet_and_info.py
lhoestq: this code is `convert_to_parquet` which is for datasets <5GB

services/worker/src/worker/job_runners/config/parquet_and_info.py
lhoestq: yep, will add a comment

services/worker/src/worker/job_runners/config/parquet_and_info.py
lhoestq: No for several reasons: API doesn't fit the lib, it may creates caching issues, it probably breaks push_to_hub etc.

services/worker/src/worker/job_runners/config/parquet_and_info.py
severo: what is the error?

severo: oh, ok, so it's because huggingface_hub is not typed? (clearer alternative could be to raise if it's not an instance of str)

lhoestq: yea a ValueError makes more sense - I'll update it

lhoestq:
I took all your comments into account :)

services/worker/src/worker/job_runners/config/parquet_and_info.py
severo: I didn't get it: you put 👍 but didn't add the test :)

severo:
Thanks! go ahead.

just: did you miss https://github.com/huggingface/datasets-server/pull/1448#discussion_r1247990666 or is it just not important?

services/worker/src/worker/job_runners/config/parquet_and_info.py
lhoestq: I added a test in the `__init__` ;)

severo: OK, perfect

PR #1459, opened 2023-06-29
fix: split-duckdb-index error when indexing columns with spaces

AndreaFrancis:
After deploying split-duckdb-index, there are some errors because of column names with spaces like:
```
duckdb.BinderException: Binder Error: Referenced column "Mean" not found in FROM clause!
Candidate bindings: "read_parquet.Model"
LINE 1: ...f_index_id, Ranking,User,Model,Results,**Mean Reward**,Std Reward FROM read_parque...
    
                                          ^
```    
This PR fixes this error.

### DIFF
--- a/services/worker/src/worker/job_runners/split/duckdb_index.py
+++ b/services/worker/src/worker/job_runners/split/duckdb_index.py
@@ -42,7 +42,7 @@ STRING_FEATURE_DTYPE = "string"
 VALUE_FEATURE_TYPE = "Value"
 DUCKDB_DEFAULT_INDEX_FILENAME = "index.duckdb"
 CREATE_SEQUENCE_COMMAND = "CREATE OR REPLACE SEQUENCE serial START 1;"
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
Good catch :)

severo (approved):

severo:
We will have to re-run the step on the erroneous datasets. Is it easy to identify them?

AndreaFrancis:
It should be using `{kind:"split-duckdb-index", error_code:"UnexpectedError", details.cause_exception:"BinderException"}` but I am seeing that there are a lot, not sure if a manual refresh will work. Maybe a backfill but since it is marked as "UnexpectedError" not sure if we can filter only the ones with this kind of issue

severo:
The alternative is to get the list of dataset names with a query in MongoDB, then to call /admin/force-refresh for each of them

severo:
This is also why it's useful to have a specific error_code for every possible exception -> it makes it easier to re-run only the affected datasets, and to monitor the errors.

PR #1468, opened 2023-07-03
Rename `/dataset-info` endpoint to `/info`

polinaeterna:
Question: do we want to show results of steps that have `pending` and `failed` keys? I assume it might be not clear for users what that means, they also sound a bit too technical. Should we explain in the docs or just don't allow access to the dataset-level aggregations (but if so, why do we even need these cache entries).


severo:
> Question: do we want to show results of steps that have pending and failed keys? I assume it might be not clear for users what that means, they also sound a bit too technical. Should we explain in the docs or just don't allow access to the dataset-level aggregations (but if so, why do we even need these cache entries).

100% agree with you. This part of the API is not clear. Let's discuss it in a new issue?

docs/source/info.mdx
```diff
@@ -0,0 +1,139 @@
+# Get dataset information
+
+Datasets Server provides an `/info` endpoint for exploring the general information about dataset, including such fields as description, citation, homepage, license and features.
+
+The `/info` endpoint accepts three query parameters:
```
severo, line 5:
```suggestion
The `/info` endpoint accepts two query parameters:
```

severo (approved):

severo (approved):
Nice

Issue #1219, opened 2023-05-22
Update old cache entries automatically


severo:
Some cache entries are very old.

For example, the following entry has been computed more than 3 months ago, and contain an error_code that is no more present in the codebase. It should have been recomputed at some point, but we didn't for some reason:

<img width="1652" alt="Capture d’écran 2023-05-22 à 10 22 29" src="https://github.com/huggingface/datasets-server/assets/1676121/c6cea7a1-ebc6-4431-8eb2-5859209d7854">

severo:
Note that in that case, we have a cache entry, but the dataset itself (and even the user) have been deleted: https://huggingface.co/datasets/MasterCorneo/sd-ti-corneo-dataset

We should also delete these stale cache entries (and their relative assets)

severo:
keep open

severo:
For the second step (creating a TTL): it's not urgent, since currently 3k cache entries are older than 3 months, to be compared with more the 1M cache entries.

PR #1469, opened 2023-07-03
feat: 🎸 backfill cache entries older than 90 days

severo:
See https://github.com/huggingface/datasets-server/issues/1219.

The idea is to have a general limitation on the duration of the cache. It will make it easier to delete unused resources (assets, cached assets, etc) later: everything older than 120 days (eg) can be deleted.


libs/libcommon/tests/test_orchestrator.py
```diff
@@ -38,6 +38,8 @@
     artifact_id_to_job_info,
 )
 
+MAX_CACHE_DAYS = 180
```
AndreaFrancis, line 41:
```suggestion
CACHE_MAX_DAYS = 180
```

services/admin/src/admin/routes/dataset_state.py
```diff
@@ -1,63 +0,0 @@
-# SPDX-License-Identifier: Apache-2.0
```
AndreaFrancis, line 1: Don't we need this endpoint?

AndreaFrancis (approved):

libs/libcommon/tests/test_orchestrator.py
severo: Thanks, changed (and the value too)

PR #1471, opened 2023-07-03
Add partial to subsequent parquet-and-info jobs

lhoestq:
Following #1448 we need all subsequent jobs to `config-parquet-and-info` to have the "partial" field:

- "config-parquet-and-info"
- "config-parquet"
- "dataset-parquet"
- "config-parquet-metadata"
- "config-info"
- "dataset-info"

For dataset level jobs, "partial" is True if there is at least one config with "partial == True".

I updated the code of all the jobs and update the migration job to add "partial" to existing entries.

I'll update the docs in another PR.

### DIFF
--- a/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230703110100_cache_add_partial_field_in_config_parquet_and_info.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/migrations/_20230703110100_cache_add_partial_field_in_config_parquet_and_info.py
@@ -17,12 +17,21 @@ class MigrationAddPartialToCacheResponse(Migration):
         # See https://docs.mongoengine.org/guide/migration.html#example-1-addition-of-a-field
         logging.info(
             "If missing, add the partial field with the default value (false) to the cached results of"
-            " config-parquet-and-info"
+            " config-parquet-and-info and subsequent steps"
         )
         db = get_db(CACHE_MONGOENGINE_ALIAS)
         db[CACHE_COLLECTION_RESPONSES].update_many(
             {
-                "kind": "config-parquet-and-info",
+                "kind": {
+                    "$in": [
+                        "config-parquet-and-info",
+                        "config-parquet",
+                        "dataset-parquet",
+                        "config-parquet-metadata",
+                        "config-info",
+                        "dataset-info",
+                    ]
+                },
                 "http_status": 200,
                 "content.partial": {"$exists": False},
             },
@@ -33,7 +42,20 @@ class MigrationAddPartialToCacheResponse(Migration):
         logging.info("Remove the partial field from all the cached results")
         db = get_db(CACHE_MONGOENGINE_ALIAS)
         db[CACHE_COLLECTION_RESPONSES].update_many(
-            {"kind": "config-parquet-and-info", "http_status": 200}, {"$unset": {"content.partial": ""}}
+            {
+                "kind": {
+                    "$in": [
+                        "config-parquet-and-info",
+                        "config-parquet",
+                        "dataset-parquet",
+                        "config-parquet-metadata",
+                        "config-info",
+                        "dataset-info",
+                    ]
+                },
+                "http_status": 200,
+            },
+            {"$unset": {"content.partial": ""}},
         )
 
     def validate(self) -> None:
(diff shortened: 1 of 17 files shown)


lhoestq:
I think I'm missing the "size" jobq - will add them before merging

severo:
- "dataset-config-names": not triggered
- "config-split-names-from-streaming": not triggered
- "split-first-rows-from-streaming": triggered, but only to get the list of splits: not affected by the 5GB limit
- "config-parquet-and-info": the step itself
- "config-parquet": updated
- "config-parquet-metadata": updated
- "split-first-rows-from-parquet": triggered, but as we only get the first 100 rows, we can assume that the partial result always contains at least 100 rows.
- "dataset-parquet": updated
- "config-info": updated
- "dataset-info": updated
- "config-split-names-from-info": triggered, but not affected by the 5GB limit
- "config-size": MISSING
- "dataset-size": MISSING
- "dataset-split-names": triggered, but not affected by the 5GB limit
- "dataset-is-valid": MISSING?
- "split-image-url-columns": MISSING?
- "split-opt-in-out-urls-scan": MISSING?
- "split-opt-in-out-urls-count": MISSING?
- "config-opt-in-out-urls-count": MISSING?
- "dataset-opt-in-out-urls-count": MISSING?
- "split-duckdb-index": triggered, but not affected by the 5GB limit, since the duckdb limit is below (100MB)

lhoestq:
Just added config-size and dataset-size.

Not sure about dataset-is-valid - idk exactly when it's used (for the front end maybe ? in that case we should add it).

Finally it doesn't matter for the URLs counts imo

severo:
> Not sure about dataset-is-valid - idk exactly when it's used (for the front end maybe ? in that case we should add it).

it's only used in https://github.com/huggingface/model-evaluator, but I don't think they need this detail.

The related thing would be to change /valid, in order to give the list of "partial" datasets, if we want to show this in the list of datasets on the Hub. Not prioritary in any case.

> Finally it doesn't matter for the URLs counts IMO

Yes, I agree. We already rely on a partial scan of the URLs

services/worker/src/worker/job_runners/dataset/info.py
```diff
@@ -74,13 +75,14 @@ def compute_dataset_info_response(dataset: str) -> Tuple[DatasetInfoResponse, fl
                 )
                 continue
             config_infos[config] = config_response["content"]["dataset_info"]
+            partial = partial or partial
```
severo, line 78:
```suggestion
            partial = partial or config_response["content"]["partial"]
```

services/worker/src/worker/job_runners/dataset/parquet.py
```diff
@@ -81,19 +82,18 @@ def compute_parquet_response(dataset: str) -> Tuple[DatasetParquetResponse, floa
                     )
                 )
                 continue
-            config_parquet_content = ConfigParquetResponse(parquet_files=response["content"]["parquet_files"])
+            config_parquet_content = ConfigParquetResponse(
+                parquet_files=response["content"]["parquet_files"], partial=False
```
severo, line 86: hmmm isn't it

```suggestion
                parquet_files=response["content"]["parquet_files"], partial=response["content"]["partial"]
```

?

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230703110100_cache_add_partial_field_in_config_parquet_and_info.py
severo, line 1: are you reusing a previous migration file? I think you have to create a new one instead.

Also: you have to add it in collector.py

jobs/mongodb_migration/tests/migrations/test_20230703110100_cache_add_partial_field_in_config_parquet_and_info.py
severo, line 1: same comment: we should create another test file

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230703110100_cache_add_partial_field_in_config_parquet_and_info.py
lhoestq: yea but it hasn't been run, so it should be fine no ?

severo: aaaaah ok :) Sorry!

severo (approved):

PR #1474, opened 2023-07-03
split-duckdb-index config

AndreaFrancis:
Moving config to worker folder since it is not used from any other project.
Also adding doc in the readme file.


severo (approved):
Thanks!

PR #1480, opened 2023-07-04
Minor fix in update_last_modified_date_of_rows_in_assets_dir

lhoestq:
FileNotFoundError can happen because of concurrent api calls, and we can ignore it

(found this error while checking some logs today)


libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
 
+import logging
```
lhoestq, line 4: github is wrong, this line doesn't exist

severo (approved):
Well spotted

Issue #688, opened 2023-01-20
Add pdf2image as preinstalled package


mariosasko:
Needed for: https://huggingface.co/datasets/jordyvl/unit-test_PDFfolder (the installation instructions for pdf2image can be found there)

severo:
Partly fixed with #691: the pdf2image dependency has been added, and we get a result: https://huggingface.co/datasets/jordyvl/unit-test_PDFfolder

But it shows a new issue (unrelated): we should be showing a gallery of images, and instead, we have a JSON. I'm working on it.

<img width="811" alt="Capture d’écran 2023-01-23 à 11 31 49" src="https://user-images.githubusercontent.com/1676121/214018544-86861fb6-756c-4b6c-9982-d0714a362f28.png">

severo:
Fixed!

https://huggingface.co/datasets/jordyvl/unit-test_PDFfolder

<img width="1142" alt="Capture d’écran 2023-01-23 à 22 28 11" src="https://user-images.githubusercontent.com/1676121/214153870-04bd1a5a-1fa6-4d4a-a5ae-b64098832508.png">

PR #1481, opened 2023-07-04
chore: update pypdf dependency in worker

AndreaFrancis:
Should fix https://github.com/huggingface/datasets-server/security/dependabot/200
But not sure where exactly we use this library, according to pypi page pypdf2 is now pypdf (Not sure if this will break something):
`NOTE: The PyPDF2 project is going back to its roots. PyPDF2==3.0.X will be the last version of PyPDF2. Development will continue with [pypdf==3.1.0](https://pypi.org/project/pyPdf/)`


severo:
> not sure where exactly we use this

I installed it because some dataset script needed it.

severo:
Here is the reference: https://github.com/huggingface/datasets-server/issues/688. So, after merging, let's refresh https://huggingface.co/datasets/jordyvl/unit-test_PDFfolder and ensure it still works

severo (approved):

AndreaFrancis:
I see they use exactly `pypdf2` version and not `pypdf` https://huggingface.co/datasets/jordyvl/unit-test_PDFfolder/blob/main/unit-test_PDFfolder.py#L5 maybe this will not work, at least a commit is sent to their repo using `pypdf` instead right?

severo:
I don't know. Maybe other @huggingface/datasets maintainers could chime in

mariosasko:
The dataset in question uses `pdf2image` (does not depend on `pypdf`/`pypdf2`) for PDF decoding, so I think we can drop this dependency :).

severo (approved):

PR #1484, opened 2023-07-04
feat: 🎸 increase RAM for /rows service

severo:
Is it the right value @lhoestq ?


lhoestq (approved):
Yup ! Though idk what kind of machines we have but this should prevent most crashes

PR #1489, opened 2023-07-05
feat: 🎸 add "difficulty" field to JobDocument

severo:
Difficulty is an integer between 0 (easy) and 100 (hard). It aims at filtering the jobs in a specific worker deployment, ie, light workers will only run jobs with difficulty <= 40.

It should make the query to MongoDB quicker than currently (a filter `type: {$in: ALLOW_LIST}`). See #1486.

For now, all the jobs for a specific step have the same difficulty, ie: 20 for dataset-size, since it's only JSON manipulation, while split-duckdb-index is 70 because it can use a lot of RAM and take a lot of time.

Maybe, one day, we could estimate the difficulty of the job, based on previous stats for the same dataset, for example.


libs/libcommon/src/libcommon/queue.py
```diff
@@ -168,6 +177,7 @@ class JobDocument(Document):
     namespace = StringField(required=True)
     priority = EnumField(Priority, default=Priority.LOW)
     status = EnumField(Status, default=Status.WAITING)
+    difficulty = IntField(required=True)
```
AndreaFrancis, line 180: Maybe as default, it could be 50?

libs/libcommon/tests/test_processing_graph.py
```diff
@@ -12,6 +12,8 @@
     ProcessingStep,
 )
 
+DIFFICULTY = 50
```
AndreaFrancis, line 15: Is this used somewhere?

severo:
For reviewers: the most important thing are the weights in https://github.com/huggingface/datasets-server/pull/1489/files#diff-f61c85fc46851b1efc82330b783439bdff6f44a21dd6560abb735ea696161e32

libs/libcommon/src/libcommon/queue.py
severo: Hmmm, I wanted to force the field to be set when creating a job. What do you think?

libs/libcommon/tests/test_processing_graph.py
severo: Seems like it's not used, indeed

severo: done

libs/libcommon/src/libcommon/config.py
```diff
@@ -208,73 +208,86 @@ class ProcessingGraphConfig:
...
                 "job_runner_version": PROCESSING_STEP_CONFIG_PARQUET_VERSION,
                 "provides_config_parquet": True,
+                "difficulty": 20,
             },
             "config-parquet-metadata": {
                 "input_type": "config",
                 "triggered_by": "config-parquet",
                 "job_runner_version": PROCESSING_STEP_CONFIG_PARQUET_METADATA_VERSION,
                 "provides_config_parquet_metadata": True,
+                "difficulty": 40,
```
lhoestq, line 245:
```suggestion
                "difficulty": 50,
```

libs/libcommon/src/libcommon/queue.py
AndreaFrancis: Sure, I was thinking to set a default to avoid doing many modifications in tests for example but it is ok for me.

libs/libcommon/tests/test_queue.py
```diff
@@ -226,11 +304,54 @@ def test_job_types_only(
...
+    )
+    if should_raise:
+        with pytest.raises(EmptyQueueError):
+            queue.start_job(difficulty_max=difficulty_max, difficulty_min=difficulty_min)
+    else:
+        job_info = queue.start_job(difficulty_max=difficulty_max, difficulty_min=difficulty_min)
+        assert job_info["params"]["dataset"] == test_dataset
+
+
+# TODO: add min max difficulty test
```
lhoestq, line 346:
```suggestion
```

lhoestq (approved):

PR #1498, opened 2023-07-06
Delete `/config-names` endpoint

polinaeterna:
Part of https://github.com/huggingface/datasets-server/issues/1086


polinaeterna:
Should we actually delete this endpoint? It's not documented and as far as I know, it's not used anywhere and it overlaps with `/splits` (`/splits` just give more information). But when `/splits` without `config` parameter is requested, a response contains `pending` and `failed` fields which means that if it is the only way to get the list of configs, these fields should be documented but we haven't agreed on what to do with them. wdyt @huggingface/datasets-server ?

severo:
I think it's OK to remove it for now because:
- nobody asked for it, or is using it
- it will require less maintenance

It would easy to add afterwards if needed

severo (approved):

services/worker/src/worker/config.py
```diff
@@ -18,7 +18,6 @@
 WORKER_CONTENT_MAX_BYTES = 10_000_000
 WORKER_DIFFICULTY_MAX = None
 WORKER_DIFFICULTY_MIN = None
-WORKER_ENDPOINT = "/config-names"
```
severo, line 21: well spotted

AndreaFrancis (approved):

Issue #1502, opened 2023-07-06
Parquet metadata OOM for tiiuae/falcon-refinedweb


lhoestq:
Logs show nothing but the kube YAML of the pod shows OOMKilled.

The job should be improved to be more memory efficient.

This issue prevents us from having pagination for this dataset.

PR #1503, opened 2023-07-06
Remove parquet index without metadata

lhoestq:
The parquet index without metadata is too slow and causes the majority of the OOMs on /rows atm.

I think it's best to remove it completely.

For now it would cause some datasets like [tiiuae/falcon-refinedweb](https://huggingface.co/datasets/tiiuae/falcon-refinedweb) and [Open-Orca/OpenOrca](https://huggingface.co/datasets/Open-Orca/OpenOrca) to have a pagination that returns an error.

PR https://github.com/huggingface/datasets-server/pull/1497 should fix it for [Open-Orca/OpenOrca](https://huggingface.co/datasets/Open-Orca/OpenOrca), and I opened and issue https://github.com/huggingface/datasets-server/issues/1502 for [tiiuae/falcon-refinedweb](https://huggingface.co/datasets/tiiuae/falcon-refinedweb)

Ultimately we will need a way for the Hub to know if pagination is available (i.e. if parquet-metadata are available) instead of relying on the /size endpoint (or whatever it's using right now)


severo:
> Ultimately we will need a way for the Hub to know if pagination is available (i.e. if parquet-metadata are available) instead of relying on the /size endpoint (or whatever it's using right now)

Feel free to open an internal issue on https://github.com/huggingface/moon-landing

AndreaFrancis (approved):

lhoestq:
I think we can do a new endpoint that combines size and readiness of the parquet metadata ?
This way it will be easy on the moon-landing side

severo:
Yes, good idea. Created https://github.com/huggingface/datasets-server/issues/1521

PR #1507, opened 2023-07-10
Reduce rows lru cache

lhoestq:
It was causing the workers memory to keep increasing and finally OOM.

If there are still memory errors after that I might remove the LRU cache altogether


severo:
> If there are still memory errors after that I might remove the LRU cache altogether

What did you observe?

lhoestq:
It's now perfectly stable so far :)

Except one unrelated OOM due to #1519

PR #1508, opened 2023-07-10
Add fully converted datasets

lhoestq:
To fully convert https://huggingface.co/datasets/Open-Orca/OpenOrca to parquet (top 1 trending dataset right now)


AndreaFrancis (approved):

julien-c:
`fully converted datasets` is datasets that were already pushed in Parquet format?

lhoestq:
Those in parquet formats don't need to be converted*, they're just copied ;)

This PR is for datasets that need to be converted to parquet. Right now it converts up to 5GB but for some datasets it's better if there's no limit.

*certain parquet datasets are actually converted anyway if they have too big row groups, but I've only seen one dataset like that so far

services/worker/src/worker/config.py
```diff
@@ -186,6 +187,7 @@ class ParquetAndInfoConfig:
     blocked_datasets: List[str] = field(default_factory=get_empty_str_list)
     commit_message: str = PARQUET_AND_INFO_COMMIT_MESSAGE
     committer_hf_token: Optional[str] = PARQUET_AND_INFO_COMMITTER_HF_TOKEN
+    fully_converted_datasets: List[str] = field(default_factory=PARQUET_AND_INFO_FULLY_CONVERTED_DATASETS.copy)
```
severo, line 190: it's clever and better than what I did in other parts (`get_empty_str_list`) because the fact that the list is empty or not is specified in the constant. I think we might want to refactor this

severo: created https://github.com/huggingface/datasets-server/issues/1522

lhoestq:
It was just a little bit above 5GB ;)

PR #1509, opened 2023-07-10
Try to improve index usage for Job collection

AndreaFrancis:
Removing some indexes already covered for other existing ones to speed query plan decision

Index, | Proposal | Index Alternative  
-- | -- | --
"dataset", | DELETE: No query using only dataset |  
("dataset", "revision", "status"), | DELETE | ("type", "dataset", "revision", "config", "split", "status", "priority"),
("type", "dataset", "status"), | KEEP |  
("type", "dataset", "revision", "config", "split", "status", "priority"), | KEEP |  
("priority", "status", "created_at", "type", "namespace"), | DELETE | ("priority", "status", "created_at", "difficulty", "namespace", "type", "unicity_id"),
("priority", "status", "created_at", "difficulty", "namespace", "type", "unicity_id"), | KEEP |  
("priority", "status", "namespace", "type", "created_at"), | DELETE | ("priority", "status", "created_at", "difficulty", "namespace", "type", "unicity_id"),
("priority", "status", "created_at", "namespace", "-difficulty"), | DELETE | ("priority", "status", "created_at", "difficulty", "namespace", "type", "unicity_id"),
("status", "type"), | KEEP |  
("status", "namespace", "priority", "type", "created_at"), | DELETE | ("priority", "status", "created_at", "difficulty", "namespace", "type", "unicity_id"),
("status", "namespace", "unicity_id", "priority", "type", "created_at"), | DELETE | ("priority", "status", "created_at", "difficulty", "namespace", "type", "unicity_id"),
"-created_at", | DELETE: No query using only created_at |  
"finished_at" | KEEP |  
("unicity", "status", "created_at"), | KEEP |


libs/libcommon/src/libcommon/queue.py
```diff
@@ -854,15 +847,15 @@ def get_pending_jobs_df(self, dataset: str, job_types: Optional[List[str]] = Non
         return self._get_df(
             [
                 job.flat_info()
-                for job in JobDocument.objects(dataset=dataset, status__in=[Status.WAITING, Status.STARTED], **filters)
+                for job in JobDocument.objects(status__in=[Status.WAITING, Status.STARTED], **filters, dataset=dataset)
```
lhoestq, line 850: does the order matter here ?

lhoestq (approved):

libs/libcommon/src/libcommon/queue.py
AndreaFrancis: I don't think it will improve performance, but I changed it because I saw that we are using the same index but in a different order in other parts of the code. But, would like to try (In other DB engines, the column order of the filter/select affects the best query plan choice).

PR #1510, opened 2023-07-10
Add priority param to force refresh endpoint

lhoestq:
Right now I always have to manually set the priority in mongo to "normal"


AndreaFrancis (approved):
Thanks! I also needed it

PR #1516, opened 2023-07-12
feat: /search endpoint

AndreaFrancis:
Second part of FTS implementation using duckdb for https://github.com/huggingface/datasets-server/issues/629

This PR introduces a new endpoint `/search` in a new project **search** service with the following parameters:
- dataset
- config
- split
- query
- offset (by default 0)
- length (by default 100)

The process of performing a search consists in:
1. Validate parameters
2. Validate authentication
3. Validate split-duckdb-index cache result in order to verify if indexing has been previously done correctly
4. Download index file from /refs/convert/parquet revision in case of missing otherwise, use it directly
5. Perform full text search using duckdb
6. Return response with similar format as /rows

```
    features: List[FeatureItem]
    rows: Any
    num_total_rows: int --> I added this new field because I think it will be used in the UI for pagination

```

Note.- In another PR, I will add a k8s Job that will periodically delete the downloaded files based on their last accessed date -> https://github.com/huggingface/datasets-server/pull/1536


AndreaFrancis:
@huggingface/datasets-server This PR is kind of ready for review, I just added a couple of test scenarios but I can add more if needed. I would like to have a first review from the team to validate the idea and then improve details.

libs/libcommon/src/libcommon/utils.py
```diff
@@ -91,6 +91,12 @@ class FeatureItem(TypedDict):
     type: Row
 
 
+class PaginatedResponse(TypedDict):
+    features: List[FeatureItem]
+    rows: Any
+    num_total_items: int
```
severo, line 97: to keep it coherent with what we have on DatasetSize, I would rename it to

```suggestion
    num_total_rows: int
```

chart/templates/services/api/deployment.yaml
severo, line 1: Is it OK to include /search in the API, instead of creating a new service for that? I'm afraid it could potentially affect the other endpoints, as the responses are computed inline instead of just returning a cached entry.

AndreaFrancis: Makes more sense to move to its separate service, I will work on that

AndreaFrancis:
Code is ready but not sure why CI is failing because of python version for /rows quality and mongodb_migration unit tests, I will check.

chart/templates/services/api/deployment.yaml
AndreaFrancis: I moved all the code to a separate project search.

severo:
The e2e tests error seems to be a bug in the code https://github.com/huggingface/datasets-server/actions/runs/5661637294/job/15339958238?pr=1516

e2e/tests/test_50_search_healthcheck.py
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2023 The HuggingFace Authors.
+
+from .utils import SEARCH_UVICORN_PORT, poll
+
+
+def test_healthcheck() -> None:
+    # this tests ensures the /healthcheck and the /metrics endpoints are hidden
+    response = poll("/healthcheck", expected_code=200, url=SEARCH_UVICORN_PORT)
```
severo, line 9:
```suggestion
from .utils import SEARCH_URL, poll


def test_healthcheck() -> None:
    # this tests ensures the /healthcheck and the /metrics endpoints are hidden
    response = poll("/healthcheck", expected_code=200, url=SEARCH_URL)
```

severo (commented):
for the e2e error

severo (commented):
.github/workflows/cd.yml and the docker compose files must be updated

services/search/src/search/routes/search.py
```diff
@@ -0,0 +1,300 @@
...
+                        hf_jwt_public_key=hf_jwt_public_key,
+                        hf_jwt_algorithm=hf_jwt_algorithm,
+                        hf_timeout_seconds=hf_timeout_seconds,
+                    )
+
+                logging.info(f"/search {dataset=} {config=} {split=} {query=} {offset=} {length=}")
+
+                with StepProfiler(method="search_endpoint", step="validate indexing was done"):
+                    # no cache data is needed to download the index file
+                    # but will help to validate if indexing was done
```
severo, line 241: why not use the content of the cached response to get the URL, instead of forging it? It seems less brittle to me

services/search/src/search/routes/search.py
```diff
@@ -0,0 +1,300 @@
...
+                            error_code=error_code,
+                            revision=revision,
+                        )
+
+                with StepProfiler(method="search_endpoint", step="download index file if missing"):
+                    index_folder = get_index_folder(duckdb_index_file_directory, dataset, config, split)
+                    repo_file_location = f"{config}/{split}/{DUCKDB_DEFAULT_INDEX_FILENAME}"
+                    index_file_location = f"{index_folder}/{repo_file_location}"
+                    index_path = Path(index_file_location)
+                    if not index_path.is_file():
```
severo, line 273: and what if the dataset has been updated but the old index is on the disk? we will search on an old version

severo: also: we don't delete the index files? It will be done by another job, right?

services/search/tests/test_app.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def test_get_search(client: TestClient) -> None:
+    # missing parameter
+    response = client.get("/search")
+    assert response.status_code == 422
+
+
+@pytest.mark.parametrize(
+    "dataset,config,split",
```
severo, line 59: also test missing query?

chart/values.yaml
```diff
@@ -466,6 +471,44 @@ rows:
     annotations: {}
   tolerations: []
 
+search:
+  # the path of the external authentication service on the hub.
+  # The string must contain `%s` which will be replaced with the dataset name.
+  hfAuthPath: "/api/datasets/%s/auth-check"
```
severo, line 477: I think we're repeating this a lot (api, rows, search). Could we factorize?

chart/values.yaml
```diff
@@ -466,6 +471,44 @@ rows:
...
+  maxAgeShort: "10"
+  # Directory where the uvicorn workers will write the prometheus metrics
+  # see https://github.com/prometheus/client_python#multiprocess-mode-eg-gunicorn
+  prometheusMultiprocDirectory: "/tmp"
+  # Hostname - it must not be set to localhost to work in Kube!
+  uvicornHostname: "0.0.0.0"
+  # Number of uvicorn workers for running the application
+  uvicornNumWorkers: "1"
+  # Application endpoint port
+  uvicornPort: 8082
```
severo, line 498: I think we could set it to 8080 (and in rows too) as it's on the pod

chart/env/prod.yaml
```diff
@@ -300,6 +300,26 @@ rows:
       cpu: 4
       memory: "8Gi"
 
+search:
+  # the timeout in seconds for the requests to the Hugging Face Hub.
+  hfTimeoutSeconds: "1.5"
+  # Number of uvicorn workers for running the application
+  # (2 x $num_cores) + 1
+  # https://docs.gunicorn.org/en/stable/design.html#how-many-workers
+  uvicornNumWorkers: "3"
```
severo, line 309: why not 9 as the other services?

chart/env/prod.yaml
```diff
@@ -300,6 +300,26 @@ rows:
...
+  nodeSelector:
+    role-datasets-server-search: "true"
+  replicas: 4
```
severo, line 312: let's increase a bit?

services/search/src/search/routes/search.py
```diff
@@ -0,0 +1,300 @@
...
+                    config = request.query_params.get("config")
+                    split = request.query_params.get("split")
+                    query = request.query_params.get("query")
+
+                    if (
+                        not dataset
+                        or not config
+                        or not split
+                        or not query
+                        or not are_valid_parameters([dataset, config, split, query])
```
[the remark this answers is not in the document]
AndreaFrancis: I added those validations again, otherwise mypy fails for optional values being used later.

services/search/src/search/routes/search.py
```diff
@@ -0,0 +1,300 @@
...
+    query_result = con.execute(
+        query=FTS_COMMAND.format(offset=offset, length=length),
+        parameters=[query],
+    )
+    pa_table = query_result.arrow()
+    con.close()
+    return (num_total_rows, pa_table)
+
+
+# TODO: Copied from api.routes.endpoint, refactor
```
lhoestq, line 108:
```suggestion
```

lhoestq (commented):
Awesome !!

Could you also update the docker compose files in tools/ ? I can also take care of it once this is merged if you want

Also I think that you can run `update_last_modified_date_of_rows_in_assets_dir` as in /rows. It sets the last modified time of the assets directories to be the last access time, which is then used when `clean_cached_assets` is called.

services/search/src/search/routes/search.py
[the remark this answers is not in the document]
severo: oh right!!!

services/search/src/search/routes/search.py
AndreaFrancis: Temporal index files will be deleted here https://github.com/huggingface/datasets-server/pull/1536 but as you said, this file also should be removed when there is a new version, maybe I can add that here? https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/duckdb_index.py#L73

chart/values.yaml
AndreaFrancis: I moved all hfX configurations to another section `hf`

AndreaFrancis:
> Could you also update the docker compose files in tools/ ? I can also take care of it once this is merged if you want


Sure, I updated docker compose files for dev and prod.


> Also I think that you can run update_last_modified_date_of_rows_in_assets_dir as in /rows. It sets the last modified time of the assets directories to be the last access time, which is then used when clean_cached_assets is called.

I added clean_cached_assets to be called on /search randomly, same as in /rows. As per our conversation, update_last_modified_date_of_rows_in_assets_dir might not be needed for now. Note also that I saw that images and audio types are being stored as binary in Duckb so, maybe no conversion will be done. I will create a separate issue to handle/enable image and audio support for search.

services/search/src/search/routes/search.py
AndreaFrancis: Not sure if I can use an URL for hf_hub_download. Doc https://huggingface.co/docs/huggingface_hub/v0.16.3/en/package_reference/file_download#huggingface_hub.hf_hub_download says that basically it is just needed  repo_type, revision, repo_id, filename, local_dir, token. Anyway, I removed DEFAULT_INDEX_NAME in favor of taking the filename from cache response and use it when calling download function.

services/search/src/search/routes/search.py
severo: we should have a way to know if the current file is the correct version. Maybe use the revision in the path?

severo:
I was wondering: what is the value of `rowIdx`? I think we want it to be the row index for the unfiltered dataset, right? ie a persistent number that identifies the row.
Is it how it's done?

AndreaFrancis:
> I was wondering: what is the value of rowIdx? I think we want it to be the row index for the unfiltered dataset, right? ie a persistent number that identifies the row.
> Is it how it's done?

No, it currently is the row index for the filtered result, I can change it to be equal to __hf_index_id so that we persist the row id.

severo:
> No, it currently is the row index for the filtered result, I can change it to be equal to __hf_index_id so that we persist the row id.

wdyt? it could be more useful to have a way to reference the original row id. Having the number in the list is easy to compute if needed.

services/search/src/search/routes/search.py
severo: hmmm ok

services/search/src/search/routes/search.py
severo: OK, you take the revision in the name of the directory, so, we're good.

severo (commented):
we need an e2e test such as def test_rows_endpoint (in test_11_api.py -> feel free to use another file if you want to separate from the API)

services/search/src/search/routes/search.py
```diff
@@ -0,0 +1,337 @@
...
+ROW_IDX_COLUMN = "__hf_index_id_row_idx"
+MAX_ROWS = 100
+UNSUPPORTED_FEATURES_MAGIC_STRINGS = ["'binary'", "Audio("]
+FTS_COMMAND_COUNT = (
+    "SELECT COUNT(*) FROM (SELECT __hf_index_id, fts_main_data.match_bm25(__hf_index_id, ?) AS score FROM"
+    " data) A WHERE score IS NOT NULL;"
+)
+
+FTS_COMMAND = (
+    "SELECT * EXCLUDE (__hf_index_id, score) FROM (SELECT __hf_index_id - 1 as __hf_index_id_row_idx, *,"
```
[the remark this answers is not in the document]
AndreaFrancis: Update: serial min value will be fixed for worker in this PR https://github.com/huggingface/datasets-server/pull/1594

libs/libapi/src/libapi/utils.py
```diff
@@ -155,16 +160,80 @@ def to_rows_list(
...
+            row_idx_column=row_idx_column,
         )
     except Exception as err:
         raise TransformRowsProcessingError(
             "Server error while post-processing the split rows. Please report the issue."
         ) from err
     return [
         {
-            "row_idx": idx + offset,
+            "row_idx": idx + offset if row_idx_column is None else row.pop(row_idx_column),  # type: ignore
```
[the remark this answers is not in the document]
AndreaFrancis: Actually, `/rows` doesn't need `row_idx_column` (it depends on the pagination offset and iteration).
I put `# type: ignore` because of mypy warning about `Mapping` without `pop` method, not sure if I could cast it to dict, I would like to perform pop in order to take advantage and remove the column to use row without it in the next line.

libs/libcommon/src/libcommon/rows_utils.py
```diff
@@ -19,14 +19,15 @@ def transform_rows(
...
+    row_idx_column: Optional[str],
 ) -> List[Row]:
     return [
         {
             featureName: get_cell_value(
                 dataset=dataset,
                 config=config,
                 split=split,
-                row_idx=offset + row_idx,
+                row_idx=offset + row_idx if row_idx_column is None else row[row_idx_column],
```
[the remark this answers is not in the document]
AndreaFrancis: `row[row_idx_column]` will be used for `/search` (because it cant depend on offset to get the id column, but it depends on `__hf_index_id`.
`offset + row_idx` is used by `/rows` (since it uses offset and will always have the same order for all rows)

services/search/src/search/routes/search.py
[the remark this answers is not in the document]
severo: > Change split-duckdb-index worker to start serial from 0 and recompute all dataset indexes

is better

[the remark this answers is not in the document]
AndreaFrancis: Fixed in https://github.com/huggingface/datasets-server/pull/1516/commits/694b00e4bb5b4c3cee6d8f5706eb148d0e77f7f7

libs/libcommon/src/libcommon/rows_utils.py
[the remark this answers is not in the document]
severo: OK, perfect

libs/libapi/src/libapi/utils.py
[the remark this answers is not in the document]
severo: OK! maybe you could change the type of Row (Dict instead of Mapping): https://github.com/huggingface/datasets-server/blob/694b00e4bb5b4c3cee6d8f5706eb148d0e77f7f7/libs/libcommon/src/libcommon/utils.py#L85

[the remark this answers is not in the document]
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1516/commits/d4babea14050b03679037cb9ae35bb7e602dc2e8

Issue #1519, opened 2023-07-17
Dataset with too big row groups makes the viewer crash


lhoestq:
eg IDEA-CCNL/laion2B-multi-chinese-subset

```python
In [1]: import fsspec; import pyarrow.parquet as pq

In [2]: url = "https://huggingface.co/datasets/IDEA-CCNL/laion2B-multi-chinese-subset/resolve/main/data/train-00000-of-00013.parquet"

In [3]: pf = pq.ParquetFile(fsspec.open(url).open())

In [4]: pf.metadata
Out[4]: 
<pyarrow._parquet.FileMetaData object at 0x11f419450>
  created_by: parquet-cpp-arrow version 7.0.0
  num_columns: 10
  num_rows: 11177146
  num_row_groups: 1
  format_version: 1.0
  serialized_size: 5880

In [5]: pf.metadata.row_group(0).total_byte_size
Out[5]: 2085382973

In [6]: pf.metadata.row_group(0)
Out[6]: 
<pyarrow._parquet.RowGroupMetaData object at 0x106609590>
  num_columns: 10
  num_rows: 11177146
  total_byte_size: 2085382973
```

lhoestq:
Ah apparently it was converted to parquet before #1497 

I just restarted the config-parquet-and-info job

lhoestq:
Actually it copied the file anyway because the check on row groups didn't run, let me fix this

severo:
Is it the case for https://datasets-server.huggingface.co/rows?dataset=bigcode/the-stack-metadata&config=bigcode--the-stack-metadata&split=train&offset=500&length=100? Or is it another issue? The request takes too much time and the reverse-proxy emits a 504 Gateway Time-out error

severo:
OK! In this case it's bigcode/the-stack-metadata, not garythung/trashnet, but I imagine it's the same. For reference, https://datasets-server.huggingface.co/rows?dataset=bigcode/the-stack-metadata&config=bigcode--the-stack-metadata&split=train&offset=500&length=100 currently takes about 9s to respond

lhoestq:
yes bigcode/the-stack-metadata should be fixed with https://github.com/huggingface/datasets-server/pull/1520 : it has too big row groups

lhoestq:
ah actually I can just refresh it I think

PR #1520, opened 2023-07-17
Always verify parquet before copying

lhoestq:
close https://github.com/huggingface/datasets-server/issues/1519


lhoestq:
https://github.com/huggingface/datasets-server/pull/1525 should fix the CI (and maybe there's a token to be refreshed)

severo (approved):

severo:
which datasets should we recompute after that? Is there a way to know?

lhoestq:
I don't think there's a way to know except with /rows that can detect too big row groups, maybe we can refresh the dataset when /rows detects it

severo:
> maybe we can refresh the dataset when /rows detects it
Could be, but instead of complicating the code for that, maybe do a one-time analysis on all the datasets that have a size over ...GB

Issue #1483, opened 2023-07-04
C4 pagination failing


lhoestq:
current error is an ApiError

```
config size could not be parsed: ValidationError: "size.config.num_bytes_original_files" must be a number
```

severo:
looking at it

severo:
I created https://github.com/huggingface/datasets-server/issues/1528 for that matter

PR #1533, opened 2023-07-20
Separate parquet metadata by split

lhoestq:
Since we added partial conversion to parquet, we introduced the new config/split/ssss.parquet paths but the parquet metadata worker was nos following it and therefore splits could overwrite each other

This affects any dataset with partial conversion and multiple splits, e.g. c4

Related tohttps://github.com/huggingface/datasets-server/issues/1483


AndreaFrancis (approved):
Do we need to increase the job runner version for backfill? Or is it compatible with the old cache entries without split?

lhoestq:
It's still compatible, I'll just manually refresh datasets with this issue

PR #1536, opened 2023-07-20
K8s job to periodically remove indexes

AndreaFrancis:
Should help delete files downloaded on https://github.com/huggingface/datasets-server/pull/1516

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -299,6 +299,23 @@ backfill:
   tolerations: []
 
 
+deleteIndexes:
+  enabled: false
+  log:
+    level: "info"
+  action: "delete-indexes"
+  error_codes_to_retry: ""
+  schedule: "*/10 * * * *"
+  # every ten minutes
+  nodeSelector: {}
+  resources:
+    requests:
+      cpu: 0
+    limits:
+      cpu: 0
+  tolerations: []
+
+
 metricsCollector:
   action: "collect-metrics"
   schedule: "*/5 * * * *"
--- a/jobs/cache_maintenance/src/cache_maintenance/delete_indexes.py
+++ b/jobs/cache_maintenance/src/cache_maintenance/delete_indexes.py
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2023 The HuggingFace Authors.
+
+import glob
+import logging
+import os
+from datetime import datetime, timedelta
+
+from libcommon.storage import StrPath
+
+
+def delete_indexes(
+    duckdb_index_cache_directory: StrPath, subdirectory: str, file_extension: str, expired_time_interval_seconds: int
+) -> None:
+    logging.info("delete indexes")
+    indexes_folder = f"{duckdb_index_cache_directory}/{subdirectory}/**/*{file_extension}"
+    logging.info(f"looking for all files with pattern {indexes_folder}")
+    for path in glob.glob(indexes_folder, recursive=True):
+        last_access_time_value = os.path.getatime(path)
+        last_access_datetime = datetime.fromtimestamp(last_access_time_value).replace(tzinfo=None)
+        now = datetime.now().replace(tzinfo=None)
+        if last_access_datetime + timedelta(seconds=expired_time_interval_seconds) <= now:
+            logging.info(f"deleting {path=}")
+            os.remove(path)
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -92,6 +92,11 @@ app.kubernetes.io/component: "{{ include "name" . }}-metrics-collector"
 app.kubernetes.io/component: "{{ include "name" . }}-backfill"
 {{- end -}}
 
+{{- define "labels.deleteIndexes" -}}
+{{ include "hf.labels.commons" . }}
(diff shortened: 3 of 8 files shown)


jobs/cache_maintenance/src/cache_maintenance/delete_indexes.py
```diff
@@ -0,0 +1,24 @@
...
+def delete_indexes(
+    duckdb_index_cache_directory: StrPath, subdirectory: str, file_extension: str, expired_time_interval_seconds: int
+) -> None:
+    logging.info("delete indexes")
+    indexes_folder = f"{duckdb_index_cache_directory}/{subdirectory}/**/*{file_extension}"
+    logging.info(f"looking for all files with pattern {indexes_folder}")
+    for path in glob.glob(indexes_folder, recursive=True):
+        last_access_time_value = os.path.getatime(path)
+        last_access_datetime = datetime.fromtimestamp(last_access_time_value).replace(tzinfo=None)
+        now = datetime.now().replace(tzinfo=None)
```
severo, line 21: nit: maybe set this at the beginning of the function

jobs/cache_maintenance/src/cache_maintenance/delete_indexes.py
```diff
@@ -0,0 +1,24 @@
...
+import logging
+import os
+from datetime import datetime, timedelta
+
+from libcommon.storage import StrPath
+
+
+def delete_indexes(
+    duckdb_index_cache_directory: StrPath, subdirectory: str, file_extension: str, expired_time_interval_seconds: int
+) -> None:
```
severo, line 14: it deletes the files but not the directories, ie currently, we have on the disk: `downloads/asoria-copy_final-e388ab6d/asoria--copy_final/train/index.duckdb`. It's the only file in `downloads/asoria-copy_final-e388ab6d/`. Shouldn't we try to delete the directories too? On the other side: it might lead to concurrency issues (/search creates the file, while the job deletes the directories). Not sure how to manage this.

---

Not related to this particular PR, but we have a somewhat similar situation for the temporal `duckdb-index` directories: many empty directories. But in that case, they should have been used only once. So these should be deleted.

chart/values.yaml
```diff
@@ -261,6 +261,8 @@ duckDBIndex:
   urlTemplate: "/datasets/%s/resolve/%s/%s"
   # the maximum size of the split parquets.
   maxParquetSizeBytes: "100_000_000"
+  # the time interval at which a downloaded index will be considered as expired and will be deleted 
+  expiredTimeIntervalSeconds: 600
```
severo, line 265: On first thought, it could be a lot more, like 1 day, or 3 days. We don't really have limits on the storage, and it would reduce the latency for datasets that are searched often.

chart/values.yaml
```diff
@@ -321,6 +323,23 @@ backfill:
   tolerations: []
 
 
+deleteIndexes:
+  enabled: false
```
severo, line 327: you can already set it to true in prod and staging (maybe here too)

chart/values.yaml
```diff
@@ -321,6 +323,23 @@ backfill:
...
+  log:
+    level: "info"
+  action: "delete-indexes"
+  error_codes_to_retry: ""
+  schedule: "*/10 * * * *"
```
severo, line 332: also: it can be a lot less frequent, to reduce IO. Maybe once a day

jobs/cache_maintenance/src/cache_maintenance/delete_indexes.py
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1536/commits/f38ba31e3b7664ee3322416f71f07444cfddf906

chart/values.yaml
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1536/commits/f38ba31e3b7664ee3322416f71f07444cfddf906

chart/values.yaml
AndreaFrancis: I changed it to run once a day https://github.com/huggingface/datasets-server/pull/1536/commits/f38ba31e3b7664ee3322416f71f07444cfddf906

chart/values.yaml
AndreaFrancis: Sure, I changed it to 3 days for prod https://github.com/huggingface/datasets-server/pull/1536/commits/f38ba31e3b7664ee3322416f71f07444cfddf906

jobs/cache_maintenance/src/cache_maintenance/delete_indexes.py
AndreaFrancis: I think I could change the logic for index downloads, instead of downloading in separate folders dataset/config/split, maybe they can be a combination in the index name as `dataset-config-split-revision-index.duckdb`  so that, this job only has to remove files without dealing with empty directories.

severo: yes, it seems better!

severo: So: in another PR

severo (approved):

PR #1544, opened 2023-07-24
fix: resources allocation and use dedicated nodes for worker light

rtrompier:


severo (approved):
OK! No overcommitting for the RAM of big workers... It will multiply the number of nodes (and the price) by two...

PR #1548, opened 2023-07-24
fix(helm): update probes for services pods

rtrompier:


severo (approved):
Nice! more explicit.

cc @AndreaFrancis  for #1516 and @albertvillanova for #1418

PR #1554, opened 2023-07-25
Align poetry version in all Docker files

albertvillanova:
Currently, the Poetry version set in jobs/cache_maintenance Docker file is different from the one set in all the other Docker files.

This PR aligns the Poetry version in jobs/cache_maintenance Docker file with all the rest.


AndreaFrancis (approved):
Thanks!

PR #1558, opened 2023-07-25
feat: 🎸 request 4Gi per /rows pod

severo:
it will allocate less pods per node. We regularly have OOM errors, with pods that use 3, 5, 6, 7 GiB RAM, and the node has not enough RAM left for them.

Capture d’écran 2023-07-25 à 12 29 26


rtrompier (commented):
LGTM

Issue #1156, opened 2023-05-10
Fix flaky executor test in services/worker when the job takes too much time


severo:
See https://github.com/huggingface/datasets-server/pull/1147#issuecomment-1541483075

severo:
cc @lhoestq 😉

severo:
We need to fix the test

PR #1567, opened 2023-07-26
Fix flaky executor test on long jobs

lhoestq:
Sometimes the executor doesn't have a chance to kill the long job before finishing

close https://github.com/huggingface/datasets-server/issues/1156


severo (approved):
OMG thanks for that!

PR #1571, opened 2023-07-27
Fix torch on macos

lhoestq:
This fixes local deployment using docker compose on macos

fixes this error when building the `worker` docker image

```
> [stage-0 13/13] RUN --mount=type=cache,target=/home/.cache/pypoetry/cache     --mount=type=cache,target=/home/.cache/pypoetry/artifacts     poetry install --no-root:                                                                          
#17 0.533 Creating virtualenv worker in /src/services/worker/.venv                                                       
#17 0.852 Installing dependencies from lock file

...

#17 10.40   AssertionError
#17 10.40 
#17 10.40   
#17 10.40 
#17 10.40   at /usr/local/lib/python3.9/site-packages/poetry/installation/executor.py:742 in _download_link
#17 10.41       738│         # to the original archive.
#17 10.41       739│         archive = self._chef.get_cached_archive_for_link(link, strict=False)
#17 10.41       740│         # 'archive' can at this point never be None. Since we previously downloaded
#17 10.41       741│         # an archive, we now should have something cached that we can use here
#17 10.41     → 742│         assert archive is not None
#17 10.41       743│ 
#17 10.41       744│         if archive.suffix != ".whl":
#17 10.41       745│             message = (
#17 10.41       746│                 f"  • {self.get_operation_message(operation)}:"
```


severo (approved):
wow, it's ugly 😛

Issue #1421, opened 2023-06-22
/rows returns `null` images for some datasets


lhoestq:
Reported in https://github.com/huggingface/datasets/issues/2526

See https://huggingface.co/datasets/lombardata/panoptic_2023_06_22 has a working /first-rows but /rows always returns `null` for images.

edit: another one https://huggingface.co/datasets/jonathan-roberts1/RSSCN7

severo:
keep open

PR #1573, opened 2023-07-27
Use cached features in /rows

lhoestq:
/rows needs the cached `features` since they're not always available in the parquet metadata.
This was causing some `Image` columns to be seen as a struct of binary data, which are not supported in the viewer (shown as "null").

Therefore I'm now passing the `features` from `config-parquet-and-info` to `config-parquet` and then to `config-parquet-metadata`. I kept it backward compatible in case a cached value doesn't have this field yet.

Therefore there's no need for a mongo migration. We can just re-run all the `config-parquet` and `config-parquet-metadata` jobs. I incremented their versions.

close https://github.com/huggingface/datasets-server/issues/1421


services/worker/src/worker/job_runners/config/parquet_metadata.py
```diff
@@ -95,6 +95,11 @@ def compute_parquet_metadata_response(
         ]
         if not parquet_file_items:
             raise ParquetResponseEmptyError("No parquet files found.")
+        content = config_parquet_best_response.response["content"]
+        if "features" in content and isinstance(content["features"], dict):
+            features = content["features"]  # config-parquet version<6 didn't have features
+        else:
+            features = None
```
severo, line 102: same comment:

maybe add a note to replace this part (features=None) later with an error. I understand you put it for backward-compatibility.

services/worker/tests/job_runners/config/test_parquet.py
severo, line 1: maybe we could test the other branch of the if/else

services/rows/src/rows/routes/rows.py
```diff
@@ -48,7 +48,7 @@
 ALL_COLUMNS_SUPPORTED_DATASETS_ALLOW_LIST: Union[Literal["all"], List[str]] = ["arabic_speech_corpus"]  # for testing
 
 # audio still has some errors when librosa is imported
-UNSUPPORTED_FEATURES_MAGIC_STRINGS = ["'binary'", "Audio("]
+UNSUPPORTED_FEATURES = [Value("binary"), Audio()]
```
severo, line 51: good! it's much more robust

libs/libcommon/src/libcommon/parquet_utils.py
```diff
@@ -30,6 +31,18 @@ class FileSystemError(Exception):
     pass
 
 
+def _clean_mongo_objects(obj: Any) -> Any:
```
severo, line 34: maybe move this to the root function in simple_cache.py that gets the content out of the MongoDB?

libs/libcommon/src/libcommon/parquet_utils.py
```diff
@@ -43,17 +56,28 @@ class ParquetFileMetadataItem(TypedDict):
...
+        def classify(feature: FeatureType) -> None:
+            nonlocal supported
+            for unsupported_feature in unsupported_features:
+                if type(unsupported_feature) == type(feature) == Value:
+                    if unsupported_feature.dtype == feature.dtype:
+                        supported = False
+                elif type(unsupported_feature) == type(feature):
+                    supported = False
+
+        _visit(feature, classify)
```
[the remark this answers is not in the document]
severo: excellent!

severo (approved):

severo:
also: I think we have to update the openapi spec

services/worker/src/worker/job_runners/dataset/parquet.py
```diff
@@ -83,7 +83,9 @@ def compute_parquet_response(dataset: str) -> Tuple[DatasetParquetResponse, floa
                 )
                 continue
             config_parquet_content = ConfigParquetResponse(
-                parquet_files=response["content"]["parquet_files"], partial=response["content"]["partial"]
+                parquet_files=response["content"]["parquet_files"],
+                partial=response["content"]["partial"],
+                features=None,
```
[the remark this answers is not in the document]
lhoestq: `config_parquet_content` is instantiated for mypy typing I believe, but we don't return the features anyway in `DatasetParquetResponse` so no need to specify it.

I'll add a comment

Issue #1576, opened 2023-07-27
/rows should return numTotalRows as /search


severo:
Needed to implement the search on the Hub

`num_total_rows`

PR #1580, opened 2023-07-28
Add num_rows_total to /rows response

severo:


libs/libcommon/src/libcommon/parquet_utils.py
```diff
@@ -81,11 +81,16 @@ class ParquetIndexWithMetadata:
     httpfs: HTTPFileSystem
     hf_token: Optional[str]
 
+    num_total_bytes: int = field(init=False)
```
AndreaFrancis, line 84: When/where will we use this new attribute?

severo: it's not used, you're right. I was not sure if adding it, just in case we need it one day, or not. Better to remove it

severo: done

severo:
fixes #1576

PR #1595, opened 2023-08-01
Fix descriptive statistics env var

severo:


chart/templates/storage-admin/deployment.yaml
```diff
@@ -17,7 +17,17 @@ spec:
...
+        {{ include "initContainerDuckDBIndex" . | nindent 8 }}
+        {{ include "initContainerDescriptiveStatistics" . | nindent 8 }}
       containers: {{ include "containerStorageAdmin" . | nindent 8 }}
       nodeSelector: {{ toYaml .Values.storageAdmin.nodeSelector | nindent 8 }}
       tolerations: {{ toYaml .Values.storageAdmin.tolerations | nindent 8 }}
-      volumes: {{ include "volumeData" . | nindent 6 }}
+      volumes:
+        {{ include "volumeCache" . | nindent 8 }}
+        {{ include "volumeData" . | nindent 8 }}
+        {{ include "volumeDuckDB" . | nindent 8 }}
```
AndreaFrancis, line 32: Maybe we need `{{ include "volumeMountParquetMetadataRW" . | nindent 8 }}` also?

AndreaFrancis (approved):

chart/templates/storage-admin/deployment.yaml
severo: no!!!

severo: In fact, it's included in `{{ include "volumeData" . | nindent 8 }}` because we didn't move the data to its own EFS, so it's the same volume as assets

PR #1601, opened 2023-08-01
refactor: 💡 clean the chart variables

severo:
- ensure coherence in the names of the chart variables
- don't provide cached-assets to admin service (it does not use it)
- remove code related to cached assets in api service (it has moved to rows)

Note: I'm not sure we're allowed to put helper templates in subdirectories (_volumes/, etc). When deployed I'll try on staging, and move them back to the templates/ root directory if it failes.


severo:
the CI error is a vulnerability with cryptography. I'll fixed it in another PR

severo:
merging to test in staging (yes, it's not optimal)

PR #1605, opened 2023-08-02
feat: 🎸 optimize the computation of metrics

### LINKED ISSUE #1604: Optimize mongo query
The metrics about the cache entries is a list of tuples `(kind, http_status, error_code, count)`.

Currently we compute with:

https://github.com/huggingface/datasets-server/blob/deb708ae737a2f8da51b74c1ca4a489c4ff39b51/libs/libcommon/src/libcommon/simple_cache.py#L490-L506

These queries are very slow (eg each `entries(kind=kind, http_status=http_status).distinct("error_code")` command can take about 10s)

An alternative is to run an aggregation:

```python
def format_group(group: Dict[str, Any]) -> CountEntry:
    kind = group["kind"]
    if not isinstance(kind, str):
        raise TypeError("kind must be a str")
    http_status = group["http_status"]
    if not isinstance(http_status, int):
        raise TypeError("http_status must be an int")
    error_code = group["error_code"]
    if not isinstance(error_code, str) and error_code is not None:
        raise TypeError("error_code must be a str or None")
    count = group["count"]
    if not isinstance(count, int):
        raise TypeError("count must be an int")
    return {"kind": kind, "http_status": http_status, "error_code": error_code, "count": count}


def get_responses_count_by_kind_status_and_error_code() -> List[CountEntry]:
    groups = CachedResponseDocument.objects().aggregate(
        [
            {
                "$group": {
                    "_id": {"kind": "$kind", "http_status": "$http_status", "error_code": "$error_code"},
                    "count": {"$sum": 1},
                }
            },
            {
                "$project": {
                    "kind": "$_id.kind",
                    "http_status": "$_id.http_status",
                    "error_code": "$_id.error_code",
                    "count": "$count",
                }
            },
        ]
    )
    return [format_group(group) for group in groups]
```

but from my tests on the prod database, it does not compute quickly either (3 minutes).

Note that we have the `(kind, http_status, error_code)` index in the database:

Capture d’écran 2023-08-02 à 11 07 00

severo:
fixes #1604.

It seems like I had to first sort by the fields.

### DIFF
--- a/libs/libcommon/src/libcommon/simple_cache.py
+++ b/libs/libcommon/src/libcommon/simple_cache.py
@@ -487,23 +487,43 @@ class CountEntry(TypedDict):
     count: int
 
 
+def format_group(group: Dict[str, Any]) -> CountEntry:
+    kind = group["kind"]
+    if not isinstance(kind, str):
+        raise TypeError("kind must be a str")
+    http_status = group["http_status"]
+    if not isinstance(http_status, int):
+        raise TypeError("http_status must be an int")
+    error_code = group["error_code"]
+    if not isinstance(error_code, str) and error_code is not None:
+        raise TypeError("error_code must be a str or None")
+    count = group["count"]
+    if not isinstance(count, int):
+        raise TypeError("count must be an int")
+    return {"kind": kind, "http_status": http_status, "error_code": error_code, "count": count}
+
+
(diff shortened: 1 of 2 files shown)


AndreaFrancis (approved):
Great!

Issue #1589, opened 2023-07-31
Update datasets dependency to 2.14


albertvillanova:
TODO:
- [x] #1614 
- [x] #1616
- [x] #1617
- [x] #1619
- [x] #1620

PR #1616, opened 2023-08-03
Fix default config name

albertvillanova:
Fix default config name and refactor:
- the function no longer uses the argument `dataset`
- it returns a 2-tuple

Fix partially #1589.


albertvillanova:
This fix makes CI number of errors pass from 37 to 21.

PR #1617, opened 2023-08-03
Fix cached filenames

albertvillanova:
Fix the filename of cached files, so that it contains the `builder.dataset_name` instead of `builder.name`.

Fix partially #1589.


albertvillanova:
This fix makes CI number of errors pass from 21 to 12.

PR #1618, opened 2023-08-03
feat: add search field to /is-valid

AndreaFrancis:
Adding `search `new field to /is-valid response, this new field should help UI identify if search is available for split viewer.
Previously, /is-valid was only available at dataset level, adding split and config level to have better granularity.

New job runners:
- split-is-valid
- config-is-valid


libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -473,8 +473,15 @@ def get_valid_datasets(kind: str) -> Set[str]:
     return set(CachedResponseDocument.objects(kind=kind, http_status=HTTPStatus.OK).distinct("dataset"))
 
 
-def is_valid_for_kinds(dataset: str, kinds: List[str]) -> bool:
-    return CachedResponseDocument.objects(dataset=dataset, kind__in=kinds, http_status=HTTPStatus.OK).count() > 0
+def is_valid_for_kinds(
```
severo, line 476: I think I didn't chose the function name well: it would be better to know that only one response needs to be a success (ie: it's an OR, not and AND).

something like:

```suggestion
def has_any_successful_response(
```

feel free to ignore or change the name to something clearer

services/worker/src/worker/job_runners/config/is_valid.py
```diff
@@ -0,0 +1,97 @@
...
+                continue
+            split_is_valid_content = response["content"]
+            preview_count += 1 if split_is_valid_content["preview"] else 0
+            viewer_count += 1 if split_is_valid_content["viewer"] else 0
+            search_count += 1 if split_is_valid_content["search"] else 0
+    except Exception as e:
+        raise PreviousStepFormatError("Previous step did not return the expected content.", e) from e
+
+    progress = (total - pending) / total if total else 1.0
+    preview = preview_count == total
```
severo, line 72: it means it is valid if all the splits are valid? I think it's not what we had before.

BTW, it might be clearer to do than to compare numbers

```
preview = preview or split_is_valid_content["preview"]
```

services/worker/src/worker/job_runners/dataset/is_valid.py
```diff
@@ -26,22 +30,54 @@ def compute_is_valid_response(dataset: str, processing_graph: ProcessingGraph) -
...
+            split_is_valid_content = response["content"]
+            preview_count += 1 if split_is_valid_content["preview"] else 0
+            viewer_count += 1 if split_is_valid_content["viewer"] else 0
+            search_count += 1 if split_is_valid_content["search"] else 0
+    except Exception as e:
+        raise PreviousStepFormatError("Previous step did not return the expected content.", e) from e
 
-    return DatasetIsValidResponse({"viewer": viewer, "preview": preview})
+    progress = (total - pending) / total if total else 1.0
+    preview = preview_count == total
```
severo, line 69: same comment: should be or, not and, and it might be clearer to use logical operators instead of a sum

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: Sure! Done in https://github.com/huggingface/datasets-server/pull/1618/commits/6f0bda0f19285519098a66d5eb90b82cc545f572

services/worker/src/worker/job_runners/dataset/is_valid.py
AndreaFrancis: Makes more sense, done in https://github.com/huggingface/datasets-server/pull/1618/commits/6f0bda0f19285519098a66d5eb90b82cc545f572

services/worker/src/worker/job_runners/config/is_valid.py
AndreaFrancis: Makes more sense, done in https://github.com/huggingface/datasets-server/pull/1618/commits/6f0bda0f19285519098a66d5eb90b82cc545f572

severo (commented):
OK, better.

I think we need to check the three levels in e2e though

PR #1619, opened 2023-08-04
Fix HfFileSystem

albertvillanova:
Fix usage of `HfFileSystem` (instead of `HTTPFileSystem`) and filename format of `data_files`.

Additionally, fix `fill_builder_info` with additional builder information: `builder_name`, `dataset_name`, `config_name` and `version`.

Fix partially #1589.


albertvillanova:
This fix makes CI number of errors pass from 12 to 7.

PR #1623, opened 2023-08-04
Enable delete-indexes job to run every 10 minutes

AndreaFrancis:
In order to try correct functionality of delete-index job, I would like to test every 10 minutes.
Then I will remove this schedule and will keep as default (once a day).


severo:
OK! Please do directly, without needing an approval. But beware: the job will possibly take more than 10 minutes to run.

PR #1628, opened 2023-08-04
fix: unquoted env vars

rtrompier:


AndreaFrancis (approved):

severo (approved):
wow thanks, it was not that easy

PR #1654, opened 2023-08-09
move cache metrics inc to orchestrator

AndreaFrancis:
Currently, we calculate cache metrics in a job that runs every X minutes but this one queries the full cache collection every time leading to mongo query targeting issues in Mongo Atlas (ratio > 1000).
In order to avoid this alert, this PR moves the metrics increase to orchestrator so that, every time a cache in recorded, it will update the metrics collection as well..

Caveats: 

- There is no way to "decrease" the counters if a cache record is deleted. It could be implemented when doing the delete but It could increase db load.

### DIFF
--- a/chart/templates/worker/_container.tpl
+++ b/chart/templates/worker/_container.tpl
@@ -8,6 +8,7 @@
   env:
   {{ include "envAssets" . | nindent 2 }}
   {{ include "envCache" . | nindent 2 }}
+  {{ include "envMetrics" . | nindent 2 }}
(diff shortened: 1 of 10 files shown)


severo:
(same for the jobs metrics)

AndreaFrancis:
as an alternative: what if we increase / decrease in simple_cache.py instead, as all the operations on the cache are done through that file?

Makes sense, I would open another PR with that approach. Also adding same logic for Jobs

PR #1658, opened 2023-08-09
Incremental cache metrics

AndreaFrancis:
Currently, we calculate cache metrics in a job that runs every X minutes but this one queries the full cache collection every time leading to mongo query targeting issues in Mongo Atlas (ratio > 1000).
 
Alert message:
`The ratio of documents scanned to returned exceeded 1000.0 on  datasets-server-prod-shard-00-00.ujrd0.mongodb.net, which typically suggests that un-indexed queries are being run. 
`

In order to avoid this alert, this PR moves the metrics increase to simple_cache so that, every time a cache is recorded, it will update the metrics collection as well.

Note.- Job metrics will be added in another PR


e2e/tests/test_31_admin_metrics.py
```diff
@@ -31,28 +31,9 @@ def test_metrics() -> None:
 
     metric_names = set(metrics.keys())
 
-    # the cache and queue metrics are computed by the background jobs. Here, in the e2e tests, we don't run them,
```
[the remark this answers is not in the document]
severo: how could we test in e2e that we have the metrics?

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -127,6 +128,19 @@ class CacheEntryDoesNotExistError(DoesNotExist):
     pass
 
 
+def increase_metrics(kind: str, http_status: HTTPStatus, error_code: Optional[str] = None) -> None:
```
severo, line 131: Since we now only change the metrics incrementally, both datasets (cache and metrics) must be aligned. Maybe we should run the complete query (snapshot) regularly to ensure we are not diverging. In particular, at startup, maybe the cache dataset is already filled while the metrics one is empty.

Not sure if it's needed, but the idea is the same as how PostgreSQL writes on the disk (https://www.postgresql.org/docs/current/wal-configuration.html): it appends logs (Write-Ahead Logging - WAL) incrementally, but also does regular Checkpoints to ensure integrity.

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -142,7 +156,10 @@ def upsert_response(
     progress: Optional[float] = None,
     updated_at: Optional[datetime] = None,
 ) -> None:
-    CachedResponseDocument.objects(kind=kind, dataset=dataset, config=config, split=split).upsert_one(
+    records = CachedResponseDocument.objects(kind=kind, dataset=dataset, config=config, split=split)
+    decrease_metrics(records=records)
```
severo, line 160: Hmmm, good point. Is it possible to have more than one result in `records`? If it's only 0 or 1, the logic is OK, if it can be >1, it would not work anymore, since upsert_one only updates the first result.

severo (commented):
Such a lot of files are modified, only to access the metrics database. Maybe we could host the cache metrics as another collection inside the cache database, it would make everything simpler. WDYT?

AndreaFrancis:
> how could we test in e2e that we have the metrics?

Actually, we were testing that there were no metrics. To test that we have the metrics, I could add another test at the end to ensure that after all jobs (or at least some) have been processed and the metrics have been increased at least for some cache kinds.

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: No, it should not be more than one since we have the constrain:
`    kind = StringField(required=True, unique_with=["dataset", "config", "split"])`

I will change to decrease only by one record.

AndreaFrancis:
> how could we test in e2e that we have the metrics?

Actually, we were testing that there were no metrics. To test that we have the metrics, I could add another test at the end to ensure that after all jobs (or at least some) have been processed and the metrics have been increased at least for some cache kinds.

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: Is it worth storing the metric history? or should we keep only the last record in the metric collection as it is currently implemented?
I think Prometheus will store already the history so maybe only option two is needed ( Correct -> Increment/decrement the current last record  that is the last captured snapshot value)

severo:
> Actually, we were testing that there were no metrics. To test that we have the metrics, I could add another test at the end to ensure that after all jobs (or at least some) have been processed and the metrics have been increased at least for some cache kinds.

Yes, the test was lame. But it would be nice to have one now.

libs/libcommon/src/libcommon/simple_cache.py
severo: Maybe we should add a test on the length of records, and raise an error if it's more than 1. Just to be sure that if we drop the constraint one day, we'll see that it breaks the logic here.

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: > Or, even, we could run the cron job more often, ie. every 5 minutes, but:: the cron job would only update one metric at a time, to reduce the query complexity. For each metric, we would store the date of the last "checkpoint", and the cron job would update the first "outdated" metric (ie. the last date was more than 1 day ago)

In this case, we could be having the same problem as we currently have: high ratio scanned keys vs returned keys, even if we reduce the group of metrics to calculate. The query complexity would reduce but anyway, we will end with a query targeting issue (e.g for split-first-rows-from-streaming we have like 158287 records, and with the aggregated record we retrieve like 10 results with a ratio of 15828 - exceeding the threashold that is 1000).


I think the first option is better and enough for now:

> So, for the cron job, we could keep the current main code, but reduce the frequency from 2 minutes to 1 day, for example (we don't need to fix more often than that, I think -- it's only to avoid the metrics to diverge too much). By the way, it would be nice to have a report, when the cron job finishes, of how much it has corrected the values.

severo: Hmmm, I think the opposite:

-> option 1: we do a big aggregate query once a day: it will raise the same warning (too many scanned keys)
-> option 2: once a day at most (for each cache type) we run a query with a filter on the specific cache type. It should use the index and not raise the warning (I think). The trick here is that we would get the cache types from the entries in the metrics collections, so: no need to query the cache collection to get the list of cache types.

wdyt?

AndreaFrancis: > -> option 2: once a day at most (for each cache type) we run a query with a filter on the specific cache type. It should use the index and not raise the warning (I think)

Currently, the problematic query already uses the appropriate index. I think there is not a performance problem with the execution plan (a bad plan or something like that), but it is a problem with the way/design we process metrics (querying everything to calculate aggregated data) and I think that is what the query ratio wants to highlight.
If we do option 2, we could end up with a lower query targeting value but anyway, it will always be more than 1000 so we could still receive the alert every time a cache kind is processed (maybe 24 times a day in the worst cases 1 by each processing step).
In my opinion, it could be better to have option 1 since anyway we will have the alert but in this case, it will be just once a day in the worst case.

BTW, I think the traditional data processing systems face this problem in a batch incremental way over periods of time, let's say saving the last_read_timestamp and computing the metrics every x time just for the records updated after last_read_timestamp. Maybe we could change to something like that in the future?

severo: OK, let's do option 1, and create a different issue to try to optimize it later.

e2e/tests/test_31_admin_metrics.py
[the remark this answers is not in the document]
AndreaFrancis: Actually, since this test runs after `test_11_api.py`, some of the cache records should have already been computed so, we could simply change the not has_metric logic to the opposite.

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -127,6 +161,31 @@ class CacheEntryDoesNotExistError(DoesNotExist):
...
+def update_metrics(kind: str, http_status: HTTPStatus, inc__total: int, error_code: Optional[str] = None) -> None:
+    CacheTotalMetricDocument.objects(kind=kind, http_status=http_status, error_code=error_code).upsert_one(
+        inc__total=inc__total
+    )
+
+
+def decrease_metrics(kind: str, dataset: str, config: Optional[str], split: Optional[str]) -> None:
+    query_set = CachedResponseDocument.objects(kind=kind, dataset=dataset, config=config, split=split)
+    if query_set.count() > 1:
+        raise MultipleCacheError("There should exist only one record for combination kind-dataset-config-split")
```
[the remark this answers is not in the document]
severo: MultipleObjectsReturned: just don't catch the exception

DoesNotExist: ignore

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -127,6 +161,31 @@ class CacheEntryDoesNotExistError(DoesNotExist):
     pass
 
 
+class MultipleCacheError(MultipleObjectsReturned):
+    pass
+
+
+def update_metrics(kind: str, http_status: HTTPStatus, inc__total: int, error_code: Optional[str] = None) -> None:
```
severo, line 168: nit: why two `__` in `inc__total`?

severo: why not `increase_by`, maybe it's clearer

severo: hmmm ok it's mongoengine syntax. Even though: maybe use `increase_by` in the method signature, then pass it to `inc__total` in the mongoengine method

libs/libcommon/src/libcommon/simple_cache.py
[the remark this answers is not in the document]
severo: Also: ultra nit, I would find it a bit clearer to have:
- `increase_metric(kind: str, http_status: HTTPStatus, error_code: Optional[str] = None)`
- `decrease_metric(kind: str, http_status: HTTPStatus, error_code: Optional[str] = None)`
- `decrease_metric_for_artifact(kind: str, dataset: str, config: Optional[str], split: Optional[str])`

`increase_metric` and `decrease_metric` would call `_update_metric(kind: str, http_status: HTTPStatus, error_code: Optional[str] = None, inc__total: int)`

jobs/cache_maintenance/src/cache_maintenance/main.py
```diff
@@ -15,15 +15,16 @@
...
 from cache_maintenance.delete_indexes import delete_indexes
-from cache_maintenance.metrics import collect_metrics
+from cache_maintenance.queue_metrics import collect_queue_metrics
 
 
 def run_job() -> None:
     job_config = JobConfig.from_env()
     action = job_config.action
-    supported_actions = ["backfill", "collect-metrics", "delete-indexes", "skip"]
+    supported_actions = ["backfill", "collect-cache-metrics", "collect-metrics", "delete-indexes", "skip"]
```
severo, line 27: maybe

```suggestion
    supported_actions = ["backfill", "collect-cache-metrics", "collect-queue-metrics", "delete-indexes", "skip"]
```

jobs/cache_maintenance/src/cache_maintenance/main.py
```diff
@@ -65,7 +66,9 @@ def run_job() -> None:
                 cache_max_days=job_config.cache.max_days,
             )
         elif action == "collect-metrics":
```
severo, line 68:
```suggestion
        elif action == "collect-queue-metrics":
```

jobs/cache_maintenance/src/cache_maintenance/cache_metrics.py
```diff
@@ -0,0 +1,29 @@
...
+        kind = metric["kind"]
+        http_status = metric["http_status"]
+        error_code = metric["error_code"]
+        new_total = metric["count"]
+
+        query_set = CacheTotalMetricDocument.objects(kind=kind, http_status=http_status, error_code=error_code)
+        current_metric = query_set.first()
+        if current_metric is not None:
+            current_total = current_metric.total
+            logging.info(
```
severo, line 24: excellent, it would be interesting to see these logs: do the metrics diverge, or does it work well?

chart/values.yaml
```diff
@@ -343,6 +343,7 @@ deleteIndexes:
 
 
 metricsCollector:
```
severo, line 345: rename to queueMetricsCollector?

chart/values.yaml
```diff
@@ -343,6 +343,7 @@ deleteIndexes:
...
+  enabled: true
   action: "collect-metrics"
```
severo, line 347:
```suggestion
  action: "collect-queue-metrics"
```

jobs/cache_maintenance/src/cache_maintenance/main.py
AndreaFrancis: I was planning to change all related names in the other PR to avoid modified files but I can do it here as well.

severo: yes, as you want. I proposed it because you renamed metrics.py to queue_metrics.py

libs/libcommon/src/libcommon/simple_cache.py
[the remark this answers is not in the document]
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1658/commits/b6650cff6125c20e7e878500bb156de7384b1cf6

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1658/commits/b6650cff6125c20e7e878500bb156de7384b1cf6

chart/values.yaml
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1658/commits/d75b8cde903ca9b317e9f7c4bcb50518e0a130c8

severo (approved):

PR #1660, opened 2023-08-10
Revert datasets authentication with DownloadConfig

### LINKED ISSUE #1659: Revert datasets authentication tweaks
Once we update `datasets` to version 2.14.4, we no longer need the authentication tweaks (where we had to use `download_config` instead of `token`) introduced by:
- #1620

albertvillanova:
Once we update `datasets` to version 2.14.4, we no longer need the authentication tweaks (where we had to use `download_config` instead of `token`) introduced by:
- #1620

Fix #1659.


AndreaFrancis (approved):
Thanks!

severo (approved):
Excellent!

Issue #1656, opened 2023-08-09
Gated datasets without authentication header return 404


severo:
It should return 401 (Unauthorized).

See for example https://datasets-server.huggingface.co/splits?dataset=severo/bigcode/the-stack from a private window.

severo:
When processing this issue, try to also test authenticated against authorized, and add examples in the openapi spec if needed

Issue #849, opened 2023-02-24
Serve openapi.json from the docs


severo:
Currently https://github.com/huggingface/datasets-server/blob/main/chart/static-files/openapi.json is a file in chart; every change in it should trigger a new version of the Chart. But, semantically, it does not belong to the chart and should be part of the <strike>API service</strike> docs.

severo:
See #1667

severo:
Mentioned here: https://github.com/huggingface/doc-builder/issues/312#issuecomment-1675099444.

Meanwhile, let's proxy /openapi.json to GitHub directly.

severo:
not important

Issue #518, opened 2022-08-24
Add tests to do proper contract testing 


severo:
Currently the e2e tests use the OpenAPI spec to do some contract testing, but it's not extensive.

The idea is to cover most of, or all, the cases documented in the OpenAPI spec.

severo:
See #1667

severo:
related to #2853

Issue #446, opened 2022-07-21
check that the openapi specification is valid


severo:
It should be triggered by `make quality` in the infra/charts/datasets-server directory and quality github action

severo:
done with https://github.com/huggingface/datasets-server/pull/509

PR #1667, opened 2023-08-10
fix: 🐛 update the OpenAPI spec

severo:
Missing:
- [x] ensure we documented all the status codes: missing: <strike>400 (BAD_REQUEST)</strike> (removed 400 from the code with [d4ee7a5](https://github.com/huggingface/datasets-server/pull/1667/commits/d4ee7a5bd32b9c1666e0f1293c8c0292a265e133)), 501 (NOT_IMPLEMENTED)
- [x] ensure the OpenAPI spec is correct (I have some issues in stoplight, some parts are not rendered). It works well in https://redocly.github.io/redoc/?url=https://github.com/huggingface/datasets-server/raw/0e3a700bb3716aad9d90897e844c9d40455a3b6f/chart/static-files/openapi.json#operation/getOptInOutUrls
- [x] move openapi outside of the chart? its size generates issues in the CI, and anyway it does not pertain the chart. See #849.
- [x] validate the openapi spec in the CI <strike>with spectral</strike> -> #446
- [x] fix the e2e test that use openapi (contract testing)

In other PRs:
- document all the X-Error-Code in OpenAPI -> #1671 
- do contract testing on the OpenAPI examples. Maybe use a new field, like `x-contract-testing-url`, to get the URL to test programmatically. Also: create permanent test datasets (like severo/private) in prod that can be used to contract test in prod? -> #518 
- generate OpenAPI spec from the code -> #1670

- test authenticated against authorized, and maybe add examples in the openapi spec -> in #1656

### DIFF
--- a/services/worker/src/worker/dtos.py
+++ b/services/worker/src/worker/dtos.py
@@ -2,7 +2,7 @@
 # Copyright 2023 The HuggingFace Authors.
 
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Mapping, Optional, TypedDict
+from typing import Any, Dict, List, Mapping, Optional, TypedDict, Union
 
 from libcommon.utils import FeatureItem, Row, RowItem, SplitHubFile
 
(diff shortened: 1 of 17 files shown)


services/worker/src/worker/dtos.py
```diff
@@ -136,7 +147,10 @@ class ConfigSize(TypedDict):
     num_columns: int
 
 
-class SplitSize(SplitItem):
+class SplitSize(TypedDict):
```
AndreaFrancis, line 150: Why to FullSplitItem instead of TypedDict?

AndreaFrancis (approved):

services/worker/src/worker/dtos.py
severo: yes, it could be. But I found that a lot of indirections are not easy to follow. Easier to just put the four properties

PR #1672, opened 2023-08-11
feat: 🎸 move openapi.json to the docs

severo:
note that we cannot serve from the deployed docs (see https://github.com/huggingface/doc-builder/issues/312#issuecomment-1675099444). We thus redirect to github. Also: we fix the github action that checks openapi (it was written opanapi)


severo:
I'll merge despite the CI error for Argo... Hopefully it will not break the deployment.

Also: I ignore the error for the openapi specification check, because it's not the point of this PR. It errors now because we fixed a typo in the github action (before: it was never called).

severo:
The argocd error was not that important. I've been able to deploy to prod, and the CI now passes, see https://github.com/huggingface/datasets-server/pull/1667

severo:
Also, in #1667, the openapi check passes

PR #1674, opened 2023-08-11
feat: 🎸 add metrics for all the volumes

### LINKED ISSUE #1561: Check the disk usage of all the storages in metrics
See https://www.notion.so/huggingface2/Disk-storage-0a4a8fcf27754c8cb7248b259dcc4b21 (internal)

severo:
fixes #1561


chart/templates/services/admin/_container.tpl
```diff
@@ -12,6 +12,10 @@
   {{ include "envCommon" . | nindent 2 }}
   {{ include "envLog" . | nindent 2 }}
   {{ include "envMetrics" . | nindent 2 }}
+  {{ include "envParquetMetadata" . | nindent 2 }}
+  {{ include "envWorker" . | nindent 2 }}
```
AndreaFrancis, line 16: Not sure if we should include all the worker configs here, maybe just the ones that are needed? (I think it is DUCKDB_INDEX_CACHE_DIRECTORY and DESCRIPTIVE_STATISTICS_CACHE_DIRECTORY)

tools/docker-compose-datasets-server.yml
```diff
@@ -28,6 +28,10 @@ services:
     build:
       context: ..
       dockerfile: services/admin/Dockerfile
+    volumes:
+      - assets:${ASSETS_STORAGE_DIRECTORY-/assets}:rw
+      - duckdb-index:${DUCKDB_INDEX_CACHE_DIRECTORY-/duckdb-index}:rw
```
AndreaFrancis, line 33: Do you know if the volume for statistics is missing here? Same for changes on line 52-55

tools/docker-compose-datasets-server.yml
```diff
@@ -28,6 +28,10 @@ services:
     build:
       context: ..
       dockerfile: services/admin/Dockerfile
+    volumes:
+      - assets:${ASSETS_STORAGE_DIRECTORY-/assets}:rw
```
AndreaFrancis, line 32: Why not just read for these volumes?

libs/libcommon/src/libcommon/prometheus.py
```diff
@@ -76,15 +100,35 @@ def update_responses_in_cache_total() -> None:
         ).set(cache_metric.total)
 
 
-def update_assets_disk_usage(assets_directory: StrPath) -> None:
+def update_disk_gauge(gauge: Gauge, directory: StrPath) -> None:
     # TODO: move to metrics, as for the other metrics (queue, cache)
-    total, used, free, percent = disk_usage(str(assets_directory))
+    total, used, free, percent = disk_usage(str(directory))
```
AndreaFrancis, line 105: In the next line, it should update the labels for the new gauge var, not for ASSETS_DISK_USAGE, I think this is the problem.

chart/templates/services/admin/_container.tpl
severo: Yes, we might be more precise. I'll change it, also: it will be clearer in the admin template

PR #1684, opened 2023-08-14
Incremental queue metrics

AndreaFrancis:


chart/templates/cron-jobs/backfill/_container.tpl
```diff
@@ -12,7 +12,6 @@
     {{ include "envCache" . | nindent 2 }}
     {{ include "envQueue" . | nindent 2 }}
     {{ include "envCommon" . | nindent 2 }}
-    {{ include "envMetrics" . | nindent 2 }}
```
severo, line 15: is envMetrics still used somewhere? Otherwise, we should also remove its template

libs/libcommon/src/libcommon/queue.py
```diff
@@ -238,6 +246,34 @@ def flat_info(self) -> FlatJobInfo:
...
+
+    Args:
+        queue (`str`): queue name
+        status (`str`): job status see libcommon.queue.Status
+        total (`int`): total of jobs
+        created_at (`datetime`): when the metric has been created.
+    """
+
+    id = ObjectIdField(db_field="_id", primary_key=True, default=ObjectId)
+    queue = StringField(required=True)
```
severo, line 264: maybe (queue, status) should be set as unique

libs/libcommon/src/libcommon/queue.py
```diff
@@ -238,6 +246,34 @@ def flat_info(self) -> FlatJobInfo:
...
 
+DEFAULT_INCREASE_AMOUNT = 1
+DEFAULT_DECREASE_AMOUNT = -1
+
+
+class JobTotalMetricDocument(Document):
+    """Jobs total metric in mongoDB database, used to compute prometheus metrics.
+
+    Args:
+        queue (`str`): queue name
```
severo, line 257: maybe job_type instead of queue, as it's how we call it in the rest of the code

libs/libcommon/src/libcommon/queue.py
```diff
@@ -449,6 +504,8 @@ def create_jobs(self, job_infos: List[JobInfo]) -> int:
                 )
                 for job_info in job_infos
             ]
+            for job in jobs:
+                increase_metric(queue=job.type, status=Status.WAITING)
```
severo, line 508: would it make sense to do only one query with an array of job_types? Or is it unnecessary optimization?

libs/libcommon/src/libcommon/queue.py
```diff
@@ -467,6 +524,8 @@ def cancel_jobs_by_job_id(self, job_ids: List[str]) -> int:
         """
         try:
             existing = JobDocument.objects(pk__in=job_ids)
+            for existing_job in existing.all():
+                update_metrics_for_updated_job(existing_job, status=Status.CANCELLED)
```
severo, line 528: same comment: one query instead of a loop?

chart/templates/cron-jobs/backfill/_container.tpl
AndreaFrancis: It is still being used in mongodb-migration because it is needed for the last migration (drop collection). I think we can remove it after this PR is merged.

libs/libcommon/src/libcommon/queue.py
AndreaFrancis: Not sure if it is more efficient to perform one __in at db level for update instead of a loop.
Also if would anyway have to keep the order-combination of job_type-status for the update (cannot do job_type__in and status__in).
Maybe with a bulk write but not sure if it is worth to add too much logic to replace a loop. wdyt?

libs/libcommon/src/libcommon/queue.py
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1684/commits/721689f02a3fd58e117a8a4bf0033a5bd883d311

libs/libcommon/src/libcommon/queue.py
AndreaFrancis: Done in https://github.com/huggingface/datasets-server/pull/1684/commits/721689f02a3fd58e117a8a4bf0033a5bd883d311

severo (approved):

PR #1685, opened 2023-08-15
Rename parquet files

lhoestq:
from `{config}/{dataset_name}-{split}{sharded_suffix}.parquet` to `{config}/{split}/{shard_idx:04d}.parquet`


services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -128,20 +128,17 @@ def __init__(
...
     @property
     def path_in_repo(self) -> str:
-        if self.partial:
-            # Using 4 digits is ok since MAX_FILES_PER_DIRECTORY == 10_000
-            return f"{self.config}/partial/{self.split}/{self.shard_idx:04d}.parquet"
-        else:
-            return f'{self.config}/{self.local_file.removeprefix(f"{self.local_dir}/")}'
+        # Using 4 digits is ok since MAX_FILES_PER_DIRECTORY == 10_000
+        partial_path = "/partial" if self.partial else ""
+        return f"{self.config}{partial_path}/{self.split}/{self.shard_idx:04d}.parquet"
```
AndreaFrancis, line 133: Maybe a dummy question but I was thinking that a "partial" was like a preview of a split, not sure if it should go inside a config.

AndreaFrancis (approved):
Thanks!

services/worker/src/worker/job_runners/config/parquet_and_info.py
severo: Same: I think it would be easier to understand as
```suggestion
        return f"{self.config}/{self.split}{partial_path}/{self.shard_idx:04d}.parquet"
```

services/worker/tests/fixtures/hub.py
```diff
@@ -494,7 +491,7 @@ def create_parquet_and_info_response(
                 "url": CI_URL_TEMPLATE.format(
                     repo_id=f"datasets/{dataset}",
                     revision="refs%2Fconvert%2Fparquet",
-                    filename=f"{config}/partial/{split}/{filename}" if partial else f"{config}/{filename}",
+                    filename=f"{config}{partial_path}/{split}/{filename}",
```
severo, line 494: if we change the order of partial and split, it should be changed here too

services/worker/src/worker/job_runners/config/parquet_and_info.py
severo: + we avoid conflict if a split is name `partial`, and we avoid the user to have to think if it's a split name or the special partial keyword

lhoestq: makes sense, I'll do the change

lhoestq:
Following internal discussion [here](https://huggingface.slack.com/archives/C04L6P8KNQ5/p1692194823626939) I switched to

```
{config}/partial-{split}/shard_idx:04d.parquet
```

severo (approved):
missing:
-> docs
-> openapi

lhoestq:
Will open a separate PR for the docs, since we might not merge it at the same time (it adds some URLs that don't exist yet)

julien-c:
was this applied retroactively to existing repos on the hub?

severo:
yes, to all of them (currently being processed)

Issue #1686, opened 2023-08-15
Enable duckdb index on gated datasets


severo:
> Currently, duckdb index is not supported for gated/private datasets. I opened a question in duckdb foundations but didn't receive a response yet, I think I will open an issue in the repo https://github.com/duckdb/foundation-discussions/discussions/16

from [Slack](https://huggingface.slack.com/archives/C04L6P8KNQ5/p1692050576450149?thread_ts=1692049509.952679&cid=C04L6P8KNQ5) (internal)

AndreaFrancis:
I was thinking that maybe we can use the same approach as split-descriptive-statistics, temporarily download the parquets using auth for gated datasets and then load it to the duckdb index.

severo:
Yes, maybe it's even faster that way. Did you compare?

AndreaFrancis:
I compared and looks to be similar, with a difference of a couple of seconds:


URL | Size MB | Time (seconds) - download | Time (seconds) - duckdb
-- | -- | -- | --
https://huggingface.co/datasets/amazon_us_reviews/resolve/refs%2Fconvert%2Fparquet/Baby_v1_00/amazon_us_reviews-train-00000-of-00002.parquet | 274 | 188.66 | 161.86
https://huggingface.co/datasets/asoria/copy_beans/resolve/refs%2Fconvert%2Fparquet/default/test/0000.parquet | 17.7 | 8.33 | 8.13
https://huggingface.co/datasets/IlyaGusev/yandex_q_full/resolve/refs%2Fconvert%2Fparquet/default/partial-train/0008.parquet | 260 | 88.7 | 96


I think we can add an exception in case it was not possible to use duckdb to load the data and try to download the parquet files.

severo:
The difference is small. I would say: until the auth is available in duckdb, it's better to just switch to downloading the files. The code will be easier to maintain.

AndreaFrancis:
Related to https://github.com/duckdb/foundation-discussions/discussions/16

julien-c:
BTW re. DuckDB maybe you can send a quick email too (on the thread we had) to make sure it's on their radar

julien-c:
@AndreaFrancis very interesting table – can you split the "download" case in `download` and `query`?

julien-c:
oh actually nevermind, i thought this was about the query time, between downloading + querying locally, vs. querying remotely 

For the workers i think we don't care that much about perf

julien-c:
> querying remotely

[...]

PR #1717, opened 2023-08-22
Download parquet in split-duckdb-index

AndreaFrancis:
Fixes https://github.com/huggingface/datasets-server/issues/1686
Using hf_download instead of loading parquet files to duckdb directly.

### DIFF
--- a/services/worker/src/worker/job_runners/split/duckdb_index.py
+++ b/services/worker/src/worker/job_runners/split/duckdb_index.py
@@ -2,10 +2,12 @@
 # Copyright 2023 The HuggingFace Authors.
 
 import logging
+import os
 from pathlib import Path
 from typing import List, Optional, Set
 
 import duckdb
+from huggingface_hub import hf_hub_download
 from huggingface_hub._commit_api import (
     CommitOperation,
     CommitOperationAdd,
@@ -53,6 +55,8 @@ CREATE_TABLE_COMMAND = "CREATE OR REPLACE TABLE data AS SELECT nextval('serial')
 INSTALL_EXTENSION_COMMAND = "INSTALL '{extension}';"
 LOAD_EXTENSION_COMMAND = "LOAD '{extension}';"
 SET_EXTENSIONS_DIRECTORY_COMMAND = "SET extension_directory='{directory}';"
+REPO_TYPE = "dataset"
+HUB_DOWNLOAD_CACHE_FOLDER = "cache"
 
 
 def compute_index_rows(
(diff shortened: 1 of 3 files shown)


services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -138,7 +141,22 @@ def compute_index_rows(
     logging.debug(CREATE_SEQUENCE_COMMAND)
     con.sql(CREATE_SEQUENCE_COMMAND)
 
-    create_command_sql = f"{CREATE_TABLE_COMMAND.format(columns=column_names)} read_parquet({parquet_urls});"
+    # see https://pypi.org/project/hf-transfer/ for more details about how to enable hf_transfer
+    os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
+    for parquet_file in parquet_file_names:
+        hf_hub_download(
```
severo, line 147: For https://huggingface.co/datasets/bigcode/the-stack, this means downloading 2TB, no?

I think we have to check the size and stop downloading if the sum of the parquet files is more than 5GB.

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -138,7 +141,22 @@ def compute_index_rows(
...
+            repo_id=dataset,
+            filename=f"{config}/{split}/{parquet_file}",
+            local_dir=duckdb_index_file_directory,
+            local_dir_use_symlinks=False,
+            token=hf_token,
+            cache_dir=duckdb_index_file_directory,
+        )
+
+    all_split_parquets = f"{duckdb_index_file_directory}/{config}/{split}/*.parquet"
+    create_command_sql = f"{CREATE_TABLE_COMMAND.format(columns=column_names)} '{all_split_parquets}';"
```
severo, line 159: We must delete the parquet files afterward, as split-descriptive-statistics does.

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: There is a validation already done here https://github.com/huggingface/datasets-server/pull/1717/files#diff-d9a2c828d7feca3b7f9e332e040ef861e842a16d18276b356461d2aa34396a8aR97 that will prevent processing splits with parquets bigger than 5 GB

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: Since this job runner inherits from SplitJobRunnerWithCache, they will be removed because `duckdb_index_file_directory` is the generated temporary `self.cache_subdirectory`.

services/worker/src/worker/job_runners/split/duckdb_index.py
severo: Oh OK! Thanks for clarifying

services/worker/src/worker/job_runners/split/duckdb_index.py
severo: OK!!!

severo (approved):

Issue #1712, opened 2023-08-21
Missing auth for `split-descriptive-statistics`


lhoestq:
```
| INFO: 2023-08-21 09:23:35,871 - root - [split-descriptive-statistics] compute JobManager(job_id=64dd8b2e4833e19c97d7c0db dataset=mozilla-foundation/common_voice_9_0 job_info={'job_id': '64dd8b2e4833e19c97d7c0db', 'type': 'split-descriptiv │
│ INFO: 2023-08-21 09:23:35,879 - root - Compute descriptive statistics for dataset='mozilla-foundation/common_voice_9_0', config='pt', split='invalidated'                                                                                      │
│ INFO: 2023-08-21 09:23:36,440 - root - Downloading remote data to a local parquet file /storage/stats-cache/74577196210609-split-descriptive-statistics-mozilla-foundation-c-1d56685e/dataset.parquet.                                         │
│ ERROR: 2023-08-21 09:23:37,038 - root - HTTP Error: HTTP GET error on 'https://huggingface.co/datasets/mozilla-foundation/common_voice_9_0/resolve/refs%2Fconvert%2Fparquet/pt/invalidated/0000.parquet' (HTTP 401)                            │
│ Traceback (most recent call last):                                                                                                                                                                                                             │
│   File "/src/services/worker/src/worker/job_manager.py", line 160, in process                                                                                                                                                                  │
│     job_result = self.job_runner.compute()                                                                                                                                                                                                     │
│   File "/src/services/worker/src/worker/job_runners/split/descriptive_statistics.py", line 404, in compute                                                                                                                                     │
│     compute_descriptive_statistics_response(                                                                                                                                                                                                   │
│   File "/src/services/worker/src/worker/job_runners/split/descriptive_statistics.py", line 323, in compute_descriptive_statistics_response                                                                                                     │
│     con.sql(                                                                                                                                                                                                                                   │
│ duckdb.HTTPException: HTTP Error: HTTP GET error on 'https://huggingface.co/datasets/mozilla-foundation/common_voice_9_0/resolve/refs%2Fconvert%2Fparquet/pt/invalidated/0000.parquet' (HTTP 401)
```

AndreaFrancis:
Maybe it is related to gated datasets, same as https://github.com/huggingface/datasets-server/issues/1686

PR #1719, opened 2023-08-23
Download parquet files with huggingface_hub instead of duckdb in `split-descriptive-statistics`

polinaeterna:
will fix https://github.com/huggingface/datasets-server/issues/1712#issuecomment-1690029285

### DIFF
--- a/services/worker/src/worker/config.py
+++ b/services/worker/src/worker/config.py
@@ -288,7 +288,6 @@ class DuckDbIndexConfig:
 
 
 DESCRIPTIVE_STATISTICS_CACHE_DIRECTORY = None
-DESCRIPTIVE_STATISTICS_TARGET_REVISION = "refs/convert/parquet"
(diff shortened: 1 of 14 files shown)


services/worker/src/worker/config.py
```diff
@@ -288,13 +288,15 @@ def from_env(cls) -> "DuckDbIndexConfig":
 
 
 DESCRIPTIVE_STATISTICS_CACHE_DIRECTORY = None
+DESCRIPTIVE_STATISTICS_TARGET_REVISION = "refs/convert/parquet"
```
severo, line 291: we already have PARQUET_AND_INFO_TARGET_REVISION, maybe reuse this?

Otherwise, if we create a new environment variable, we have to add it to the README, to the docker-compose files, and to the Helm chart (yes, it's a lot)

severo (approved):

AndreaFrancis (approved):

Issue #1713, opened 2023-08-21
Search doesn't always use Image type


lhoestq:
It seems the feature type is not loaded correctly, resulting in a binary type that is ignored in the viewer

e.g.

https://datasets-server.huggingface.co/search?dataset=lambdalabs/pokemon-blip-captions&config=lambdalabs--pokemon-blip-captions&split=train&query=red&offset=0&limit=100

```

  "features": [
    {
      "feature_idx": 0,
      "name": "image",
      "type": {
        "bytes": {
          "dtype": "binary",
          "_type": "Value"
        },
        "path": {
          "dtype": "string",
          "_type": "Value"
        }
      }
    },
    {
      "feature_idx": 1,
      "name": "text",
      "type": {
        "dtype": "string",
        "_type": "Value"
      }
    }
  ],
  "rows": [
    {
      "row_idx": 0,
      "row": {
        "image": null,
        "text": "a drawing of a green pokemon with red eyes"
      },
      "truncated_cells": [
        
      ]
    },
    {
      "row_idx": 1,
      "row": {
        "image": null,
        "text": "a green and yellow toy with a red nose"
      },
      "truncated_cells": [
        
      ]
    },
```

cc @AndreaFrancis 

another example: https://datasets-server.huggingface.co/search?dataset=jmhessel/newyorker_caption_contest&config=explanation&split=train&query=hospital&offset=0&limit=100

another example: https://datasets-server.huggingface.co/search?dataset=ademax/ocr_scan_vi_01&config=default&split=train&query=pho&offset=0&limit=100

lhoestq:
It's simply the `Features` metadata with the `Image()` type that are not used (not present in the arrow schema from the data from the duckdb index file)

lhoestq:
fixed: https://datasets-server.huggingface.co/search?dataset=lambdalabs/pokemon-blip-captions&config=lambdalabs--pokemon-blip-captions&split=train&query=red&offset=0&limit=100

will update the other datasets

PR #1725, opened 2023-08-24
Use features in search

lhoestq:
The search endpoint was using the feature types from the arrow table returned by duckdb, which doesn't contain any metadata about the Image type.

So I added a `features` field to the `split-duckdb-index` job to store the feature types that the search endpoint can use to correctly load the image data.

I added a migration job to fill this field with `None` and I will simply update the `split-duckdb-index` job for image datasets.

Fix https://github.com/huggingface/datasets-server/issues/1713


jobs/mongodb_migration/src/mongodb_migration/migrations/_20230824154999_cache_add_features_field_in_split_duckdb_index.py
```diff
@@ -0,0 +1,46 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 The HuggingFace Authors.
```
AndreaFrancis, line 2:
```suggestion
# Copyright 2023 The HuggingFace Authors.
```

jobs/mongodb_migration/src/mongodb_migration/migrations/_20230824154999_cache_add_features_field_in_split_duckdb_index.py
```diff
@@ -0,0 +1,46 @@
...
+from libcommon.constants import CACHE_COLLECTION_RESPONSES, CACHE_MONGOENGINE_ALIAS
+from libcommon.simple_cache import CachedResponseDocument
+from mongoengine.connection import get_db
+
+from mongodb_migration.check import check_documents
+from mongodb_migration.migration import Migration
+
+
+# connection already occurred in the main.py (caveat: we use globals)
+class MigrationAddFeaturesToSplitDuckdbIndexCacheResponse(Migration):
```
AndreaFrancis, line 15: Not critical: Can you add a test for this migration? To ensure the new field is added

AndreaFrancis (approved):
Thanks!

severo (approved):
Excellent. Can you update the openapi spec as well? (it's in docs/source)

lhoestq:
Will update the spec with a real example once it's deployed

lhoestq:
Just deleted the cached responses of datasets with an `"image": Image()` type to refresh them (around 13k documents).

Not sure how to get the list of datasets with an `Image()` type in a column named differently than `image` though without scanning all the db

severo:
> Not sure how to get the list of datasets with an Image() type in a column named differently than image though without scanning all the db

Do you have an idea @AndreaFrancis 

> Will update the spec with a real example once it's deployed

Could you do it?

severo:
> (fixed by freeing some inodes)
cached-assets now use an EFS volume instead of the limited NFS, so this issue should disappear for some time. Now, only the assets (/first-rows) are stored in the NFS.

severo:
> Will update the spec with a real example once it's deployed

https://github.com/huggingface/datasets-server/pull/1787

Issue #1406, opened 2023-06-20
Use S3 + cloudfront for assets and cached-assets


severo:
related to #1072

internal: see https://github.com/huggingface/infra/issues/605#issue-1758616648

severo:
keep open

PR #1726, opened 2023-08-24
Cached assets to s3

AndreaFrancis:
The first part of https://github.com/huggingface/datasets-server/issues/1406 
Migration for cached-assets enabling only for a list of datasets initially (I added "asoria/image" dataset for the first testing, later, we can remove this logic and apply it for all datasets).
Note that the new logic: validates if the file exists in the bucket and if not, uploads it.
This might increase latency, but will see how it works once it is deployed on prod, for now, I just added a couple of tests to ensure behavior.
All /rows and /search resources will be stored in` /cached-assets` subfolder in the bucket and the infra team will help us configure a TTL policy at the folder level to delete the files created more than a day before.
Notice that I decided to apply the same approach as Audio for Images because of:
- We are doing mostly a IO-bound task
- According to best practices, it is better to use multithreading when working with IO-bound tasks
- This doc helped me to understand that we are taking the best approach initially for parallelizing requests to S3 https://superfastpython.com/threadpool-vs-pool-in-python/

> You should use the ThreadPool for IO-bound tasks in Python in general.
> 	- Reading or writing a file from the hard drive.
> 	- Downloading or uploading a file.

Also, I compared working /rows + s3 with threa_map vs multiprocessing.Pool and had the following results (might be affected by my internet connection but at least it can give us an idea about comparison):

**For an audio dataset https://huggingface.co/datasets/asoria/audio-sample:**
Multithreading (thread_map)
- First time (all files are uploaded to the bucket)  = 7s
- Second time (only validation for existing files is done) = 4s

Multiprocessing (multiprocessing.Pool)
- First time (all files are uploaded to the bucket)  = 14s
- Second time (only validation for existing files is done) = 5s

**For an Image dataset https://huggingface.co/datasets/asoria/image:**
Multithreading (thread_map)
- First time (all files are uploaded to the bucket)  = 6s
- Second time (only validation for existing files is done) = 3s

Multiprocessing (multiprocessing.Pool)
- First time (all files are uploaded to the bucket)  = 15s
- Second time (only validation for existing files is done) = 7s


libs/libapi/src/libapi/utils.py
```diff
@@ -244,3 +259,49 @@ def clean_cached_assets(
...
+                use_s3_storage=use_s3_storage,
+                s3_bucket=cached_assets_s3_bucket,
+                s3_access_key_id=cached_assets_s3_access_key_id,
+                s3_secret_access_key=cached_assets_s3_secret_access_key,
+                s3_region=cached_assets_s3_region,
+                s3_folder_name=cached_assets_s3_folder_name,
+            )
+            for (featureName, fieldType) in features.items()
+        }
+        for row_idx, row in enumerate(rows)
```
lhoestq, line 306: Later we can parallelize the uploads to S3, e.g. using thread_map (being added for audio conversions in https://github.com/huggingface/datasets-server/pull/1743/files#diff-aaecea079c2cfdad76e443bdae72f57f785551f1ac1573c7e760728b5d54c988R69)

lhoestq (commented):
Nice ! I just added two suggestions:
- one about uploading the files in parallel - this can be done using thread_map in `transform_rows` (note that I have a PR open that adds it for audio, let's see which PR to merge first)
- one about using a Resource for S3 to avoid recreating the boto client over and over, and to avoid passing all the s3 parameters everywhere

libs/libapi/src/libapi/utils.py
AndreaFrancis: As explained in the description, I kept thread_map also for images, looks to be faster than multiprocessing for our request.

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -78,21 +83,53 @@ def create_image_file(
...
         column=column,
         assets_directory=assets_directory,
     )
-    makedirs(dir_path, ASSET_DIR_MODE, exist_ok=True)
-    file_path = dir_path / filename
-    if overwrite or not file_path.exists():
-        image.save(file_path)
+    if use_s3_storage:
+        key = f"{s3_folder_name}/{url_dir_path}/{filename}"
+        s3_client = boto3.client(
```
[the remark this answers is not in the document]
AndreaFrancis: I created a wrapper for the S3 client so it is initialized at the beginning only.

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -115,25 +152,84 @@ def create_audio_files(
...
     )
-    makedirs(dir_path, ASSET_DIR_MODE, exist_ok=True)
-    wav_file_path = dir_path / wav_filename
-    mp3_file_path = dir_path / mp3_filename
+    if not use_s3_storage:
+        wav_file_path = dir_path / wav_filename
+        mp3_file_path = dir_path / mp3_filename
+        makedirs(dir_path, ASSET_DIR_MODE, exist_ok=True)
+    else:
+        random_str = f"{random.randrange(10**13, 10**14)}"  # nosec B311
```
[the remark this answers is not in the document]
AndreaFrancis: I applied it in the create_asset_file

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -112,21 +192,31 @@ def create_audio_file(
     filename: str,
     assets_directory: StrPath,
     overwrite: bool = True,
+    # TODO: Once assets and cached-assets are migrated to S3, this parameter is no more needed
+    use_s3_storage: bool = False,
+    s3_client: Optional[S3Client] = None,
```
severo, line 197: instead of number of flat arguments, maybe we could have `storage_options: DirectoryStorageOptions | S3StorageOptions`, and pack all the S3 related options inside `S3StorageOptions`.

Also if we migrate to S3 in production, we can let the support for directory-based storage too?

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -65,6 +70,70 @@ class ImageSource(TypedDict):
...
+        column=column,
+        assets_directory=assets_directory,
+    )
+
+    if not use_s3_storage:
+        file_path = dir_path / filename
+        makedirs(dir_path, ASSET_DIR_MODE, exist_ok=True)
+    else:
+        payload = (
+            str(uuid4()),
```
severo, line 105: if we use uuid, no need for the rest of the payload, right? let's just set the prefix using the uuid.

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -65,6 +70,70 @@ class ImageSource(TypedDict):
...
+        )
+        prefix = sha1(json.dumps(payload, sort_keys=True).encode(), usedforsecurity=False).hexdigest()[:8]
+        file_path = Path(assets_directory).resolve() / f"{prefix}-{filename}"
+
+    if overwrite or not file_path.exists():
+        fn(file_path=file_path)
+
+    if use_s3_storage and s3_client is not None and s3_bucket is not None:
+        object_key = f"{s3_folder_name}/{url_dir_path}/{filename}"
+        create_object = overwrite or not s3_client.exists_in_bucket(s3_bucket, object_key)
```
severo, line 120: we are checking if the file exists, but maybe it's an outdated version? How could we ensure we always have the correct version file?

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -65,6 +70,70 @@ class ImageSource(TypedDict):
...
+        if create_object:
+            s3_client.upload_to_bucket(str(file_path), s3_bucket, object_key)
+            os.remove(file_path)
```
severo, line 123: shouldn't we always delete the file?

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -65,6 +70,70 @@ class ImageSource(TypedDict):
...
+
+    return f"{assets_base_url}/{url_dir_path}/{filename}"
```
severo, line 125: how do we manage the difference between 
```
        object_key = f"{s3_folder_name}/{url_dir_path}/{filename}"
```

and 

```
     return f"{assets_base_url}/{url_dir_path}/{filename}"
```
?

libs/libcommon/tests/viewer_utils/test_features.py
```diff
@@ -294,6 +299,93 @@ def test_others(
...
+    datasets: Mapping[str, Dataset],
+    cached_assets_directory: StrPath,
+) -> None:
+    dataset = datasets["image"]
+    feature = dataset.features["col"]
+    with mock_s3():
+        bucket_name = "bucket"
+        access_key_id = "access_key_id"
+        secret_access_key = "secret_access_key"
+        s3_client = S3Client(
```
severo, line 312: maybe move this as a fixture in conftest.py?

severo: (maybe even the boto3 connection itself)

libs/libcommon/tests/viewer_utils/test_features.py
severo, line 1: good. Maybe we could also add specific unit tests for asset.py, instead of only having derived tests as done here.

libs/libapi/src/libapi/config.py
```diff
@@ -39,6 +39,34 @@ def from_env(cls) -> "UvicornConfig":
 API_MAX_AGE_SHORT = 10  # 10 seconds
 
 
+CACHED_ASSETS_S3_BUCKET = "hf-datasets-server-statics"
```
severo, line 42: would it make sense to have one S3 bucket for assets, and another one for cached-assets? Maybe it's necessary to apply a different TTL policy for them.

If so, maybe change the bucket name to `hf-datasets-server-cached-assets` and change or remove the folder name , below

services/rows/tests/routes/test_rows.py
```diff
@@ -377,34 +390,57 @@ def test_create_response(ds: Dataset, app_config: AppConfig, cached_assets_direc
...
+                unsupported_columns=[],
+                num_rows_total=10,
+            )
+        assert response["features"] == [{"feature_idx": 0, "name": "image", "type": {"_type": "Image"}}]
+        assert response["rows"] == [
+            {
+                "row_idx": 0,
+                "row": {
+                    "image": {
+                        "src": "http://localhost/cached-assets/ds_image/--/default/train/0/image/image.jpg",
```
severo, line 428: do we want to proxy the images through the Datasets Server? else: it should be the URL of the S3 bucket, or the CDN in front of it

severo: also: to make it easier to manage the response, possibly we need two keys:
1. `src`: the online resource
2. `path`: the path inside the storage (be it S3 or another filesystem)

tools/docker-compose-base.yml
```diff
@@ -42,6 +42,12 @@ services:
       API_HF_TIMEOUT_SECONDS: ${API_HF_TIMEOUT_SECONDS-0.2}
       API_MAX_AGE_LONG: ${API_MAX_AGE_LONG-120}
       API_MAX_AGE_SHORT: ${API_MAX_AGE_SHORT-10}
+      CACHED_ASSETS_S3_BUCKET: ${CACHED_ASSETS_S3_BUCKET-}
```
severo, line 45: We generally repeat the default value here. I'm not sure if it's the best way to do it, but we should try to stay consistent, so: add them here, or remove them from elsewhere, depending on what seems more logical

tools/docker-compose-base.yml
severo, line 1: shouldn't we also update the .github directory to take this into account in the CI?

chart/values.yaml
```diff
@@ -113,6 +113,13 @@ secrets:
   spawningToken:
     fromSecret: true
     secretName: "spawning-token"
+  cachedAssetsS3:
+    accessKeyId:
+      fromSecret: true
+      secretName: "cached-assets-s3-access-key-id"
```
severo, line 119: are they created in our infra?

chart/values.yaml
```diff
@@ -255,6 +262,11 @@ cachedAssets:
   # When cleaning the cached assets directory: maximum number of rows to discard.
   maxCleanedRowsNumber: 10000
 
+cachedAssetsS3:
+  bucket: "hf-datasets-server-statics"
```
severo, line 266: see comment below (use one bucket for cached assets, and another one for assets?)

chart/templates/_env/_envCachedAssetsS3.tpl
```diff
@@ -0,0 +1,31 @@
...
+- name: CACHED_ASSETS_S3_REGION
+  value: {{ .Values.cachedAssetsS3.region | quote }}
+- name: CACHED_ASSETS_S3_FOLDER_NAME
+  value: {{ .Values.cachedAssetsS3.folderName | quote }}
+- name: CACHED_ASSETS_S3_ACCESS_KEY_ID
+  {{- if .Values.secrets.cachedAssetsS3.accessKeyId.fromSecret }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.secrets.cachedAssetsS3.accessKeyId.secretName | quote }}
+      key: API_HF_JWT_ADDITIONAL_PUBLIC_KEYS
```
severo, line 16: is this value correct?

chart/templates/_env/_envCachedAssetsS3.tpl
```diff
@@ -0,0 +1,31 @@
...
+      optional: false
+  {{- else }}
+  value: {{ .Values.secrets.cachedAssetsS3.accessKeyId.value | quote }}
+  {{- end }}
+- name: CACHED_ASSETS_S3_SECRET_ACCESS_KEY
+  {{- if .Values.secrets.cachedAssetsS3.secretAccessKey.fromSecret }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.secrets.cachedAssetsS3.secretAccessKey.secretName | quote }}
+      key: API_HF_JWT_ADDITIONAL_PUBLIC_KEYS
```
severo, line 26: is this value correct?

libs/libapi/src/libapi/config.py
AndreaFrancis: According to what we discussed with @rtrompier, one bucket is enough since applying TTL policy at folder level will be possible.
wdyt?

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: I haven't realized this issue, Don't we have currently the same problem? I think we could probably add the dataset version to the filename wdyt?

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: No, just when working with S3 storage. I put it only here for backward compatibility with assets and /rows and /search for datasets not enabled for s3 storage (Those not included in CACHED_ASSETS_S3_SUPPORTED_DATASETS/all). 
I think we will be able to delete it once everything is migrated to S3.

services/rows/tests/routes/test_rows.py
AndreaFrancis: As discussed with @rtrompier once the CDN is enabled in prod, it will be transparent for us to use the current URL for cached-assets like for example: https://datasets-server.huggingface.co/cached-assets/asoria/image/--/default/train/0/image/image.jpg

> also: to make it easier to manage the response, possibly we need two keys

Not sure if it is a good time to add more fields to the response. How the new `path` field could be useful from the client side?

libs/libapi/src/libapi/config.py
severo: OK! And is it easy to get the size (number of files, total number of bytes) per directory too?

services/rows/tests/routes/test_rows.py
severo: OK. Indeed the path is something internal.

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -3,29 +3,40 @@
...
 DATASETS_SERVER_MDATE_FILENAME = ".dss"
 SUPPORTED_AUDIO_EXTENSION_TO_MEDIA_TYPE = {".wav": "audio/wav", ".mp3": "audio/mpeg"}
 
 
-def create_asset_dir(
+def get_and_create_dir_path(
     dataset: str, config: str, split: str, row_idx: int, column: str, assets_directory: StrPath
-) -> Tuple[Path, str]:
+) -> Path:
     dir_path = Path(assets_directory).resolve() / dataset / DATASET_SEPARATOR / config / split / str(row_idx) / column
```
severo, line 29: why don't we use the url_dir_path (or the function get_url_dir_path)?

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -79,48 +86,74 @@ class AudioSource(TypedDict):
...
+            row_idx=row_idx,
+            column=column,
+            assets_directory=assets_directory,
+        )
+        / filename
+    )
 
-    src = f"{assets_base_url}/{url_dir_path}/{filename}"
+    # create file locally
     asset_file = fn(file_path=file_path, src=src, overwrite=overwrite)  # type: ignore
```
severo, line 140: what is the issue with the type?

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -67,71 +78,102 @@ class ImageSource(TypedDict):
...
+    filename: str,
+    file_path: Path,
+    overwrite: bool = True,
+    s3_client: Optional[S3Client] = None,
+    s3_bucket: Optional[str] = None,
+    s3_folder_name: Optional[str] = None,
+) -> None:
+    if s3_client is not None and s3_bucket is not None:
+        object_key = f"{s3_folder_name}/{url_dir_path}/{filename}"
+        create_object = overwrite or not s3_client.exists_in_bucket(s3_bucket, object_key)
```
severo, line 100: nit: no need for the intermediate variable

libs/libcommon/src/libcommon/viewer_utils/asset.py
lhoestq: No need to fix this is this PR imo. We always do overwrite=True anyway

libs/libapi/src/libapi/config.py
AndreaFrancis: I think so, I was able to see details by folder 
Screenshot from 2023-09-08 08-26-33

chart/values.yaml
AndreaFrancis: I think so https://github.com/huggingface/infra/blob/8978f9832b5b8f2248bb228b6554f89c0af36a83/projects/hub/kubernetes_app_datasets_server.tf#L136

services/api/poetry.lock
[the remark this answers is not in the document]
AndreaFrancis: Yes, I used the last version in main and then applied poetry update libcommon, same for the rest

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: Makes sense, done here https://github.com/huggingface/datasets-server/pull/1726/files#diff-4e03e1cb11067daafd8bc2c042d1e166ec0a88c4538a712e9757d125b5c6afbfR26

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: I removed it

libs/libcommon/src/libcommon/s3_client.py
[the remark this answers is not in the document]
AndreaFrancis: I added some tests here https://github.com/huggingface/datasets-server/pull/1726/files#diff-9fe3368dac6d7fbbaa51515b6c5380980c524df0593a1ea852a598cfa5847b0cR12

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: I think I can create another issue to track this behavior.

> We always do overwrite=True
Notice that I changed to overwrite=False in /rows and /search when working with S3 storage. Is that ok? I mean, I think it could help reduce latency when uploading files. Also, I don't think it will make sense to take advantage of Cloudfront when having overwrite=True right?

libs/libcommon/tests/viewer_utils/test_features.py
AndreaFrancis: I created a couple of tests for assets but just for the new implementation (s3 storage) https://github.com/huggingface/datasets-server/pull/1726/files#diff-1f215a99045381472ee6e586ca88cb287fd8ea1e96ddfdbf3278a4537cdc05efR17
I think I can add more in another PR

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: Mypy does not recognize common parameters for it: `Unexpected keyword argument "file_path" , Unexpected keyword argument "src" and Unexpected keyword argument "overwrite"` errors appeared.
I should have called just `fn(file_path, src, overwrite)` but  as `save_image` and `save_audio` have a different set of parameters, it failed because of a "duplicate" assignation of parameters.
Honesty, I don't know how to properly solve this in mypy.

severo: ok, nevermind

libs/libcommon/src/libcommon/viewer_utils/asset.py
severo: ok

severo (approved):

rtrompier (approved):
LGTM, but we don't have yet a S3 & Cloudfront for Ephemeral Env.
We are currently working on a big refactoring for Terraform projects, but once done, I will try to find a solution for that.

lhoestq (commented):
Awesome ! Mostly one question regarding the S3 base url:

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -68,71 +77,97 @@ class ImageSource(TypedDict):
...
-        config=config,
-        split=split,
-        row_idx=row_idx,
-        column=column,
-        assets_directory=assets_directory,
+    storage_options: Union[DirectoryStorageOptions, S3StorageOptions],
+    fn: Callable[[Path, str, bool], SupportedSource],
+) -> SupportedSource:
+    # get url dir path
+    assets_base_url = storage_options.assets_base_url
```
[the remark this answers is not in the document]
lhoestq: It seems that it's the same as the one that points to the disk, is it expected ? (app_config.cached_assets.base_url)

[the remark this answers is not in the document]
AndreaFrancis: Yes, it will be the same as app_config.cached_assets.base_url (https://datasets-server.huggingface.co/cached-assets).
In this case, the prod DNS is not yet pointing to CloudFront for that URL so probably, the first test dataset images/audios won't be available via the public domain, just using the CloudFront https://d1j3nh2g9qznw6.cloudfront.net/cached-assets/*.
Once we confirm that everything is working as expected infra team will be able to update the Datasets Server DNS record to point on Cloudfront right @rtrompier?
I think for the first tests, we can at least ensure that the images/audio are being stored in S3 and are accessible via CloudFront.

lhoestq (approved):
Looks all good to me ! Let's see how it goes :)

Issue #1729, opened 2023-08-25
/search num_rows_total field is incoherent


severo:
In the following example, num_rows_total=1 (ie. the total number of results for that search) while the `rows` field is an array of 100 rows

https://datasets-server.huggingface.co/search?dataset=loubnabnl/gpt4-1k-annotations&config=default&split=train&query=pokemon&offset=0&limit=100

AndreaFrancis:
Good catch! Thanks for reporting!

AndreaFrancis:
Fixed on https://github.com/huggingface/datasets-server/pull/1730

PR #1730, opened 2023-08-25
fix: Change score alias name in FST query

AndreaFrancis:
Fix for https://github.com/huggingface/datasets-server/issues/1729

### DIFF
--- a/services/search/src/search/routes/search.py
+++ b/services/search/src/search/routes/search.py
@@ -56,13 +56,13 @@ MAX_ROWS = 100
 UNSUPPORTED_FEATURES = [Value("binary"), Audio()]
 
 FTS_COMMAND_COUNT = (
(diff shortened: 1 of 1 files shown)


severo (commented):
Ohhhh OK, it was due to a column being called `score` in the dataset.

Note: in https://datasets-server.huggingface.co/search?dataset=loubnabnl/gpt4-1k-annotations&config=default&split=train&query=pokemon&offset=0&limit=100, the score field looks weird (called: `score:1` instead of `score`). Is it related, and is it fixed by this PR?
Capture d’écran 2023-08-25 à 13 58 03

AndreaFrancis:
I tried adding a new column named "score:1" in duckdb and the fix worked:
Screenshot from 2023-08-25 14-22-27

severo (approved):

PR #1732, opened 2023-08-25
feat: 🎸 create step dataset-hub-cache

severo:
A new step, specific to the Hub (ie. it will not be backward-compatible), to help the Hub have a cache of the informations it needs for each dataset


lhoestq (approved):
cool ! lgtm :)

AndreaFrancis (approved):
LGTM!
It makes me wonder if maybe it is better to handle the data needed for the hub in another environment (maybe another db? and/or a model like a data warehouse?).
We are storing the same operational data with hub query data, could this affect somehow the performance in the future?

severo:
>  It makes me wonder if maybe it is better to handle the data needed for the hub in another environment (maybe another db? and/or a model like a data warehouse?).
> We are storing the same operational data with hub query data, could this affect somehow the performance in the future?

I think it's OK for now, with the SSE that we will implement. cc @julien-c @Pierrci for your information (duplication of the data between moon-landing and datasets-server)

Issue #1710, opened 2023-08-18
Enable Duckdb index on nested texts


lhoestq:
E.g. for dialog datasets with features
```python
Features({
    "conversations": [{"from": Value("string"), "value": Value("string")}]
})
```

like https://huggingface.co/datasets/LDJnr/Puffin

lhoestq:
I just checked and the duckdb FTS index does support nested columns natively so it should be pretty easy to add

PR #1735, opened 2023-08-28
Search for nested data

lhoestq:
Duckdb does support indexing nested data so we're all good on that side already.
Therefore I simply improved the indexable column detection to check for nested data.

The old code used to index all the columns as long as there is at least one non-nested string column though, so we just need to refresh the datasets with the "no indexable columns" error.

Close https://github.com/huggingface/datasets-server/issues/1710

### DIFF
--- a/services/worker/src/worker/job_runners/split/duckdb_index.py
+++ b/services/worker/src/worker/job_runners/split/duckdb_index.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2023 The HuggingFace Authors.
 
+import copy
 import logging
 import os
 from pathlib import Path
@@ -130,8 +131,9 @@ def compute_index_rows(
         column_names = ",".join('"' + column + '"' for column in list(features.keys()))
 
         # look for indexable columns (= possibly nested columns containing string data)
+        # copy the features is needed but will be fixed with https://github.com/huggingface/datasets/pull/6189
         indexable_columns = ",".join(
-            '"' + column + '"' for column in get_indexable_columns(Features.from_dict(features))
+            '"' + column + '"' for column in get_indexable_columns(Features.from_dict(copy.deepcopy(features)))
         )
         if not indexable_columns:
             raise NoIndexableColumnsError("No string columns available to index.")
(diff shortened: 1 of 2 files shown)


lhoestq:
oops forgot to mark as "ready for review"

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -166,7 +180,7 @@ def compute_index_rows(
 
     # TODO: by default, 'porter' stemmer is being used, use a specific one by dataset language in the future
     # see https://duckdb.org/docs/extensions/full_text_search.html for more details about 'stemmer' parameter
-    create_index_sql = CREATE_INDEX_COMMAND.format(columns=column_names)
```
AndreaFrancis, line 169: We were indexing all columns, why it didn't work with nested strings?

AndreaFrancis (approved):

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -112,16 +130,12 @@ def compute_index_rows(
...
-            column
-            for column, feature in features.items()
-            if "dtype" in feature
-            and "_type" in feature
-            and feature["dtype"] == STRING_FEATURE_DTYPE
-            and feature["_type"] == VALUE_FEATURE_TYPE
-        ]
-        if not string_columns:
+        # look for indexable columns (= possibly nested columns containing string data)
+        # copy the features is needed but will be fixed with https://github.com/huggingface/datasets/pull/6189
```
severo, line 134: remove the comment, since https://github.com/huggingface/datasets/pull/6189 has been merged?

severo:
For now, [LDJnr/Puffin](https://huggingface.co/datasets/LDJnr/Puffin) cannot be searched into. I'll refresh it.

severo:
It now works! Congrats @lhoestq

services/worker/src/worker/job_runners/split/duckdb_index.py
lhoestq: will removed when updating `datasets` to a version that includes the fix

severo: oh right, it's in another repository, sorry

PR #1743, opened 2023-08-30
Support audio in rows and search

lhoestq:
Using this code I can run `to_rows_list` in less than 2sec on 100 wav files of 1.5MB (from 20sec before).
I chose to stop converting to WAV and only send the MP3 file to the user, and to parallelize the audio conversions.

This should enable the viewer on audio datasets for both /rows and /search :)
Though to try this feature on /rows I first enabled it on `arabic_speech_corpus` and will enable it on all datasets if it works fine

Note: this can probably be improved later, e.g. by avoiding unnecessary conversions

### DIFF
--- a/services/search/src/search/routes/search.py
+++ b/services/search/src/search/routes/search.py
@@ -13,7 +13,7 @@ from typing import List, Optional, Tuple
 
 import duckdb
 import pyarrow as pa
-from datasets import Audio, Features, Value
+from datasets import Features, Value
 from huggingface_hub import hf_hub_download
 from libapi.authentication import auth_check
 from libapi.exceptions import (
--- a/services/search/src/search/routes/search.py
+++ b/services/search/src/search/routes/search.py
@@ -53,7 +53,7 @@ logger = logging.getLogger(__name__)
 
 ROW_IDX_COLUMN = "__hf_index_id"
 MAX_ROWS = 100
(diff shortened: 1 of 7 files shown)


services/search/src/search/routes/search.py
```diff
@@ -53,7 +53,7 @@
 
 ROW_IDX_COLUMN = "__hf_index_id"
 MAX_ROWS = 100
-UNSUPPORTED_FEATURES = [Value("binary"), Audio()]
+UNSUPPORTED_FEATURES = [Value("binary")]
```
AndreaFrancis, line 56: Will you remove Audio from rows.py?

AndreaFrancis (commented):
I couldn'd see where arabic_speech_corpus is being enabled

services/search/src/search/routes/search.py
lhoestq: I'm keeping Audio() in the /rows UNSUPPORTED_FEATURES so that it's still unsupported in general, except for arabic_speech_corpus

lhoestq:
> I couldn't see where arabic_speech_corpus is being enabled

I didn't need to do any change: in rows.py it already supports Audio() only for this dataset (hardcoded in the variable ALL_COLUMNS_SUPPORTED_DATASETS_ALLOW_LIST)

PR #1751, opened 2023-09-04
Add blockedDatasets variable

lhoestq:
and block `alexandrainst/nota`, see [internal slack thread](https://huggingface.slack.com/archives/C04L6P8KNQ5/p1693548446825959)

The idea is to never process anything for blocked datasets. I also chose to not store anything in the cache in this case

### DIFF
--- a/chart/templates/_env/_envWorker.tpl
+++ b/chart/templates/_env/_envWorker.tpl
@@ -2,6 +2,8 @@
 # Copyright 2022 The HuggingFace Authors.
 
 {{- define "envWorker" -}}
+- name: WORKER_BLOCKED_DATASETS
+  value: {{ .Values.worker.blockedDatasets | quote}}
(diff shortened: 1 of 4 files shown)


severo (commented):
Should we also delete existing cache entries? Maybe in a cronjob, or in a migration every time the block list has changed?

Also: should we remove the other block list (only used in config-parquet-and-info) ?

lhoestq:
I think the cache entries should be deleted yes, but not a big deal for now given that there is only one dataset.

> Also: should we remove the other block list (only used in config-parquet-and-info) ?

let's keep it for now, but ideally yes we should remove it imo

severo:
OK, created two issues: #1753 #1754

lhoestq:
Thanks ! May I merge this one ? I'd like to deploy after this one

severo:
yes, sorry!

Issue #1758, opened 2023-09-04
[split-duckdb-index] OOM for big datasets


lhoestq:
e.g. dataset=c4, config=en, split=train has a partial parquet export of 5GB

There also seems to be more than 12k cache entries with the same error code, over more than 3k datasets. This includes lots of big datasets (c4, oscar, wikipedia and all the variants by users) as well as many image and audio datasets.

lhoestq:
I had to increase the RAM to 29GB to make the indexing work on C4 (en, train).
The resulting duckdb file is 9GB.

Will work on issue https://github.com/huggingface/datasets-server/issues/1891 which should hopefully fix this one

PR #1759, opened 2023-09-04
More memory for workers

lhoestq:
8GB -> 16GB

Fixes (hopefully) https://github.com/huggingface/datasets-server/issues/1758

Idk what kind of pods we have though, we might have to double check it's not too much

It should be enough to perform the indexing of datasets (up to 5GB which is the max)

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -276,10 +276,10 @@ admin:
   resources:
     requests:
       cpu: 1
-      memory: "512Mi"
+      memory: "1Gi"
     limits:
       cpu: 4
-      memory: "4Gi"
+      memory: "8Gi"
 
 hf:
   timeoutSeconds: "1.5"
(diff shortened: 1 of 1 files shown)


severo (approved):
OK! Note that it will double the number of nodes. @rtrompier @XciD @McPatate do you know if there are instances that provide more RAM for a better price?

severo:
for reference: https://github.com/huggingface/infra/blob/main/projects/hub/env/prod-us-east-1.tfvars#L201 (internal)

severo:
As we already have 34 nodes, and the max in https://github.com/huggingface/infra/blob/main/projects/hub/env/prod-us-east-1.tfvars#L201 is 50, it will not fit. We have to reduce the number of pods.

lhoestq:
Reducing the number of pods is fine for now. Later we can maybe dedicate high RAM machines only for export, index and stats jobs. Let me update the PR

severo (approved):

severo:
ah ah https://github.com/huggingface/datasets-server/commit/0430d6d12a25da7b4ca4a05fde73dd8fff0e2d44

PR #1768, opened 2023-09-05
More rows worker

lhoestq:
they were too much spammed and were OOMing a lot this morning


albertvillanova:
Spam?! How do you know that?

lhoestq:
Actually not sure it was spam, from the logs I don't see an activity that is that big

PR #1770, opened 2023-09-05
Less rows workers and less uvicorn workers

lhoestq:
Following https://github.com/huggingface/datasets-server/pull/1769

since deployed failed again because there are no nodes available. Sureley because /rows can't run on nodes made for dataset workers.

I also reduced the amount of uvicorn workers in case it was the reason the pods were OOMing (too many parallel calls)


severo:
Yes, the pods created for them are in a different pool of nodes (machines). We have one pool for workers, one for light workers, one for search, one for rows, and one for the rest, I think. See https://github.com/huggingface/infra/blob/main/projects/hub/env/prod-us-east-1.tfvars (internal)

Issue #1772, opened 2023-09-05
/rows OOMs when the arrow data are too big


lhoestq:
it should raise an error instead, to not impact the other dataset pages

Issue #1527, opened 2023-07-18
/rows: raise an error if a dataset has too big row groups


lhoestq:
It can happen if a dataset was converted to parquet before the recent row group size optimization, e.g. garythung/trashnet

Currently it makes the worker crash.

We could also refresh the parquet export of the dataset when this happens

severo:
See https://github.com/huggingface/datasets-server/pull/1520#issuecomment-1640684785

> Could be, but instead of complicating the code for that, maybe do a one-time analysis on all the datasets that have a size over ...GB

PR #1773, opened 2023-09-05
Set maximum allowed arrow data for /rows to avoid OOM

lhoestq:
Return an error if the Arrow data to load from the Parquet row groups are bigger than 300MB

This impacts /rows and the first-rows-from-parquet job

Fix https://github.com/huggingface/datasets-server/issues/1772
Fix https://github.com/huggingface/datasets-server/issues/1527


chart/templates/services/rows/_container.tpl
```diff
@@ -29,6 +29,8 @@
     value: {{ .Values.rows.uvicornNumWorkers | quote }}
   - name: API_UVICORN_PORT
     value: {{ .Values.rows.uvicornPort | quote }}
+  - name: ROWS_INDEX_MAX_ARROW_DATA_IN_MEMORY
```
severo, line 32: we need it in the worker container too

severo: also: in the two docker-compose files + possibly in the .github CI files

libs/libcommon/src/libcommon/parquet_utils.py
```diff
@@ -139,9 +157,24 @@ def query(self, offset: int, length: int) -> pa.Table:
...
             )
 
+        with StepProfiler(
+            method="parquet_index_with_metadata.row_groups_size_check", step="check if the rows can fit in memory"
+        ):
+            row_groups_size = sum(
+                [row_group_readers[i].read_size() for i in range(first_row_group_id, last_row_group_id + 1)]
+            )
+            if row_groups_size > self.max_arrow_data_in_memory:
+                raise TooBigRows(
```
severo, line 167: maybe mention in the `query` function docstring that it can raise this exception, and when

services/worker/src/worker/job_runners/split/first_rows_from_parquet.py
```diff
@@ -192,6 +192,7 @@ def __init__(
             httpfs=HTTPFileSystem(headers={"authorization": f"Bearer {self.app_config.common.hf_token}"}),
             unsupported_features=[],
             all_columns_supported_datasets_allow_list="all",
+            max_arrow_data_in_memory=app_config.rows_index.max_arrow_data_in_memory,
```
severo, line 195: catch the TooBigRows exception and raise a CacheableError instead, so that we have an X-Error-Code header.

severo:
feel free to deploy tomorrow!

PR #1778, opened 2023-09-06
feat: 🎸 use EFS instead of NFS for cached assets

severo:
Depends on https://github.com/huggingface/infra/pull/680 being deployed first (internal).


severo:
merging, and trying on staging

PR #1784, opened 2023-09-06
Add /sse/hub-cache to update the Hub's backend cache

severo:
TODO:

- [x] setup replicaset in mongo server, because it's required to be able to watch the collection changes
- [x] feature: send an SSE on every update of a `dataset-hub-cache` cache entry
- [x] feature: ^ only if some value has changed (?)
- [x] feature: On the first call to /sse/hub-cache, emit one SSE per `dataset-hub-cache` cache entry (initial synchronization)
- [x] <strike>document in OpenAPI (no need to add to the docs, as it's internal)</strike> -> No: it's not supported by the OpenAPI specification (see https://github.com/OAI/OpenAPI-Specification/issues/396)
- [x] add sseApi in https://github.com/huggingface/infra-deployments/blob/main/apps/datasets-server/datasets-server-prod.yaml and https://github.com/huggingface/infra-deployments/blob/main/apps/datasets-server/datasets-server-dev.yaml (internal)?
- [x] document the usage of /sse/hub-cache in `services/sse-api/README.md` since it's not possible to do so in the OpenAPI spec

Maybe:
- [ ] add e2e tests
- [ ] monitor the overhead that it generates?
- [ ] How to make features 1 and 3 work well together? What if a `dataset-hub-cache` cache entry is updated in the middle of the initial synchronization SSE? How do we test this?
- [ ] protect the endpoint so only the Hub can request it?


severo:
This video shows a client (Postman) that listens to the Server-Sent Events on the new endpoint `/sse/hub-cache` during the e2e tests.

The `operation` field helps have an idea of what is happening:
- `insert` when the cache entry has been created
- `update` when an existing cache entry has been modified (i.e.. updated and the relevant contents have changed)
- `delete`(not showcased in the e2e tests) when an existing cache entry is removed
- `init`: special case, just after the request, where all the existing cache entries are sent as SSE as well.

https://github.com/huggingface/datasets-server/assets/1676121/6bc3a2be-7915-42e1-9ba2-5b3b86322efe

severo:
cc @Kakulukian @julien-c: is it what we expect from the moon-landing side?

Kakulukian:
> cc @Kakulukian @julien-c: is it what we expect from the moon-landing side?

it sounds like it, but I'm not seeing the init event in your video

severo:
It is present. When I stop the client, and launch the request again, we get an init event, because we had one dataset in the cache at that point.

.github/workflows/s-sse-api.yml
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 The HuggingFace Authors.
```
AndreaFrancis, line 2:
```suggestion
# Copyright 2023 The HuggingFace Authors.
```

severo:
OK, merging

chart/templates/services/sse-api/_container.tpl
```diff
@@ -0,0 +1,45 @@
...
+- name: "{{ include "name" . }}-sse-api"
+  image: {{ include "services.sseApi.image" . }}
+  imagePullPolicy: {{ .Values.images.pullPolicy }}
+  env:
+  {{ include "envCache" . | nindent 2 }}
+  {{ include "envQueue" . | nindent 2 }}
+  {{ include "envCommon" . | nindent 2 }}
+  {{ include "envHf" . | nindent 2 }}
+  {{ include "envLog" . | nindent 2 }}
+  {{ include "envNumba" . | nindent 2 }}
```
AndreaFrancis, line 14: Do we need numba config here?

chart/templates/services/sse-api/_container.tpl
```diff
@@ -0,0 +1,45 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2023 The HuggingFace Authors.
+
+{{- define "containerSseApi" -}}
+- name: "{{ include "name" . }}-sse-api"
+  image: {{ include "services.sseApi.image" . }}
+  imagePullPolicy: {{ .Values.images.pullPolicy }}
+  env:
+  {{ include "envCache" . | nindent 2 }}
+  {{ include "envQueue" . | nindent 2 }}
```
AndreaFrancis, line 10: Do we use queue here? I saw only cache resource is initialized in the main app

services/sse-api/src/sse_api/watcher.py
```diff
@@ -0,0 +1,208 @@
...
+    async def _init_loop(self, suscriber: str) -> None:
+        """
+        publish an event for each initial dataset-hub-cache cache entry.
+
+        TODO: we don't want to send to all the suscribers
+        """
+
+        async for document in self._collection.find(
+            filter={"kind": HUB_CACHE_KIND},
+            projection={"dataset": 1, "content": 1, "http_status": 1},
```
AndreaFrancis, line 130: Did you validate that this kind of query has a matching index in the collection?

services/sse-api/Makefile
```diff
@@ -0,0 +1,20 @@
+# environment variables for the commands (docker compose, poetry)
+export COMPOSE_PROJECT_NAME := sse-api
+export MONGO_PORT := 27035
+export CACHE_MONGO_URL := mongodb://localhost:${MONGO_PORT}
+export QUEUE_MONGO_URL := mongodb://localhost:${MONGO_PORT}
```
AndreaFrancis, line 5: Same here, do we use this config for sse? I mean, it only connects to the cache collection right?

AndreaFrancis (commented):
Great! I just put a couple of comments related to queue config in SSE. I think it is not needed since it will just use cache collection right?

services/sse-api/src/sse_api/watcher.py
severo: I think so, because three indexes start with `kind`, which is the only field used in the filter, and I think mongo can always sort by `_id`, right?

https://github.com/huggingface/datasets-server/blob/049ec7af688dbbf7ec56c3d9a779a8a8680b013a/libs/libcommon/src/libcommon/simple_cache.py#L113-L118

Anyway, I'm interested in your analysis

Issue #1255, opened 2023-05-25
Fix audio data in pagination of audio datasets


lhoestq:
Currently pagination is only enabled for testing purposes on [arabic_speech_corpus](https://huggingface.co/datasets/arabic_speech_corpus) but times out because the "transform to list" step that writes the audio files to disk takes too much time.

Currently it writes both MP3 and WAV - but we should find which one is faster and only write this one.

This is not super high prio for now though, since there aren't a lot of audio dataset with pagination

severo:
keep open

severo:
keep open

lhoestq:
So it's technically fixed but not enabled for all the datasets yet because most of them need to be converted to parquet (job failed because of OOM). Will re-run them once #1759 is deployed

severo:
> Will re-run them once https://github.com/huggingface/datasets-server/pull/1759 is deployed

Done

lhoestq:
Awesome ! Though after more digging I found that while the endpoint does work now, it's still too slow: 10sec to convert the audio data. We'll have to find a more clever way. For example I can simply not convert anything (if it's wav, keep it as wav - if it's mp3, keep it as mp3)

severo:
Done.

PR #1788, opened 2023-09-07
Don't convert mp3 or wav

lhoestq:
Following https://github.com/huggingface/datasets-server/pull/1743

Fixes (hopefully) #1255 if the speed is finally good enough.


severo (approved):
Nice!!!

PR #1791, opened 2023-09-08
feat: 🎸 disable the viewer for org CyberHarem

severo:


lhoestq:
Do we really want to block it ? We jut send them a message to ask the to add the NFAA tag so maybe let's just wait for their response first

lhoestq:
Ah, just saw the slack thread - as you prefer

severo:
What if we unblock them once they added the tag?

severo:
slack thread (internal): https://huggingface.slack.com/archives/C042H6N5JR0/p1694159493341069

lhoestq:
> What if we unblock them once they add the tag?

ok, I'll let you know

giadilli:
Let's give the authors 48h to respond and then eventually we will disable the viewer. Thanks everyone

lhoestq:
ok, reverting the block for now

Issue #1443, opened 2023-06-28
Raise specific errors (and error_code) instead of UnexpectedError


severo:
The following query on the production database gives the number of datasets with at least one cache entry with error_code "UnexpectedError", grouped by the underlying "cause_exception". 

For the most common ones (`DatasetGenerationError`, `HfHubHTTPError`, `OSError`, etc.) we would benefit from raising a specific error with its error code. It would allow to:
- retry automatically, if needed
- show an adequate error message to the users
- even: adapt the way we show the dataset viewer on the Hub

`null` means it has no `details.cause_exception`. These cache entries should be inspected more closely. See https://github.com/huggingface/datasets-server/issues/1123 in particular, which is one of the cases where no cause exception is reported.

```
db.cachedResponsesBlue.aggregate([
    {$match: {error_code: "UnexpectedError"}},
    {$group: {_id: {cause: "$details.cause_exception", dataset: "$dataset"}, count: {$sum: 1}}},
    {$group: {_id: "$_id.cause", count: {$sum: 1}}},
    {$sort: {count: -1}}
])
{ _id: 'DatasetGenerationError', count: 1964 }
{ _id: null, count: 1388 }
{ _id: 'HfHubHTTPError', count: 1154 }
{ _id: 'OSError', count: 433 }
{ _id: 'FileNotFoundError', count: 242 }
{ _id: 'FileExistsError', count: 198 }
{ _id: 'ValueError', count: 186 }
{ _id: 'TypeError', count: 160 }
{ _id: 'ConnectionError', count: 146 }
{ _id: 'RuntimeError', count: 86 }
{ _id: 'NonMatchingSplitsSizesError', count: 83 }
{ _id: 'FileSystemError', count: 62 }
{ _id: 'ClientResponseError', count: 52 }
{ _id: 'ArrowInvalid', count: 45 }
{ _id: 'ParquetResponseEmptyError', count: 43 }
{ _id: 'RepositoryNotFoundError', count: 41 }
{ _id: 'ManualDownloadError', count: 39 }
{ _id: 'IndexError', count: 28 }
{ _id: 'AttributeError', count: 16 }
{ _id: 'KeyError', count: 15 }
{ _id: 'GatedRepoError', count: 13 }
{ _id: 'NotImplementedError', count: 11 }
{ _id: 'ExpectedMoreSplits', count: 9 }
{ _id: 'PermissionError', count: 8 }
{ _id: 'BadRequestError', count: 7 }
[...]

PR #1797, opened 2023-09-11
fix: split-descriptive-statistics fails for datasets with list features

AndreaFrancis:
While reviewing https://github.com/huggingface/datasets-server/issues/1443 I found that for the first error AttributeError, there is a minor bug in split-descriptive-statistics, it fails when features include list (sequences).
After this fix, we should force-refresh cached records for:
```
db.cachedResponsesBlue.countDocuments({error_code:"UnexpectedError", "details.copied_from_artifact":{$exists:false}, "details.cause_exception":"AttributeError",kind:"split-descriptive-statistics"})
9748

```


polinaeterna (approved):
awesome, thank you for catching this!

PR #1805, opened 2023-09-14
Replace deprecated typing types

albertvillanova:
Since Python 3.9 (this project is in Python 3.9.15), some types defined in `typing` module are deprecated.

This PR replaces deprecated `typing` types.


severo (approved):
OK, good to know!

nit: is there a way to check this with `make quality`? I don't think we ever want to check this by hand again.

PR #1807, opened 2023-09-14
Validate job runner

lhoestq:
We should be able to create an invalid job runners (e.g. for tests but most importantly because we need to do it to kill zombie jobs). Therefore I moved the check_config_exists() validation check to a dedicated `.validate()` method that is called only when the job will be run.

This should fix the kill_zombies failures that make all our workers crash in loop now.

### DIFF
--- a/services/worker/src/worker/job_manager.py
+++ b/services/worker/src/worker/job_manager.py
@@ -105,6 +105,7 @@ class JobManager:
 
     def run_job(self) -> JobResult:
         try:
+            self.job_runner.validate()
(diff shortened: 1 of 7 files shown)


severo (approved):
Oh, ok, it makes sense. Thanks

AndreaFrancis:
I don't quite understand why the previous changes didn't break the executor tests. I mean, test_executor_stops_on_long_job was supposed to simulate a long job right? Also, not sure why the checks could make the zombie killer fail because it only validates that the parent contains the split and config (that should always exist).

PR #1816, opened 2023-09-18
Compute descriptive statistics for string type

polinaeterna:
- treat as category if num of unique values <= 40
- compute numerical stats over string length otherwise
UPD: decreased the number from 40 to 30, any objections? :))


services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -37,30 +37,40 @@
 INTEGER_DTYPES = ["int8", "int16", "int32", "int64", "uint8", "uint16", "uint32", "uint64"]
 FLOAT_DTYPES = ["float16", "float32", "float64"]
 NUMERICAL_DTYPES = INTEGER_DTYPES + FLOAT_DTYPES
+STRING_DTYPES = ["string"]  # , "large_string"] ?
```
severo, line 40: maybe allow large_string too here (adding a test too)

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -244,11 +254,51 @@ def compute_categorical_statistics(
...
 
+def compute_string_statistics(
+    con: duckdb.DuckDBPyConnection,
+    column_name: str,
+    n_bins: int,
+    n_samples: int,
+    table_name: str,
+) -> Union[CategoricalStatisticsItem, NumericalStatisticsItem]:
+    (n_unique,) = con.sql(f"SELECT COUNT(DISTINCT {column_name}) FROM {table_name};").fetchall()[0]  # nosec
+    if n_unique <= 40:
```
severo, line 270: nit: use an intermediate CONSTANT, defined in the first lines of the file

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -244,11 +254,51 @@ def compute_categorical_statistics(
...
+        frequencies: dict[str, int] = dict(con.sql(categorical_counts_query).fetchall())
+        nan_count = frequencies.pop(None, 0)  # type: ignore
+        nan_proportion = np.round(nan_count / n_samples, DECIMALS).item() if nan_count != 0 else 0.0
+        return CategoricalStatisticsItem(
+            nan_count=nan_count,
+            nan_proportion=nan_proportion,
+            n_unique=len(frequencies),
+            frequencies=frequencies,
+        )
+    # compute numerical stats over string lengths (min, max, ..., hist)
```
severo, line 284: smart: we re-use the code for numerical statistics.

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -386,7 +441,27 @@ def compute_descriptive_statistics_response(
     con.sql("INSTALL httpfs")
     con.sql("LOAD httpfs")
     con.sql("SET enable_progress_bar=true;")
+    con.sql(f"CREATE TABLE {DATA_TABLE_NAME} as SELECT * FROM read_parquet('{local_parquet_glob_path}');")  # nosec
```
polinaeterna, line 444: this takes more RAM but should make the querying faster

severo: OK! I was wondering why using this. Will it break our workers on big datasets (GBs or TBs of parquet files)?

severo: Or are the statistics always computed on the first 5GB anyway?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -386,7 +441,27 @@ def compute_descriptive_statistics_response(
...
 
+    if string_features:
+        logging.info(f"Compute statistics for categorical columns {categorical_features}")
```
severo, line 447:
```suggestion
        logging.info(f"Compute statistics for string columns {string_features}")
```

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: I think it's OK if hardcoded for now

polinaeterna:
also i realized why it would be helpful to have feature as another level of granularity - the way it works now we will have to recompute all the datasets, just to add one feature. any ideas how to mitigate this? @severo @AndreaFrancis 
@lhoestq suggested to update instead of rewrite the result if it exists in cache but force recomputing also must be supported

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: (not for this PR): if we limit to the first 5GB, and if we use duckdb, and if we already have a duckdb file for the first 5GB (used for search and filter), why not using this file instead of the parquet files?

severo:
> also i realized why it would be helpful to have feature as another level of granularity - the way it works now we will have to recompute all the datasets, just to add one feature. any ideas how to mitigate this? @severo @AndreaFrancis

yes, we could add the column/feature level to dataset/config/split. I think we had some discussion already, but couldn't find it. Maybe create a dedicate issue for that.

> @lhoestq suggested to update instead of rewrite the result if it exists in cache but force recomputing also must be supported

hmmm, due to the imbalance between strings and other types, I would say the overhead will not be significant on this PR (we will recompute all the datasets anyway)

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -244,11 +254,51 @@ def compute_categorical_statistics(
...
 
 
+def compute_string_statistics(
+    con: duckdb.DuckDBPyConnection,
+    column_name: str,
+    n_bins: int,
+    n_samples: int,
+    table_name: str,
+) -> Union[CategoricalStatisticsItem, NumericalStatisticsItem]:
+    (n_unique,) = con.sql(f"SELECT COUNT(DISTINCT {column_name}) FROM {table_name};").fetchall()[0]  # nosec
```
AndreaFrancis, line 269: Maybe convert the SQL command a constant too

AndreaFrancis: Same for all the SQL we use everywhere

services/worker/src/worker/job_runners/split/descriptive_statistics.py
AndreaFrancis: We only create the duckdb index file if there are string columns

AndreaFrancis: We only create the duckdb index file if there are string columns

polinaeterna: i thought an index file is a file with search index, not with dataset data, no?

AndreaFrancis: > i thought an index file is a file with search index, not with dataset data, no?

In our case, it has the data + the index

polinaeterna: > Or are the statistics always computed on the first 5GB anyway?

ah actually only on the datasets < 5gb in size

lhoestq:
> hmmm, due to the imbalance between strings and other types, I would say the overhead will not be significant on this PR (we will recompute all the datasets anyway)

Yea now it will compute stats for literally all the datasets ! 🤯

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: > In our case, it has the data + the index

ah okay. but only string columns data, right?

lhoestq (approved):
Awesome ! Nothing to add to the previous comments

Let's monitor the number of crashed due to OOM when this is deployed.

Let's also increment the job version ?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -386,7 +441,27 @@ def compute_descriptive_statistics_response(
     con.sql("INSTALL httpfs")
```
lhoestq, line 441: do we have to free the memory at the end of the job by deleting the table that was created ?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -374,10 +424,15 @@ def compute_descriptive_statistics_response(
...
+        if isinstance(feature, dict) and feature.get("_type") == "Value" and feature.get("dtype") in STRING_DTYPES
+    }
+    if not categorical_features and not numerical_features and not string_features:
         raise NoSupportedFeaturesError(
             "No columns for statistics computation found. Currently supported feature types are: "
-            f"{NUMERICAL_DTYPES} and ClassLabel. "
+            f"{NUMERICAL_DTYPES}, {STRING_DTYPES} and ClassLabel. "
         )
 
     con = duckdb.connect(":memory:")  # we don't load data in local db file, use local parquet file instead
```
lhoestq, line 438: do we have to free the memory at the end of the job by deleting the table that was created ?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
lhoestq: no, all of them

services/worker/src/worker/job_runners/split/descriptive_statistics.py
lhoestq: actually everything is cleared when con.close() is called so we're all good

polinaeterna: no, since it's in-memory, it's freed automatically when connection is closed.

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: as you wish

severo (approved):

PR #1821, opened 2023-09-19
feat: 🎸 add HIGH priority level

severo:
And set the "refresh" jobs to HIGH priority in the admin space.

This way, the manual actions we do at https://huggingface.co/spaces/datasets-maintainers/datasets-server-admin-ui are effective more quickly.


severo:
Merging, as it's simple

Issue #1285, opened 2023-06-01
Create a cronjob to delete dangling cache entries


severo:
from the cache database

related to https://github.com/huggingface/datasets-server/issues/1284

severo:
keep open

AndreaFrancis:
See https://github.com/huggingface/datasets-server/issues/1550#issuecomment-1707471830
There were some datasets that could not be backfilled because they didn't exist on the hub. We should remove those as well as their related assets.

AndreaFrancis:
Firs part was implemented here: https://github.com/huggingface/datasets-server/pull/1822 for datasets that no longer exist in the hub.
After calling DELETE /admin/obsolete-cache, 3920 datasets were deleted with 86780 cache records.
See the detailed list in the response 
[obsolete_cache.csv](https://github.com/huggingface/datasets-server/files/12702277/obsolete_cache.csv)

severo:
The backfill cron job now loops on the cache entries, so: it will delete all the obsolete cache entries. Done

Issue #1284, opened 2023-06-01
Create a cron job to clean the dangling assets and cached assets


severo:
Related to https://github.com/huggingface/datasets-server/issues/1122

severo:
keep open

severo:
As we now use signed URLs, only the assets referenced by a cache entry will be accessible. The other ones, if any, will "just" waste space on S3

severo:
Also: the files on S3 should now be deleted when we delete a dataset (see https://github.com/huggingface/datasets-server/blob/c22f563f350a36148a1deaaf12531d4da991a677/libs/libcommon/src/libcommon/orchestrator.py#L678)

PR #1822, opened 2023-09-19
Delete cache and assets for non existent datasets

AndreaFrancis:
Related to https://github.com/huggingface/datasets-server/issues/1285 and https://github.com/huggingface/datasets-server/issues/1284
This PR will fix only the obsolete cache by non existent datasets, the ones related to obsolete configs/splits will be treated in another PR


chart/templates/cron-jobs/cache-cleaner/job.yaml
```diff
@@ -0,0 +1,29 @@
...
+          {{- include "dnsConfig" . | nindent 10 }}
+          {{- include "image.imagePullSecrets" . | nindent 6 }}
+          nodeSelector: {{ toYaml .Values.cacheCleaner.nodeSelector | nindent 12 }}
+          tolerations: {{ toYaml .Values.cacheCleaner.tolerations | nindent 12 }}
+          containers: {{ include "containerCacheCleaner" . | nindent 12 }}
+          securityContext: {{ include "securityContext" . | nindent 12 }}
+          initContainers: 
+            {{ include "initContainerAssets" . | nindent 12 }}
+            {{ include "initContainerCachedAssets" . | nindent 12 }}
+          volumes: {{ include "volumeDuckDBIndex" . | nindent 12 }}
```
severo, line 28: it should be assets and cached assets instead

jobs/cache_maintenance/tests/test_clean_cache.py
```diff
@@ -0,0 +1,71 @@
...
+            hf_token="hf_token",
+            assets_directory=assets_directory,
+            cached_assets_directory=cached_assets_directory,
+        )
+
+    assert asset_file.is_file()
+    assert cached_asset_file.is_file()
+    assert has_some_cache(dataset=dataset)
+
+    with patch("cache_maintenance.cache_cleaner.get_supported_dataset_infos", return_value=[]):
```
severo, line 59: This part frightens me: if the code does the same, for some reason, the following part returns an empty list, we will delete everything:

https://github.com/huggingface/datasets-server/blob/cbd38812d1d08da24f10ce6501e4c520687b2faf/libs/libcommon/src/libcommon/dataset.py#L172

We could add a safeguard: cancel the action if we get an empty list. Even a bit safer: cancel if the list of datasets is under a threshold (`20,000`, for example). Imagine an issue with the pagination that would return only 10,000 results.

lhoestq:
I think I'd add a step that requires a human to validate the deletion instead of a cron job. This way we can show what will be deleted before approving the deletion.

Could it be part of the admin app maybe ?

chart/templates/cron-jobs/cache-cleaner/job.yaml
AndreaFrancis: Thanks, I fixed it on https://github.com/huggingface/datasets-server/pull/1822/commits/819c0cd20e8b99df071842f5f8d2047a6d5832a8

jobs/cache_maintenance/tests/test_clean_cache.py
AndreaFrancis: Sure, I added a validation on https://github.com/huggingface/datasets-server/pull/1822/commits/819c0cd20e8b99df071842f5f8d2047a6d5832a8

AndreaFrancis:
> Could it be part of the admin app maybe ?

Ok, so maybe two new endpoints would be required on admin service?
one to show the detailed list and number of estimated cached records to delete and a second one to execute the deletion?

jobs/cache_maintenance/src/cache_maintenance/cache_cleaner.py
```diff
@@ -0,0 +1,59 @@
...
+
+import logging
+from typing import Optional
+
+from libcommon.dataset import get_supported_dataset_infos
+from libcommon.simple_cache import delete_dataset_responses, get_all_datasets
+from libcommon.storage import StrPath
+from libcommon.viewer_utils.asset import delete_asset_dir
+
+MINIMUM_SUPPORTED_DATASETS = 20000
```
severo, line 12: nit

```suggestion
MINIMUM_SUPPORTED_DATASETS = 20_000
```

jobs/cache_maintenance/tests/test_clean_cache.py
```diff
@@ -0,0 +1,70 @@
...
+from huggingface_hub.hf_api import DatasetInfo
+from libcommon.simple_cache import has_some_cache, upsert_response
+
+from cache_maintenance.cache_cleaner import clean_cache
+
+
+@pytest.mark.parametrize(
+    "dataset_infos,minimun_supported_datasets,should_keep",
+    [
+        ([DatasetInfo(id="dataset")], 1, True),  # do not delete, dataset it is still supported
```
severo, line 20:
```suggestion
        ([DatasetInfo(id="dataset")], 1, True),  # do not delete, dataset is still supported
```

severo:
> Ok, so maybe two new endpoints would be required on admin service?

indeed. And it we create a cronjob, it will call the admin endpoints.

severo:
In my review, I also wondered how we would handle the concurrency. I think the function will take a long time to proceed, in particular the first one, and it is possible for a dataset with the same name to be created at the same time. Its cached entries and assets could be deleted.

As the probability seems very low, and as it would be fixed by the daily backfill anyway, I assume it's a detail that we can ignore, right?

AndreaFrancis:
> As the probability seems very low, and as it would be fixed by the daily backfill anyway, I assume it's a detail that we can ignore, right?

I think so, anyway we do the backfill when some API request does not find the cache right?

lhoestq:
> In my review, I also wondered how we would handle the concurrency. I think the function will take a long time to proceed, in particular the first one, and it is possible for a dataset with the same name to be created at the same time. Its cached entries and assets could be deleted.

Maybe we only delete the cache responses if they haven't changed ? And do it by batch of e.g. 10k ?

in pseudo code

```python

job_start_time = get_datetime()
existing_datasets = ...
supported_datasets = ...
datasets_to_delete = ...
# wait for validation from 
CachedResponseDocument.objects(dataset=dataset, updated_at__leq=job_start_time).delete()

**Comment by lhoestq** [comment]
Ah we might not be able to get the list of datasets that were effectively removed that way - please ignore if it's too complicated.

As sylvain mentioned the backfill would do its job in this case

**Review by lhoestq — APPROVED** [review]
Looks all good !

File: services/admin/src/admin/routes/obsolete_cache.py [hunk 5]
```diff
@@ -0,0 +1,140 @@
...
+) -> tuple[int, int]:
+    supported_dataset_names = get_supported_dataset_names(hf_endpoint=hf_endpoint, hf_token=hf_token)
+    if len(supported_dataset_names) < MINIMUM_SUPPORTED_DATASETS:
+        raise UnexpectedError(f"only {len(supported_dataset_names)} datasets were found")
+
+    existing_datasets = get_all_datasets(DATASET_CONFIG_NAMES_KIND)
+    datasets_to_delete = existing_datasets.difference(supported_dataset_names)
+    deleted_cache_records = 0
+
+    for dataset in datasets_to_delete:
```
lhoestq, line 96: (nit) Maybe add a log.info that mentions the dataset being deleted ?

services/admin/src/admin/routes/obsolete_cache.py
```diff
@@ -0,0 +1,140 @@
...
+    supported_dataset_infos = get_supported_dataset_infos(hf_endpoint=hf_endpoint, hf_token=hf_token)
+    return {dataset_info.id for dataset_info in supported_dataset_infos}
+
+
+def get_obsolete_cache(
+    hf_endpoint: str,
+    hf_token: Optional[str] = None,
+) -> list[DatasetCacheReport]:
+    supported_dataset_names = get_supported_dataset_names(hf_endpoint=hf_endpoint, hf_token=hf_token)
+    existing_datasets = get_all_datasets(DATASET_CONFIG_NAMES_KIND)
```
severo, line 48: why do we filter on `kind="dataset-config-names"`? To get a faster result? We might miss some obsolete cache entries

services/admin/src/admin/routes/obsolete_cache.py
severo: and maybe even return datasets_to_delete, instead of just the number -> to show in the admin UI

services/admin/src/admin/app.py
```diff
@@ -130,6 +139,31 @@ def create_app() -> Starlette:
...
+                hf_endpoint=app_config.common.hf_endpoint,
+                max_age=app_config.admin.max_age,
+                assets_directory=assets_directory,
+                cached_assets_directory=cached_assets_directory,
+                external_auth_url=app_config.admin.external_auth_url,
+                organization=app_config.admin.hf_organization,
+                hf_timeout_seconds=app_config.admin.hf_timeout_seconds,
+                hf_token=app_config.common.hf_token,
+            ),
+            methods=["DELETE"],
```
severo, line 165: nice

severo (commented):
Cool. I understand that the integration into admin UI is still missing. Do you want to do it here or in another PR?

services/admin/src/admin/routes/obsolete_cache.py
AndreaFrancis: Yes it was because of performance. Do you think is better to run over the whole collection?

severo: I don't know. It depends on the difference in performance and in the results

AndreaFrancis: Done

services/admin/src/admin/routes/obsolete_cache.py
AndreaFrancis: Ok, I kept the same response format as GET /obsolete-cache to have detailed information about which datasets and how many cached records were deleted for each of them.

AndreaFrancis:
> Cool. I understand that the integration into admin UI is still missing. Do you want to do it here or in another PR?

Yes! I would like to send the UI code in another PR

PR #1832, opened 2023-09-21
unblock cyberharem

lhoestq:
they're now using the NFAA tag.

I'll also delete the cache entries with `DatasetInBlockListError `


severo:
> I'll also delete the cache entries with DatasetInBlockListError 

just for that dataset, right?

lhoestq:
> just for that dataset, right?

yes

lhoestq:
deleted 2k cache entries

PR #1834, opened 2023-09-21
Obsolete cache Tab in admin UI

AndreaFrancis:
Adding tab for obsolete cache actions (get list and delete)
Screenshot from 2023-09-21 13-41-11

### DIFF
--- a/front/admin_ui/app.py
+++ b/front/admin_ui/app.py
@@ -401,6 +401,8 @@ The cache is outdated or in an incoherent state. Here is the plan to backfill th
     dataset_status_button.click(get_dataset_status_and_backfill_plan, inputs=[token_box, dataset_name], outputs=[cached_responses_table, jobs_table, backfill_message, backfill_plan_table, backfill_execute_button, backfill_execute_error])
     backfill_execute_button.click(execute_backfill_plan, inputs=[token_box, dataset_name], outputs=[cached_responses_table, jobs_table, backfill_message, backfill_plan_table, backfill_execute_button, backfill_execute_error])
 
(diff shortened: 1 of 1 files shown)


front/admin_ui/app.py
```diff
@@ -261,6 +315,51 @@ def query_jobs(pending_jobs_query):
             return {pending_jobs_query_result_df: gr.update(value=pd.DataFrame({"Error": [f"❌ {str(error)}"]}))}
         return {pending_jobs_query_result_df: gr.update(value=result)}
 
+    def refresh_dataset(
```
severo, line 318: I don't understand: is it an unintended copy/paste?

front/admin_ui/app.py
```diff
@@ -128,6 +136,52 @@ def auth(token):
                 auth_error: gr.update(value=f"❌ Unauthorized (user '{user['name']} is not a member of '{ADMIN_HF_ORGANIZATION}')")
             }
 
+    def call_obsolete_cache(token, delete):
+        headers = {"Authorization": f"Bearer {token}"}
+        obsolete_cache_endpoint = f"{DSS_ENDPOINT}/admin/obsolete-cache"
+        response = (
+            requests.delete(obsolete_cache_endpoint, headers=headers, timeout=240)
```
severo, line 143: it's a very big timeout. As it's an admin endpoint, I think it's OK to just set a large value instead of trying to paginate the results, for example (not worth it)

severo (approved):
One detail to fix

albertvillanova (approved):
Thanks.

front/admin_ui/app.py
AndreaFrancis: Sorry, it was a mistake, I fixed it.

Issue #1694, opened 2023-08-16
Load a parquet export with `pyarrow.parquet.ParquetDataset`


lhoestq:
`read()` fails because it tries to load the `index.duckdb` file as a parquet file

```python
from huggingface_hub import HfFileSystem
import pyarrow.parquet as pq

ds = pq.ParquetDataset("datasets/squad@~parquet", filesystem=HfFileSystem()).read()
```

raises

```
ArrowInvalid: Could not open Parquet input source 'datasets/squad@~parquet/plain_text/train/index.duckdb': Parquet magic bytes not found in footer. Either the file is corrupted or this is not a parquet file.
```

Shall we rename the `index.duckdb` file ? I think we can try renaming it to `_index.duckdb` or `.index.duckdb`

-----------------------

Also `read_pandas()` also fails (this is what pandas uses when calling `pd.read_parquet`)

```python
from huggingface_hub import HfFileSystem
import pyarrow.parquet as pq

df = pq.ParquetDataset("datasets/squad@~parquet", filesystem=HfFileSystem()).read_pandas()
```

raises

```
EntryNotFoundError: 404 Client Error. (Request ID: Root=1-64dd05ce-3f58691a7b2e8fd67c76425b)

Entry Not Found for url: https://huggingface.co/api/datasets/squad/tree/~parquet/_common_metadata?expand=True.
Tree Entry not found: _common_metadata
```

severo:
Or switch to refs/convert/duckdb

lhoestq:
What would be the best way to move the duckdb indexes (to a new branch or a new directory) ?

Shall we increment the version of the duckdb-index job with the new location and implement an "update" step in case the index is already there just to move it ?

severo:
First: do we want to move/rename the file inside the same branch (`refs/convert/parquet`) or move to a different branch (`refs/convert/duckdb`). btw: we now convert all the splits to duckdb (before: we only did it for text datasets).

What do you think @julien-c @gary149?

albertvillanova:
Also: https://huggingface.slack.com/archives/C0311GZ7R6K/p1686665484050679

And I repeat what I said there:
> I think it is important that the directory containing the parquet files for a specific config+split, just contains the parquet files and no other files.

severo:
see #2413

AndreaFrancis:
Now that the index file has been removed, I  see another issue:
[...]

PR #1854, opened 2023-09-25
remove indexable columns limitation in split-duckdb-index

AndreaFrancis:
Needed for https://github.com/huggingface/datasets-server/pull/1418
Option 1: Since we need all the split datasets with the duckdb file for /filter, I propose just removing the limitation in the job runner as a noninvasive change.

Pending:

- [X] update tests
- [X] migration to add default field as has_fts=True for the existing cache records with status 200 

~~Option 2:~~
~~As an alternative, I think we could split the tasks into two separate runners:~~
~~- split-duckdb-data: in charge of data ingestion for the duckdb file only.~~
~~- split-duck-index: in charge of indexing the existing table only if indexable columns exist (might need to first download the previous one and generate a new one + the index).~~
~~But there could be open questions: Will we keep only one file in the repo? Should we have two separate files in the repo? (one for data and another for data+index)?~~

### DIFF
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -58,6 +58,9 @@ from mongodb_migration.migrations._20230703110100_cache_add_partial_field_in_con
 from mongodb_migration.migrations._20230705160600_queue_job_add_difficulty import (
     MigrationQueueAddDifficultyToJob,
 )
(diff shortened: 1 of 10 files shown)


severo:
somewhat related, if we change the name of the file: https://github.com/huggingface/datasets-server/issues/1694

severo:
I think option 1, as you coded it, is good. Only missing: the migration script (note: we might have to stop all the workers before running it to avoid creating cache entries with the old format).

severo (approved):

PR #1864, opened 2023-09-26
Reduce rows lru cache in /rows

lhoestq:
someone is querying an image dataset ioclab/animesfw a lot at different offsets which fills the lru cache of image data and leads to OOM


severo (approved):
OK. Maybe we can tweak this value later to get a bit more advantage from the LRU, maye by incrementing the RAM for the /rows pods

Issue #1833, opened 2023-09-21
Descriptive stats: wrong class label proportions when the labels are `-1` (no label)


lhoestq:
e.g. for https://huggingface.co/datasets/glue/viewer/cola/test

the response contains

```json
{
    "column_name": "label",
    "column_type": "class_label",
    "column_statistics": {
        "nan_count": 0,
        "nan_proportion": 0.0,
        "n_unique": 1,
        "frequencies": {
            "acceptable": 1063
        }
    }
}
```

cc @polinaeterna

PR #1870, opened 2023-09-27
fix: account for possible -1 value in ClassLabel feature in statistics computation

polinaeterna:
will fix https://github.com/huggingface/datasets-server/issues/1833

It appeared that `ClassLabel` feature might have `-1` (int) value which means "no label", I didn't know that so I end up having the last value of `ClassLabel.names` instead of it lol.

What would be the best way to represent it? It's not a null value. Now I convert it to string form (like `"-1"`). I thought of converting it to something like `"no_label"` or `"(no label)"` (as the viewer currently displays them in rows) but in theory it might match to the existing label specified in `ClassLabel.names` explicitly. Here is an example of response:

```python
 {
        "nan_count": 0,
        "nan_proportion": 0.0,
        "n_unique": 2,  # num unique values stays the same as when there is no `no label` values
        "frequencies": {
        	"cat": 10
        	"dog": 5,
        	"-1": 10,
        }
    }
```
Alternatively, I can add a new field "num_no_labels" to store this number instead of having it as a key in "frequencies".

wdyt?


lhoestq:
Maybe this ?

```
 {
        "nan_count": 0,
        "nan_proportion": 0.0,
        "nolabel_proportion": 0.0,
        "n_unique": 2,  # num unique values stays the same as when there is no `no label` values
        "frequencies": {
        	"cat": 10
        	"dog": 5,
        }
    }
```

polinaeterna:
@lhoestq i like it this way! the question is is it okay to the frontend to change the fields set? @severo

severo:
Sure!

severo:
nit: no_label instead of nolabel

polinaeterna:
@severo i also renamed `n_unique` to `num_unique` - is it okay? because we have `num_examples`, to be consistent.
can revert though

severo:
> @severo i also renamed n_unique to num_unique - is it okay? because we have num_examples, to be consistent.
can revert though

hmmm: it would require handling both field names while we recompute all the cache entries. I would prefer avoiding the pain

polinaeterna:
@severo ouch true, thanks, reverting then

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -36,6 +37,8 @@
 # the maximum number of unique values in a string column so that it is considered to be a category,
 # otherwise it's treated as a string
 MAX_NUM_STRING_LABELS = 30
+# datasets.ClassLabel feature uses -1 to encode `no label` value
+NO_LABEL_VALUE = -1
```
severo, line 41: https://github.com/huggingface/datasets/blob/faedc670ca896584d0f8edcb1fd9c13d1d6cc903/src/datasets/features/features.py#L1077

not exported as a constant, so: ok to define it here instead

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -490,13 +509,15 @@ def compute_descriptive_statistics_response(
     # compute for ClassLabels (we are sure that these are discrete categories)
     if categorical_features:
         logging.info(f"Compute statistics for categorical columns {categorical_features}")
+        categorical_features = Features.from_dict(categorical_features)
     for feature_name, feature in tqdm(categorical_features.items()):
         logging.debug(f"Compute statistics for ClassLabel feature '{feature_name}'")
-        class_label_names = feature["names"]
+        # class_label_feature = feature["names"]
```
severo, line 515: remove dead code

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -239,27 +244,39 @@ def compute_categorical_statistics(
...
     logging.debug(f"Compute categories counts for {column_name}.\n{categorical_counts_query}")
-    ids_to_counts: dict[int, int] = dict(
+    ids2counts: dict[int, int] = dict(
         con.sql(categorical_counts_query).fetchall()
-    )  # dict {idx: num_samples}; idx might be also None for null values
-    nan_count = ids_to_counts.pop(None, 0)  # type: ignore
-    labels_to_counts: dict[str, int] = {class_label_names[cat_id]: freq for cat_id, freq in ids_to_counts.items()}
-    n_unique = len(labels_to_counts)
+    )  # dict {idx: num_samples}; idx might be also None for null values and -1 for `no label`
+    nan_count = ids2counts.pop(None, 0)  # type: ignore
```
severo, line 257: to fix the type:ignore issue, maybe dict[Optional[int], int] instead of dict[int, int]

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -239,27 +244,39 @@ def compute_categorical_statistics(
...
-    class_label_names: list[str],
+    class_label_feature: ClassLabel,
     n_samples: int,
 ) -> CategoricalStatisticsItem:
     categorical_counts_query = COMPUTE_CATEGORIES_COUNTS_COMMAND.format(
         column_name=column_name, data_table_name=table_name
     )
     logging.debug(f"Compute categories counts for {column_name}.\n{categorical_counts_query}")
-    ids_to_counts: dict[int, int] = dict(
+    ids2counts: dict[int, int] = dict(
```
severo, line 254: nit: why changing this?

services/worker/tests/job_runners/split/test_descriptive_statistics.py
```diff
@@ -271,9 +280,9 @@ def descriptive_statistics_string_text_expected(datasets: Mapping[str, Dataset])
     [
         ("descriptive_statistics", None),
         ("descriptive_statistics_string_text", None),
-        ("gated", None),
-        ("audio", "NoSupportedFeaturesError"),
-        ("big", "SplitWithTooBigParquetError"),
+        # ("gated", None),
+        # ("audio", "NoSupportedFeaturesError"),
+        # ("big", "SplitWithTooBigParquetError"),
```
severo, line 285: why are these commented?

remove them if not needed anymore

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: for nothing actually, i think usually people do it with `2` (for example `datasets` has `ClassLabel.str2int()`, `ClassLabel.int2str()`)
can revert

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: oh thank you that worked!

services/worker/tests/job_runners/split/test_descriptive_statistics.py
polinaeterna: no they are needed, i comment them to run only the tests i need locally and always forget to uncomment... thanks for noticing!

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: it's not important, but it diverts a bit while reviewing (as we try not to change the style/format of JSON files for example, I prefer when "form-only" changes are in their own PR)

polinaeterna:
i don't understand why e22 tests fail with timeout :( any ideas? @severo @AndreaFrancis @lhoestq

severo:
> any ideas?

no, but let's create a dedicated issue for that

polinaeterna:
also, should I increment the version of the worker or better rerun manually only for datasets that has `ClassLabel` feature? @severo @AndreaFrancis

severo:
> rerun manually only for datasets that has ClassLabel feature

would be a lot better!

lhoestq:
> also, should I increment the version of the worker or better rerun manually only for datasets that have ClassLabel feature? @severo @AndreaFrancis

no need imo, we can refresh the main ones manually

polinaeterna:
@severo since the e2e tests failures are not related to this pr, can we merge?

severo (approved):

Issue #1857, opened 2023-09-26
Publish more statistics about Hub datasets usage


severo:
- dataset count (private/public, with script/without script)
- validity (/is-valid) of the first 20 trending datasets

severo:
I assigned you @lhoestq as discussed. Feel free to remove yourself if you don't plan to work on it soon

lhoestq:
on it

PR #1874, opened 2023-09-27
Add Hub dashboard in admin app

lhoestq:
A simple dashboard in gradio to show the number of datasets per type (parquet, json, dataset script etc) and the viewer coverage of trending datasets (can be extended later if needed).

The idea is to follow if users are still using dataset scripts often or if usage is going to our goal: less and less datasets defined with code.

The second idea is to see if the trending datasets are well supported by the datasets-server, which is one of our main priorities.

I used two endpoints for this:
- `/is-valid`
- `/admin/num-dataset-infos-by-builder-name` that I created

Fix https://github.com/huggingface/datasets-server/issues/1857


services/admin/src/admin/routes/num_dataset_infos_by_builder_name.py
```diff
@@ -0,0 +1,86 @@
...
+            # if auth_check fails, it will raise an exception that will be caught below
+            auth_check(
+                external_auth_url=external_auth_url,
+                request=request,
+                organization=organization,
+                hf_timeout_seconds=hf_timeout_seconds,
+            )
+            num_datasets_infos = CachedResponseDocument.objects(http_status=200, kind="dataset-info").count()
+            num_parquet_datasets_infos = CachedResponseDocument.objects(
+                http_status=200, kind="dataset-info", content__dataset_info__default__builder_name="parquet"
```
AndreaFrancis, line 40: Maybe all these requests should be inside simple_cache.py and also parameterize the builder_name or maybe use a method that returns summarized information (count of records grouped by builder_name).
Also, If I understand well, are you querying the builder_name for the "default" config? If so, what about the datasets that don't have that config?

front/admin_ui/app.py
```diff
@@ -182,6 +190,67 @@ def delete_obsolete_cache(token):
...
+        else:
+            out[home_dashboard_trending_datasets_infos_by_builder_name_table] = gr.update(visible=True, value=pd.DataFrame({
+                "Error": [f"❌ Failed to fetch dataset infos from {DSS_ENDPOINT} (error {response.status_code})"]
+            }))
+        response = requests.get(f"{HF_ENDPOINT}/api/trending?type=dataset&limit=20", timeout=60)
+        if response.status_code == 200:
+            trending_datasets = [repo_info["repoData"]["id"] for repo_info in response.json()["recentlyTrending"]]
+
+            def get_is_valid_response(dataset: str):
+                return requests.get(f"{DSS_ENDPOINT}/admin/dataset-responses?kind=dataset-is-valid&dataset={dataset}", headers=headers, timeout=60)
```
AndreaFrancis, line 221: I didn't understand, why not call directly to https://datasets-server.huggingface.co/is-valid?dataset=<dataset>

AndreaFrancis (commented):
Great!

front/admin_ui/app.py
severo: I agree, in this case, no need for another endpoint, the response will be the same

services/admin/src/admin/routes/num_dataset_infos_by_builder_name.py
severo: > count of records grouped by builder_name

I understand that each dataset with its own script will have a different builder name, so it makes sense to only check a list of reserved builder names.

We could optimize the code a bit, but maybe don't spend too much time in this, as it's only for our own admin purpose

lhoestq: Ah yes indeed ! It's because I started working on this endpoint to do something else first and reverted it - will remove

front/admin_ui/app.py
lhoestq: Ah yes indeed ! It's because I started working on this endpoint to do something else first and reverted it - will remove

services/admin/src/admin/routes/num_dataset_infos_by_builder_name.py
lhoestq: > Also, If I understand well, are you querying the builder_name for the "default" config? If so, what about the datasets that don't have that config?

Usually datasets with a builder name like parquet/csv etc have the "default" config (<0.1% don't) and it simplifies the mongo request. Alternatively I could query the config-level cache though the query is a bit longer

> We could optimize the code a bit, but maybe don't spend too much time in this, as it's only for our own admin purpose

Yes that's what I had in mind

severo (approved):

AndreaFrancis (approved):

PR #1881, opened 2023-09-28
Clean HF datasets cache

lhoestq:
Every 3h for files older than 3h

Also contrary to the delete-duckdb-index job I simply scan the mtime of the directories instead of the mtime of every single file

to be merged after I'm done removing the current 60TB of cache (done !)


lhoestq:
I just deleted the old cache files and there is <1TB left. I think we can enable this job now :)

jobs/cache_maintenance/src/cache_maintenance/clean_hf_datasets_cache.py
```diff
@@ -0,0 +1,50 @@
...
+
+from libcommon.storage import StrPath
+
+
+def clean_hf_datasets_cache(hf_datasets_cache: StrPath, expired_time_interval_seconds: int) -> None:
+    """
+    Delete temporary cache directories from job runners with datasets cache.
+    """
+    logging.info("delete hf datasets cache")
+    pattern = f"{hf_datasets_cache}/*"
```
severo, line 19: what if `hf_datasets_cache` happens to be `""` for some configuration reason? How could we avoid deleting everything below the root? Maybe at least ensure it's non-empty? Maybe some other idea to avoid deleting wrong files by error?

severo (approved):
see if. you can add some protection, at least against the value being an empty string?

jobs/cache_maintenance/src/cache_maintenance/clean_hf_datasets_cache.py
lhoestq: added a sanity check that raises an error if len(hf_datasets_cache) < 10

severo: OK, thanks!

PR #1882, opened 2023-09-28
cached assets on s3 for all datasets

AndreaFrancis:


severo (approved):
OK. I was wondering if it's OK to delete the alternative option (fielsystem). But I think it's easy to set it back if needed, right?

Can we also deprecate some environment variables at the same time, or are they all needed yet

lhoestq (approved):
let's go !

AndreaFrancis:
> OK. I was wondering if it's OK to delete the alternative option (fielsystem). But I think it's easy to set it back if needed, right?

I think so, but the problem is with DNS configuration, not sure if there is a way to redirect some datasets to EFS and others to S3.
Maybe @rtrompier  could help us confirm.

severo:
Oh, yes, I hadn't this in mind. OK to do that. Should we remove /cached-assets from the nginx reverse proxy configuration?

severo:
(I'm wondering how we will handle the tests: end to end, unit)

AndreaFrancis:
> Oh, yes, I hadn't this in mind. OK to do that. Should we remove /cached-assets from the nginx reverse proxy configuration?

I think so, I created this issue for tracking https://github.com/huggingface/datasets-server/issues/1885

severo:
Currently the docker compose files don't use S3. I think it means we cannot use /rows and /search with docker for images and audio, right?
Should we keep the switch between directory and S3 in the code, to support docker compose (and thus: the init tests and the e2e tests)? The alternatives are: a) to include S3 in docker compose (and thus: use or mock S3 for tests), b) to remove docker compose completely and only use k8s/helm (it still means that we need S3 for the tests)

rtrompier:
> I think so, but the problem is with DNS configuration, not sure if there is a way to redirect some datasets to EFS and others to S3.

Nop we can't. The switch is based on url path in cloudfront.
But if you can have a different path between datasets, it should be possible.

lhoestq:
Another alternative: use a a storage agnostic tool like fsspec to store the images (currently we use boto if I'm not mistaken).
In prod it would point to s3://bucket-name and in docker-compose it would point to a disk shared with the reverse proxy

severo:
> Another alternative: use a a storage agnostic tool like fsspec to store the images (currently we use boto if I'm not mistaken). In prod it would point to s3://bucket-name and in docker-compose it would point to a volume shared with the reverse proxy

I like this option

Issue #1823, opened 2023-09-20
Avoid providing "mixed" cache while refreshing a dataset


severo:
When a dataset is being refreshed, or even after having been refreshed, its cache can be incoherent.

## Case 1

Reported by @AndreaFrancis: during a dataset refresh, the search was enabled in the viewer, but the "first rows" were not shown. The `split-duckdb-index` step had finished, while the `split-first-rows-from...` steps were still being computed. The data shown by the viewer were not coherent, and it would be hard for a user to understand what was occurring.

## Case 2

Seen with https://huggingface.co/datasets/jbrendsel/ECTSum. All the steps had been computed, and the `config-size` step had an error (`PreviousStepFormatError`). The erroneous cache entry remained after fixing the issue and refreshing the dataset. The reason is a previous step (`dataset-config-names`) now gives an error, and the rest of the cache entries are never cleaned, so they stay in the database even if useless. See https://github.com/huggingface/datasets-server/issues/1582#issuecomment-1727151243 and https://github.com/huggingface/datasets-server/issues/1285.

## Proposal

Give each "DAG" execution an identifier, and use the same identifier for all the endpoints. To ensure coherence between the responses, change the identifier we use in the API only when all the steps have been refreshed. If we send the identifier in the response, the client can also use the identifier value to check the coherence. If there is no previous identifier in the database, use the identifier of the current incomplete DAG (the dataset viewer already handles an incomplete state).

Once the new DAG execution has finished, we can delete the cache entries with other identifiers.

AndreaFrancis:
> Once the new DAG execution has finished, we can delete the cache entries with other identifiers.

Pros:
- We will have always something to show in the viewer

Cons:
- I think we could end up with our collections growing considering that these are some of the implementation details:
We would have to add an extra field on cache and job collection maybe "dag_version", this new field should be added in the unique constraints index by kind.
We will probably have to add an index for this new field
Double the amount of data in our collections in the worst cases (duplicated content)
Will probably have to run a new job runner or something like that to save the current_dag_version and update it when all the DAG is completed

Maybe initially a simpler approach:  Once a job_runner finishes, delete all its children and backfill
Pros:
- Dont need too much change in our current infrastructure
- It could be good to have in the db only what we need 
Cons:
- Dataset viewer would end up waiting for the new information but it used to be quick. We have delays only when the queue is overloaded.

WDYT @huggingface/datasets-server

severo:
[...]

PR #1884, opened 2023-09-28
feat: 🎸 add POST /admin/recreate-dataset?dataset=.&priority=

severo:
see
https://github.com/huggingface/datasets-server/issues/1823#issuecomment-1739117350

Missing:
- tests (it's destructive, so: better to ensure we don't affect other datasets for example)
- add a tab in the admin UI


services/admin/src/admin/routes/recreate_dataset.py
```diff
@@ -0,0 +1,84 @@
...
+                external_auth_url=external_auth_url,
+                request=request,
+                organization=organization,
+                hf_timeout_seconds=hf_timeout_seconds,
+            )
+            # try to get the revision of the dataset (before deleting the jobs and the cache, in case it fails)
+            revision = get_dataset_git_revision(dataset=dataset, hf_endpoint=hf_endpoint, hf_token=hf_token)
+            # delete all the jobs and all the cache entries for the dataset (for all the revisions)
+            cancelled_jobs = Queue().cancel_dataset_jobs(dataset=dataset)
+            deleted_cached_responses = delete_dataset_responses(dataset=dataset)
```
AndreaFrancis, line 60: I was thinking of integrating this cache deletion with a couple of tasks:
- delete assets 
- delete uploaded files in refs/convert/parquet

severo: yes, it would be a lot better:
- having a complete inventory of related assets/files in external places for a dataset
- being able to clean them for a dataset, a config, a split

I think we can still go on with this PR, and fix this later when implemented.

libs/libcommon/src/libcommon/queue.py
```diff
@@ -883,6 +898,7 @@ def finish_job(self, job_id: str, is_success: bool) -> bool:
         job.update(finished_at=get_datetime(), status=finished_status)
         update_metrics_for_type(job_type=job.type, previous_status=previous_status, new_status=finished_status)
         release_locks(owner=job_id)
+        # ^ bug: the lock owner is not set to the job id anymore when calling start_job()!
```
severo, line 901: cc @lhoestq : the owner is now a UUID, we should fix that

services/admin/src/admin/routes/recreate_dataset.py
severo: I created https://github.com/huggingface/datasets-server/issues/1887, feel free to comment or edit the description

libs/libcommon/src/libcommon/queue.py
severo: https://github.com/huggingface/datasets-server/issues/1888

services/admin/src/admin/routes/recreate_dataset.py
```diff
@@ -0,0 +1,92 @@
...
+    processing_graph: ProcessingGraph,
+    assets_directory: StrPath,
+    cached_assets_directory: StrPath,
+    hf_endpoint: str,
+    hf_token: Optional[str] = None,
+    external_auth_url: Optional[str] = None,
+    organization: Optional[str] = None,
+    hf_timeout_seconds: Optional[float] = None,
+) -> Endpoint:
+    async def force_refresh_endpoint(request: Request) -> Response:
```
AndreaFrancis, line 40:
```suggestion
    async def recreate_dataset_endpoint(request: Request) -> Response:
```

services/admin/src/admin/routes/recreate_dataset.py
```diff
@@ -0,0 +1,92 @@
...
+            revision = get_dataset_git_revision(dataset=dataset, hf_endpoint=hf_endpoint, hf_token=hf_token)
+            # delete all the jobs and all the cache entries for the dataset (for all the revisions)
+            cancelled_jobs = Queue().cancel_dataset_jobs(dataset=dataset)
+            deleted_cached_responses = delete_dataset_responses(dataset=dataset)
+            if deleted_cached_responses is not None and deleted_cached_responses > 0:
+                # delete assets
+                delete_asset_dir(dataset=dataset, directory=assets_directory)
+                delete_asset_dir(dataset=dataset, directory=cached_assets_directory)
+            # add the first processing steps to the queue
+            for processing_step in processing_graph.get_first_processing_steps():
```
AndreaFrancis, line 70: Maybe try to use something similar to [try_backfill_dataset_then_raise](https://github.com/huggingface/datasets-server/blob/main/libs/libapi/src/libapi/utils.py#L109)?
What if the dataset does not exist, maybe we would be appending unnecessary jobs.
Not sure, feel free to ignore it if does not make sense.

AndreaFrancis (approved):

services/admin/src/admin/routes/recreate_dataset.py
severo: done

services/admin/src/admin/routes/recreate_dataset.py
severo: You're right! I forgot we add code for that (even if I wrote it). I made as in /webhook.

severo:
thanks for the reviews @AndreaFrancis

PR #1889, opened 2023-09-29
fix: remove file in cached-assets even if not uploaded

AndreaFrancis:
Local files remain in local cached-assets when they already exist on the bucket, we should remove them anyway.
I will improve the logic in another PR: create files ONLY when they don't exist on the bucket.


severo (approved):

AndreaFrancis:
I will merge this PR, looks like e2e is taking too much time, but I don't think this change is related somehow.

PR #1898, opened 2023-10-02
Delete unnecessary condition on expected_error_code

albertvillanova:
As reported by @AndreaFrancis (see: https://github.com/huggingface/datasets-server/pull/1418#discussion_r1340621923), the condition on `expected_error_code` can be removed because it equals `None`.


severo (approved):
Indeed, these are remainings of previous tests, and it's simpler when removing this code

PR #1899, opened 2023-10-02
feat: 🎸 delete the dataset and raise if blocked

### LINKED ISSUE #1897: Reduce amount of jobs for blocked datasets
Currently most of the jobs are for datasets inside the `open-llm-leaderboard` org, which are all blocked 

https://github.com/huggingface/datasets-server/blob/f77e375a49503693c2537c7a4b35229551b37b95/chart/env/prod.yaml#L149

We could directly set the value for all the steps to "blocked" error?

Also: currently the blocked list is for `config-parquet-and-info`. We might want to apply it to the whole DAG.

severo:
in libcommon, not only in the workers.

Fixes #1897


libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -577,10 +615,58 @@ def _create_plan(self) -> None:
...
     processing_graph: ProcessingGraph
+    blocked_datasets: InitVar[list[str]]
+
+    def __post_init__(self, blocked_datasets: list[str]) -> None:
+        try:
+            raise_if_blocked(dataset=self.dataset, blocked_datasets=blocked_datasets)
+        except DatasetInBlockListError:
+            logging.warning(f"The dataset {self.dataset} is in the block list, we delete it from the Datasets Server.")
+            self.remove_dataset()
+            raise
```
lhoestq, line 662: Maybe only raise when calling backfill or finish_job or set_revision or any other function that would create a job ? Otherwise you can't create one to call `remove_dataset` on a blocked dataset

Anyway in general a `__post_init__` raise is not always convenient e.g. for testing or if you need to call some read methods even on a blocked dataset for some reason.

lhoestq (commented):
Cool ! Just one comment:

libs/libcommon/src/libcommon/utils.py
```diff
@@ -144,3 +147,32 @@ def is_image_url(text: str) -> bool:
...
+            The list of blocked datasets. If empty, no dataset is blocked.
+            Unix shell-style wildcards are supported, e.g. "open-llm-leaderboard/*"
+
+    Returns:
+        `None`
+    Raises the following errors:
+        - [`libcommon.exceptions.DatasetInBlockListError`]
+          If the dataset is in the list of blocked datasets.
+    """
+    for blocked_dataset in blocked_datasets:
```
AndreaFrancis, line 173: Like in other PRs, what if by mistake we add "*" to the blocket_datasets? Maybe add a validation here?

AndreaFrancis (approved):
Thanks! I just put a comment

libs/libcommon/src/libcommon/orchestrator.py
severo: > Otherwise you can't create one to call remove_dataset on a blocked dataset

I expose `remove_dataset` for other admin endpoints (but I don't use it now, so...)

> Anyway in general a __post_init__ raise is not always convenient

yes, you're right, it's a bad practise to raise in the constructor (or like here, in __post_init__). I'll change this.

libs/libcommon/src/libcommon/utils.py
severo: ahah, good point.

severo: https://github.com/huggingface/datasets-server/pull/1899/commits/60645e6800405b712af53f3a55d0a0a62cbaefa8

libs/libcommon/src/libcommon/orchestrator.py
severo: https://github.com/huggingface/datasets-server/pull/1899/commits/125789acd3fa8db89b699f01d20465011bfc848f

lhoestq (approved):
Cool !

Another sanity check idea: disallow "*" in user/org names and only allow it in dataset names for datasets within a namespace.
This time I would allow to raise in `CommonConfig.__post_init__` lol

severo:
oh, OK, sorry, I rushed too much this time. I thought you were waiting for me to fix and merge.

severo:
> Another sanity check idea: disallow "*" in user/org names and only allow it in dataset names for datasets within a namespace.

It's much better -> https://github.com/huggingface/datasets-server/pull/1901

Take your time to review! :P

PR #1900, opened 2023-10-02
Fix class label computation check and return counts for all labels, not only those found in split

polinaeterna:
There might be cases when the number of unique values found in a set is not equal to the number of classes predefined in the `ClassLabel` feature (for example, in test sets might be only `no label` values). So I check that the number of unique values is not greater than the number of classes stored in a feature instead of checking their equality as before. 

**Also return all classes and their counts in response, not only those found (to make different splits aligned - but for strings-classes the behavior is different because we don't have predefined classes there).**

For example, for [`glue`](https://huggingface.co/datasets/glue), `ax` config, `test` set the result will be: 
```python
{
    "column_name": "label",
    "column_type": "class_label",
    "column_statistics": {
        "nan_count": 0,
        "nan_proportion": 0.0,
        "no_label_count": 1104,
        "no_label_proportion": 1.0,
        "n_unique": 3,
        "frequencies": {  # without this change it would be an empty dict
            "entailment": 0,
            "neutral": 0,
            "contradiction": 0
        }
    }
}
```


lhoestq (approved):
Nice ! lgtm :)

severo (approved):
Good catch

severo:
I merged, to be able to deploy all the finished PR now.

Issue #1891, opened 2023-09-29
Create "heavy" workers with a lot of RAM


severo:
Aside from light and normal workers.

To make the most efficient use of them, maybe rework how we assign the "difficulty" field of the jobs. Currently it only depends on the job type. Maybe we want another field to ask for more RAM, or another way to decide which type of worker should process the job.

severo:
related: https://github.com/huggingface/datasets-server/issues/1892

PR #1903, opened 2023-10-02
Add bonus difficulty if dataset is big

lhoestq:
Add a bonus difficulty of 20 for `duckdb-index` if the dataset is bigger than 3GB (from the config info).
This is defined in the orchestrator.

This can be used to defined "heavy workers" that would have enough RAM to process them (up to 29GB of memory for datasets like C4 that have their first 5GB converted to parquet)

The main changes are in
libs/libcommon/src/libcommon/config.py
libs/libcommon/src/libcommon/constants.py
libs/libcommon/src/libcommon/orchestrator.py
libs/libcommon/src/libcommon/processing_graph.py

TODO:
- [x] tests

Related to https://github.com/huggingface/datasets-server/issues/1891


libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -593,8 +633,8 @@ def parse_id(id: str) -> tuple[str, str, Optional[str], Optional[str], str]:
         parts = parts[2:]
         config = None
         split = None
-        if len(parts) > 1:
-            config = parts[1]
-            if len(parts) > 2:
-                split = parts[2]
+        if len(parts) > 0:
```
severo, line 636: it was wrong? and had no consequences until now?

libs/libcommon/src/libcommon/config.py
```diff
@@ -35,7 +36,9 @@
     PROCESSING_STEP_SPLIT_OPT_IN_OUT_URLS_COUNT_VERSION,
     PROCESSING_STEP_SPLIT_OPT_IN_OUT_URLS_SCAN_VERSION,
 )
-from libcommon.processing_graph import ProcessingGraphSpecification
+
+if TYPE_CHECKING:
+    from libcommon.processing_graph import ProcessingGraphSpecification
```
severo, line 41: why the change?

libs/libcommon/tests/test_orchestrator.py
```diff
@@ -345,3 +346,80 @@ def test_remove_dataset() -> None:
     pending_jobs_df = Queue().get_pending_jobs_df(dataset=DATASET_NAME)
     assert len(pending_jobs_df) == 0
     assert has_some_cache(dataset=DATASET_NAME) is False
+
+
+@pytest.mark.parametrize("is_big", [False, True])
+def test_after_joplan_gives_bonus_difficulty(is_big: bool) -> None:
```
severo, line 352: nit

```suggestion
def test_after_job_plan_gives_bonus_difficulty(is_big: bool) -> None:
```

severo (approved):

libs/libcommon/src/libcommon/config.py
lhoestq: To avoid circular dependencies

libs/libcommon/src/libcommon/processing_graph.py
lhoestq: It was wrong and had no consequences since this function was only used on jobs at dataset level in tests

severo: OK, thanks for checking!

libs/libcommon/src/libcommon/config.py
severo: hmmm, ok. I imagine that's why I used the "specification" field at that time. Maybe it's better to move `ProcessingStepSpecification` and `ProcessingGraphSpecification` to a types file, no?

lhoestq: I think it's fine.

It's common for config files to import other types using TYPE_CHECKING to make sure every file can still import the config

severo: OKOK

PR #1827, opened 2023-09-21
Update pandas dependency to 2.1.1 version

albertvillanova:
Update `pandas` dependency to 2.1.1 version, their latest patch release: https://github.com/pandas-dev/pandas/releases/tag/v2.1.1

Note that:
- several pandas versions were installed
- most of them were 2.0.2: there was a newer 2.0.3 patch release
- one was 2.1.0: for this, the newer 2.1.1 patch release is recommended

PR #1911, opened 2023-10-03
Revert "Update pandas to 2.1.1 and duckdb to 0.9.0 (#1827)"

### LINKED ISSUE #1910: Revert update of duckdb
Revert update of duckdb because indexes computed with prior version are not compatible with the latest version:
- #1827

If we eventually update duckdb, we will need to recompute all the indexes.

Reported by @lhoestq

albertvillanova:
This reverts commit 93bcf55fafe4bc23d3dd7dc75dc7c8ff216f7df0.

Fix #1910.

Note that we cannot update `pandas` patch release 2.1.1 if we do not update `duckdb`. See discussion in: https://github.com/huggingface/datasets-server/pull/1827#issuecomment-1729090034


lhoestq (approved):
Thanks !

Issue #1912, opened 2023-10-03
Some audio files in /first-rows are 0 bytes


lhoestq:
Some audio files in /first-rows are 0 bytes, e.g.
https://datasets-server.huggingface.co/assets/ccmusic-database/acapella_eval/--/default/song1/0/song/audio.wav (edited)

PR #1915, opened 2023-10-03
Fix audi oasset from streaming

lhoestq:
Some audio files were not correctly created when the data came from streaming and is decoded (i.e. is a numpy array representing the audio sample)

Fix https://github.com/huggingface/datasets-server/issues/1912

### DIFF
--- a/libs/libcommon/src/libcommon/viewer_utils/features.py
+++ b/libs/libcommon/src/libcommon/viewer_utils/features.py
@@ -120,12 +120,11 @@ def audio(
             raise ValueError(
                 f"An audio sample should have a 'path' with a valid extension but got '{audio_file_extension}'."
             )
-    elif "array" in value:
+    elif ("path" in value and value["path"] is None) or "array" in value:
         audio_file_extension = ".wav"
     else:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


severo (approved):

lhoestq:
Removed 2k `split-first-rows-from-streaming` cache entries by filtering on `updated_at` and using a list of audio datasets (the ones with a column named "audio")

severo:
OK. Do you rely on the daily backfill to create the jobs?

severo:
Also: I think we can use a mongo query to get all the `split-first-rows-from-streaming` responses that have an `Audio` feature, which would capture all the cache entries to refresh, right?

lhoestq:
> Also: I think we can use a mongo query to get all the split-first-rows-from-streaming responses that have an Audio feature, which would capture all the cache entries to refresh, right?

Yes but it's not as easy as it seems :) I did the other ones I had in mind manually

lhoestq:
I mean, I tried using `content__features__match` but it seems you need to provide the exact value of `type` (which can include a specific sampling_rate value) and I didn't spend more time figuring it out

severo:
OK

PR #1922, opened 2023-10-04
fix: 🐛 fix vulnerability (pillow)

severo:
see https://github.com/huggingface/datasets-server/security/dependabot/272

also update libcommon in front/admin

### DIFF
--- a/front/admin_ui/poetry.lock
+++ b/front/admin_ui/poetry.lock
@@ -237,6 +237,46 @@ files = [
 ]
 
 [[package]]
+name = "boto3"
+version = "1.28.59"
+description = "The AWS SDK for Python"
+category = "main"
+optional = false
+python-versions = ">= 3.7"
+files = [
+    {file = "boto3-1.28.59-py3-none-any.whl", hash = "sha256:7623b52c135becf145f762a9cc4203a1fb30055bb1cc7a254f82e5f7954d44a1"},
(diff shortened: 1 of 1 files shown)


front/admin_ui/poetry.lock
```diff
@@ -236,6 +236,46 @@ files = [
...
+[package.dependencies]
+botocore = ">=1.31.59,<1.32.0"
+jmespath = ">=0.7.1,<2.0.0"
+s3transfer = ">=0.7.0,<0.8.0"
+
+[package.extras]
+crt = ["botocore[crt] (>=1.21.0,<2.0a0)"]
+
+[[package]]
+name = "botocore"
```
albertvillanova, line 260: I am wondering why `boto3` and `botocore` are added by upgrading pillow?

albertvillanova (commented):
I would mention this is a follow-up of the PR:
- #1853

front/admin_ui/poetry.lock
severo: I run `poetry update libcommon` at the same time, because it was not up to date

albertvillanova: And `s3transfer`...

albertvillanova: I see... I think it is because of the indirect dependency on `libcommon`...

severo: yes, it's what I mentioned above

> I run poetry update libcommon at the same time, because it was not up to date

albertvillanova (approved):

PR #1926, opened 2023-10-04
feat(chart): use spot instance in staging env

rtrompier:
Try to use karpenter in staging env for a better optimisation of our resources consumptions.

### DIFF
--- a/chart/env/staging.yaml
+++ b/chart/env/staging.yaml
@@ -168,6 +168,17 @@ cacheMaintenance:
       cpu: 100m
     limits:
       cpu: 1
+  nodeSelector:
+    project: "datasets-server"
+  tolerations:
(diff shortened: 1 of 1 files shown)


lhoestq:
Can you give more details ? I think most of us are not familiar with karpenter

rtrompier:
> Can you give more details ? I think most of us are not familiar with karpenter

It's an infra tool to manager resources into our Kube Cluster. It will be transparent for you.
Karpenter will help us to choose the best node depending on your request specification on your pods.
It can choose to start a new node, use an existing one, replace an existing one by another one with a better performance, ...
I also configured it to use `spot` instance, to reduce our cost for staging env.

More info here: https://karpenter.sh/

PR #1928, opened 2023-10-04
assets to s3

AndreaFrancis:
Moving assets to s3.
Second part of https://github.com/huggingface/datasets-server/issues/1406


libs/libcommon/src/libcommon/config.py
```diff
@@ -42,12 +42,14 @@
 
 ASSETS_BASE_URL = "assets"
 ASSETS_STORAGE_DIRECTORY = None
+ASSETS_S3_FOLDER_NAME = "assets"
```
severo, line 45: it's a bit weird to have two configuration variables (ASSETS_STORAGE_DIRECTORY and ASSETS_S3_FOLDER_NAME), while we will only use one of them.

But I understand that this will be fixed in another PR where we will abstract S3 or FS storages in a common interface.

chart/templates/worker/_container.tpl
```diff
@@ -7,6 +7,7 @@
   imagePullPolicy: {{ .Values.images.pullPolicy }}
   env:
   {{ include "envAssets" . | nindent 2 }}
+  {{ include "envS3" . | nindent 2 }}
```
severo, line 10: no other containers need these environment variables? the jobs, the api/admin services, the storage machine?

services/worker/pyproject.toml
```diff
@@ -58,6 +58,7 @@ flake8 = "^3.9.2"
 flake8-pep585 = "^0.1.7"
 flake8-unused-arguments = "^0.0.13"
 isort = "^5.12.0"
+moto = "^4.2.5"
```
severo, line 61: (maybe) install in the dev dependencies of libcommon instead

severo (commented):
Another comment: could you describe how we will migrate from NFS to S3?

chart/templates/worker/_container.tpl
AndreaFrancis: S3Client is initialized only in `worker` `rows` and `search`. I think we don't need it in other containers.

libs/libcommon/src/libcommon/config.py
AndreaFrancis: Yes, I think using `fsspec` would help us create the file directly in a specified location and ASSETS_STORAGE_DIRECTORY will be removed.

chart/templates/worker/_container.tpl
severo: OK, thanks for checking

services/worker/pyproject.toml
severo: (feel free to ignore)

AndreaFrancis:
> Another comment: could you describe how we will migrate from NFS to S3?

In Prod:
- We will need to stop the workers while @rtrompier  helps us move/copy the files from NFS to S3
- Need to enable DNS for assets to S3
- Merge and deploy this PR
The last two steps can be done at the same time or independently.

severo:
but the assets URLs in `splits-first-rows-...` responses will still be valid once migrated to S3?

services/worker/pyproject.toml
AndreaFrancis: I think I will leave it as it is now, anyway once we use `fsspec` instead we will probably remove all the references to boto3 and moto3.

AndreaFrancis:
> but the assets URLs in splits-first-rows-... responses will still be valid once migrated to S3?

I see that the way we store the URLS is like https://datasets-server.huggingface.co/assets/asoria/image/--/default/train/0/image/image.jpg
So, same as in rows https://datasets-server.huggingface.co/cached-assets/asoria/image/--/default/train/0/image/image.jpg the switch in the DNS to point https://datasets-server.huggingface.co/assets to s3 /assets should work  correctly as well.
Can you confirm @rtrompier?

severo:
> We will need to stop the workers while @rtrompier helps us move/copy the files from NFS to S3

is there a way to copy everything, then when we stop the workers, only do a last sync for the few files that have changed? It would reduce the downtime for the workers.

rtrompier:
> > but the assets URLs in splits-first-rows-... responses will still be valid once migrated to S3?
> 
> I see that the way we store the URLS is like https://datasets-server.huggingface.co/assets/asoria/image/--/default/train/0/image/image.jpg So, same as in rows https://datasets-server.huggingface.co/cached-assets/asoria/image/--/default/train/0/image/image.jpg the switch in the DNS to point https://datasets-server.huggingface.co/assets to s3 /assets should work correctly as well. Can you confirm @rtrompier?

Yes I confirm. I will only redirect the traffic on `/assets/*` on S3.

> is there a way to copy everything, then when we stop the workers, only do a last sync for the few files that have changed? It would reduce the downtime for the workers.

It's what I'm trying to do. 
I will start a first transfert will all files, and once deployed, I will start an incremental transfert.
I come back to you when I found the best solution ;)

severo (approved):

severo:
hmmmm are we ready to deploy this????????

severo:
I mean: is the migration finished?

AndreaFrancis:
> hmmmm are we ready to deploy this????????

Yes, we will be working on the deploy with @rtrompier

severo:
how long do you think it will take? (because meanwhile we cannot deploy anything)

rtrompier:
> how long do you think it will take? (because meanwhile we cannot deploy anything)

Should be quick. We just performing some tests on staging env before.

PR #1934, opened 2023-10-05
Factorize getting the request parameters length and offset

albertvillanova:
Factorize getting the request parameters `length` and `offset` from the service routes (/rows, /search and /filter) to `libapi`.
Additionally, align their error messages.


severo (approved):
Excellent, good refactoring

PR #1937, opened 2023-10-05
Do not pass None headers in e2e tests

albertvillanova:
This PR refactors e2e tests so that we do not pass unnecessary None headers (this is their default value).


severo (approved):
indeed. Thanks

Issue #86, opened 2021-10-25
Support timestamp and date feature types


severo:
- [ ] add a date type
- [ ] detect dates, ie `allenai/c4` -> `timestamp` , `curiosity_dialogs`

severo:
see also (internal): https://huggingface.slack.com/archives/C0311GZ7R6K/p1680784227964169

lhoestq:
just got
```
bson.errors.InvalidDocument: cannot encode object: datetime.date(2016, 5, 20), of type: <class 'datetime.date'>
```
when running the first rows job for https://huggingface.co/datasets/ClimatePolicyRadar/global-stocktake-documents

It seems the column `date: date32[day]` can't be stored in mongo as is

lhoestq:
I think we store dates using a nice formatting, maybe like "YYYY-MM-DD" for date32 and "YYYY-MM-DDTHH:MM:SSZ" for date64 ?

severo:
Another report: https://huggingface.co/datasets/poloclub/diffusiondb/discussions/2#6374ff55b93cbdf65675f563 (old!)

lhoestq:
Another report: https://huggingface.co/datasets/aborruso/pnrr/discussions/1

lhoestq:
another one: https://huggingface.co/datasets/hyperdemocracy/us-congress-bills/discussions/1#64fd5e06b3eee10ba54990d3

severo:
It should not generate an error since https://github.com/huggingface/datasets-server/pull/1938.

But beyond not raising an exception, we want to fully support the type, in order to get dates histogram for example.

lhoestq:
Now getting this during `upsert_response_params` when `CachedResponseDocument.upsert_one` is called

```
bson.errors.InvalidDocument: cannot encode object: datetime.date(2023, 6, 13), of type: <class 'datetime.date'>
```

PR #1938, opened 2023-10-05
fix: adding support for other data types in orjson_dumps

AndreaFrancis:
Currently, we have 105 cache records with error types like:

```
  'Type is not JSON serializable: Timedelta',
  'Type is not JSON serializable: Timestamp',
  'Type is not JSON serializable: datetime.timedelta',
  'Type is not JSON serializable: numpy.int64',
  'Type is not JSON serializable: numpy.ndarray'
```

The problem is at https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/utils.py#L120 trying to serialize the values.

After this PR, I will force refresh the affected splits (most of them are for `split-first-rows-from-streaming`)

Also there is an error for "split-descriptive-statistics":

`"Dict key must be str"
`
This was due to non-string keys in ClassLabels like ints. After this PR, I will force refresh the affected splits (49 records)


libs/libcommon/src/libcommon/utils.py
```diff
@@ -113,7 +114,13 @@ def orjson_default(obj: Any) -> Any:
         # the bytes are encoded with base64, and then decoded as utf-8
         # (ascii only, by the way) to get a string
         return base64.b64encode(obj).decode("utf-8")
-    raise TypeError
+    if isinstance(obj, np.integer):
+        return int(obj)
+    if isinstance(obj, np.floating):
+        return float(obj)
+    if isinstance(obj, np.ndarray):
+        return obj.tolist()
```
albertvillanova, line 122: To support the serialization of `numpy` types, why don't you use the `orjson.OPT_SERIALIZE_NUMPY` option in `orjson.dumps`?

albertvillanova (approved):
Thanks. Just a question below.

libs/libcommon/src/libcommon/utils.py
AndreaFrancis: Thanks for the suggestion! I will send a new commit with this option.

lhoestq (approved):
Nice !

The datetime formatting is like `b'"2022-01-15T00:00:00"'` and numpy arrays is like `b'[0,1,2,3,4]'` which is perfect

lhoestq:
Cc @severo in case it requires changes in the front-end (might work out of the box no?)

lhoestq:
This should fix https://github.com/huggingface/datasets-server/issues/86 btw :)

severo:
By default, it will return a string. It should work

albertvillanova (commented):
Just another comment: the datetime and the Timestamp are not currently aligned
- datetime: 
  ```
  b'"2017-01-01T12:10:20.345000"'
  ```
- Timestamp: 
  ```
  b'"2017-01-01 12:10:20.345000"'
  ```

If the alignment is important, this could be achieved with:
```python
def orjson_default(obj):
    ...
    if isinstance(obj, pd.Timestamp):
        return obj.to_pydatetime()
```

AndreaFrancis:
> Just another comment: the datetime and the Timestamp are not currently aligned

Thanks @albertvillanova! I added your suggestion.

AndreaFrancis:
Fixed:
```
Atlas atlas-x5jgb3-shard-0 [primary] datasets_server_cache> db.cachedResponsesBlue.countDocuments({"details.copied_from_artifact":{$exists:false}, error_code:"UnexpectedError", "content.error":/Dict key must be str/i})
0
Atlas atlas-x5jgb3-shard-0 [primary] datasets_server_cache> db.cachedResponsesBlue.countDocuments({"details.copied_from_artifact":{$exists:false}, error_code:"UnexpectedError", "content.error":/Type is not JSON serializable/i})
0

```

severo:
you rock!

PR #1939, opened 2023-10-05
force download when previous file is obsolete

AndreaFrancis:
For https://datasets-server.huggingface.co/search?dataset=xnli&config=all_languages&split=train&offset=0&length=1&query=language

```

OSError: Consistency check failed: file should be of size 301736502 but has size 293192246 (0002.parquet).
We are sorry for the inconvenience. Please retry download and pass `force_download=True, resume_download=False` as argument.
If the issue persists, please let us know by opening an issue on https://github.com/huggingface/huggingface_hub.

```

See that there is an error in hf_hub_download, It looks like the library tried to compare the file with a previous one in the local cache. Maybe the data changed or there was a previously downloaded obsolete/corrupted file. 
I think we should force-download each time to always have a refreshed parquet file.

### DIFF
--- a/libs/libapi/src/libapi/duckdb.py
+++ b/libs/libapi/src/libapi/duckdb.py
@@ -83,6 +83,7 @@ def download_index_file(
         local_dir_use_symlinks=False,
         token=hf_token,
         cache_dir=cache_folder,
+        force_download=True,
     )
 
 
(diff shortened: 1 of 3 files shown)


AndreaFrancis:
I will merge this PR since looks like the e2e failed check is not related to this code.
I was able to run the tests locally 
Screenshot from 2023-10-09 08-32-05

severo:
> I will merge this PR since looks like the e2e failed check is not related to this code.

https://twitter.com/tmcw/status/1651646574873681925 <= we're not the only ones with half-broken e2e tests... But we have to fix this

PR #1942, opened 2023-10-06
Remove boto3 as dev dependency

albertvillanova:
After the merge of #1418, `boto3` is no longer necessary as a dev dependency in rows and search services.

Note that the dependency on `boto3` is indirectly set by the dependency on `libcommon`.


AndreaFrancis (approved):
Thanks!

PR #1954, opened 2023-10-10
Fix UnexpectedError for non-integer length/offset

albertvillanova:
Fix `UnexpectedError` for non-integer length/offset and raise `InvalidParameterError` instead.


severo (approved):
Good addition!

PR #1965, opened 2023-10-11
Factorize getting required request parameters

albertvillanova:
Factorize getting the required request parameters `dataset`, `config`, `split`, `query` and `where` (from the service admin and the endpoints /rows, /search and /filter) to `libapi`.
Additionally, align their error messages.


AndreaFrancis (approved):
Nice!

Issue #1895, opened 2023-09-29
e2e tests fail with `RuntimeError: Poll timeout` after 2+ hours  of running


polinaeterna:
Started after [this commit](https://github.com/huggingface/datasets-server/commit/3c6c79b6184295d52518a7499cbfe1616077a753) but doesn't seem to be related to it?

```
2023-09-28T20:39:19.0725016Z =========================== short test summary info ============================
2023-09-28T20:39:19.0725373Z FAILED tests/test_11_api.py::test_auth_e2e[public-none-200-None] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0725700Z FAILED tests/test_11_api.py::test_auth_e2e[public-token-200-None] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0726149Z FAILED tests/test_11_api.py::test_auth_e2e[public-cookie-200-None] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0726464Z FAILED tests/test_11_api.py::test_auth_e2e[gated-token-200-None] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0726782Z FAILED tests/test_11_api.py::test_auth_e2e[gated-cookie-200-None] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0727091Z FAILED tests/test_11_api.py::test_endpoint[/splits-dataset] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0727511Z FAILED tests/test_11_api.py::test_endpoint[/splits-config] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0727836Z FAILED tests/test_11_api.py::test_endpoint[/first-rows-split] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0728146Z FAILED tests/test_11_api.py::test_endpoint[/parquet-dataset] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0728454Z FAILED tests/test_11_api.py::test_endpoint[/parquet-config] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0728758Z FAILED tests/test_11_api.py::test_endpoint[/info-dataset] - RuntimeError: Poll timeout
2023-09-28T20:39:19.0729056Z FAILED tests/test_11_api.py::test_endpoint[/info-config] - RuntimeError: Poll timeout
[...]

PR #1968, opened 2023-10-11
clean duckdb index files

AndreaFrancis:
Currently, we share the duckdb  EFS volume for:
- downloads (247 G)
- job runner index processing
capture_d___e__cran_2023-10-03_a___15 14 59

For downloads folder cleaning, job "delete-indexes" used to do the work but it looks like the volume is being filled because of remaining data when a job runner crashes because of long-running (zombie) and https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/_job_runner_with_cache.py#L66 is never called.

This PR add will apply same logic as for existing for hf_datasets_cache but:
- will change the subfolder where the job runner indexes are created under /storage/duckdb-index/job_runner
- will apply an expiration time for  storage/duckdb-index/job_runner folder to 3 hours in prod
- will apply an expiration time for  storage/duckdb-index/downloads folder to 3 days in prod
- will reuse logic in hf_datasets_cache for both cleaning (downloads and job_runner folders)


jobs/cache_maintenance/src/cache_maintenance/clean_files_and_directories.py
```diff
@@ -8,19 +8,14 @@
...
 from typing import Any
 
-from libcommon.storage import StrPath
 
-
-def clean_hf_datasets_cache(hf_datasets_cache: StrPath, expired_time_interval_seconds: int) -> None:
+def clean_files_and_directories(pattern: str, expired_time_interval_seconds: int) -> None:
     """
-    Delete temporary cache directories from job runners with datasets cache.
+    Delete temporary cache directories that matches the given pattern.
```
severo, line 14:
```suggestion
    Delete temporary cache directories that match the given pattern.
```

jobs/cache_maintenance/src/cache_maintenance/config.py
```diff
@@ -30,29 +30,40 @@ def from_env(cls) -> "BackfillConfig":
 
 
 DUCKDB_INDEX_CACHE_DIRECTORY = None
-DUCKDB_INDEX_SUBDIRECTORY = "downloads"
-DUCKDB_INDEX_EXPIRED_TIME_INTERVAL_SECONDS = 10 * 60  # 10 minutes
-DUCKDB_INDEX_FILE_EXTENSION = ".duckdb"
+DUCKDB_INDEX_DOWNLOADS_SUBDIRECTORY = "downloads"
+DUCKDB_INDEX_JOB_RUNNER_SUBDIRECTORY = "job_runner"
```
severo, line 34: nit:

move these hardcoded values to libcommon.constants?

jobs/cache_maintenance/src/cache_maintenance/config.py
```diff
@@ -30,29 +30,40 @@ def from_env(cls) -> "BackfillConfig":
...
     @classmethod
     def from_env(cls) -> "DuckDbConfig":
         env = Env(expand_vars=True)
         with env.prefixed("DUCKDB_INDEX_"):
             return cls(
                 cache_directory=env.str(name="CACHE_DIRECTORY", default=DUCKDB_INDEX_CACHE_DIRECTORY),
-                subdirectory=env.str(name="SUBDIRECTORY", default=DUCKDB_INDEX_SUBDIRECTORY),
-                expired_time_interval_seconds=env.int(
-                    name="EXPIRED_TIME_INTERVAL_SECONDS", default=DUCKDB_INDEX_EXPIRED_TIME_INTERVAL_SECONDS
+                downloads_subdirectory=env.str(
```
severo, line 53: maybe it's simpler and easier to extend to just have one `EXPIRED_TIME_INTERVAL_SECONDS` instead of `DUCKDB_INDEX_DOWNLOADS_EXPIRED_TIME_INTERVAL_SECONDS`, `DUCKDB_INDEX_JOB_RUNNER_EXPIRED_TIME_INTERVAL_SECONDS` and `DATASETS_BASED_EXPIRED_TIME_INTERVAL_SECONDS`. Same for the directory.

And the two values are set as environment variables (have only one action: clean-directory)?

Feel free to ignore if it's too much change

chart/env/prod.yaml
```diff
@@ -181,7 +180,7 @@ mongodbMigration:
 
 cacheMaintenance:
   action: "skip"
-  # ^ allowed values are {backfill, collect-queue-metrics, collect-cache-metrics, delete-indexes, post-messages, skip}
+  # ^ allowed values are {backfill, collect-queue-metrics, collect-cache-metrics, clean-duckdb-index-downloads, clean-duckdb-index-job-runner, post-messages, skip}
```
severo, line 183: I think clean-hf-datasets-cache is missing.

It's a lot of different actions for the same thing. Maybe use only one `clean-directory` action, and differentiate using environment variables? or is it too risky?

jobs/cache_maintenance/src/cache_maintenance/config.py
lhoestq: imo it's nice to have the prefixes

lhoestq:
Btw you might want to include the minor changes from https://github.com/huggingface/datasets-server/pull/1969 to improve logging

jobs/cache_maintenance/src/cache_maintenance/config.py
severo: OK. Let's ignore the comment

jobs/cache_maintenance/src/cache_maintenance/config.py
AndreaFrancis: done

jobs/cache_maintenance/src/cache_maintenance/config.py
AndreaFrancis: I thought it was better to factorize the parameters so that we can use them for other folders quickly (Maybe for statistics?).
Maybe have a look at the last commit? Let me know if I have to rollback.

severo: thanks for refactoring to have only one action, it's exactly what I had in mind

AndreaFrancis:
I think e2e is taking too long again because of https://github.com/huggingface/datasets-server/issues/1895
I will merge, I verified that the current changes are not affecting e2e:
Screenshot from 2023-10-12 11-34-43

PR #1971, opened 2023-10-12
Add date type support in cache responses

lhoestq:
Mongo only supports dates with time (not date alone) because of bson. 
So the `date` python type from `date32` pyarrow type couldn't be saved in mongo.

I fixed it by saving the `date` the same way as `string` objects.

Fix https://github.com/huggingface/datasets-server/issues/86

e.g. in first-rows for [aborruso](https://huggingface.co/aborruso)[/pnrr](https://huggingface.co/datasets/aborruso/pnrr)
```json
{
  "dataset": "__DUMMY_DATASETS_SERVER_USER__/pnrr",
  "config": "default",
  "split": "train",
  "features": [
    {
      "feature_idx": 0,
      "name": "Programma",
      "type": {
        "dtype": "string",
        "_type": "Value"
      }
    },
    ...
    {
      "feature_idx": 50,
      "name": "Data di Estrazione",
      "type": {
        "dtype": "date32",
        "_type": "Value"
      }
    }
  ],
  "rows": [
    {
      "row_idx": 0,
      "row": {
        "Programma": "PNRR",
        ...
        "Data di Estrazione": "2023-06-13"
      },
      "truncated_cells": []
    },
    ...
    ]
}
```

EDIT: also added support for time, timedelta and decimal

I also checked binary but it is already supported, though its format is not ideal (we get a list of integers from mongo)

### DIFF
--- a/libs/libcommon/src/libcommon/resources.py
+++ b/libs/libcommon/src/libcommon/resources.py
@@ -80,7 +80,7 @@ class MongoResource(Resource):
     mongoengine_alias: str
     server_selection_timeout_ms: int = 30_000
 
-    _client: MongoClient = field(init=False)
+    _client: MongoClient = field(init=False, repr=False)
 
     def allocate(self) -> None:
         try:
--- a/libs/libcommon/src/libcommon/simple_cache.py
+++ b/libs/libcommon/src/libcommon/simple_cache.py
@@ -35,6 +35,8 @@ from libcommon.utils import JobParams, get_datetime
 
 
 class DateCodec(TypeCodec):  # type: ignore[misc]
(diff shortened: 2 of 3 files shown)


AndreaFrancis (approved):
Awesome!

severo (commented):
OK.

It means that these four types are now converted to string. Could we document this conversion in the docs? Only openapi maybe. It's important for the client that wants to parse the response.

Do you want to work on it @lhoestq? Else I can also take care of it.

lhoestq:
sure

severo:
Also: did you launch some dataset refreshes?

lhoestq:
nope not yet

lhoestq:
It works well :)

The type si marked as "unknown" though

image

PR #1975, opened 2023-10-12
Make authentication asynchronous

albertvillanova:
Make authentication asynchronous by using `httpx` instead of `requests`.


severo:
OK. What does it fix? (so that I learn something today)

severo (approved):
maybe (in another PR) do the same in services/admin/src/admin/authentication.py?

also: switch from requests to httpx in e2e?

PR #1976, opened 2023-10-12
wip: fsspec storage instead of boto3

AndreaFrancis:


services/worker/tests/job_runners/split/test_first_rows_from_streaming.py
```diff
@@ -135,13 +123,13 @@ def test_compute(app_config: AppConfig, get_job_runner: GetJobRunner, hub_public
...
-        # The assumption is that the dataset exists and is accessible with the token
-        ("gated", False, "InfoError", "FileNotFoundError"),
-        ("private", False, "InfoError", "FileNotFoundError"),
+        # ("jsonl", False, None, None),
+        # ("gated", True, None, None),
+        # ("private", True, None, None),
+        # # should we really test the following cases?
+        # # The assumption is that the dataset exists and is accessible with the token
+        # ("gated", False, "InfoError", "FileNotFoundError"),
+        # ("private", False, "InfoError", "FileNotFoundError"),
```
lhoestq, line 132: re-enable this test

```suggestion
        ("jsonl", False, None, None),
        ("gated", True, None, None),
        ("private", True, None, None),
        # should we really test the following cases?
        # The assumption is that the dataset exists and is accessible with the token
        ("gated", False, "InfoError", "FileNotFoundError"),
        ("private", False, "InfoError", "FileNotFoundError"),
```

libs/libcommon/src/libcommon/storage_options.py
```diff
@@ -2,18 +2,11 @@
...
 from dataclasses import dataclass
 
-from libcommon.s3_client import S3Client
-from libcommon.storage import StrPath
+from libcommon.storage_client import StorageClient
 
 
 @dataclass
-class DirectoryStorageOptions:
+class StorageOptions:
```
lhoestq, line 9: storage_option is already a concept in `fsspec` (it refers to the kwargs passed to a filesystem)

Maybe we can rename this "PublicAssetsStorage" or something like that ?

services/admin/src/admin/routes/recreate_dataset.py
```diff
@@ -58,8 +57,8 @@ async def recreate_dataset_endpoint(request: Request) -> Response:
             deleted_cached_responses = delete_dataset_responses(dataset=dataset)
             if deleted_cached_responses is not None and deleted_cached_responses > 0:
                 # delete assets
-                delete_asset_dir(dataset=dataset, directory=assets_directory)
-                delete_asset_dir(dataset=dataset, directory=cached_assets_directory)
+                cached_assets_storage_client.delete_dataset_directory(dataset)
+                assets_storage_client.delete_dataset_directory(dataset)
```
lhoestq, line 61: would be cool to have a test to make sure that it's working fine

lhoestq (approved):
Nice ! It looks all good to me :)

I just added some minor comments

libs/libcommon/src/libcommon/config.py
```diff
@@ -40,72 +40,72 @@
...
+                folder_name=env.str(name="FOLDER_NAME", default=ASSETS_FOLDER_NAME),
+                storage_protocol=env.str(name="STORAGE_PROTOCOL", default=ASSETS_STORAGE_PROTOCOL),
+                storage_root=env.str(name="STORAGE_ROOT", default=ASSETS_STORAGE_ROOT),
             )
 
 
-S3_BUCKET = "hf-datasets-server-statics"
 S3_ACCESS_KEY_ID = None
 S3_SECRET_ACCESS_KEY = None
-S3_REGION = "us-east-1"
```
severo, line 68: the region is not needed anymore? what occurs if we use a bucket in another region?

libs/libcommon/src/libcommon/config.py
```diff
@@ -40,72 +40,72 @@
 if TYPE_CHECKING:
     from libcommon.processing_graph import ProcessingGraphSpecification
 
-ASSETS_BASE_URL = "assets"
-ASSETS_STORAGE_DIRECTORY = None
-ASSETS_S3_FOLDER_NAME = "assets"
+ASSETS_BASE_URL = "http://localhost/assets"
```
severo, line 43: a lot of changes in the env vars. Could you update the README file?

In particular: I understand that the S3 bucket name must be passed as `ASSETS_STORAGE_ROOT`, right?

libs/libcommon/src/libcommon/config.py
```diff
@@ -40,72 +40,72 @@
...
-CACHED_ASSETS_S3_FOLDER_NAME = "cached-assets"
+CACHED_ASSETS_BASE_URL = "http://localhost/cached-assets"
+CACHED_ASSETS_FOLDER_NAME = "cached-assets"
+CACHED_ASSETS_STORAGE_ROOT = "/storage"
+CACHED_ASSETS_STORAGE_PROTOCOL = "file"
 
 
 @dataclass(frozen=True)
 class CachedAssetsConfig:
     base_url: str = ASSETS_BASE_URL
```
severo, line 95:
```suggestion
    base_url: str = CACHED_ASSETS_BASE_URL
```

services/admin/src/admin/app.py
```diff
@@ -49,20 +47,27 @@ def create_app() -> Starlette:
...
     statistics_cache_directory = init_statistics_cache_dir(app_config.descriptive_statistics.cache_directory)
 
-    if not exists(assets_directory):
-        raise RuntimeError("The assets storage directory could not be accessed. Exiting.")
-
-    if not exists(cached_assets_directory):
-        raise RuntimeError("The cached-assets storage directory could not be accessed. Exiting.")
-
     processing_graph = ProcessingGraph(app_config.processing_graph)
+    cached_assets_storage_client = StorageClient(
```
severo, line 56: what if we try to access the storage when creating the client, to exit in case of a configuration error, as before (see above: `raise RuntimeError("The assets storage directory could not be accessed. Exiting.")`)

tools/docker-compose-base.yml
```diff
@@ -47,7 +47,7 @@ services:
       CACHED_ASSETS_S3_ACCESS_KEY_ID: ${CACHED_ASSETS_S3_ACCESS_KEY_ID-}
       CACHED_ASSETS_S3_SECRET_ACCESS_KEY: ${CACHED_ASSETS_S3_SECRET_ACCESS_KEY-}
       CACHED_ASSETS_S3_REGION: ${CACHED_ASSETS_S3_REGION-us-east-1}
```
severo, line 49: the above variables don't exist anymore, if I understand correctly.

libs/libcommon/src/libcommon/storage_options.py
AndreaFrancis: Thanks! It makes sense, I changed to `PublicAssetsStorage`

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -1,43 +1,26 @@
...
-    makedirs(dir_path, ASSET_DIR_MODE, exist_ok=True)
-    return dir_path
-
-
 def get_url_dir_path(dataset: str, revision: str, config: str, split: str, row_idx: int, column: str) -> str:
     return f"{dataset}/{DATASET_SEPARATOR}/{revision}/{DATASET_SEPARATOR}/{config}/{split}/{str(row_idx)}/{column}"
 
 
-def get_unique_path_for_filename(assets_directory: StrPath, filename: str) -> Path:
-    return Path(assets_directory).resolve() / f"{str(uuid4())}-{filename}"
```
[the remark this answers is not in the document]
AndreaFrancis: It was used before to create the temporal image/audio but it is no more used since the data is written directly with fsspec.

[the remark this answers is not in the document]
severo: excellent!

libs/libcommon/src/libcommon/config.py
AndreaFrancis: I think it is not needed when using fsspec. See https://github.com/fsspec/s3fs/blob/2c074502c2d6a9be0d3f05eb678f4cc5add2e7e5/s3fs/core.py#L171 params.
Anyway, I will add the region as part of client_kwargs to prevent issues.

libs/libcommon/src/libcommon/config.py
AndreaFrancis: > In particular: I understand that the S3 bucket name must be passed as ASSETS_STORAGE_ROOT, right?
Yes!
Sure! I totally forgot to add the new vars to the readme.

services/admin/src/admin/app.py
AndreaFrancis: I added a very basic validation after client creation (try to list files) in case of errors, a StorageClientInitializeError will be thrown.

tools/docker-compose-base.yml
AndreaFrancis: Thanks! I removed them.

services/admin/src/admin/app.py
severo: OK. Just to be sure: will listing files always be quick?

AndreaFrancis: Yes, in this case, we will always have only "cached-assets" and "assets" folders so, it will be super fast to get them listed.

libs/libcommon/src/libcommon/viewer_utils/features.py
```diff
@@ -78,7 +78,7 @@ def image(
             f"but got {str(value)[:300]}{'...' if len(str(value)) > 300 else ''}"
         )
     # attempt to generate one of the supported formats; if unsuccessful, throw an error
-    for ext in [".jpg", ".png"]:
+    for ext, format in [(".jpg", "JPEG"), (".png", "PNG")]:
```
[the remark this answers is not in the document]
AndreaFrancis: Yes, previously we were using the local file_path to save the image but now there are two types of objects:
- fsspec.implementations.local.LocalFileOpener -> When the storage protocol is "file"
- s3fs.core.S3File -> When the storage protocol is "s3"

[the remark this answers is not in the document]
severo: OK, thanks

libs/libcommon/src/libcommon/config.py
```diff
@@ -40,72 +40,75 @@
...
     @classmethod
     def from_env(cls) -> "S3Config":
         env = Env(expand_vars=True)
         with env.prefixed("S3_"):
             return cls(
-                bucket=env.str(name="BUCKET", default=S3_BUCKET),
                 access_key_id=env.str(name="ACCESS_KEY_ID", default=S3_ACCESS_KEY_ID),
                 secret_access_key=env.str(name="SECRET_ACCESS_KEY", default=S3_SECRET_ACCESS_KEY),
-                region=env.str(name="REGION", default=S3_REGION),
+                region_name=env.str(name="REGION_NAME", default=S3_REGION_NAME),
```
severo, line 86: you have to fix the name in the Helm chart (and maybe the docker files?)

severo (approved):
a detail

libs/libcommon/src/libcommon/config.py
severo: and also ensure the README is right

AndreaFrancis:
e2e is failed because of https://github.com/huggingface/datasets-server/pull/2027

libs/libcommon/src/libcommon/config.py
AndreaFrancis: done

services/admin/src/admin/routes/recreate_dataset.py
AndreaFrancis: I added a test for recreate_dataset https://github.com/huggingface/datasets-server/pull/2028, then, I will update it here but using fsspec instead

services/admin/src/admin/app.py
lhoestq: Another quick validation is simply to to call `fs.exists()` on the directory

services/admin/src/admin/routes/recreate_dataset.py
AndreaFrancis: Done

PR #1978, opened 2023-10-13
Make admin authentication asynchronous

albertvillanova:
Make admin service authentication asynchronous by using `httpx` instead of `requests`.

Related to:
- #1975

### DIFF
--- a/services/admin/src/admin/authentication.py
+++ b/services/admin/src/admin/authentication.py
@@ -3,27 +3,10 @@
 
 from typing import Literal, Optional
 
-import requests
+import httpx
+from libapi.authentication import RequestAuth
 from libapi.exceptions import ExternalAuthenticatedError, ExternalUnauthenticatedError
-from requests import PreparedRequest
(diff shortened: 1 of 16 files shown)


services/admin/src/admin/authentication.py
```diff
@@ -50,7 +33,8 @@ def auth_check(
     if organization is None or external_auth_url is None:
         return True
     try:
-        response = requests.get(external_auth_url, auth=RequestAuth(request), timeout=hf_timeout_seconds)
+        async with httpx.AsyncClient() as client:
+            response = await client.get(external_auth_url, auth=RequestAuth(request), timeout=hf_timeout_seconds)
```
severo, line 37: this means that we now pass the "cookie", while we didn't before. I think it's OK though

severo (approved):

services/admin/src/admin/authentication.py
albertvillanova: Yes, I thought it won't break anything...

PR #1979, opened 2023-10-13
Remove requests as dependency of libapi

albertvillanova:
Remove `requests` as `libapi` dependency by using `httpx` instead.


severo (approved):
nice

PR #1982, opened 2023-10-13
Remove unused httpfs duckdb extension

polinaeterna:
it's not used for files download anymore, we use `huggingface_hub` directly

**+ I added some logging to check what's going on with the stats computing in this PR to make it fast, to be removed!**


severo (approved):

AndreaFrancis (approved):
Great!

PR #1984, opened 2023-10-16
Run duckdb queries async in separate thread

albertvillanova:
Run `duckdb` queries asynchronously in a separate thread.

### DIFF
--- a/services/search/src/search/routes/filter.py
+++ b/services/search/src/search/routes/filter.py
@@ -6,6 +6,7 @@ import re
 from http import HTTPStatus
 from typing import Optional
 
+import anyio
 import duckdb
 import pyarrow as pa
 from datasets import Features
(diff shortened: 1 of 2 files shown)


severo (approved):
I don't know much about the threads. Does it use more CPUs? Should we adapt how we setup the CPU resources for the pods?

albertvillanova:
The main reason for this PR is that DuckDB queries are not asynchronous and they block the event loop. By running the non-async functions in a separate thread, we avoid blocking the event loop in the main thread and we allow run the rest of async functions.

In relation with the the CPU usage:
- the async framework uses a thread pool that automatically sets the maximum number of workers accordingly with the number of processors of the machine: the threads are enqueued in this thread pool
- I am assuming that the DuckDB SELECT queries are IO-bound: this seems reasonable to me (I would say SELECT queries to a DB are IO-bound in general) and I searched information for DuckDB in particular but could not find any confirmation/refutation

PR #1986, opened 2023-10-16
adding clean stats cache cron job

AndreaFrancis:
Same as we do now for `duckdb-index` and `hf-datasets-cache` (cleaning periodically obsolete/old files), we can sometimes keep files in stats because of job runner crash or zombie killer.
Adding a cron job to delete files/folders in /storage/stats-cache that are older than 3 hours ago.

### DIFF
--- a/chart/templates/cron-jobs/clean-stats-cache/job.yaml
+++ b/chart/templates/cron-jobs/clean-stats-cache/job.yaml
@@ -1,0 +1,27 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2023 The HuggingFace Authors.
+
+{{- if and .Values.images.jobs.cacheMaintenance .Values.cleanStatsCache.enabled }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  labels: {{ include "labels.cleanStatsCache" . | nindent 4 }}
+  name: "{{ include "name" . }}-job-clean-stats-cache"
+  namespace: {{ .Release.Namespace }}
+spec:
+  schedule: {{ .Values.cleanStatsCache.schedule | quote }}
+  jobTemplate:
+    spec:
+      ttlSecondsAfterFinished: 3600
+      template:
+        spec:
+          restartPolicy: OnFailure
+          {{- include "dnsConfig" . | nindent 10 }}
+          {{- include "image.imagePullSecrets" . | nindent 6 }}
+          nodeSelector: {{ toYaml .Values.cleanStatsCache.nodeSelector | nindent 12 }}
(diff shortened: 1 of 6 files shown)


chart/templates/cron-jobs/clean-stats-cache/job.yaml
```diff
@@ -0,0 +1,27 @@
...
+    spec:
+      ttlSecondsAfterFinished: 3600
+      template:
+        spec:
+          restartPolicy: OnFailure
+          {{- include "dnsConfig" . | nindent 10 }}
+          {{- include "image.imagePullSecrets" . | nindent 6 }}
+          nodeSelector: {{ toYaml .Values.cleanStatsCache.nodeSelector | nindent 12 }}
+          tolerations: {{ toYaml .Values.cleanStatsCache.tolerations | nindent 12 }}
+          containers: {{ include "containerCleanHfDatasetsCache" . | nindent 12 }}
```
lhoestq, line 23:
```suggestion
          containers: {{ include "containerCleanStatsCache" . | nindent 12 }}
```

lhoestq (approved):
Apart from my first comment it looks all good to me ! cc @polinaeterna as well

polinaeterna (commented):
thank you! all good for me except for I'm not sure about naming - `cleanDuckdbIndexJobRunner` and `cleanStatsCache` do basically the same thing (for their corresponding workers) but named differently. Maybe name the latter one `cleanStatsJobRunner` (or even `cleanDescriptiveStatisticsJobRunner` so that @severo is happy 😄 )

AndreaFrancis:
> thank you! all good for me except for I'm not sure about naming - cleanDuckdbIndexJobRunner and cleanStatsCache do basically the same thing (for their corresponding workers) but named differently. Maybe name the latter one cleanStatsJobRunner (or even cleanDescriptiveStatisticsJobRunner so that @severo is happy 😄 )

I would like to leave it as it is right now, maybe will have to standardize it in another PR? mostly it was because of folder names (clean+folder_name):
- /storage/duckdb-index/downloads -> cleanDuckdbIndexDownloads
- /storage/duckdb-index/job_runner -> cleanDuckdbIndexJobRunner
- /storage/hf-datasets-cache -> cleanHfDatasetsCache
- /storage/stats-cache -> cleanStatsCache

Maybe `duckdb-index` should be renamed to `duckdb-index-cache` but I would rather handle it in another PR, wdyt?

polinaeterna:
@AndreaFrancis okay, sure! as soon as renaming doesn't add any complexity further, I don't mind :) and anyway it's not that important

Issue #1981, opened 2023-10-13
Overwrite image and audio cached assets when necessary


lhoestq:
Currently we never overwrite image and audio cached assets, which can lead to outdated data in the viewer like here:
https://huggingface.co/datasets/ccmusic-database/instrument_timbre_eval/discussions/1

In particular we have `overwrite=False` here:

https://github.com/huggingface/datasets-server/blob/b84c8210ec023664c4780ab348b0232468afe116/libs/libapi/src/libapi/response.py#L36-L42

cc @AndreaFrancis @severo

AndreaFrancis:
It depends on the TTL for cached-assets (which is 1 day) so, the files will be refreshed after that time.
If we are forced to overwrite each time, it isn't worth having a CloudFront layer right?
If the dataset had a change in a lower interval than the TTL, I think we should recreate all related assets to each processing step:
- first-rows should delete all the assets before processing (so that the obsolete files will be refreshed)
- config-parquet should delete all the cached-assets before processing

lhoestq:
I think first-rows is fine since it has `overwrite=True`:

https://github.com/huggingface/datasets-server/blob/b84c8210ec023664c4780ab348b0232468afe116/services/worker/src/worker/job_runners/split/first_rows_from_streaming.py#L283-L289

And yes config-parquet-metadata (which is the job that provides /rows) can take care of removing the outdated cached assets, good idea

AndreaFrancis:
Related to https://github.com/huggingface/datasets-server/issues/1823
Probably, if we recreate the dataset it won't work because https://github.com/huggingface/datasets-server/blob/main/services/admin/src/admin/routes/recreate_dataset.py#L61 is pointing still to the old cached-assets EFS but it should delete all files in S3.
Maybe I can add it as part of https://github.com/huggingface/datasets-server/pull/1976

AndreaFrancis:
https://huggingface.co/datasets/ccmusic-database/instrument_timbre_eval is now serving the correct files. What I saw is that there is a time period for CloudFront https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html 
`By default, each file automatically expires after 24 hours`

So, if we change/update an S3 file in a lower time interval, for assets and cached-assets, we will have to invalidate the cache or something like this https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html

AndreaFrancis:
[...]

PR #1988, opened 2023-10-16
adding revision for assets creation path

AndreaFrancis:
Fix for https://github.com/huggingface/datasets-server/issues/1981


AndreaFrancis:
e2e execution evidence:
Screenshot from 2023-10-16 17-00-16
Screenshot from 2023-10-16 17-00-33

lhoestq (approved):
LGTM ! and the already existing assets will keep working indeed since we store the full URLs :)

lhoestq:
> Since all our cache records in [cachedResponsesBlue](https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/simple_cache.py#L129) collection have a revision value, should we change it to a required and non Optional same as in [Jobs](https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/queue.py#L178)?

Yes I think so

> For assets, if a new version is processed, we will be storing the previous files in S3, we need to find a way to clean them or, maybe in https://github.com/huggingface/datasets-server/issues/1981 it will make more sense since a "versioned" approach was proposed.

The revision for assets is set to "main" so it should overwrite the files every time, but IIUC it will have the same issue with cloudfront still serving the old files ? In this case we should use the right revision (git commit sha) and have a mechanism to clean the outdated directory in the first-rows jobs maybe ? or a cron job, not sure what's the easiest

AndreaFrancis:
> Yes I think so

I would like to send another PR before merging this one to ensure that now dataset "version" is mandatory, so that I can remove the optionals here.


> The revision for assets is set to "main" so it should overwrite the files every time, but IIUC it will have the same issue with cloudfront still serving the old files ? In this case we should use the right revision (git commit sha) and have a mechanism to clean the outdated directory in the first-rows jobs maybe ? or a cron job, not sure what's the easiest

No, for assets it will apply the revision as well https://github.com/huggingface/datasets-server/pull/1988/files#diff-4072a6f63a7fdcc185cb1170651d53216c3db8db2b5d2374db9e7974d648e2e5R229 so, no problem with CloudFront.
About the mechanism to cleaning outdated directories yes, I think I can propose another PR to clean them periodically until the Versioned DAG approach is implemented (https://github.com/huggingface/datasets-server/issues/1823).

severo:
> I would like to send another PR before merging this one to ensure that now dataset "version" is mandatory, so that I can remove the optionals here.

good idea

docs/source/openapi.json
```diff
@@ -2270,12 +2270,12 @@
                           "row_idx": 0,
                           "row": {
                             "imageA": {
-                              "src": "https://datasets-server.huggingface.co/assets/huggan/horse2zebra/--/default/train/0/imageA/image.jpg",
+                              "src": "https://datasets-server.huggingface.co/assets/huggan/horse2zebra/main/default/train/0/imageA/image.jpg",
```
severo, line 2273: If we do this, we can no longer parse the URL. The `--` separator ensures we can get the dataset name, canonical or with a namespace.

If we keep it, the URL would be:

```
                              "src": "https://datasets-server.huggingface.co/assets/huggan/horse2zebra/--/main/default/train/0/imageA/image.jpg",
```

We can also remove it if it is useless, but let's be sure it does not break anything

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -31,8 +29,8 @@ def get_and_create_dir_path(assets_directory: StrPath, url_dir_path: str) -> Pat
     return dir_path
 
 
-def get_url_dir_path(dataset: str, config: str, split: str, row_idx: int, column: str) -> str:
-    return f"{dataset}/{DATASET_SEPARATOR}/{config}/{split}/{str(row_idx)}/{column}"
+def get_url_dir_path(dataset: str, revision: str, config: str, split: str, row_idx: int, column: str) -> str:
+    return f"{dataset}/{revision}/{config}/{split}/{str(row_idx)}/{column}"
```
severo, line 33: do we explicitly support revisions with `/` in them, like `refs/convert/parquet`? I think we will only have `main` and commit SHAs, but I wonder if it would break something or not if we have this kind of revision (it would create more sub-directories, and we will not be able to parse the revision from the URL anymore, unless we add DATASET_SEPARATOR before and after it).

services/rows/src/rows/routes/rows.py
```diff
@@ -123,29 +116,10 @@ async def rows_endpoint(request: Request) -> Response:
...
-                                dataset=dataset,
-                                cached_assets_directory=cached_assets_directory,
-                                keep_first_rows_number=keep_first_rows_number,
-                                keep_most_recent_rows_number=keep_most_recent_rows_number,
-                                max_cleaned_rows_number=max_cleaned_rows_number,
-                            )
                 with StepProfiler(method="rows_endpoint", step="transform to a list"):
                     response = create_response(
                         dataset=dataset,
+                        revision=revision,  # type: ignore
```
severo, line 122: why the "type: ignore"? It feels weird: if it can be None, we should generate an error instead

services/worker/src/worker/job_runners/dataset/dataset_job_runner.py
```diff
@@ -30,7 +30,7 @@ def __init__(
         if job_info["params"]["revision"] is None:
             raise ParameterMissingError("'revision' parameter is required")
         self.dataset = job_info["params"]["dataset"]
-        self.revision = job_info["params"]["revision"]
+        self.dataset_git_revision = job_info["params"]["revision"]
```
severo, line 33: was it a bug? does it mean we should fix some cache entries?

AndreaFrancis: Yes, it was a bug. But I think it does not affect the cache entries since the field self.dataset_git_revision was not used anywhere (not in job runner nor in cache record).
I started using it in this PR for first-rows-from-streaming and first-rows-from-parquet job runners.

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: Not sure which type of revisions we could have (so far we don't have any in the db with `/` but I think I will add the separator after and before to ensure we don't break anything.

services/rows/src/rows/routes/rows.py
AndreaFrancis: Makes sense. In this case, we use an Optional revision (initialized at the beginning of the function) in case of error but since we use `create_response` only when `revision = rows_index.revision` is success I think we don't need other validation (Since dataset_git_revision is mandatory now and it is read from the cache in https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/parquet_utils.py#L290
`mypy` was alerting because we had `try_backfill_dataset_then_raise` block before calling  `create_response` (that is why it didn't fail for /search and /filter) but I fixed the order of execution.

severo (approved):

PR #1991, opened 2023-10-17
Limit max RAM that duckdb can use

polinaeterna:
Current value of `max_memory` in prod is 105.7Gb (despite the pod has only 34Gb but it's set automatically by duckdb as 80% of ram and ram of the node in total is 128Gb)
image
I want to try to lower it manually since we assume there might be some memory issues with duckdb causing the  `split-descriptive-statistics` process to stuck at the step of loading data into in-memory db.
(also want to try to load data to local db file, this is the only difference with `duckdb-index` step)

### DIFF
--- a/services/worker/src/worker/job_runners/split/descriptive_statistics.py
+++ b/services/worker/src/worker/job_runners/split/descriptive_statistics.py
@@ -481,7 +481,13 @@ def compute_descriptive_statistics_response(
     logging.info(f"Original number of threads={n_threads}")
     con.sql("SET threads TO 8;")
     n_threads = con.sql("SELECT current_setting('threads')").fetchall()[0][0]
-    logging.info(f"Number of threads={n_threads}")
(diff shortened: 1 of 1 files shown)


lhoestq (approved):

severo:
Maybe report to duckdb, no?

severo:
Also, at some point, use an env var to set the value

polinaeterna:
> Maybe report to duckdb, no?

@severo I don't know, it's kinda an expected default behaviour, 80% of RAM size

severo:
hmmm, OK. The pod only has "requested/limit" RAM, but the Python app is unaware of that limit. Thus DuckDB gets the total RAM of the node and sets the maximum usage to 80% of that RAM, which is too much for the pod, and it OOMs. @rtrompier @XciD maybe you have recommendations on how to manage this?

Issue #1742, opened 2023-08-29
Support Search for datasets on the first 5GB of big datasets


lhoestq:
If a dataset is in parquet format the viewer shows the full dataset but search is disabled.

In this case it would be nice to support search anyway, at least on the first 5GB.

lhoestq:
related to https://github.com/huggingface/datasets-server/issues/1758

lhoestq:
I think I'll add new jobs `duckdb_index_size` for the front end to know whether `search` can be performed on the full dataset or only on the first 5GB.

IMO the viewer should display "(First 5GB)" if the index is partial - even if the parquet export is full.

It can also use the `num_rows` from indexing instead of the `num_rows` from the parquet export to determine the number of pages and to display in the configs and splits dropdown. What do you think @severo ?

severo:
I agree: we now have two sources of truth depending on the service (/rows, or /search+/filter), but we have only the size for the first one (parquet). We should have the information for both (size, number of rows, if it is partial) for Parquet and DuckDB.
I don't know how we should make the API evolve, though. Do you want to work on a proposal?

lhoestq:
I'm on it ;)

PR #1992, opened 2023-10-17
Partial duckdb index

lhoestq:
For big datasets, instead of using all the parquet files, use only the first ones until the limit of 5GB is reached.
It's used for both `search` and `filter`.

## Indexing

I added three fields in the duckdb-index job:
- partial (if it used a partial parquet export or if it didn't index the full parquet export)
- num_rows
- num_bytes

## New jobs

I also added two jobs and their responses can be used in the front-end instead of the `size` endpoint:
- config-duckdb-index-size
- dataset-duckdb-index-size

Both jobs return the same response format as the config-size and the dataset-size jobs, but with information about the size of the indexed data instead of the size of the parquet-exported data.

## Update of search and filter

I also added the `partial` field in the responses of `/search` and `/filter`, in case they use a partial index.

## Index file

The file name is `index.duckdb` if it indexes all the rows from the parquet export, otherwise it's named `partial-duckdb.index`. It is placed in the same folder as the parquet export in `refs/convert/parquet`.

Fix https://github.com/huggingface/datasets-server/issues/1742

TODO
- [x] tests
- [x] docs
- [x] openapi
- [x] update /search and /filter for partial indexes


severo:
Reviewing...

About the new jobs: from the frontend point of view, I prefer calling fewer endpoints with more info in them. Would it be OK to move this info to the size cache entries?
An alternative would be to create another endpoint, dedicated to the Hub frontend, that contains all the info we need for a split on page load

wdyt @lhoestq?

services/worker/src/worker/config.py
```diff
@@ -273,9 +273,7 @@ def from_env(cls) -> "DuckDbIndexConfig":
                 committer_hf_token=env.str(name="COMMITTER_HF_TOKEN", default=DUCKDB_INDEX_COMMITTER_HF_TOKEN),
                 target_revision=env.str(name="TARGET_REVISION", default=DUCKDB_INDEX_TARGET_REVISION),
                 url_template=env.str(name="URL_TEMPLATE", default=DUCKDB_INDEX_URL_TEMPLATE),
-                max_parquet_size_bytes=env.int(
-                    name="MAX_PARQUET_SIZE_BYTES", default=DUCKDB_INDEX_MAX_PARQUET_SIZE_BYTES
-                ),
+                max_dataset_size=env.int(name="MAX_DATASET_SIZE", default=DUCKDB_INDEX_MAX_DATASET_SIZE),
```
severo, line 276: could you update the README?

lhoestq:
> About the new jobs: from the frontend point of view, I prefer calling fewer endpoints with more info in them. Would it be OK to move this info to the size cache entries?

I also like the idea of calling fewer endpoints. But in that case shouldn't it call something like `dataset-and-index-size` rather than `size` to get the size of the dataset and index ?

libs/libapi/tests/test_duckdb.py
```diff
@@ -0,0 +1,58 @@
...
+    duckdb_index_file_directory.mkdir(parents=True, exist_ok=True)
+
+    dataset = "dataset"
+    revision = "revision"
+    target_revision = "refs/convert/parquet"
+    config = "config"
+    split = "split"
+    default_filename = "index.duckdb"
+    split_directory = f"partial-{split}" if partial_split else split
+    filename = f"partial-{default_filename}" if partial_index else default_filename
```
severo, line 28: OH ok... So we can have:

`train/index.duckdb`: normal
`train/partial-index.duckdb`: the dataset is originally in Parquet format and contains more than 5GB data: the Parquet "conversion" is complete, while the duckdb index is only on the first 5GB
`partial-train/partial-index.duckdb`: if the Parquet files have been created only on the first 5GB of the data
`partial-train/index.duckdb`: ... we should never have this, right?

services/search/src/search/routes/filter.py
```diff
@@ -123,15 +123,22 @@ async def filter_endpoint(request: Request) -> Response:
                             error_code=duckdb_index_cache_entry["error_code"],
                             revision=revision,
                         )
+
+                    # check if the index is on the full dataset or if it's partial
```
severo, line 127: we have the info in `duckdb_index_cache_entry`, no?

libs/libapi/src/libapi/response.py
```diff
@@ -58,3 +62,52 @@ def create_response(
         "num_rows_total": num_rows_total,
         "num_rows_per_page": MAX_NUM_ROWS_PER_PAGE,
     }
+
+
+def create_maybe_partial_response(
```
severo, line 67: wouldn't it make sense to add the partial field to PaginatedResponse?

severo (commented):
comments on first part of the review

libs/libapi/tests/test_duckdb.py
lhoestq: `partial-train/index.duckdb` we can have this if the index contains all the data from the partial parquet export

services/search/src/search/routes/filter.py
lhoestq: not always - I added the `partial` field for this PR and didn't increase the job version

libs/libapi/src/libapi/response.py
lhoestq: No because /rows always returns complete data (the rows from 0 to 100 are complete, whereas a search may ignore some rows because it only scans the first 5GB)

libs/libapi/tests/test_duckdb.py
severo: OK. Hence, is it `partial-train/partial-index.duckdb` that will never appear?

services/search/src/search/routes/filter.py
severo: yes, but if I understand correctly, if the duckdb has the `partial` keyword in its name, it also has the field in `duckdb_index_cache_entry`

libs/libapi/src/libapi/response.py
severo: We can set `partial: true` for `/rows`. Ideally, all the endpoints that return "rows" would use the same format.

libs/libapi/tests/test_duckdb.py
severo: well, it should not occur, but... it could, if the parquet files have been created on the first 5GB of data, and the result is N parquet files, where at most N-1 parquet files sum at least 5GB. I don't think it's possible, but... OK

severo:
> I also like the idea of calling fewer endpoints. But in that case shouldn't it call something like dataset-and-index-size rather than size to get the size of the dataset and index ?

OK, we can go with these new steps, and then I can create another one that synthesizes everything. It would keep things atomic, and these jobs are ultra-quick and the responses don't occupy much storage.

services/worker/src/worker/dtos.py
```diff
@@ -169,6 +169,63 @@ class ConfigSizeResponse(TypedDict):
     partial: bool
 
 
+class SplitDuckdbIndex(SplitHubFile):
+    features: Optional[dict[str, Any]]
+    has_fts: bool
+    # The 2 following fields can be None in old cache entries
```
severo, line 175:
```suggestion
    # The following fields can be None in old cache entries
```

services/worker/src/worker/job_runners/config/duckdb_index_size.py
```diff
@@ -0,0 +1,143 @@
...
+        dataset=dataset,
+        config=config,
+    )
+    content = split_names_response.response["content"]
+    if "splits" not in content:
+        raise PreviousStepFormatError("Previous step did not return the expected content: 'splits'.")
+
+    try:
+        total = 0
+        pending = 0
```
AndreaFrancis, line 54: Looks like this var is never used

services/search/src/search/routes/filter.py
lhoestq: If it doesn't have the `partial` keyword in its name, the split directory name can still have it have it

libs/libapi/src/libapi/response.py
lhoestq: So `/rows` would return `partial: false` all the time.

Ah actually there is one case where it can return `partial: true`: if we request the last page and if we can't show the rest of the dataset.

Let me add `partial` to `/rows` too then

lhoestq: I think I'll just add `partial: true` for all pages of `/rows` if the dataset is partially exported

lhoestq:
I took all your comments into account :)

One big change is I added `partial` to `/rows`. It now has `partial: true` if the parquet export for the split is partial

services/search/src/search/routes/filter.py
severo: ok

services/worker/src/worker/dtos.py
```diff
@@ -169,6 +169,63 @@ class ConfigSizeResponse(TypedDict):
     partial: bool
 
 
+class SplitDuckdbIndex(SplitHubFile):
+    features: Optional[dict[str, Any]]
+    has_fts: bool
+    # The following fields can be None in old cache entries
```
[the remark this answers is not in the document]
lhoestq: I don't think we need an issue or a PR. We should keep the Optional unless we ever need to increase the indexing version.

lhoestq:
I factorized the partial checks, added the `bytes` suffix when needed and added a migration to make the `partial` field non-optional (though the `num_rows` and `num_bytes` will still be optional)

jobs/mongodb_migration/src/mongodb_migration/migrations/_20231106193200_cache_add_partial_field_in_split_duckdb_index.py
```diff
@@ -0,0 +1,60 @@
...
+        db = get_db(CACHE_MONGOENGINE_ALIAS)
+        db[CACHE_COLLECTION_RESPONSES].update_many(
+            {
+                "kind": "split-duckdb-index",
+                "http_status": 200,
+                "content.partial": {"$exists": False},
+            },
+            {
+                "$set": {
+                    "content.partial": True,
```
severo, line 31: isn't it False?

severo:
and a missing rename in worker

jobs/mongodb_migration/src/mongodb_migration/migrations/_20231106193200_cache_add_partial_field_in_split_duckdb_index.py
lhoestq: Good catch ! And actually it should be `None` since it may have indexed a partial parquet dump so we have to keep the `Optional`

severo (approved):

PR #2001, opened 2023-10-19
Disable dataset scripts

lhoestq:
The `config-parquet-and-info` step will now raise a `DatasetWithScriptNotSupportedError` for datasets with a script, except those in the allow list. This will prevent users from running arbitrary code using dataset scripts.

The error message is shown to the user on the website and it says
```python
            raise DatasetWithScriptNotSupportedError(
                "The dataset viewer doesn't support this dataset because it runs "
                "arbitrary python code. Please open a discussion in the discussion tab "
                "if you think this is an error and tag @lhoestq and @severo."
            )
```

The allow list is hardcoded for now: `DATASET_SCRIPTS_ALLOW_LIST = ["canonical"]`
The keyword "canonical" means all the datasets without namespaces.

We can add other datasets to the allow list, and it supports `fnmatch`, for example to support all the datasets from `huggingface` we can add `huggingface/*` to the allow list.

cc @severo @XciD

### DIFF
--- a/services/worker/src/worker/job_runners/config/parquet_and_info.py
+++ b/services/worker/src/worker/job_runners/config/parquet_and_info.py
@@ -1129,7 +1129,7 @@ def compute_config_parquet_and_info_response(
         dataset_scripts_allow_list (`list[str]`):
             List of datasets for which we support dataset scripts.
             Unix shell-style wildcards also work in the dataset name for namespaced datasets, for example `some_namespace/*` to refer to all the datasets in the `some_namespace` namespace.
-            The keyword `{{ALL_DATASETS_WITH_NO_NAMESPACE}}` can be used to refer to all the datasets without namespaces.
+            The keyword `{{ALL_DATASETS_WITH_NO_NAMESPACE}}` can be used to refer to all the datasets without namespace.
     Returns:
         `ConfigParquetAndInfoResponse`: An object with the config_parquet_and_info_response
           (dataset info and list of parquet files).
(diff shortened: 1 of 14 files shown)


services/worker/src/worker/utils.py
```diff
@@ -372,3 +378,35 @@ def is_dataset_script_error() -> bool:
...
+    def raise_unsupported_dataset_with_script_or_init(
+        self: HubDatasetModuleFactoryWithScript,
+        name: str,
+        revision: Optional[Union[str, datasets.Version]] = None,
+        download_config: Optional[DownloadConfig] = None,
+        download_mode: Optional[Union[datasets.DownloadMode, str]] = None,
+        dynamic_modules_path: Optional[str] = None,
+    ) -> None:
+        for allowed_pattern in allow_list:
+            if (allowed_pattern == "canonical" and "/" not in name) or fnmatch(name, allowed_pattern):
```
severo, line 395: maybe a more specific keyword instead of "canonical", as this org exists? (https://huggingface.co/Canonical)

even if it's handled by your code, it would feel less bug-prone.

You can use a forbidden character for example (see https://github.com/huggingface/moon-landing/blob/main/server/lib/Names.ts - internal)

severo (commented):
we need to raise the exception in every job that uses the datasets library to run the user script

severo (approved):

services/worker/src/worker/utils.py
lhoestq: renamed `{{ALL_DATASETS_WITH_NO_NAMESPACE}}`

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -88,6 +89,8 @@
 MAX_FILES_PER_DIRECTORY = 10_000  # hf hub limitation
 MAX_OPERATIONS_PER_COMMIT = 500
 
+DATASET_SCRIPTS_ALLOW_LIST = ["canonical"]
```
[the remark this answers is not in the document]
lhoestq: I used an environment variable that is used in the common config for workers

severo (approved):

PR #2011, opened 2023-10-25
Remove no_max_size_limit_datasets (no more full parquet conversion for Open-Orca/OpenOrca to fix the viewer)

lhoestq:
The full Open-Orca/OpenOrca is too big to be processed correctly by most services, so let it work like all the other datasets (use the first 5GB)


severo (approved):
good idea

PR #2013, opened 2023-10-25
Factorize getting request parameters

albertvillanova:
Factorize getting request parameters for both required and non-required ones, by using the function `get_request_parameter` and merging `get_required_request_parameter` into it, with the addition of the parameters "required" and "default".

Use `get_request_parameter` to get the "cursor", "priority" and "all" parameters.


severo (approved):
Better, thanks

Capture d’écran 2023-10-25 à 18 00 24

🧌 nice refactoring

PR #2016, opened 2023-10-26
add cron job to clean duckdb index cache

AndreaFrancis:
We have a cache folder in `/storage/duckdb-index`, this folder is used for cached downloads in `/search` and `/filter` while calling `  hf_hub_download` https://github.com/huggingface/datasets-server/blob/ef8a1cb34ba661b859962ac8d6c138374a0b0190/libs/libapi/src/libapi/duckdb.py#L78 (which needs a cache directory).
I noticed that we have old files in `/storage/duckdb-index/cache` which are using about 185G:
image

We can apply the same cleaning as we do for other folders so I added a new cron job.
I think another approach to avoid using `hf_hub_download` cache is to use `hf_transfer` directly, but I will investigate more about the advantages, meanwhile, the cron job will avoid increasing the storage usage.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -18,7 +18,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.20.0
+version: 1.21.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -213,7 +213,7 @@ backfill:
 cleanDuckdbIndexCache:
   nodeSelector:
     role-datasets-server: "true"
-  expiredTimeIntervalSeconds: 10_800 # 3 hours
+  expiredTimeIntervalSeconds: 259_200 # 3 days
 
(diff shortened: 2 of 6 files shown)


lhoestq:
The index files are the ones used by `/search` and `/filter` ? If so maybe it's fine to keep them longer on disk to not have to redownload them all the time and keep `/search` and `/filter` fast ?

AndreaFrancis:
> The index files are the ones used by /search and /filter ? If so maybe it's fine to keep them longer on disk to not have to redownload them all the time and keep /search and /filter fast ?

Yes but, it is like we have a copy of the files first in the cache (This is the way hf_hub_download works) and then, we move the file to /storage/duckdb-index/downloads so, maybe I can enable local_dir_use_symlinks=True https://huggingface.co/docs/huggingface_hub/package_reference/file_download#huggingface_hub.hf_hub_download.local_dir_use_symlinks and configure a long expiration time for the folder (like 3 days same as /downloads https://github.com/huggingface/datasets-server/blob/main/chart/env/prod.yaml#L221)

severo (approved):
OK! The only thing to decide is how long we should keep the files (@lhoestq mentioned having them for more than. 3 hours)

AndreaFrancis:
> OK! The only thing to decide is how long we should keep the files (@lhoestq mentioned having them for more than. 3 hours)

I changed it to 3 days for now, let's see how it goes and we can set it a new value later.

PR #2017, opened 2023-10-26
Use get_request_parameter in endpoint

albertvillanova:
Refactor the "endpoint" route in the "api" service and use libapi `get_request_parameter`.

### DIFF
--- a/services/api/src/api/routes/endpoint.py
+++ b/services/api/src/api/routes/endpoint.py
@@ -13,6 +13,7 @@ from libapi.exceptions import (
     MissingRequiredParameterError,
     UnexpectedApiError,
 )
+from libapi.request import get_request_parameter
 from libapi.utils import (
     Endpoint,
     are_valid_parameters,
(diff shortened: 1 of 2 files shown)


services/api/src/api/routes/endpoint.py
```diff
@@ -296,3 +183,30 @@ async def processing_step_endpoint(request: Request) -> Response:
...
+            and (error_message := "Parameters 'dataset' and 'config' are required")
+            and are_valid_parameters([dataset, config])
+        ):
+            return dataset, config, None, input_type
+        elif (
+            input_type == "dataset"
+            and (error_message := "Parameter 'dataset' is required")
+            and are_valid_parameters([dataset])
+        ):
+            return dataset, None, None, input_type
```
lhoestq, line 211: Not a big fan of the `and (error_message := "Parameter 'dataset' is required")` syntax. Maybe let's update `error_message` more explicitly ?

```suggestion
    for input_type in input_types:
        if input_type == "split":
            if are_valid_parameters([dataset, config, split]):
                return dataset, config, split, input_type
            else:
                error_message = "Parameters 'dataset', 'config' and 'split' are required"
        elif input_type == "config":
            if are_valid_parameters([dataset, config]):
                return dataset, config, None, input_type
            else:
                error_message = "Parameters 'dataset' and 'config' are required"
        elif input_type == "dataset":
            if are_valid_parameters([dataset]):
                return dataset, None, None, input_type
            else:
                error_message = "Parameter 'dataset' is required"
```

lhoestq (approved):
Thanks for simplifying this ! Just one nit:

services/api/src/api/routes/endpoint.py
albertvillanova: Thanks, @lhoestq. I totally agree. :hugs:

PR #2018, opened 2023-10-27
Download duckdb index async in separate thread

albertvillanova:
Download `duckdb` index asynchronously in separate thread, and make `get_index_file_location_and_download_if_missing` asynchronous.


severo (approved):
gooood

PR #2021, opened 2023-10-27
Remove duckdb as libapi dependency

albertvillanova:
Remove `duckdb` from libapi by moving the function `duckdb_connect` to the "search" service. 

In the future, if this function is required by other components, I would suggest to create a new duckdb-specific lib, so that we decouple API and DuckDB functionalities.

Additionally update all components that depend on `libapi` and remove their dependency on `duckdb` if they do not need it. These are:
- admin
- api
- rows
- sse-api


severo (approved):
Good refactoring, thanks

PR #2024, opened 2023-10-27
2023 reduce number comments

### LINKED ISSUE #2023: Reduce the amount of comments after parquet conversion
See https://huggingface.co/datasets/open-source-metrics/datasets-dependents/discussions/1#6537d488b33b913282d076de

> Would it make sense to just update the original message to add the different revisions that have been converted?

> I agree that editing original might be better. Translating this same situation to GitHub, I would not be happy with a bot that pings for every single commit. For example, the doc-builder bot does not write a comment for every commit in a PR, it just does a single comment and link will work. I don't think most users will want to get a new notification for every single commit. We can also not silence notifications at a repo level, so as a user i would be inclined to silence the whole org.

Let's just create a discussion that points to the refs/convert/parquet page, and ignore if a discussion already exists.

severo:
fixes #2023 

note: `__PARQUET_CONVERTER_USER__` will be shown as `parquet-converter` in prod, without the bold formatting (due to the `__` characters).

Capture d’écran 2023-10-27 à 14 53 26

---

Capture d’écran 2023-10-27 à 15 01 00

### DIFF
--- a/jobs/cache_maintenance/src/cache_maintenance/discussions.py
+++ b/jobs/cache_maintenance/src/cache_maintenance/discussions.py
@@ -8,21 +8,16 @@ from urllib import parse
 
 from huggingface_hub import HfApi
 from huggingface_hub.constants import REPO_TYPE_DATASET
-from libcommon.simple_cache import (
-    DatasetWithRevision,
-    get_datasets_with_last_updated_kind,
-)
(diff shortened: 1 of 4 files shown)


severo:
OK. Merging, we can still revert if we want to spam again :P

PR #2025, opened 2023-10-27
Make create_response async and run _transform_row map async in separate thread

albertvillanova:
Make `create_response` (and inner functions) asynchronous and run `_transform_row` map asynchronously in separate thread.


severo (approved):
another one, nice. Wondering how it reflects in the response time

PR #2027, opened 2023-10-27
fix duckdb circular import

AndreaFrancis:
When deploying to staging (and also locally), the following error appeared for /search service:
Screenshot from 2023-10-27 12-34-59


severo (approved):

albertvillanova (commented):
Sorry, I had not noticed it...
Thanks for the fix!  🤗

PR #2031, opened 2023-10-30
Use ruff for CI quality

albertvillanova:
Use `ruff` for CI quality: 10-100x faster than existing linters (like Flake8) and formatters (like Black).

Note this is most useful when developing: `make style` now is much faster and additionally runs the linter (before `flake8` was only called by `make quality`).


severo:
Did you check how it integrates with vscode? We might want to update https://github.com/huggingface/datasets-server/blob/7281b0eae583f29fb2c78fe0d7e606d3509ba4f7/.vscode/monorepo.code-workspace#L71-L75

.github/workflows/_quality-python.yml
```diff
@@ -40,14 +40,12 @@ jobs:
         run: |
           poetry env use "${{ env.python-version }}"
           poetry install
-      - name: Run black
-        run: poetry run black --check tests src
-      - name: Run isort
```
severo, line 45: so ruff replaces black, flake8, and isort? Nice

severo (approved):
Excellent:
- faster
- easier to maintain

albertvillanova:
> Did you check how it integrates with vscode? We might want to update 

No idea... I have been looking for it, but didn't find anything applicable...

severo:
feel free to merge, and then I can work on the specific support of ruff in vscode

PR #2039, opened 2023-10-31
Revert "Revert assets fsspec (#2036)"

AndreaFrancis:
This reverts commit 5716638a683d83652da672f0f59bf68935d0a93e.


AndreaFrancis:
I will add fsspec changes again. I confirmed with @rtrompier  that the problem with staging env for files not being solved (Like https://datasets-server.us.dev.moon.huggingface.tech/cached-assets/asoria/image/--/059f77f/--/default/train/0/image/image.jpg) is due to external DNS updating the DNS record to the ALB endpoint so, it shouldn't be the case with prod.

PR #2043, opened 2023-11-02
Explain single quote string SQL syntax in filter docs

### LINKED ISSUE #2038: How to pass single quote in /filter endpoint "where" parameter?
See `https://huggingface.co/datasets/albertvillanova/lm_en_dummy2/viewer/default/train?f[meta][value]='{'file': 'file_4.txt'}'`

From `https://datasets-server.huggingface.co/filter?dataset=albertvillanova/lm_en_dummy2&config=default&split=train&where=meta='{'file': 'file_4.txt'}'`, we get:

```
{"error":"Parameter 'where' is invalid"}
```

We want to search he value `{'file': 'file_4.txt'}` in the column `meta`

albertvillanova:
Explain single quote string SQL syntax in filter docs.

Close #2038.


severo (approved):
Excellent, thanks a lot

Issue #2044, opened 2023-11-02
Support ruff in vscode


severo:
See https://github.com/huggingface/datasets-server/pull/2031#issuecomment-1790455688 and https://github.com/huggingface/datasets-server/pull/2031#issuecomment-1790718911

severo:
I'm having some difficulties to make it work. But I'll share here something that helps a lot and that I didn't know. We can see the logs related to a specific extension here:

<img width="758" alt="Capture d’écran 2023-11-03 à 21 32 01" src="https://github.com/huggingface/datasets-server/assets/1676121/5508b045-1b94-45ee-9438-6f6e6928f489">

severo:
This is the issue (https://github.com/astral-sh/ruff-vscode/issues/240): monorepo is not well supported for now, we have to set a Python interpreter for the root, even if it does not make sense for the project.

severo:
Fixed

PR #2054, opened 2023-11-03
Use ruff in vscode

severo:
Can you test if it works for you?

For me, after some struggle, it works well. A nice improvement is that it will sort the imports.

https://github.com/huggingface/datasets-server/assets/1676121/ba698454-b6df-42d6-b402-676b6d030e8c


severo:
fixes #2044

PR #2058, opened 2023-11-06
Stop worker loop executor on SIGTERM

### LINKED ISSUE #1999: Better handling of k8s SIGTERM when stopping workers
We've observed recently that worker pods can take more than 1h to get killed by kubernetes (see internal [slack thread](https://huggingface.slack.com/archives/C04L6P8KNQ5/p1697641144036899) for example).
According to the k8s docs, at one point it sends a TERM and then after a grace period a SIGKILL on all the running processes.
If this keeps happening we need to improve the way we handle those to kill the pods faster.

One idea is to catch SIGTERM in the worker executor (main python process of workers) that would stop the worker loop (its subprocess where the actual code runs).

albertvillanova:
Fix #1999.

### DIFF
--- a/services/worker/src/worker/executor.py
+++ b/services/worker/src/worker/executor.py
@@ -117,6 +117,9 @@ class WorkerExecutor:
         if exceptions:
             raise RuntimeError(f"Some async tasks failed: {exceptions}")
 
+    def stop(self, worker_loop_executor: OutputExecutor) -> None:
+        worker_loop_executor.stop()
+
     def get_state(self) -> Optional[WorkerState]:
         worker_state_file_path = self.state_file_path
         if not os.path.exists(worker_state_file_path):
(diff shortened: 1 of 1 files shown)


severo (approved):
I guess it's what we need (not a specialist).

Not sure if we really want it: do we want a test to be sure?

lhoestq (approved):
Awesome ! Would be cool to have a test indeed

albertvillanova:
Yes, I agree we should test this. However I don't know how to send a SIGTERM within a test, without terminating the test itself...

Any hint?

lhoestq:
Maybe run a subprocess from the test ?

albertvillanova:
I am afraid I was not able to implement the test. I would suggest:
- either somebody else pushes the test to this PR
- or we merge it as it is and create an issue to implement the test in a subsequent PR

severo:
I don't think I would be able to do it quickly either. Maybe @lhoestq. What do you think?

lhoestq:
Let's merge it, we'll see if it works the next time we deploy anyway

PR #2060, opened 2023-11-06
invalidate the datasets cache after 3 years instead of 3 months

severo:
Quick fix to avoid deleting old cache entries (in particular, to be sure to keep the old dataset cache entries for datasets with a dataset script, now that we don't support scripts anymore)

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -167,7 +167,7 @@ mongodb:
 
 cache:
   # Maximum number of days to keep the cached API responses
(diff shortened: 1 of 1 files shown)


severo:
is it worth the time?

lhoestq:
as you want ;)

severo:
I'll open a "good first issue" issue for that

severo:
https://github.com/huggingface/datasets-server/issues/2061

Issue #1831, opened 2023-09-21
Descriptive stats: `duckdb.IOException: IO Error: No files found that match the pattern ...`


lhoestq:
Happens for [imagenet-1k](https://huggingface.co/datasets/imagenet-1k):

```
"Traceback (most recent call last):\n",
" File \"/src/services/worker/src/worker/job_manager.py\", line 168, in process\n job_result = self.job_runner.compute()\n",
" File \"/src/services/worker/src/worker/job_runners/split/descriptive_statistics.py\", line 467, in compute\n compute_descriptive_statistics_response(\n", "
File \"/src/services/worker/src/worker/job_runners/split/descriptive_statistics.py\", line 396, in compute_descriptive_statistics_response\n cat_column_stats: CategoricalStatisticsItem = compute_categorical_statistics(\n",
" File \"/src/services/worker/src/worker/job_runners/split/descriptive_statistics.py\", line 232, in compute_categorical_statistics\n categories: List[Tuple[int, int]] = con.sql(\n",
"duckdb.IOException: IO Error: No files found that match the pattern \"/storage/stats-cache/48010914651841-split-descriptive-statistics-imagenet-1k-3acf5c2b/default/test/*.parquet\"\n"
```

cc @polinaeterna

severo:
Not the case anymore on https://huggingface.co/datasets/imagenet-1k

severo:
Not the case anymore on https://huggingface.co/datasets/imagenet-1k

severo:
https://datasets-server.huggingface.co/statistics?dataset=imagenet-1k&config=default&split=train

gives

> {"error":"IO Error: No files found that match the pattern \"/storage/stats-cache/43040372179808-split-descriptive-statistics-imagenet-1k-d5e07eb7/default/train/*.parquet\""}

polinaeterna:
i think it's the case for partial datasets, i can work on this

PR #2062, opened 2023-11-06
Fix cache directory for partial datasets for `split-descriptive-statistics` runner

polinaeterna:
fixes https://github.com/huggingface/datasets-server/issues/1831

fix glob pattern: preserve 'partial' in a split dir name


polinaeterna:
@severo they all either have it or not and i take the first one just to get this value. the same is done in [`duckdb-index` worker](https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/duckdb_index.py#L132) actually

severo (approved):

PR #2040, opened 2023-10-31
remove assets and cached assets

AndreaFrancis:

PR #2063, opened 2023-11-06
fix reverse proxy

AndreaFrancis:
After deploying https://github.com/huggingface/datasets-server/pull/2040 in staging, reverse-proxy component failed (I could not see the specific error but suspect it is because of an alias to a no longer existent location /storage).

### DIFF
--- a/chart/nginx-templates/default.conf.template
+++ b/chart/nginx-templates/default.conf.template
@@ -29,14 +29,6 @@ server {
     return 307 https://raw.githubusercontent.com/huggingface/datasets-server/main/${OPENAPI_FILE};
   }
 
-  location /assets/ {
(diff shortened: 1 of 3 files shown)


AndreaFrancis:
> So, possibly, we have to keep the configuration in the nginx configuration but enable it conditionally, depending on the value, no?

Yes, I am not familiar with how to achieve it with Nginx but I would like to propose it in another PR, is it ok @severo ?

severo:
OK, but we have to do it quickly because our e2e tests will be broken meanwhile, right?

AndreaFrancis:
> OK, but we have to do it quickly because our e2e tests will be broken meanwhile, right?

Probably not, since we don't test image/audio content. But sure, I am working on it (Just don't want to block deploys)

PR #2068, opened 2023-11-07
try to remove assets and cached-assets

AndreaFrancis:
Second try to delete all dependencies to assets and cached assets.
In https://github.com/huggingface/datasets-server/pull/2040, reverse-proxy failed (Maybe because of a wrong nginx configuration file, I didn't find any log error).


chart/nginx-templates/default.conf.template
```diff
@@ -24,16 +24,24 @@ server {
...
   server_name ${HOST};
+  set $assets_directory ${ASSETS_DIRECTORY};
+  set $cached_assets_directory ${CACHED_ASSETS_DIRECTORY};
 
   location /openapi.json {
     return 307 https://raw.githubusercontent.com/huggingface/datasets-server/main/${OPENAPI_FILE};
   }
 
   location /assets/ {
+    if ($assets_directory = false) {
```
AndreaFrancis, line 35: Not sure if this is the best way to do it:
- If ASSETS_DIRECTORY is set, it should take the alias with the value but if not, it should not consider the alias. We need this logic because in prod, we don't need this redirect, but for e2e and local dev env, we use local storage.
Same for cached-assets.

chart/templates/reverse-proxy/deployment.yaml
```diff
@@ -26,15 +26,10 @@ spec:
         checksum/config: {{ include (print $.Template.BasePath "/reverse-proxy/configMap.yaml") . | sha256sum }}
     spec:
       {{- include "image.imagePullSecrets" . | nindent 6 }}
-      initContainers:
```
AndreaFrancis, line 29: In https://github.com/huggingface/datasets-server/pull/2040 I forgot to remove this line, maybe that was the problem when deploying reverse-proxy? I didn't have any errors when running helm lint though.

severo: I'm not a YAML expert, it seems like an empty sequence should have `[]` indeed, so:

```
initContainers: []
```

would have been valid, while

```
initContainers: 
```

seems to be an invalid syntax.

https://stackoverflow.com/questions/5110313/how-do-i-create-an-empty-array-in-yaml

severo (approved):

chart/templates/reverse-proxy/deployment.yaml
AndreaFrancis: I will try with initContainers: []

severo: hmmm, no, no need to add it, if it's empty we can just avoid having it in the code. I just wanted to say that `initContainers: ` seems to be an invalid syntax

PR #2073, opened 2023-11-08
add service for workers

severo:
We need to access the /metrics endpoint in `servicemonitor`, and it surely requires a _service_ to _monitor_.


XciD:
You can use a [`PodMonitor`](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#podmonitor), it avoid the need for service

severo:
Closing, since we will not add a service

Issue #2009, opened 2023-10-24
Are URLs in rows response sanitized?


severo:
see https://github.com/huggingface/moon-landing/pull/7798#discussion_r1369813236 (internal)

> Is "src" validated / sanitized?
> if not there is a potential XSS exploit here (you can inject javascript code in an image src)

> Are S3 object names sanitized? If no, it should be the case in dataset-server side

severo:
cc @AndreaFrancis do you want to look at this?

AndreaFrancis:
We generate the images/audios URLs at these locations:
- For images, it's created at https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/viewer_utils/asset.py#L60.
- For audio, the URL is generated at https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/viewer_utils/asset.py#L90.

(Maybe I can refactor to get the src from a single function).

I think the URLs we generate are automatically sanitized, because we forge them with the following parameters:

Base URL: We retrieve this from [here](https://github.com/huggingface/datasets-server/blob/main/chart/templates/_common/_helpers.tpl#L200) for assets (used in firs-rows) and [here](https://github.com/huggingface/datasets-server/blob/main/chart/templates/_common/_helpers.tpl#L207) for cached assets (used in /rows, /filter and /search).
Directory Path: It's a combination of {dataset_name}/{DATASET_SEPARATOR}/{revision}/{DATASET_SEPARATOR}/{config}/{split}/{str(row_idx)}/{column}.
Filename: The filename typically follows the format image.<ext> and audio.<ext> (Only image.jpg, image.png, audio.wav, audio.mp3 are supported).
Not sure if  an additional validation may  be necessary, as these processes are always managed on the datasets-server side.
An src value always looks like: "https://datasets-server.huggingface.co/cached-assets/asoria/bluey/--/abf99ba2de1d5c93e777016bb2634248c7f366e5/--/config/train/47/image/image.jpg" and as you can it just contains scheme, netloc and path components (No query part, we need add any related value).

severo:
Maybe ensure that dataset_name, split, and especially config and column are sanitized, as these are user defined.

PR #2080, opened 2023-11-09
sanitize src url

AndreaFrancis:
Will close https://github.com/huggingface/datasets-server/issues/2009
I added a parse quote to input vulnerable variables like dataset, config, split, and column name. Also I added a basic test to ensure src is sanitized.
If someone wants to try with other values, I can add those to enhance the test coverage.

### DIFF
--- a/libs/libcommon/src/libcommon/viewer_utils/asset.py
+++ b/libs/libcommon/src/libcommon/viewer_utils/asset.py
@@ -4,6 +4,7 @@
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import TypedDict
+from urllib import parse
 
 from PIL import Image  # type: ignore
 from pydub import AudioSegment  # type:ignore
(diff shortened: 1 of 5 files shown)


libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -37,6 +34,13 @@ class AudioSource(TypedDict):
     type: str
 
 
+def generate_asset_src(
+    base_url: str, dataset: str, revision: str, config: str, split: str, row_idx: int, column: str, filename: str
+) -> tuple[str, str]:
+    dir_path = f"{parse.quote(dataset)}/{DATASET_SEPARATOR}/{revision}/{DATASET_SEPARATOR}/{parse.quote(config)}/{parse.quote(split)}/{str(row_idx)}/{parse.quote(column)}"
```
severo, line 40: what does it mean for dataset="asoria/mydataset" for example? I imagine that the slash is quoted. Is that what we want?

AndreaFrancis: I think by default, "/" won't be quoted https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote but I can set a specific `safe` parameters to prevent.

severo: currently we keep it, and it creates a subdirectory. Do we want to stick with this?

AndreaFrancis: I think it is not a big deal, only existing cached assets would be affected (Because /rows, /filter, and /search will upload image/audio with the new path).
 Do you find any inconvenience with the change @lhoestq?

PR #2081, opened 2023-11-09
Stats resources

lhoestq:
@polinaeterna I fixed the n_threads and max_memory settings of DuckDB (they should be different based on the pod type)

I also set requests == limits for cpu to see if it fixes the DuckDB hanging issue (and more generally any hanging issue that we can also observe when killing pods or in parquet conversion jobs)

### DIFF
--- a/services/worker/src/worker/job_runners/split/descriptive_statistics.py
+++ b/services/worker/src/worker/job_runners/split/descriptive_statistics.py
@@ -479,15 +479,15 @@ def compute_descriptive_statistics_response(
 
     con = duckdb.connect(str(local_parquet_directory / DATABASE_FILENAME))  # load data in local db file
     con.sql("SET enable_progress_bar=true;")
+    
+    # DuckDB uses n_threads = num kubernetes cpu (limits)
     n_threads = con.sql("SELECT current_setting('threads')").fetchall()[0][0]
-    logging.info(f"Original number of threads={n_threads}")
-    con.sql("SET threads TO 8;")
-    n_threads = con.sql("SELECT current_setting('threads')").fetchall()[0][0]
-    logging.info(f"Current number of threads={n_threads}")
+    logging.info(f"Number of threads={n_threads}")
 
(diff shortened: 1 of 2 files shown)


services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -479,15 +479,15 @@ def compute_descriptive_statistics_response(
...
-    con.sql("SET threads TO 8;")
-    n_threads = con.sql("SELECT current_setting('threads')").fetchall()[0][0]
-    logging.info(f"Current number of threads={n_threads}")
+    logging.info(f"Number of threads={n_threads}")
 
+    # However DuckDB uses max_memory = memory of the entite kubernetes node so we lower it
     max_memory = con.sql("SELECT current_setting('max_memory');").fetchall()[0][0]
     logging.info(f"Original {max_memory=}")
-    con.sql("SET max_memory TO '28gb';")
+    con.sql(f"SET max_memory TO '{28 if n_threads >= 8 else 10}gb';")
```
severo, line 490: hmmmmmmmmm! OK. Wondering if we can avoid hardcoding this or have some safeguard to have it always coherent with the values in the chart? Maybe through the CI? Or have comments in both parts of the code that refer to the other one, so that we see the warning when changing a value

severo (approved):

severo:
I think it will help so much. The current state of the medium workers:

Capture d’écran 2023-11-09 à 23 05 17

A lot of them are being recreated because they failed for some reason

services/worker/src/worker/job_runners/split/descriptive_statistics.py
lhoestq: Let's see if it helps first, then yea I think we should avoid hardcoded stuff like thus

severo: -> #2091 to keep track of this

severo: we don't set `max_memory` anymore in the code, so... let's close

PR #2100, opened 2023-11-13
cron job to delete obsolete cache

AndreaFrancis:
Currently, deleting obsolete cache is an admin endpoint. Adding cron job configuration to have it running every 3 hours (Admin UI option will still be available).


chart/values.yaml
```diff
@@ -435,6 +435,22 @@ cleanStatsCache:
   expiredTimeIntervalSeconds: 600
   subfolderPattern: "*"
 
+deleteObsoleteCache:
+  enabled: true
+  log:
+    level: "info"
+  action: "delete-obsolete-cache"
+  error_codes_to_retry: ""
```
severo, line 443: is this required?

chart/templates/cron-jobs/delete-obsolete-cache/job.yaml
```diff
@@ -0,0 +1,27 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 The HuggingFace Authors.
```
severo, line 2:
```suggestion
# Copyright 2023 The HuggingFace Authors.
```

chart/values.yaml
```diff
@@ -435,6 +435,22 @@ cleanStatsCache:
...
+  schedule: "0 */3 * * *"
+  # every 3 hours
+  nodeSelector: {}
+  resources:
```
severo, line 447: you don't set the resources in prod either. Will the job have enough resources? (I'm not sure if it's important, I think we don't need much anyway)

severo (approved):
nothing blocking

chart/values.yaml
AndreaFrancis: I think it will be enough as the default (Like in cleanHfDatasetsCache, cleanDuckdbIndexCache, cleanDuckdbIndexDownloads, cleanDuckdbIndexJobRunner, cleanStatsCache, etc).
Maybe at the very first time, it would have needed more resources but since it will run every 3 hours I think it will me removing at most about 100 datasets which is not critical. 
I fixed the nodeSelector in the prod.yaml (I forgot it).

chart/values.yaml
AndreaFrancis: No, I removed it in the other jobs as well.

PR #2101, opened 2023-11-13
add cacheable ParquetResponseEmptyError in first-rows-from-parquet

AndreaFrancis:
Currently, we have `{ _id: { cause: 'ParquetResponseEmptyError' }, count: 303 }`, records in cache collection with UnexpectedError because of `ParquetResponseEmptyError`.
Actually, we have two definitions of `ParquetResponseEmptyError`: one in parquet_utils.py https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/parquet_utils.py#L29 and another in https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/exceptions.py#L424.
`SplitFirstRowsFromParquetJobRunner` should throw the cacheable version of `ParquetResponseEmptyError` but currently it is not handling the noncacheable exception leading to an `UnexpectedError`.
After this PR, I will refresh the cache of the 303 records.

### DIFF
--- a/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py
+++ b/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py
@@ -15,8 +15,7 @@ from libcommon.exceptions import (
     TooBigContentError,
     TooManyColumnsError,
 )
-from libcommon.parquet_utils import Indexer, TooBigRows
-from libcommon.parquet_utils import ParquetResponseEmptyError as EmptyError
+from libcommon.parquet_utils import EmptyParquetMetadataError, Indexer, TooBigRows
 from libcommon.processing_graph import ProcessingGraph, ProcessingStep
 from libcommon.public_assets_storage import PublicAssetsStorage
 from libcommon.storage import StrPath
@@ -79,7 +78,7 @@ def compute_first_rows_response(
             config=config,
             split=split,
         )
-    except EmptyError:
+    except EmptyParquetMetadataError:
         raise ParquetResponseEmptyError("No parquet files found.")
 
     # validate the features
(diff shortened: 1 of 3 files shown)


services/worker/src/worker/job_runners/split/first_rows_from_parquet.py
```diff
@@ -10,11 +10,13 @@
     PROCESSING_STEP_SPLIT_FIRST_ROWS_FROM_STREAMING_VERSION,
 )
 from libcommon.exceptions import (
+    ParquetResponseEmptyError,
     RowsPostProcessingError,
     TooBigContentError,
     TooManyColumnsError,
 )
 from libcommon.parquet_utils import Indexer, TooBigRows
+from libcommon.parquet_utils import ParquetResponseEmptyError as EmptyError
```
severo, line 19: maybe rename ParquetResponseEmptyError in libcommon.parquet_utils to prevent future confusions?

severo: Or directly raise libcommon.exceptions.ParquetResponseEmptyError there? It would remove the need to put the following code in a try/except structure

AndreaFrancis: > Or directly raise libcommon.exceptions.ParquetResponseEmptyError there? It would remove the need to put the following code in a try/except structure

I thought the same but, I was not sure if it is appropriate to return a `CacheableError` in parquet_utils that could be exposed in /rows.


> maybe rename ParquetResponseEmptyError in libcommon.parquet_utils to prevent future confusions?

I like more this approach

severo: > I thought the same but, I was not sure if it is appropriate to return a CacheableError in parquet_utils that could be exposed in /rows.

You're right

PR #2102, opened 2023-11-14
add an index recommended by Atlas

severo:
Capture d’écran 2023-11-14 à 11 54 53

### DIFF
--- a/libs/libcommon/src/libcommon/simple_cache.py
+++ b/libs/libcommon/src/libcommon/simple_cache.py
@@ -143,6 +143,7 @@ class CachedResponseDocument(Document):
             ("kind", "http_status", "_id"),
             ("kind", "_id"),  # < recommended by Atlas
             ("details.cause_exception", "error_code", "details.copied_from_artifact"),  # < recommended by Atlas
+            ("error_code", "kind", "details.copied_from_artifact"),  # < recommended by Atlas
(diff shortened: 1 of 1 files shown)


AndreaFrancis:
I was running yesterday a couple of ad-hoc queries for https://github.com/huggingface/datasets-server/issues/1443 (including error_code, kind, and details.copied_from_artifact).
But not sure if this is the only reason for Atlas advice, do you know if it is required by other parts of the code?
Should we create indexes for the ad-hoc queries?

severo:
Yes, I was wondering if we should do it or not (the other recommendations also come from ad-hoc queries, I imagine). The indexes are not that big, and it will speed up the queries, even if we run them only from time to time. WDYT?

AndreaFrancis:
> The indexes are not that big, and it will speed up the queries, even if we run them only from time to time. WDYT?
I agree storage is not a big problem. 
I'm not an expert in MongoDB, but I've learned from using other databases that sometimes when you add more indexes, it can make the database run less efficiently (because of query plan analizer taking more time to find the best index). Also, we should make sure our code clearly shows how we're using it for the datasets server project. This way, if someone else wants to use our project as a model or build on it, they'll find it easier to understand.

severo:
OK. I'll merge since I already did the same for the other recommendations, and the comments give a hint of why we added them. If we want to remove them later, it's easy to know which ones are "optional".

PR #2117, opened 2023-11-15
Remove build/egg directories from worker vendor trec-car-tools

albertvillanova:
After updating the worker service to use Python 3.9.18, I got this update in its `trec-car-tools` vendor (file `services/worker/vendors/trec-car-tools/python3/trec_car_tools.egg-info/PKG-INFO`):
```
Requires-Dist: cbor>=1.0.0
Requires-Dist: numpy>=1.11.2
```

This PR removes unnecessary vendor `build` and `.egg-info` directories (which are automatically recreated at install).


severo:
I'm not sure we should change the vendored dependency. I also had the case but I ignored the changes. What do you think?

albertvillanova:
I see your point...

First, note this is only metadata about the installed package `trec-car-tools`:
```
Requires-Dist: cbor>=1.0.0
Requires-Dist: numpy>=1.11.2
```

Also note its installation indeed has these requirements. See: https://github.com/huggingface/datasets-server/blob/f894794b6cff251f9cf24d8fc4a4c7e700c2b3be/services/worker/vendors/trec-car-tools/python3/setup.py#L15

Apparently Python 3.9.15 was not adding this metadata, whereas Python 3.9.18 does.

Also note that the entire `vendors/trec-car-tools/python3/trec_car_tools.egg-info` directory is created at installation. You can check this with:
```shell
rm -fr vendors/trec-car-tools/python3/trec_car_tools.egg-info
make install
```
- the entire directory is recreated after installation

Not a strong opinion about the convenience of adding or not this additional metadata about the installed package.

severo:
OK, thanks for the detailed answer. Let's merge, there's no reason to ignore that, indeed. It would bother us everytime, and it's surely a useful addition.

severo (approved):

albertvillanova:
@severo I was thinking that other option would be to Git-untrack the directory `services/worker/vendors/trec-car-tools/python3/trec_car_tools.egg-info` and modify `.gitignore` to add:
```
*.egg-info/
```

This is commonly done. See: https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Python.gitignore#L24C1-L24C12

albertvillanova:
And also remove the `services/worker/vendors/trec-car-tools/python3/build` direcotry, which is also recreated at install.

severo:
OK, feel free to update the PR, it seems cleaner indeed

severo (approved):

PR #2124, opened 2023-11-16
Split parquet schema mismatch

AndreaFrancis:
We have 11 cache records for split-first-rows-from-parquet because of ArrowInvalid when split parquets have a different schema.
See 
Screenshot from 2023-11-16 15-02-48

This PR contains:
- Intercept ArrowInvalid exception when trying to concatenate shared parquet files in parquet_utils.py
- Separated tests from test_rows.py to test_response.py and test_parquet_utils.py since there was not a specific rows logic inside.
- Added a test in test_parquet_utils.py to simulate schema mismatch error.


libs/libcommon/src/libcommon/exceptions.py
```diff
@@ -496,6 +497,19 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
         )
 
 
+class SplitParquetSchemaMismatchError(CacheableError):
+    """The split parquet files have different schema."""
```
severo, line 501: same comment as below: the error message could be more explicit for the end user

severo (commented):
By the way: in which case do we have that error?

AndreaFrancis:
> By the way: in which case do we have that error?
The following datasets have this error:

- mikeee/chroma-paraphrase-multilingual-mpnet-base-v2
- antonkulaga/openai_6000_chunk_papers
- longevity-genie/openai_6000_chunk_modules
- mariosasko/test_tags
- kye/all-lucidrain-code-python-tokenized-8192
- jitx/distillation_code_sample
- Lolz14/mine
- Rami/utd_reddit
- eco4cast/neon4cast-scores
- texonom/md-chroma-instructor-xl
- hysts-samples/sample-user-preferences

It happens when for the same split, the user provided parquet files with different columns, for instance: dataset=kye/all-lucidrain-code-python-tokenized-8192 config=default split=train has two source files but their schemas (column names) are different:
Screenshot from 2023-11-17 07-19-51
First file has: python_code, repo_name and file_path columns.
Second file has: repo_name, file_path, input_ids, attention_mask columns.



> Maybe change the wording thinking in the final user. I don't think I would have understood what a "schema" is. Also, the error message should give a hint at what the user can do to fix the issue.

Maybe: "The column names of the split Parquet files do not match." ?

severo:
Maybe we can dig a bit into why this is occurring. In the case of https://huggingface.co/datasets/kye/all-lucidrain-code-python-tokenized-8192/commits/main, it seems the user wanted to replace the file, but due to the different suffix, we have two files, instead of one. The solution in that case would be to delete the first parquet file which is obsolete.

I'm not sure how we detect this is the case, and how to reflect this in the error message.

Maybe we can look at the other datasets to see if it's the same situation or another use case

AndreaFrancis:
>  The solution in that case would be to delete the first parquet file which is obsolete.

Yes, but I think it would be better to let the user fix the problem (In case we want to do it, we have be near 100% sure that the cause is an obsolete file)

The following datasets look to have the same case (trying to upload and replace a file):
https://huggingface.co/datasets/kye/all-lucidrain-code-python-tokenized-8192/commits/main
https://huggingface.co/datasets/mariosasko/test_tags/commits/main
https://huggingface.co/datasets/jitx/distillation_code_sample/commits/main
https://huggingface.co/datasets/Rami/utd_reddit/commits/main
https://huggingface.co/datasets/hysts-samples/sample-user-preferences/commits/main

I'm not sure if we should fix this issue directly on the datasets server, as it might cause unintended user actions. It might be better to update our guides to help users with update actions. We could also suggest specific fixes for certain errors. These changes should be applied to the whole datasets server, possibly in a new pull request. WDYT? I can open an issue to suggest ways for users to fix their problems. We could link each error with possible solutions to simplify issue resolution. I'm not sure if this should be implemented on the API side or the UI, or perhaps in Moonlanding.

severo:
OK. It's pretty rare, in any case (about 10 datasets out of 70K). I think we should make the error message clearer (it can be longer) so that the user has a hint that it's an issue with having incoherent files in the same split. Then, on the Hub, we propose the user to open a discussion, so, we can fix on the issues on that basis.

severo (approved):

PR #2127, opened 2023-11-17
add /croissant API endpoint

lhoestq:
Added /croissant API

It returns a dataset's JSON Croissant representation.

In particular it contains one recordSet per dataset configuration, with the feature types and parquet files list.

It uses `git+https` instead of pure HTTP locations for the files because Croissant requires hash verifications and they're already provided by git (cc @marcenacp for suggesting this)
Since Croissant implementations might support HTTP more easily than git, we might change to HTTP locations in the future and provide the parquet files hashes.

Datasets that are partially exported to parquet are supported and their recordSet descriptions mention "(first 5GB)".

Unsupported columns are skipped for now, e.g. nested columns (lists, structs), audio etc. (not yet supported in Croissant)
Supported types are strings, integers, and floats.
It is mentioned in the recordSet description which columns are skipped.

Train/test/validation splits are not taken into account yet.

The endpoint will be available at e.g.
```
https://datasets-server.huggingface.co/croissant?dataset=rotten_tomatoes
```

Will add documentation in another PR as soon as the API is stable, and then see how to add it in the dataset pages headers


services/api/src/api/routes/croissant.py
```diff
@@ -0,0 +1,238 @@
...
+                        blocked_datasets=blocked_datasets,
+                        hf_timeout_seconds=hf_timeout_seconds,
+                        cache_max_days=cache_max_days,
+                    )
+                content = info_result["content"]
+                http_status = info_result["http_status"]
+                error_code = info_result["error_code"]
+                revision = info_result["dataset_git_revision"]
+                if http_status == HTTPStatus.OK:
+                    infos = list(islice(content["dataset_info"].values(), MAX_CONFIGS))
```
severo, line 218: Maybe for another version, as it's a POC, but if we truncate the response, we should the user know, with a specific field in the response, or a HTTP header, or something (or even, return an error if it's > 100, instead of truncating)

services/api/src/api/routes/croissant.py
```diff
@@ -0,0 +1,238 @@
...
+
+def get_croissant_from_dataset_infos(dataset: str, infos: list[Mapping[str, Any]], partial: bool) -> Mapping[str, Any]:
+    distribution = [
+        {
+            "@type": "sc:FileObject",
+            "name": "repo",
+            "description": "The Hugging Face git repository.",
+            "contentUrl": f"https://huggingface.co/datasets/{dataset}",
+            "encodingFormat": "git+https",
+            "sha256": "https://github.com/mlcommons/croissant/issues/80",
```
severo, line 58: Is this a TODO?

services/api/src/api/routes/croissant.py
```diff
@@ -0,0 +1,238 @@
...
+class ParquetFile(TypedDict):
+    dataset: str
+    config: str
+    split: str
+    url: str
+    filename: str
+    size: int
+
+
+HF_TO_CRROISSANT_VALUE_TYPE = {
```
severo, line 40: nit: remove one `R`?

services/api/src/api/routes/croissant.py
```diff
@@ -0,0 +1,238 @@
...
+)
+from libcommon.processing_graph import ProcessingGraph
+from libcommon.prometheus import StepProfiler
+from starlette.requests import Request
+from starlette.responses import Response
+
+MAX_CONFIGS = 100
+
+
+class ParquetFile(TypedDict):
```
severo, line 31: it's not used, right?

services/api/src/api/routes/croissant.py
```diff
@@ -0,0 +1,238 @@
...
+        features = Features.from_dict(info["features"])
+        fields: list[dict[str, Any]] = []
+        distribution_name = f"parquet-files-for-config-{config}"
+        distribution.append(
+            {
+                "@type": "sc:FileSet",
+                "name": distribution_name,
+                "containedIn": "repo",
+                "encodingFormat": "application/x-parquet",
+                "includes": f"{config}/*/*.parquet",
```
severo, line 73: it does not list the parquet files? What is this useful for? And we don't refer to the `refs/convert/parquet` branch? (sorry I don't know the Croissant format)

services/api/src/api/routes/croissant.py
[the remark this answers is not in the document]
lhoestq: Yes we can have the hashes and use them once it's supported

services/api/src/api/routes/croissant.py
```diff
@@ -0,0 +1,238 @@
...
+    "int64": "sc:Integer",
+    "float32": "sc:Float",
+    "float64": "sc:Float",
+    "bool": "sc:Boolean",
+}
+
+
+def get_croissant_from_dataset_infos(dataset: str, infos: list[Mapping[str, Any]], partial: bool) -> Mapping[str, Any]:
+    distribution = [
+        {
```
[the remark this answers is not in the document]
lhoestq: definitely an option in the future !

services/api/src/api/routes/croissant.py
lhoestq: Yes it will be replaced by actual hashes later, once it's supported in mlcroissant

services/api/src/api/routes/croissant.py
lhoestq: Good catch, I forgot to add the branch x)

Currently no need to list the parquet files URLs, since mlcroissant supports globbing in git repos

severo: OK. And if we have the `config-info` in cache, it always means that we also have the parquet files, since it depends on `config-parquet-and-info`, right?

severo: (you might get both the parquet files URLs, and the info, from `config-parquet-and-info`, by the way

lhoestq: > OK. And if we have the config-info in cache, it always means that we also have the parquet files, since it depends on config-parquet-and-info, right?

yes exactly

> (you might get both the parquet files URLs, and the info, from config-parquet-and-info, by the way)

I used "dataset-info" to be able to do one single call to the db

severo: yes, good idea, in particular if we don't need the detailed list of config parquet files

services/api/src/api/routes/croissant.py
severo: Sure. To be put in 
> Will add documentation in another PR as soon as the API is stable, and then see how to add it in the dataset pages headers

severo (approved):

PR #2128, opened 2023-11-20
upgrade huggingface-hub to temporary branch

severo:


severo:
merging. I'll revert as soon as we confirm it works (or fails) on OpenOrca

PR #2130, opened 2023-11-20
Use `polars` to compute stats for three example big datasets

polinaeterna:
> I tried using polars instead of duckdb to see if it works for bigger datasets and does not stuck and in a local container it works for me, so now I want to check if it's the same on prod but don't want to spend too much time on this in case it doesn't work, that's why I omitted tests (it's only for 3 datasets anyway).


polinaeterna:
@severo @lhoestq do you think we can try to deploy this and see if that works for three datasets `["Open-Orca/OpenOrca", "vivym/midjourney-messages", "HuggingFaceH4/ultrachat_200k"]`. And if it does, I'd clean up the code, check tests and fix some details (like bins for integers which I didn't adjust here)

AndreaFrancis:
Sure! btw, looks like the failed e2e is not related to this PR, maybe just retry.

AndreaFrancis (approved):
nit: Maybe add a small/basic test?

severo:
What is the reason of the PR? is it to try if it's faster and does not fail as duckdb does?

front/admin_ui/poetry.lock
severo, line 1: nit: not related to the PR

severo:
> nit: Maybe add a small/basic test?

yes, would it be possible to run the statistics both on pandas and polars in the tests, and ensure the result is the same?

polinaeterna:
> What is the reason of the PR? is it to try if it's faster and does not fail as duckdb does?

@severo yes, sorry, I should have added a description. I tried using `polars` instead of `duckdb` to see if it works for bigger datasets and does not stuck and in a local container it works for me, so now I want to check if it's the same on prod but don't want to spend too much time on this in case it doesn't work...

front/admin_ui/poetry.lock
polinaeterna: yes, thanks, i don't know where it comes from :D so random

severo:
Hmmm, OK. Feel free to merge and deploy, then ensure it works quickly. But: how will you ensure it's faster? We have some metrics (StepProfiler), but we don't get the dataset name. And the logs don't give the duration of the worker.

polinaeterna:
@severo well we will see if the jobs work at all :smiling_face_with_tear: (for now for these datasets they crushed / exceeded max duration)

severo:
gogogo then

lhoestq (approved):
Let's see if this works !

severo:
I think it introduces a breaking change for `string_text` histograms: the last value in bin_edges is `null`. It generates an issue in the dataset viewer because I didn't expect that value: https://github.com/huggingface/moon-landing/issues/8177. 

See https://datasets-server.huggingface.co/statistics?dataset=Open-Orca/OpenOrca&config=default&split=train

```
bin_edges":[3.0,4.0,5.0,6.0,7.0,8.0,9.0,10.0,11.0,12.0,null]
```

Issue #2048, opened 2023-11-03
Improve selection of the next job to priorize datasets with few configs/splits


severo:
Internal Slack conversation: https://huggingface.slack.com/archives/C04L6P8KNQ5/p1698323643019099?thread_ts=1698323528.797149&cid=C04L6P8KNQ5

> `opus_euconst` is squatting the pods, and the small datasets are waiting for too long before being processed.
![image](https://github.com/huggingface/datasets-server/assets/1676121/976df781-fda8-4072-9133-9a2cdd273eb6)
> maybe switch it to "low" prior ?
> It could be a good solution for these big datasets, indeed! If number of configs or splits is over some threshold -> change the priority to low.
> I like it, because it would not change our bottleneck query, which is to find the next job to process.

severo:
we already prioritize the datasets that have few started jobs, so... I think we can close

PR #2132, opened 2023-11-20
lower job priority if configs/splits exceed threashold

AndreaFrancis:
Fix for https://github.com/huggingface/datasets-server/issues/2048

I set MAX_CONFIGS_THRESHOLD_TO_LOWER_PRIORITY and MAX_SPLITS_THRESHOLD_TO_LOWER_PRIORITY constants = 500 but this could be changed if someone considers a better value.
I thought about the possibility of having these values as env variables but we would have to validate them and I don't think this is something we will be changing frequently.

### DIFF
--- a/libs/libcommon/src/libcommon/orchestrator.py
+++ b/libs/libcommon/src/libcommon/orchestrator.py
@@ -10,11 +10,7 @@ import pandas as pd
 
 from libcommon.constants import ERROR_CODES_TO_RETRY
 from libcommon.exceptions import DatasetInBlockListError
-from libcommon.processing_graph import (
-    ProcessingGraph,
-    ProcessingStep,
-    ProcessingStepDoesNotExist,
-)
+from libcommon.processing_graph import ProcessingGraph, ProcessingStep, ProcessingStepDoesNotExist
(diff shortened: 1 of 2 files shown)


severo:
I think it's a good idea to lower the priority depending on the number of splits. But I don't know of we can do it with the current concepts.

Indeed, of it's HIGH, generally it's because we launched it on purpose via the admin page. I don't think we want to move the priority to LOW, since it's done on purpose.

If it's already LOW: no effect.

And if it's NORMAL,... Ok, why not.

But maybe we should not mix theses priority levels, and instead penalize the datasets that have a lot of pending jobs, in the mongo query? Or change the priority enum to an int, as we do with difficulty?

AndreaFrancis:
> But maybe we should not mix theses priority levels, and instead penalize the datasets that have a lot of pending jobs, in the mongo query? Or change the priority enum to an int, as we do with difficulty?

I think we can achieve the penalization only with an int value to use it on the order by statement right?
I will probably need to: If adding/changing the data type needs a new index or a change on one of the existing ones. Also I think the impact will be bigger, if it is ok I can propose another PR and close this one. WDYT @severo?

PR #2133, opened 2023-11-21
Escape names in Croissant.

marcenacp:
Reason: `/` are used in the syntax as delimiters.

I ended up testing on the first dataset names and configs, and found more conflicts (like ` ` or `=` in the names), so I ended up applying a regular expression corresponding to [the one used by Croissant](https://github.com/mlcommons/croissant/blob/475998a89716a8a022817f1126cf3dd1a3526418/python/mlcroissant/mlcroissant/_src/structure_graph/base_node.py#L20).

### DIFF
--- a/services/api/src/api/routes/croissant.py
+++ b/services/api/src/api/routes/croissant.py
@@ -1,8 +1,8 @@
 import logging
+import re
 from collections.abc import Mapping
 from http import HTTPStatus
 from itertools import islice
-import re
 from typing import Any, Optional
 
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
Thanks for the fix ! Though it would be cool to be able to have
- slashes in dataset ids (pretty common namespace / repo situation like github, hf, etc.)
- spaces in column names (this can be the case in many data format)

Spaces in config names would be cool but less important than the two previous points

PR #2137, opened 2023-11-21
POC separate queue

AndreaFrancis:
While trying to optimize the queue, I realized most of the data in Job collection is unnecessary to get the next waiting job.
See from a total of 8562 records:
Status | Count | %
-- | -- | --
waiting | 2835 | 33.11%
success | 4614 | 53.89%
error | 840 | 9.81%
started | 24 | 0.28%
cancelled | 249 | 2.91%

This means that the get_next_waiting function would only need about 34% of the data in the collection.
We have seen some months ago (and whenever we have a high load in the queue) that MongoDB complains about `Scanned Objects / Returned has gone above 1000.`
I propose to split the queue data into two different collections:
One for the active jobs (started and waiting) and another for the finished ones (success, error, and canceled). 
This approach would help to optimize the query duration and also would target the index definition in a better way.
  
There is a lot to do yet regarding this PR:
- Evaluate if it is needed MongoDB migrations for the new collection
- Fix tests
- Fix style
- Modify indexes based on requirements

Before proceeding, let me know wdyt @huggingface/datasets-server


severo:
The finished jobs are kept only to be able to report metrics, and are deleted after 10 minutes. If we're able to change the way we collect metrics, we could even get rid of them completely

AndreaFrancis:
> The finished jobs are kept only to be able to report metrics, and are deleted after 10 minutes. If we're able to change the way we collect metrics, we could even get rid of them completely

Yes, it is only used in count_finished_jobs, which is needed for the cron job to collect queue metrics.
This job helped us do a cleaning/regularization of the metrics, but the value of the metric is even not totally confident since the table has a TTL, I would say we can safely stop storing the finished jobs, or as I proposed in this PR, create a new collection (Maybe it can have even less attributes like unicity_id, job_type and status) and compute metrics from there.

severo:
as the workers can now generate metrics, I would say we should delete the jobs and register a new metric on every finished job

AndreaFrancis:
As per our last conversation with @severo, I will close this PR in favor of another one that will remove the finished jobs and then we will store the metrics for them in a different Prometheus Counter (Instead of our incremental metrics).

PR #2141, opened 2023-11-22
Add penalization field to queue

AndreaFrancis:
Will close https://github.com/huggingface/datasets-server/issues/2048
Adding a penalization field to the job so that it is used as part of the next waiting job function.
The penalization heuristic is:
- By default, First steps (like dataset-config-names) don't have a penalization
- When generating children jobs, the new penalization will be incremental, that is, the sum of: parent penalization + (# of configs or splits // PENALIZATION_FACTOR) where PENALIZATION_FACTOR is by default 100 
- Now, the prioritization for the next waiting job will be penalization + create_at

See the example for a dataset without penalization with only 1 config with 1 split (no one job has a penalization):
Copy of Demo1

See examples of penalization for a dataset with 102 configs; the first one has 506 splits, and the second one has 1 split:

Demo1

Update: penalization = parent_penalization + (# of configs or splits // PENALIZATION_FACTOR)


libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -306,7 +304,7 @@ def __post_init__(self) -> None:
                         name_field="split",
                     )  # Note that we use the cached content even the revision is different (ie. maybe obsolete)
                 for split_name in split_names:
-                    self.update(next_processing_step, config, split_name)
+                    self.update(next_processing_step, config, split_name, len(split_names))
```
severo, line 307: I think we should set the penalization to a value that reflects the total complexity of the dataset, not just the config's complexity.

For example, set it to the total number of splits of the dataset when we have it. Otherwise, if a dataset has 1000 configs with 2 splits each, the split-level jobs will have a penalization value of 2. It would be better to set it to 2000 (or the value available at that time, since we will progressively know the total number of splits the dataset contain).

AndreaFrancis: > Otherwise, if a dataset has 1000 configs with 2 splits each, the split-level jobs will have a penalization value of 2.

Actually, it will be the sum of job penalization +  parent penalization
https://github.com/huggingface/datasets-server/pull/2141/files#diff-45fde89a57a9a9a207231137c66a1ddf75b5830ba0eb7f504f87d87f921afb27R368

In this case, for a dataset with 1000 configs each with 2 splits, parents will have a penalization of 1000, and children will be 1002. Maybe it is enough since it will be used in the sort condition, also I thought I would add a default penalization by step to increase by 1 instead of 0 so that, we always have a descending value (In a small dataset with 1 config and 1 split, no job will have penalization).

severo: OK, I should have read the code more carefully. It seems a better approach, by the way, because if a dataset has two configs, one with two splits, and the other one with 1000 splits, the first config should not be penalized by the second one.

severo: > add a default penalization by step to increase by 1 instead of 0

I didn't get it, could you elaborate more on what it would change?

AndreaFrancis: I was suggesting it to have a progressive penalization (parents having more than children). But nevermind, I think it is better not to, because otherwise, it will always affect querying the predecessors before the children's processing steps (which is not what we want).

severo: OK? Let's start with the number of configs and splits, it should already help a lot

severo (commented):
As we spoke on Slack, it would be slightly better to replace `len(splits)` with `len(splits)-1`, and same with config, so that one split from one config has no penalization.

Also: don't we need a migration script to add the default value (not sure it's required)?

AndreaFrancis:
> Also: don't we need a migration script to add the default value (not sure it's required)?

Yes, I will add the migration, thanks!

lhoestq:
Is there a risk that some datasets are never processed if the queue stays non-empty ?

lhoestq:
Also I feel like datasets with <= 10 configs and <= 10 splits might be fast enough to process to have same priority as any other dataset

AndreaFrancis:
> Is there a risk that some datasets are never processed if the queue stays non-empty ?

I think so. If we found this behavior, we could have a mechanism that evaluates jobs with high penalization waiting for a certain period of time an try to reduce their penalization (maybe a cron job).

> Also I feel like datasets with <= 10 configs and <= 10 splits might be fast enough to process to have the same priority as any other dataset

Do you mean applying penalization by grouping by # of configs/splits? i.e. 0-10 configs/splits => 1 points, 10 - 100 configs/splits => 2 points, > 100 configs/splits => 3 points and so on?

lhoestq:
> I think so. If we found this behavior, we could have a mechanism that evaluates jobs with high penalization waiting for a certain period of time an try to reduce their penalization (maybe a cron job).

yea indeed, maybe we can see if we ever need to do this

> Do you mean applying penalization by grouping by # of configs/splits? i.e. 0-10 configs/splits => 1 points, 10 - 100 configs/splits => 2 points, > 100 configs/splits => 3 points and so on?

oh I like the idea - this way for most datasets we still order by created_at

AndreaFrancis:
>  Why did you think we needed this if we already have a limit to let small datasets be processed without being blocked by a big dataset ?

I think even having the namespace filtering, the queue is being processed sequentially, I mean for a big dataset (with many configs) it will be processed one by one (it won't block the queue but because of created_at, it will avoid small datasets to be processed).
But maybe I am wrong and @severo can help us clarify.

severo:
Currently, the tests in https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/tests/test_queue.py wrt the priority are:
- https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/tests/test_queue.py#L169: the oldest job is select first
- https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/tests/test_queue.py#L196: if a dataset already has a started job, we prefer starting a job from another dataset
- https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/tests/test_queue.py#L233: if a namespace already has a started job, we prefer starting a job from another dataset
- https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/tests/test_queue.py#L279: the "priority" field respects: HIGH > NORMAL > LOW

I think points 2 and 3 should avoid the current behavior, as mentioned by @lhoestq in https://github.com/huggingface/datasets-server/pull/2141#issuecomment-1827996923

So: maybe the first step is to try to reproduce the issue in the tests

AndreaFrancis:
> So: maybe the first step is to create a test that reproduces the issue (small datasets blocked by a big dataset?)

I simulated the scenario of many workers trying to read the next waiting job in this test https://github.com/huggingface/datasets-server/pull/2171. What I realized is that there were no disadvantages against big datasets and small datasets (nonblocking).
Maybe it was obvious but what is interesting for me is that the number of attempts to start a non-started job increases proportionally with the amount of workers, see https://github.com/huggingface/datasets-server/pull/2171/files#diff-50d332d1aa73d93eb7f9cc97c6423c1326e11d8bffadab918c98cce0f287b2feR243
And since there is a delay between each attempt https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/loop.py#L117, most of the time worker are just trying to get an available job to process (wasting time).
See https://grafana.huggingface.tech/d/0uOH3n-Vk/code-profiling?orgId=1&var-method=loop&var-step=sleep&var-job=datasets-server%2Fprod-datasets-server-worker&var-context=None
image
95% of the time, our workers are sleeping for about 7 seconds.

AndreaFrancis:
I couldn't reproduce the issue for small datasets being blocked by big ones (in a parallel scenario), but I put some helpful information in this PR https://github.com/huggingface/datasets-server/pull/2171

AndreaFrancis:
> The StepProfiler("loop", "sleep") is only entered when a worker has to sleep. But maybe in general it doesn't have to sleep so it's not an issue ? Maybe we need to enter the StepProfiler("loop", "sleep") even if it doesn't sleep in order to take the times it doesn't sleep into account in the stats

I would say that it would be helpful to measure the loop-start step but separate between the jobs that were processed and the ones that were already started but not sure if this is possible with our step profiler.

AndreaFrancis:
As explained in https://github.com/huggingface/datasets-server/pull/2192, this isn't the issue with the slow queue for the next waiting job. The real problem we should first fix is blocked workers by long-running jobs that are never finished.
 I will close this PR an focus in the other one.

PR #2143, opened 2023-11-23
allow dataset script on a new dataset

severo:
asked here (internal): https://huggingface.slack.com/archives/C0311GZ7R6K/p1700727747983759


severo:
cc @clefourrier

PR #2144, opened 2023-11-23
Give insights to the user when we raise TooBigRows

### LINKED ISSUE #2139: Give insights to the user when we raise TooBigRows in /rows, /search, or /filter
https://huggingface.co/datasets/danielz01/xView2/discussions/2#655dc3ccde919d41622b51bd

Currently, we only show

```
Error code:   UnexpectedApiError
```

We should say that the rows contents are too heavy and that we don't show them for that reason (> 400MB in that case)

albertvillanova:
Fix #2139.

### DIFF
--- a/libs/libapi/src/libapi/exceptions.py
+++ b/libs/libapi/src/libapi/exceptions.py
@@ -116,6 +116,15 @@ class SearchFeatureNotAvailableError(ApiError):
         super().__init__(message, HTTPStatus.BAD_REQUEST, "SearchFeatureNotAvailableError", cause, True)
 
 
+class TooBigContentError(ApiError):
+    """The content size in bytes is bigger than the supported value."""
+
+    def __init__(self, message: str, cause: Optional[BaseException] = None):
+        super().__init__(
+            message, HTTPStatus.INTERNAL_SERVER_ERROR, "TransformRowsProcessingError", cause=cause, disclose_cause=True
+        )
+
+
 class TransformRowsProcessingError(ApiError):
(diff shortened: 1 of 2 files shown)


albertvillanova:
@severo thanks for your review.

> Can it also be raised in search/ and filter/?

The `TooBigRows` error (maybe we should rename it to `TooBigRowsError`) is raised by `ParquetIndexWithMetadata.query` and this is only used by:
- services/rows
- services/worker::job_runners.split.first_rows_from_parquet

The "search" and "filter" services use the DuckDB instead.

> Should we simply make TooBigRows inherit from ApiError? Or should we instead catch it in these other services?

I spent some time thinking about this but did not get a clear solution:
- The `TooBigRows` error is defined and raised by "parquet_utils", a module in "libs/libcommon", so it does not make sense that it raises an `ApiError` (defined in "libs/libapi")
- We might move the module "parquet_utils" to "libs/libapi", but the we are forced to make "services/worker" (that also uses "parquet_utils") to depned on "libs/libapi" (with all the other dependencies that are not needed by this worker)

severo:
OK, thanks for explaining. It makes me wonder: it's somewhat incoherent not to check the same on filter and search, no? we could end up sending >500MB in one response on these endpoints?

albertvillanova:
Yes, I thought the same, but we should first find out how to check that when reading from DuckDB file (or before writing the DuckDB file?).

Maybe open a different issue and rename the one associated with this PR?

severo:
yes, sounds perfect

severo (approved):

PR #2157, opened 2023-11-28
Update aiohttp in worker pyproject

albertvillanova:
After checking locally the files, I realized we forgot to update `aiohttp` in the worker pyproject file.

Related to #2156.

CC: @severo


severo:
indeed, thanks, I forgot to check if we explicitly depended on aiohttp in pyproject.toml files. It does not change the installed version (poetry.lock), but this way we ensure previous versions are not supported anymore.

PR #2175, opened 2023-11-30
remove finished jobs

AndreaFrancis:
See https://github.com/huggingface/datasets-server/pull/2137#issue-2005097421 for context.
The next waiting job operation could be affected because of useless job records (finished). This PR proposes removing the jobs in SUCCESS, ERROR, and CANCELLED status so that the query can improve performance and profile better indexing approaches.
- A waiting job becomes started (as usual)
- A started job no longer becomes a success/error/cancelled status, the record will be deleted from the job collection
- Since we don't need the finished job records, TTL index is no longer necessary and the same for `finished_at` field
-  In an upcoming PR, I will add metrics for finished jobs but using Prometheus Counter https://github.com/huggingface/datasets-server/pull/2137#issuecomment-1832674005

Depends on:
https://github.com/huggingface/datasets-server/pull/2178
https://github.com/huggingface/datasets-server/pull/2179


jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -276,4 +278,17 @@ def get_migrations(self) -> list[Migration]:
                 version="20231106193200",
                 description="add 'partial', 'num_rows' and 'num_bytes' fields for 'split-duckdb-index' cache records",
             ),
+            MigrationDeleteJobsByStatus(
+                status_list=["success", "error", "cancelled"],
+                version="20231201074900",
+                description="delete jobs with success, error and cancelled status",
+            ),
+            MigrationQueueDeleteTTLIndex(
```
severo, line 286: remark: I think that we will have to deploy twice to have this index deleted (since during the first sync, old pods will still be running and will recreate the deleted index)

severo: we could instead:
1. create another PR that only removes the index from the code in queue.py, and deploy
2. in this PR: delete the index + the rest of unrelated changes

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -276,4 +278,17 @@ def get_migrations(self) -> list[Migration]:
...
+                version="20231201112000",
+                description="delete the TTL index on the 'finished_at' field in the queue database",
+                field_name="finished_at",
+            ),
+            MigrationRemoveFieldFromJob(
```
severo, line 291: same kind of comment: we will delete the field, but: as the current pods will still finish jobs meanwhile, we will have to run the migration again. Not sure how to do it in two PRs...

Anyway: we can just try and run twice (and fix afterwards if something broke)

severo: (and same comment for the deletion of jobs with a finished status: they will still exist).

Maybe the best strategy would be:
- a PR that removes the "finished_at" field + the index, in the code
- a PR that only runs the three migrations

libs/libcommon/src/libcommon/queue.py
```diff
@@ -544,11 +530,10 @@ def cancel_jobs_by_job_id(self, job_ids: list[str]) -> int:
         """
         try:
             existing = JobDocument.objects(pk__in=job_ids)
-            previous_status = [(job.type, job.status) for job in existing.all()]
-            existing.update(finished_at=get_datetime(), status=Status.CANCELLED)
-            for job_type, status in previous_status:
-                update_metrics_for_type(job_type=job_type, previous_status=status, new_status=Status.CANCELLED)
-            return existing.count()
+            for job in existing.all():
+                decrease_metric(job_type=job.type, status=job.status)
```
severo, line 534: maybe not very important, but instead of N -1 operations, we could do -N directly by grouping by job type.

maybe in another pr

libs/libcommon/src/libcommon/queue.py
```diff
@@ -708,6 +693,7 @@ def _start_newest_job_and_cancel_others(self, job: JobDocument) -> JobDocument:
             # retry for 2 seconds
             with lock(key=job.unicity_id, owner=lock_owner, sleeps=[0.1] * RETRIES, ttl=LOCK_TTL_SECONDS):
                 # get all the pending jobs for the same unicity_id
+                # TODO: remove status__in=[Status.WAITING, Status.STARTED] filter
```
severo, line 696: in a followup PR?

libs/libcommon/src/libcommon/queue.py
```diff
@@ -884,31 +859,30 @@ def finish_job(self, job_id: str, is_success: bool) -> bool:
...
+    def delete_dataset_jobs(self, dataset: str) -> int:
         """
-        Cancel all the jobs for a given dataset.
+        Delete all the jobs for a given dataset.
 
         Args:
             dataset (`str`, required): dataset name
 
         Returns:
             `int`: the number of canceled jobs
```
severo, line 876:
```suggestion
            `int`: the number of deleted jobs
```

libs/libcommon/src/libcommon/queue.py
```diff
@@ -884,31 +859,30 @@ def finish_job(self, job_id: str, is_success: bool) -> bool:
...
+        # TODO: remove status__in=[Status.WAITING, Status.STARTED] filter
         jobs = JobDocument.objects(dataset=dataset, status__in=[Status.WAITING, Status.STARTED])
         previous_status = [(job.type, job.status, job.unicity_id) for job in jobs.all()]
         jobs_to_cancel = len(previous_status)
-        jobs.update(finished_at=get_datetime(), status=Status.CANCELLED)
         for job_type, status, unicity_id in previous_status:
-            update_metrics_for_type(job_type=job_type, previous_status=status, new_status=Status.CANCELLED)
+            decrease_metric(job_type=job_type, status=status)
             release_lock(key=unicity_id)
+        jobs.delete()
```
severo, line 885: maybe let this row above (instead of the previous `jobs.update`) as I assume we want to release the locks only once the jobs are deleted

severo (approved):

libs/libcommon/src/libcommon/queue.py
AndreaFrancis: Yes, since we will remove the unfinished jobs, it won't be necessary to query by the available status. But we can remove the filter until all finished jobs have been removed from the queue (See migration in this PR).

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: I created the following PRs that should be merged/deployed in this order:
https://github.com/huggingface/datasets-server/pull/2178 (removing  "finished_at" field + the index, in the code
https://github.com/huggingface/datasets-server/pull/2179 (migrations)
and then https://github.com/huggingface/datasets-server/pull/2175 (this)

PR #2176, opened 2023-11-30
enable viewer for MERA benchmark

polinaeterna:
The viewer for the [MERA benchmark](https://huggingface.co/datasets/ai-forever/MERA) is not working because it is a script-based dataset but this is a cool benchmark for LLMs evaluation in Russian and I believe it's okay to add it to the allow list for now (along with encouraging the authors to push it as parquet in the future as well).

https://huggingface.co/datasets/ai-forever/MERA/discussions/1


polinaeterna:
i changed my mind nevermind

lhoestq:
It seems that it can be fixed using YAML configuration, since the dataset is just a bunch of JSONL files

PR #2180, opened 2023-12-04
Do not raise error if audio file has no extension

### LINKED ISSUE #2096: TransformRowsProcessingError in /rows with simple audio dataset
https://huggingface.co/datasets/GregoryVandromme/vandromme_dataset/viewer/default/test

https://datasets-server.huggingface.co/rows?dataset=GregoryVandromme/vandromme_dataset&config=default&split=test

```
{"error":"Server error while post-processing the split rows. Please report the issue."}
```

Reported here: https://huggingface.co/datasets/GregoryVandromme/vandromme_dataset/discussions/1

albertvillanova:
This PR enables support for audio no-script datasets on the Hub: audio files downloaded from the Hub have no extension, e.g. the audio column embedded in the Parquet file contains:
```
{'bytes': b'RIFF\x86Y\x0c\x00WAVEfmt \x10...',
 'path': '1b607f504e9c4489e199198847691e2b6b60d99ef50614becb059ac71c22d737'}
```

Currently we are raising `RowsPostProcessingError`: 
```
Traceback (most recent call last):
 File \"/src/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py, line 137, in compute_first_rows_response
   transformed_rows = transform_rows(
 File \"/src/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py, line 42, in transform_rows
   return [
 File \"/src/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py, line 43, in <listcomp>
   {
 File \"/src/services/worker/src/worker/job_runners/split/first_rows_from_parquet.py, line 44, in <dictcomp>
   featureName: get_cell_value( 
 File \"/src/libs/libcommon/src/libcommon/viewer_utils/features.py, line 203, in get_cell_value
   return audio( 
 File \"/src/libs/libcommon/src/libcommon/viewer_utils/features.py, line 126, in audio
   raise ValueError(
ValueError: An audio sample should have a 'path' with a valid extension but got ''.
```

Fix #2096.


lhoestq (commented):
If we don't infer the audio format correctly we might end up doing unnecessary conversions to ".wav" (the default), as per

```python
ext = audio_file_extension if audio_file_extension in [".wav", ".mp3"] else ".wav"
```

This would slow down /rows significantly for example.

Should we set the target `ext` using the file format that we can infer from the header of the file instead (magic bytes) ?

I'm fine with merging this change right now and improving that later though

albertvillanova:
Thanks for your review, @lhoestq.

I understand your concern raised, but note that:
- audio files with extension are not impacted
- audio files without extension are currently not supported (we are raising an error); this PR enables supporting them.

I agree the performance optimization for audio files without extension could be addressed in a subsequent PR.

albertvillanova:
Follow-up issue:
- #2181

lhoestq (approved):
LGTM then :)

PR #2191, opened 2023-12-07
More rows workers

lhoestq:
`/rows` workers get killed since this morning when someone started using it to scrape coyo-700m

image

image


severo (approved):

lhoestq:
we're back to 5req/sec so we're good in the end

PR #2192, opened 2023-12-07
logs and stop executor on failure

AndreaFrancis:
Sometimes, a long job gets finished by another worker (still investigating the cause), and heartbeat logs the issue forever.
Once the heartbeat fails for some reason, the executor should stop to avoid blocking the worker:
image

This scenario keeps being logged forever until the worker restarts, which is why, in each deployment, we have almost the same number of started jobs compared to the number of workers, but when getting into this case, the number of active workers reduces.
I also changed the log level to INFO when finishing a job to investigate the root cause of the message: `"WARNING: 2023-12-07 17:44:14,644 - root - Heartbeat skipped because job 657202cb9c5adfceb6840d88 doesn't exist in the queue." `

Note.- I have already seen this behavior before: https://github.com/huggingface/datasets-server/pull/2175, but I didn't realize that this was the cause of blocked workers.

### DIFF
--- a/libs/libcommon/src/libcommon/queue.py
+++ b/libs/libcommon/src/libcommon/queue.py
@@ -1006,11 +1006,7 @@ class Queue:
         This is used to keep track of running jobs.
         If a job doesn't have recent heartbeats, it means it crashed at one point and is considered a zombie.
         """
-        try:
-            job = self.get_job_with_id(job_id)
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
Cool, I also saw this behavior but couldn't find the cause. Hopefully this helps, in addition to making the restart faster

PR #2214, opened 2023-12-15
increase limits memory for heavy workers

AndreaFrancis:
With 34Gi it fails because of OOMs when processing `split-duckdb-index` for some Wikipedia configs like 20231101.es and 20231101.en when creating index step.

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -412,7 +412,7 @@ workers:
         memory: "34Gi"
       limits:
         cpu: 8
-        memory: "34Gi"
+        memory: "44Gi"
     tolerations: []
   - deployName: "medium"
     prometheusMultiprocDirectory: "/tmp"


severo (approved):

AndreaFrancis:
For Wikipedia 20231101.es config it worked allocating 38G and generating a 10.5G partial index file
See https://datasets-server.huggingface.co/search?dataset=wikimedia%2Fwikipedia&config=20231101.es&split=train&query=Argentina&length=10

lhoestq:
Awesome ! And search works well :)
https://huggingface.co/datasets/wikimedia/wikipedia/viewer/20231101.es/train?q=pokemon

Note that it only searches the first 5GB of the dataset (the duckdb index is a partial one on the first 5GB) even though we don't mention that it's on the first 5GB in the UI yet

severo:
-> I opened https://github.com/huggingface/moon-landing/issues/8429

AndreaFrancis:
For Wikipedia 20231101.en config it worked allocating 33G and generating a 8.33G partial index file
See https://datasets-server.huggingface.co/search?dataset=wikimedia%2Fwikipedia&config=20231101.en&split=train&query=alchemy&length=10

severo:
> Note that it only searches the first 5GB of the dataset (the duckdb index is a partial one on the first 5GB) even though we don't mention that it's on the first 5GB in the UI yet

We now show the 5GB limit. Thanks @lhoestq for asking

Capture d’écran 2023-12-18 à 13 22 58

PR #2219, opened 2023-12-19
Remove abstractions

severo:
Hardcode step names, instead of creating an abstraction around what a step provides (it's too artificial, and we only have one DAG, so: no need to look for genericity)


AndreaFrancis (approved):
I agree with hadcoding job runner version since it is only set in the config.
But for step names, we could at least set them in constants (to avoid typing errors) and use them anywhere now that the steps provided by the processing graph were deleted.
We could use these constants in processing graph configuration, API configuration, workers, and tests.

severo:
Yes, you're right, I thought the same in my bed last night :) I want to simplify the code, but this is too much, we need to maintain type safety

AndreaFrancis (approved):

PR #2220, opened 2023-12-20
Remove abstractions

severo:
- import shared instance of processing_graph from libcommon.processing_graph, instead of instanciating a new one from a (fake) configuration, and passing it through function arguments.
- fix tests locally (`/storage` directory does not exist on every machine; and should not be used for tests)


AndreaFrancis (approved):
Much simpler! Thanks!

Issue #2218, opened 2023-12-19
JobManagerCrashedError jobs are never retried


AndreaFrancis:
Currently, we have 7768 jobs with error_code `JobManagerCrashedError`. Some of them are caused by zombie killer set crashes.

```
Atlas atlas-x5jgb3-shard-0 [primary] datasets_server_cache> db.cachedResponsesBlue.aggregate([{$match:{error_code:"JobManagerCrashedError","details.copied_from_artifact":{$exists:false}}},{$group:{_id:{kind:"$kind"},count:{$sum:1}}},{$sort:{count:-1}}])
[
  { _id: { kind: 'split-duckdb-index' }, count: 3658 },
  { _id: { kind: 'split-descriptive-statistics' }, count: 1872 },
  { _id: { kind: 'config-parquet-and-info' }, count: 1765 },
  { _id: { kind: 'split-first-rows-from-streaming' }, count: 322 },
  { _id: { kind: 'split-first-rows-from-parquet' }, count: 72 },
  { _id: { kind: 'split-opt-in-out-urls-scan' }, count: 60 },
  { _id: { kind: 'dataset-config-names' }, count: 21 }
]

```

But most of them are set as crashed when deploying and are never retried, even if they are fast and straightforward to process.
Should we retry those jobs in backfill?  I think we should differentiate the ones that are easy to process against those that are difficult (primarily because of OOMs), maybe retry once or twice, and set a different error so that we can identify which of them are caused by limited resources.

severo:
yes, I agree; they should be retried. I don't think we can easily differentiate jobs that were running during a deployment, from jobs that crashed, so... add a max number of retries?

severo:
See https://huggingface.co/datasets/ayymen/Pontoon-Translations/discussions/2#659451dca41c3cbad528f88d.

AndreaFrancis:
Retry logic was implemented on https://github.com/huggingface/datasets-server/pull/2221

PR #2221, opened 2023-12-20
Retry JobManagerCrashedError and limit a certain number of retries

AndreaFrancis:
Proposal for https://github.com/huggingface/datasets-server/issues/2218

When saving a new cache result, the "retries" counter will be increased if there already exists another record for the same revision (and kind+dataset+config+split combination). 
If it is the first time it is by default 0.
If there already exists a cache for a revision and there is a new cache result because of a new revision (commit), the counter will start in 0 again.
When creating a new job by backfill process, it will retry jobs if the error_code is in the error_codes_to_retry AND the number of retries is less than `MAX_RETRY_ATTEMPTS.`


lhoestq (approved):
Nice ! One nit: maybe rename "attempts" to "retries" ? I think having "retries: 0" is better than "attempts: 0" (since a job may succeed on first attempt)

severo (commented):
I'm wondering why we need to have a field in JobDocument.

It's only used in worker/job_manager.py, to prepare the cache entry data. But we could instead add the logic when upserting a cache entry, no?

Because the condition on the number of attempts is done on the cache entry field.

AndreaFrancis:
@severo  I removed the field on the Job document, and it looks better. Thanks for the suggestion!
@lhoestq Given that now (with the last commit) the counter will increase each upsert to the cache collection, I think it is better to keep it as "attempts" instead of "retries" what do you think? I mean, now the logic is that whenever a job is run, it will by default put 1 in the counter so it does not make sense to consider it a retry.

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -244,6 +245,7 @@ def upsert_response(
         progress=progress,
         updated_at=updated_at or get_datetime(),
         job_runner_version=job_runner_version,
+        inc__attempts=1,
```
severo, line 248: Hmmmm... this would increment the number of attempts on every update, even if it's the first update for a new commit.
Something is still missing to get what we want:
- store the number of attempts for the current revision
- increment it on every new attempt on the current revision
- set it to 1 (if we keep the word attempt, or to 0 if we switch to retries, which I prefer as it's clearer to me: the default is 0, it's easy) when we set a new revision

Could we have tests in test_simple_cache.py to ensure it works as expected?

severo (changes requested):

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -259,19 +262,37 @@ def upsert_response_params(
...
+    )
+    try:
+        previous_response = get_response_metadata(kind=kind, dataset=dataset, config=config, split=split)
+    except Exception:
+        previous_response = None
+    retries = (
+        0
+        if previous_response is None
+        else 0
+        if previous_response["dataset_git_revision"] != revision
```
severo, line 279: maybe use a ` or ` since the first two branches return 0

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: I changed to "retries" and added a test to verify these scenarios:
- Firs time a cache in saved (retries=0)
- Second time a cache is save for the same revision (retries=1)
- Third time a cache is saved but for a different revision (retries=0)

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -259,19 +262,37 @@ def upsert_response_params(
...
     updated_at: Optional[datetime] = None,
 ) -> None:
+    dataset, config, split, revision = (
+        job_params["dataset"],
+        job_params["config"],
+        job_params["split"],
+        job_params["revision"],
+    )
+    try:
+        previous_response = get_response_metadata(kind=kind, dataset=dataset, config=config, split=split)
```
severo, line 272: maybe directly

```
CachedResponseDocument.objects(kind=kind, dataset=dataset, config=config, split=split, dataset_git_revision=revision).only('retries').get()
```

and catch DoesNotExist

libs/libcommon/src/libcommon/state.py
```diff
@@ -88,6 +89,7 @@ def is_error_to_retry(self) -> bool:
             and (
                 self.cache_entry_metadata["http_status"] >= 400
                 and self.cache_entry_metadata["error_code"] in self.error_codes_to_retry
+                and self.cache_entry_metadata["retries"] <= MAX_RETRY_ATTEMPTS
```
severo, line 92: I think it's `<`. Else we could retry 4 times if MAX_RETRY_ATTEMPTS=3.

libs/libcommon/tests/test_simple_cache.py
```diff
@@ -125,6 +127,52 @@ def assert_metric(http_status: HTTPStatus, error_code: Optional[str], kind: str,
...
+    assert cached_entry_metadata is not None
+    assert cached_entry_metadata["retries"] == 0
+    assert CacheTotalMetricDocument.objects().count() == 1
+    # insert for the second time
+    upsert_response_params(
+        kind=kind,
+        job_params=job_params,
+        content=content,
+        http_status=HTTPStatus.OK,
+    )
```
severo, line 156: OK. that means that if we force-refresh from the admin endpoint, it will increment the value too. I think it's ok.

libs/libcommon/tests/test_backfill.py
```diff
@@ -362,6 +363,37 @@ def test_plan_compute_all(processing_graph: ProcessingGraph, up_to_date: list[st
     )
 
 
+@pytest.mark.parametrize(
+    "retries,max_retries,should_create",
+    [
+        (0, 1, True),
+        (3, 1, False),
+        (2, 4, True),
```
severo, line 371:
```suggestion
        (3, 3, False),
        (2, 4, True),
```

severo: just to check we handle the limit correctly

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -291,4 +292,7 @@ def get_migrations(self) -> list[Migration]:
             MigrationRemoveFieldFromJob(
                 field_name="finished_at", version="20231201112600", description="remove 'finished_at' field from queue"
             ),
+            MigrationAddRetriesToCacheResponse(
+                version="20240104085000", description="add 'retries' field to cache records"
+            ),
```
severo, line 297: when deploying, we might want to stop the workers, in order to be sure to add the field for all the cache entries

severo (commented):
Details

libs/libcommon/tests/test_backfill.py
AndreaFrancis: Done

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: Makes sense, I will take care of it.

lhoestq:
Nice !

Issue #2208, opened 2023-12-14
Add a collection with datasets infos


severo:
While working on enabling private datasets (#39) under conditions (isPro, isEnterprise), I thought we missed a place where we control the access to the dataset.

I think the first step in the DAG, instead of dataset-config-names, should be more about the dataset characteristics: if it's private or public, maybe if it's gated (not sure if it's useful info), if the user is pro or if the org is enterprise, if the viewer is disabled through the README (see https://github.com/huggingface/datasets-server/issues/2207), if the dataset is in the block list.

All that information could go to a new step called `dataset-status` or something similar.

The content could be:

```json
{
  "dataset": "namespace/dataset",
  "private": true,
  "proUser": false,
  "enterpriseOrg": true,
  "disabledFromReadme": false,
  "gated": false,
  "blocked": false,
}
```

And a second step, called `dataset-enabled`, that would depend on `dataset-status`, and would return:
- 200 `{enabled: true}` if all the conditions are met
- 404 if we don't want to disclose the existence of the dataset, or if it does not exist
- 501 if it's not implemented
- 403? 404? if the dataset viewer is not enabled (private dataset, no pro user/enterprise org)

Then, the following steps would propagate the error if so, or if 200, will act as currently.

I think it's clearer to have two different steps: one to collect the data, another one to take a decision on this basis. We could also have everything in one cache entry, but I think the logic for maintenance would be harder (we would have to add info like: is that dataset private, is the user pro, etc. in the error details, or in the content, etc. to be able to check them regularly)

severo:
Note that an alternative for the first step (`dataset-status`) would be to store this info in a new collection (`datasets`), not as the result of a step. Indeed, it's for internal use, and not meant to be the response of an endpoint.

lhoestq:
We would end up with erroring cache entries for all the private datasets, which is a lot of mongo storage for nothing no ?

Imo these two should not be part of the queue/worker logic

severo:
Indeed, good point.

So: maybe store that data in a separate collection as mentioned here:

https://github.com/huggingface/datasets-server/issues/2208#issuecomment-1855916266

And only create jobs and cache entries if some conditions are met (basically: don't create if the dataset is private and is not enabled by a paid plan).

severo:
I think it's not needed (see https://github.com/huggingface/datasets-server/pull/2224). Let's open again if I'm wrong.

PR #2224, opened 2023-12-21
Support private datasets

severo:
This PR changes the logic a bit: operations.py replaces dataset.py and is the module that must be used to update or delete a dataset (ie: do not create jobs or cache entries directly or through DatasetOrchestrator). Indeed, operations.py checks if the dataset is supported (public, or private + pro user or enterprise org. Also checks if the repo is disabled, if the viewer is disabled, or if the dataset is in the list of blocked datasets). It also means that behind that layer, we don't check anymore if the dataset is supported. For example, DatasetOrchestrator is now free from these checks.

Remaining:
- [x] fix tests on hub-ci
  - a) the database has not been updated. See https://huggingface.slack.com/archives/C035N5UE2JU/p1703193635197549 (internal)
  - b) the app token has no right to access the private datasets (see https://github.com/huggingface/moon-landing/pull/8573 <- internal)
- [x] add e2e tests about the private datasets (normal, pro, enterprise). In particular: make sure both app tokens have the appropriate rights (read datasets, bypass gates, write in the refs/convert/parquet branch, etc. for private datasets)
- [x] change the logic of the backfill cronjob. Options:
  a) <strike>check support and completeness (backfill) of each public and private dataset on the Hub (requires getting the list of all the datasets). </strike>
  b) only check if the datasets we currently have in the datasets server are still supported and are complete (backfill). If a dataset is missing, we will create it when we receive the first request on an API endpoint for it. <- let's do this
- [x] change the logic of the obsolete cache cron job (it's currently disabled, see TODO in the code)
- [x] replace the assets/cached-assets URLs with S3 signed URLs (when the configured storage is S3). We will have to change a bit how we handle /first-rows, since it's the only endpoint that returns rows that is not computed on the fly. See https://github.com/huggingface/datasets-server/issues/2280
  - [x] cached-assets: done
  - [x] assets: see https://github.com/huggingface/datasets-server/issues/2280
    - [ ] change the configuration to restrict access to `/assets/*` to signed URLs.  
- [ ] change the settings of the webhook, to receive webhooks on the private datasets too. Otherwise, we will not update the dataset.

Not crucial:
- [x] when we finish a job: should we ensure the dataset still exists and is still supported? Not important: in the worst case, it will be cleaned in the next backfill cron job.
- [x] maybe? cache the info about the supported datasets to avoid calling the Hub all the time? See https://github.com/huggingface/datasets-server/issues/2208 for a proposal. -> I think it's not required.

Unrelated to this PR:
- when we delete a dataset, we should also delete the files (assets, cached_assets, parquet files, etc.) ->
  - [x] done for assets and cached_assets.
  - [x] invalidate the cloudfront cache? no need: the URLs are valid only for 1h...
  - [ ] The parquet files and duckdb index would be deleted if the repo has disappeared, but not if it went "unsupported" (public -> private, disabled repo, disabled viewer, or end of pro/enterprise subscription for private dataset).
  - [ ] the parquet metadata is not deleted either

Note that the following PR on moonlanding should not be needed anymore: https://github.com/huggingface/moon-landing/pull/8262


severo:
I think it's ready for review... Good luck!

libs/libapi/src/libapi/utils.py
```diff
@@ -108,44 +105,39 @@ def try_backfill_dataset_then_raise(
...
+        logging.debug("Cache entry not found but some jobs are still in progress, so it could exist in the future")
         raise ResponseNotReadyError(
             "The server is busier than usual and the response is not ready yet. Please retry later."
         )
-    elif dataset_orchestrator.has_pending_ancestor_jobs(processing_step_names=processing_step_names):
-        # some jobs are still in progress, the cache entries could exist in the future
+    logging.debug("No pending job that could create the expected cache entry")
+    if has_some_cache(dataset=dataset):
+        logging.debug(
+            "Some cache entries exist, so the dataset is supported, but that cache entry will never be created"
```
AndreaFrancis, line 121: Maybe it's a dummy question, but I don't understand why that cache entry will never be created?

AndreaFrancis (approved):
It looks good to me, and thank you for integrating the storage logic in the backfilling process.
If I understand correctly, the obsolete cache job is no longer needed since it will be handled in the backfill process, right?
The only edge case I found is regarding the external URL generated for assets and cached assets. Since we handle them by revision, if one dataset was public and then becomes private, older ULRs will still be working, but not sure if this is dangerous.

severo:
>  If I understand correctly, the obsolete cache job is no longer needed since it will be handled in the backfill process, right?

Yes, exactly

severo:
> The only edge case I found is regarding the external URL generated for assets and cached assets. Since we handle them by revision, if one dataset was public and then becomes private, older ULRs will still be working, but not sure if this is dangerous

Interesting, I forgot this point. I'm not sure what to do. Should we add an authentication layer to the access the assets and cached assets, to be sure that only the logged users with access to the dataset will be able to access them?

If so: how can we add this layer? Directly in S3 (I don't know how) or through some proxy?

cc @Kakulukian @XciD maybe?

libs/libapi/src/libapi/utils.py
severo: At this point, we know that:
- some cache entries exist for the dataset,
- but there is no pending jobs for the specific step, or for any of its ancestors,
so: we're not in the situation where the cache entry is being created. Even if other jobs exist, they will not create the cache entry we need.

XciD:
> > The only edge case I found is regarding the external URL generated for assets and cached assets. Since we handle them by revision, if one dataset was public and then becomes private, older ULRs will still be working, but not sure if this is dangerous
> 
> Interesting, I forgot this point. I'm not sure what to do. Should we add an authentication layer to access the assets and cached assets (such as https://datasets-server.huggingface.co/assets/mnist/--/mnist/train/0/image/image.jpg, to give an example for a public dataset) to be sure that only the logged users with access to the private dataset will be able to access them?
> 
> We were already aware of the issue with images made publicly available for gated datasets, but we decided it was OK, since it's only a small extract of the data.
> 
> But for private dataset, the logic could be different.
> 
> If we want to add authentication: how could we do? Directly in S3 (I don't know how) or through some proxy?
> 
> cc @Kakulukian @XciD @rtrompier maybe?

You may need to do the same as the hub, signing S3 urls and send them back to the client.

severo:
> You may need to do the same as the hub, signing S3 urls and send them back to the client.

wow, looking at https://github.com/huggingface/moon-landing/issues/2207 and all the following issues/PR I could find, it's not simple...

Maybe we have to re-think how we handle the assets in the datasets-server.

severo:
let's first work on https://github.com/huggingface/datasets-server/pull/2244, then see how to add support for the private datasets:
- revert https://github.com/huggingface/datasets-server/pull/2244/commits/5c72d20345909e23158908b8b9dc1a3fefc5b6f2
- most importantly, find how to handle the assets privately

severo:
I'm wondering... maybe we "just" need to use a proxy to access the assets (images or audio), as it's done on images in the gallery of a private dataset.
For example, https://huggingface.co/datasets/severo/private-images/blob/main/1.jpg shows the image

Capture d’écran 2024-01-08 à 18 40 00

The image element is:

```html
<img src="/datasets/severo/private-images/resolve/main/1.jpg" alt="1.jpg">
```

And the URL `"/datasets/severo/private-images/resolve/main/1.jpg` is a proxy, which uses the cookie to redirect to a temporary public URL:

https://cdn-lfs-us-1.huggingface.co/repos/4b/4f/4b4f83f49b2d0658d06831f0998a1070162262a05b34d6cfcfa00fb2fea8e28b/b14aacda1579e2a7df44945758ce8212160ab43d5accfde0fca4ff30d5c393ae?response-content-disposition=inline%3B+filename*%3DUTF-8%27%271.jpg%3B+filename%3D%221.jpg%22%3B&response-content-type=image%2Fjpeg&Expires=1704994818&Policy=eyJTdGF0ZW1lbnQiOlt7IkNvbmRpdGlvbiI6eyJEYXRlTGVzc1RoYW4iOnsiQVdTOkVwb2NoVGltZSI6MTcwNDk5NDgxOH19LCJSZXNvdXJjZSI6Imh0dHBzOi8vY2RuLWxmcy11cy0xLmh1Z2dpbmdmYWNlLmNvL3JlcG9zLzRiLzRmLzRiNGY4M2Y0OWIyZDA2NThkMDY4MzFmMDk5OGExMDcwMTYyMjYyYTA1YjM0ZDZjZmNmYTAwZmIyZmVhOGUyOGIvYjE0YWFjZGExNTc5ZTJhN2RmNDQ5NDU3NThjZTgyMTIxNjBhYjQzZDVhY2NmZGUwZmNhNGZmMzBkNWMzOTNhZT9yZXNwb25zZS1jb250ZW50LWRpc3Bvc2l0aW9uPSomcmVzcG9uc2UtY29udGVudC10eXBlPSoifV19&Signature=PIVlp0FUZEh0%7EyDAdRDKs7ndrKXGqAHG1tnjw4cjbHKmwa10rFeVPYZRsHYR9-mHMQQ9k0nJDD1kZmB3mmSM-yCMz6jWQ4Nx7PztRk6aTM0wCwLnP-OJd9nGmyGNmWsdM0mLnOIhGp55OQJ1RMTg6Z%7EDMI2RqEv-k9vQqpuGVwVE%7Et8LYzbxwTQky8Dhd4c7pbopeUBRlT-wBu8Uf31DvaP2epNxXNmnHmXdzRXjxDW20U2zIxsH24y-ZogY-4V1ij9sUamci%7EN-%7Ezk57OSlMoHvjheMaBkguKQaf2NbWpTVX6qKbc1NPAJERV3UxFOCYjrXLkVOuNqnRf%7EjDkL7FQ__&Key-Pair-Id=KCD77M1F0VK2B

But the image or audio could still be delivered by the proxy itself, right?

severo:
I'm not sure if I would be able to understand how to manage the _signing S3 URLs_ and ensure we handle them without security issues.

Also, @rtrompier, I'm wondering what we should change to the current setup. If I understand well, currently, the bucket contents are public, and directly accessible thanks to the cloudflare configuration (proxy on /assets/...)

XciD:
>And the URL "/datasets/severo/private-images/resolve/main/1.jpg is a proxy, which uses the cookie to redirect to a temporary public URL:

Not a proxy, it's the hub that sign the url.

> I'm not sure if I would be able to understand how to manage the signing S3 URLs and ensure we handle them without security issues.

We can't do a quick huddle tomorrow if you want

> Also, @rtrompier, I'm wondering what we should change to the current setup. If I understand well, currently, the bucket contents are public, and directly accessible thanks to the cloudfront configuration (proxy on /assets/...)

Bucket will became private and only accessible with signed url

severo:
OK, thanks for your comment, and @rtrompier for your explanation on Slack. So... it's very doable! It means recomputing the image/audio URLs for all the datasets, even the public ones, but I think it's not important at all.

Small detail for us: as the URLs will be permanent, we will have to compute new ones (ie invalidate/refresh the cache) when a dataset changes from public to private.

libs/libcommon/tests/test_operations.py
```diff
@@ -58,6 +64,21 @@ def test_get_revision_timeout() -> None:
         )
 
 
+@pytest.mark.parametrize(
+    "name, expected_pro, expected_enterprise",
```
AndreaFrancis, line 68:
```suggestion
    "name,expected_pro,expected_enterprise",
```

AndreaFrancis (approved):

severo:
before merging/deploying: wait for https://github.com/huggingface/moon-landing/pull/8801 to be merged and deployed

severo:
e2e tests are flaky. Not sure if it's an issue with how we wait for the services to be ready before testing, or with something else. Anyway, not related to the PR

severo:
e2e failures:

- attempt # 1
```
tests/test_11_api.py::test_pro_user_private_gated ResponseNotReady
ResponseNotFound
FAILED
```
- attempt # 2
```
tests/test_11_api.py::test_auth_gated[token-200-None] ResponseNotReady
ResponseNotReady
ResponseNotReady
ResponseNotReady
ResponseNotFound
FAILED
```
- attempt # 3
```
tests/test_11_api.py::test_pro_user_private ResponseNotReady
ResponseNotFound
FAILED
```

lhoestq (approved):
Yayyy !

PR #2227, opened 2023-12-29
Fix stats for datasets bigger than 5Gb on first n parquets

polinaeterna:
Currently for datasets with lots of parquet files which were copied from the main branch statistics are not computed with `SplitWithTooBigParquetError` (example: `bigcode/starcoderdata`). While we can take first n parquet files instead, as it's done for `duckdb-index` runner. 

\+ also updated openapi.json with a change not from this PR (bool feature in stats)
\+ added check for "partial" field for /search and /filter endpoints in tests - or were they omitted intentionally?

Questions:
- I added `"partial"` field for statistics runner too but not sure how to update the cache. Previously datasets with partial exports (with `partial-` in directory) worked fine so I had to set `None` as default value to populate cache entry in migration because I cannot set either `True` or `False` (there are already both cases in cache).
- I noticed that `/search` Hub's don't have info about some keys that are returned in response (`num_rows_total`, `num_rows_per_page` and `partial`. Is it intentional? If not, I can add it.


services/worker/README.md
```diff
@@ -92,7 +92,7 @@ Set environment variables to configure the `duckdb-index` worker (`DUCKDB_INDEX_
...
 - `DUCKDB_INDEX_COMMIT_MESSAGE`: the git commit message when the worker uploads the duckdb index file to the Hub. Defaults to `Update duckdb index file`.
 - `DUCKDB_INDEX_COMMITTER_HF_TOKEN`: the HuggingFace token to commit the duckdb index file to the Hub. The token must be an app token associated with a user that has the right to 1. create the `refs/convert/parquet` branch (see `DUCKDB_INDEX_TARGET_REVISION`) and 2. push commits to it on any dataset. [Datasets maintainers](https://huggingface.co/datasets-maintainers) members have these rights. The token must have permission to write. If not set, the worker will fail. Defaults to None.
-- `DUCKDB_INDEX_MAX_DATASET_SIZE_BYTES`: the maximum size in bytes of the dataset's parquet files to index. Datasets with bigger size are ignored. Defaults to `100_000_000`.
+- `DUCKDB_INDEX_MAX_DATASET_SIZE_BYTES`: size in bytes by which the number of first `n` parquet files to index is determined, satisfies the following condition: `sum(parquet_files[0] + ... + parquet_files[n-1]) <= DUCKDB_INDEX_MAX_DATASET_SIZE_BYTES & sum(parquet_files[0] + ... + parquet_files[n]) > DUCKDB_INDEX_MAX_DATASET_SIZE_BYTES`. Defaults to `100_000_000`.
```
polinaeterna, line 95: idk how to explain this more clearly....... but this is exactly what we do (from my understanding) and not what is said everywhere (that this number is the exact maximum)

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -527,19 +526,24 @@ def compute_descriptive_statistics_response(
...
-            f"Statistics computation is limited to split parquets under {max_parquet_size_bytes} bytes. "
-            f"Current size of sum of split parquets is {split_parquets_size} bytes."
-        )
+    num_parquet_files_to_index = 0
+    total_bytes = 0
+    split_directory = split_parquet_files[0]["url"].rsplit("/", 2)[1]
+    partial_parquet_export = parquet_export_is_partial(split_parquet_files[0]["url"])
+    for parquet_file_id, parquet_file in enumerate(split_parquet_files):
+        num_parquet_files_to_index += 1
+        parquet_file_size_bytes = parquet_file["size"]
```
polinaeterna, line 535: btw why in `duckdb-index` worker the size is taken from `config-parquet-metadata` step and read again with `pyarrow.parquet` @lhoestq ? 
i mean this: https://github.com/huggingface/datasets-server/blob/348567390f4dfbcf5d93aec462f07c46d09cd9ad/services/worker/src/worker/job_runners/split/duckdb_index.py#L128

docs/source/openapi.json
```diff
@@ -6105,7 +6132,132 @@
...
                           }
                         }
-                      ]
+                      ],
+                      "partial": false
+                    }
+                  },
+                  "A split (lmsys/lmsys-chat-1m) with a boolean column": {
+                    "summary": "Statistics on a boolean column 'redacted'.",
+                    "description": "Try with https://datasets-server.huggingface.co/statistics?dataset=lmsys/lmsys-chat-1m&config=default&split=train.",
```
AndreaFrancis, line 6141: Looks like `lmsys/lmsys-chat-1m` is a gated dataset, maybe point to a public one so that it is easier for the users to see the data?

jobs/mongodb_migration/tests/migrations/test_20240112164500_cache_add_partial_field_in_split_descriptive_statistics.py
```diff
@@ -0,0 +1,81 @@
...
+def assert_unchanged(dataset: str, kind: str) -> None:
+    db = get_db(CACHE_MONGOENGINE_ALIAS)
+    entry = db[CACHE_COLLECTION_RESPONSES].find_one({"dataset": dataset, "kind": kind})
+    assert entry is not None
+    assert "partial" not in entry["content"]
+
+
+def test_cache_add_partial(mongo_host: str) -> None:
+    kind = "split-descriptive-statistics"
+    with MongoResource(database="test_cache_add_has_fts", host=mongo_host, mongoengine_alias="cache"):
```
AndreaFrancis, line 31:
```suggestion
    with MongoResource(database="test_cache_add_partial", host=mongo_host, mongoengine_alias="cache"):
```

chart/env/prod.yaml
```diff
@@ -163,7 +163,7 @@ rowsIndex:
   maxArrowDataInMemory: "300_000_000"
 
 descriptiveStatistics:
-  maxParquetSizeBytes: "5_000_000_000"
+  maxDatasetSizeBytes: "10_000_000_000"
```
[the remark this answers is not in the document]
severo: Yes, I think it's better to keep the value unchanged here, and update in another dedicated PR. In particular, it means finding which datasets to recompute.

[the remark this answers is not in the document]
severo: So

```suggestion
  maxDatasetSizeBytes: "5_000_000_000"
```

chart/templates/_env/_envWorker.tpl
```diff
@@ -110,8 +110,8 @@
 # specific to 'split-descriptive-statistics'
 - name: DESCRIPTIVE_STATISTICS_HISTOGRAM_NUM_BINS
   value: {{ .Values.descriptiveStatistics.histogramNumBins | quote }}
-- name: DESCRIPTIVE_STATISTICS_MAX_PARQUET_SIZE_BYTES
-  value: {{ .Values.descriptiveStatistics.maxParquetSizeBytes | quote }}
+- name: DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES
```
severo, line 113: or even `MAX_SPLIT_SIZE_BYTES`? to be crystal clear that it's the size in bytes of the split's data.

docs/source/statistics.mdx
```diff
@@ -54,9 +54,10 @@ curl https://datasets-server.huggingface.co/statistics?dataset=glue&config=cola&
 </curl>
 </inferencesnippet>
 
-The response JSON contains two keys:
-* `num_examples` - number of samples in a split
+The response JSON contains three keys:
+* `num_examples` - number of samples in a split or number of samples in the first N Gb of data (see `partial` field below).
```
severo, line 58: nit: not more or less correct, as N can be any value, but I think it's easier to relate to current limit of 5 GB

```suggestion
* `num_examples` - number of samples in a split or number of samples in the first N GB of data (see `partial` field below).
```

docs/source/statistics.mdx
```diff
@@ -54,9 +54,10 @@ curl https://datasets-server.huggingface.co/statistics?dataset=glue&config=cola&
...
 * `statistics` - list of dictionaries of statistics per each column, each dictionary has three keys: `column_name`, `column_type`, and `column_statistics`. Content of `column_statistics` depends on a column type, see [Response structure by data types](./statistics#response-structure-by-data-type) for more details
+* `partial` - `"true"` if statistics are computed on the first 5 Gb of data, not on the full dataset, `"false"` otherwise.
```
severo, line 60:
```suggestion
* `partial` - `"true"` if statistics are computed on the first 5 GB of data, not on the full split, `"false"` otherwise.
```

tools/docker-compose-datasets-server.yml
```diff
@@ -197,7 +197,7 @@ services:
       CONFIG_NAMES_MAX_NUMBER: ${CONFIG_NAMES_MAX_NUMBER-3_000}
       DESCRIPTIVE_STATISTICS_CACHE_DIRECTORY: ${DESCRIPTIVE_STATISTICS_CACHE_DIRECTORY-/stats-cache}
       DESCRIPTIVE_STATISTICS_HISTOGRAM_NUM_BINS: ${DESCRIPTIVE_STATISTICS_HISTOGRAM_NUM_BINS-10}
-      DESCRIPTIVE_STATISTICS_MAX_PARQUET_SIZE_BYTES: ${DESCRIPTIVE_STATISTICS_MAX_PARQUET_SIZE_BYTES-100_000_000}
+      DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES: ${DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES-5_000_000_000}
```
severo, line 200: let's keep the same value? Or does it help to change it?

```suggestion
      DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES: ${DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES-100_000_000}
```

services/worker/README.md
```diff
@@ -102,13 +102,15 @@ Set environment variables to configure the `duckdb-index` worker (`DUCKDB_INDEX_
 Set environment variables to configure the `descriptive-statistics` worker (`DESCRIPTIVE_STATISTICS_` prefix):
 
 - `DESCRIPTIVE_STATISTICS_CACHE_DIRECTORY`: directory to which a dataset in parquet format is downloaded. Defaults to empty.
+- `DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES`: size in bytes by which the number of first `n` parquet files to compute statistics over is determined, satisfies the following condition: `sum(parquet_files[0] + ... + parquet_files[n-1]) <= DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES & sum(parquet_files[0] + ... + parquet_files[n]) > DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES`. Defaults to `100_000_000`.
 - `DESCRIPTIVE_STATISTICS_HISTOGRAM_NUM_BINS`: number of histogram bins (see examples below for more info).
 - `DESCRIPTIVE_STATISTICS_MAX_PARQUET_SIZE_BYTES`: maximum size in bytes of the dataset's parquet files to compute statistics. Datasets with bigger size are ignored. Defaults to `100_000_000`.
-
+variable
```
severo, line 108:
```suggestion

```

services/worker/README.md
severo: is it the sum of the parquet files? in which case, it could contain a lot more than 5GB of uncompressed "data".

services/worker/README.md
```diff
@@ -102,13 +102,15 @@ Set environment variables to configure the `duckdb-index` worker (`DUCKDB_INDEX_
 Set environment variables to configure the `descriptive-statistics` worker (`DESCRIPTIVE_STATISTICS_` prefix):
 
 - `DESCRIPTIVE_STATISTICS_CACHE_DIRECTORY`: directory to which a dataset in parquet format is downloaded. Defaults to empty.
+- `DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES`: size in bytes by which the number of first `n` parquet files to compute statistics over is determined, satisfies the following condition: `sum(parquet_files[0] + ... + parquet_files[n-1]) <= DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES & sum(parquet_files[0] + ... + parquet_files[n]) > DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES`. Defaults to `100_000_000`.
```
severo, line 105: here too: DATASET -> SPLIT? to make it clear that it's the size of a split's data?

services/worker/README.md
```diff
@@ -102,13 +102,15 @@ Set environment variables to configure the `duckdb-index` worker (`DUCKDB_INDEX_
...
 - `DESCRIPTIVE_STATISTICS_HISTOGRAM_NUM_BINS`: number of histogram bins (see examples below for more info).
 - `DESCRIPTIVE_STATISTICS_MAX_PARQUET_SIZE_BYTES`: maximum size in bytes of the dataset's parquet files to compute statistics. Datasets with bigger size are ignored. Defaults to `100_000_000`.
-
+variable
 #### How descriptive statistics are computed 
 
 Descriptive statistics are currently computed for the following data types: strings, floats, and ints (including `ClassLabel` int). 
-Response has two fields: `num_examples` and `statistics`. `statistics` field is a list of dicts with three keys: `column_name`, `column_type`, and `column_statistics`.
+Response has three fields: `num_examples`, `statistics` and `partial`. `partial` indicates if statistics are computed over first ~`DESCRIPTIVE_STATISTICS_MAX_DATASET_SIZE_BYTES` of a dataset. If it equals to `False`, that means that `num_examples` correspond to the number of examples in this first chunk of data, not of the full dataset.
```
severo, line 112: wow, it's complex. Is it possible to change this, so that, if `partial: False`, we are sure that the statistics are for the full split? Else, it's not useful for the Hub, it does not bring enough information.

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -527,19 +526,24 @@ def compute_descriptive_statistics_response(
...
+        total_bytes += parquet_file_size_bytes
+        if total_bytes > max_dataset_size_bytes:
```
severo, line 537: it means that we will compute the stats for more than the limit, right? (off by one issue?)

services/worker/README.md
severo: > If it equals to `False`, that means that `num_examples` correspond to the number of examples in this first chunk of data, not of the full dataset. 

hmmm, from the code, I understand that if it's False, then all the split's data have been used, no?

libs/libcommon/src/libcommon/exceptions.py
```diff
@@ -463,13 +463,6 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
         )
 
 
-class SplitWithTooBigParquetError(CacheableError):
```
severo, line 466: To automatically recompute the stats for the big splits, we could add `SplitWithTooBigParquetError` to the list of retryable errors, and the first backfill cron job run after the deployment should create the missing jobs.

And later, once the database is free from this error code, we could remove it.

jobs/mongodb_migration/src/mongodb_migration/migrations/_20240112164500_cache_add_partial_field_in_split_descriptive_statistics.py
```diff
@@ -0,0 +1,54 @@
...
+        db = get_db(CACHE_MONGOENGINE_ALIAS)
+        db[CACHE_COLLECTION_RESPONSES].update_many(
+            {
+                "kind": "split-descriptive-statistics",
+                "http_status": 200,
+                "content.partial": {"$exists": False},
+            },
+            {
+                "$set": {
+                    "content.partial": None,
```
severo, line 31: is it useful to have this field set to None, as it adds no information? Or should we compute all the statistics?

jobs/mongodb_migration/tests/migrations/test_20240112164500_cache_add_partial_field_in_split_descriptive_statistics.py
polinaeterna: omg thanks for the catch!

libs/libcommon/src/libcommon/exceptions.py
severo: I think we should do it. See https://github.com/huggingface/datasets-server/blob/a5c69da73ba8fe9e419e6ca1996134ab645ccf42/libs/libcommon/src/libcommon/constants.py#L38-L45

We can add the error code there, with a comment to remove it later

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: to be sure not to forget this, maybe open an issue for duckdb-index?

polinaeterna: this is actually related to the idea that i misunderstood - that we take size of raw data, while `size` key here corresponds to parquet file size itself. 

would be nice if we stored raw data size too though, to simplify this code but okay :D

libs/libcommon/src/libcommon/exceptions.py
```diff
@@ -463,6 +463,14 @@ def __init__(self, message: str, cause: Optional[BaseException] = None):
         )
 
 
+# TODO: to be deleted after cache with this error is recomputed
```
severo, line 466: I think we don't need it here. Only in the `ERROR_CODES_TO_RETRY` list

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: oops i just saw this comment. yes! this is copied from duckdb-index. that's why i was struggling so much with the description :D but now I just put the word `"approximately"` everywhere, i think it's enough: d9c11fb, e1c45af

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -503,44 +514,41 @@ def compute_descriptive_statistics_response(
...
-        )
 
-    split_parquets_size = sum(parquet_file["size"] for parquet_file in split_parquet_files)
-    if split_parquets_size > max_parquet_size_bytes:
-        raise SplitWithTooBigParquetError(
-            f"Statistics computation is limited to split parquets under {max_parquet_size_bytes} bytes. "
-            f"Current size of sum of split parquets is {split_parquets_size} bytes."
-        )
+    if not features:
+        raise FeaturesError("No features found in cache. ")
```
polinaeterna, line 542: not sure about this error, maybe better `PreviousStepFormatError`?

polinaeterna (commented):
i fixed an issue with taking compressed data size instead of raw size [e1c45af](https://github.com/huggingface/datasets-server/pull/2227/commits/e1c45af6afe675ae3785560132457d8b2808f041)

should be fine now

jobs/mongodb_migration/src/mongodb_migration/migrations/_20240112164500_cache_add_partial_field_in_split_descriptive_statistics.py
```diff
@@ -0,0 +1,65 @@
...
+        db = get_db(CACHE_MONGOENGINE_ALIAS)
+        db[CACHE_COLLECTION_RESPONSES].update_many(
+            {
+                "kind": "split-descriptive-statistics",
+                "http_status": 200,
+                "content.partial": {"$exists": False},
+            },
+            {
+                "$set": {
+                    "content.partial": False,
```
polinaeterna, line 31: oh no i realized that it's actually not always the case :(( some partial datasets worked before this PR - what's not working is when parquet files in parquet export are copied from the main branch and there are too many of them. For example, [CATMuS/medieval](https://huggingface.co/datasets/CATMuS/medieval/tree/refs%2Fconvert%2Fparquet/default/train) is not working (this PR aims to fix the latter case) but partial [OpenOrca](https://huggingface.co/datasets/Open-Orca/OpenOrca/tree/refs%2Fconvert%2Fparquet/default/partial-train) works. I don't know if it's easy to construct the rule that would account for these cases correctly here in migration somehow

polinaeterna: so basically i need to put `partial: True` for splits with 200 HTTP code in stats and `partial-` in directory name (so for example, as far as i understand, when `kind: "config-parquet" & content.partial: True`), something like this...

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -479,17 +485,22 @@ def compute_descriptive_statistics_response(
         histogram_num_bins (`int`):
             (Maximum) number of bins to compute histogram for numerical data.
             The resulting number of bins might be lower than the requested one for integer data.
-        max_parquet_size_bytes (`int`):
-            The maximum size in bytes of the dataset's parquet files to compute statistics.
-            Datasets with bigger size are ignored.
+        max_split_size_bytes (`int`):
+            If raw uncompressed split data is larger than this value, the statistics will are computed
```
severo, line 489:
```suggestion
            If raw uncompressed split data is larger than this value, the statistics are computed
```

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: I'm not sure it can exist. But I would say: create a new exception (`EmptyFeaturesError`?) so that we know where it comes from in case it raises.

severo: Or do we ensure in config-parquet-and-info that it's never empty? In that case, yes, it would be `PreviousStepFormatError` I think

severo: Hmmm: we don't at https://github.com/huggingface/datasets-server/blob/42d945dc5702d1a3016a699993078bb01fdf6774/services/worker/src/worker/job_runners/config/parquet_and_info.py#L1260

So: it's theorically possible for the features dict to be empty

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: ok

polinaeterna:
@severo i think i addressed your comments, hopefully migration is correct now...

severo (approved):
nice!

jobs/mongodb_migration/src/mongodb_migration/migrations/_20240112164500_cache_add_partial_field_in_split_descriptive_statistics.py
```diff
@@ -0,0 +1,95 @@
...
+        # See https://docs.mongoengine.org/guide/migration.html#example-1-addition-of-a-field
+        logging.info(
+            "If missing, add the 'partial' field with the default value None"
+            " to the cached results of split-descriptive-statistics job runner"
+        )
+        db = get_db(CACHE_MONGOENGINE_ALIAS)
+        partial_configs_entries = db[CACHE_COLLECTION_RESPONSES].find(
+            {
+                "kind": "config-parquet",
+                "content.partial": True,
```
AndreaFrancis, line 27: Maybe should filter `"http_status": 200` so that it can help speed up the query.

AndreaFrancis (approved):

Issue #2229, opened 2024-01-02
Upgrade to datasets@2.16.1


severo:
We're still using 2.15.0 from November! We could upgrade to 2.16.1. Release notes: https://github.com/huggingface/datasets/releases/tag/2.16.0 and https://github.com/huggingface/datasets/releases/tag/2.16.1

Main changes to to in datasets-server:
- remove custom code to disable datasets script, and simply pass `trust_remote_code=False` (and `True` for datasets in allowlist)
- check that webdatasets work well, and show it in the docs
- https://github.com/huggingface/datasets/issues/6397:  DatasetNotFoundError and DataFilesNotFoundError are now two different exceptions: see how we should handle them in the code
- https://github.com/huggingface/datasets/pull/6449 / https://github.com/huggingface/datasets/issues/6442#issuecomment-1822561211: ensure it fixes https://huggingface.co/datasets/severo/doc-image-4 (it should show another column with the labels). If so: update the docs and the datasets collection
- something else?

Maybe upgrade to huggingface_hub@0.20.1 at the same time (https://github.com/huggingface/huggingface_hub/releases/tag/v0.20.1 and https://github.com/huggingface/huggingface_hub/releases/tag/v0.20.0) as:
- it brings improvements in globs (required for `load_dataset` speedup)
- we are currently pinning a specific commit, which is brittle

PR #2231, opened 2024-01-03
Update `datasets` to 2.16.1

lhoestq:
...and hfh to 0.20.1

This enables:

- Webdataset support for the Dataset Viewer
- Faster data files resolution, which makes listing configs, splits and generating parquet export faster
- Less pressure on the HF back-end (less /tree calls on the Hub, and they no longer use expand=true)

In addition to the update:

- I removed the #2001 hot-fix and replaced it with a proper use of `trust_remote_code`
- I set `HF_DATASETS_TRUST_REMOTE_CODE=0` by default
- I set `datasets.config.USE_PARQUET_EXPORT = False` since we don't need to query parquer exports when loading datasets (this is just to avoid unnecessary calls)

Fix https://github.com/huggingface/datasets-server/issues/2229

TODO:

- [x] fix tests

TODO after the update + deploy:

- [ ] re-generate webdatasets on the hub, e.g.
  - [ ] https://huggingface.co/datasets/timm/imagenet-12k-wds
  - [ ] https://huggingface.co/datasets/pixparse/cc12m-wds
  - [ ] https://huggingface.co/datasets/pixparse/cc3m-wds


lhoestq:
Not sure why I'm getting auth errors in the e2e tests :/
Cc @severo maybe ?

severo:
Looking at it

severo:
I think it's this: https://github.com/huggingface/datasets-server/pull/2233

severo:
and, even if the token is old, the value in the database was different and never changed, before https://github.com/huggingface/moon-landing/commit/aab4a9b05562e1256c902552fdfb5eed78a21f03 (which was used yesterday to delete the user sessions collection on hub-ci, which replaced the value with the correct one)

cc @coyotte508 for visibility

libs/libcommon/src/libcommon/__init__.py
```diff
@@ -1,2 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
+
+from datasets import config as _datasets_config
+
+_datasets_config.USE_PARQUET_EXPORT = False
```
severo, line 6: maybe add your comment in the code, to understand why we set it to False

severo (approved):
YEAH

lhoestq:
Still got an e2e test fail on `main` because of auth: https://github.com/huggingface/datasets-server/actions/runs/7420424182/job/20191781984

re-running...

PR #2234, opened 2024-01-04
chore: 🤖 use the same packages in all the docker imges

severo:
we could surely build a base image, if somebody wants to do it...

it should fix the following error message we see in the admin service logs:

```
/src/services/admin/.venv/lib/python3.9/site-packages/pydub/utils.py:170: RuntimeWarning: Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work 
  warn("Couldn't find ffmpeg or avconv - defaulting to ffmpeg, but may not work", RuntimeWarning)
```

### DIFF
--- a/jobs/cache_maintenance/Dockerfile
+++ b/jobs/cache_maintenance/Dockerfile
@@ -16,7 +16,9 @@ ENV PYTHONFAULTHANDLER=1 \
 
 # System deps:
 RUN apt-get update \
-    && apt-get install -y build-essential python3-dev make \
+    && apt-get install -y build-essential unzip wget make \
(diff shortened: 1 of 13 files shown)


AndreaFrancis (approved):

AndreaFrancis:
I think in the future, we could also separate libcommon in different libraries according to the content of the utilities. For admin, I don't think we really need `pydub`, but maybe it is referenced because of libcommon even though it only needs to know about datasets server logic for graph and processing steps.

severo:
Exactly. When I started the project, I had more libraries, and it was more painful, so I grouped everything in one... But if needed we could split again.

PR #2236, opened 2024-01-05
add filter field to is-valid

### LINKED ISSUE #2235: Add field "filter" to /is-valid endpoint
Currently, in moon-landing, we use the "search" field to enable the search and the filter. But on https://huggingface.co/datasets/Andyrasika/ShoeSandalBootimages/viewer/default/train?p=1, for example, the search is disabled because we have no text columns, while the filter works (see https://datasets-server.huggingface.co/filter?dataset=Andyrasika/ShoeSandalBootimages&config=default&split=train&where=label=1&length=1).

See the related internal issue: https://github.com/huggingface/moon-landing/issues/8590

severo:
fixes #2235

### DIFF
--- a/services/worker/src/worker/job_runners/split/is_valid.py
+++ b/services/worker/src/worker/job_runners/split/is_valid.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2023 The HuggingFace Authors.
 
+import contextlib
 import logging
 
 from libcommon.constants import CONFIG_HAS_VIEWER_KINDS, SPLIT_HAS_PREVIEW_KINDS, SPLIT_HAS_SEARCH_KINDS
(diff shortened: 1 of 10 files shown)


services/worker/src/worker/job_runners/split/is_valid.py
```diff
@@ -48,19 +49,20 @@ def compute_is_valid_response(dataset: str, config: str, split: str) -> IsValidR
         kinds=SPLIT_HAS_PREVIEW_KINDS,
     )
 
-    try:
+    filter = False
+    search = False
+    with contextlib.suppress(Exception):
```
lhoestq, line 54: (nit) in terms of coding style we use try/except instead of suppress, but not important

lhoestq (approved):
Cool ! thanks for the fix

services/worker/src/worker/job_runners/split/is_valid.py
severo: ah ok. I'll revert

severo: indeed, the code is easier to read after that change

PR #2239, opened 2024-01-08
add missing 'filter' field in IsValidResponse

severo:


severo:
merging, because it's straightforward

PR #2240, opened 2024-01-08
add .ogg as supported audio format

severo:


severo:
merging, since it's simple

PR #2242, opened 2024-01-08
add filter column + ignore errors (fixes #2241)

### LINKED ISSUE #2241: The admin UI is not able to handle some gated datasets
See 
Capture d’écran 2024-01-08 à 13 47 41
Capture d’écran 2024-01-08 à 13 48 27

As I've not been accepted, I cannot see the status of the dataset in datasets-server.

I think that the admin UI should show a special value in the dataframe for the gated dataset that misses access, instead of hiding the result for all the datasets

severo:
fixes #2241

### DIFF
--- a/front/admin_ui/app.py
+++ b/front/admin_ui/app.py
@@ -67,8 +67,8 @@ with gr.Blocks() as demo:
                 gr.Markdown("### Dataset infos")
                 home_dashboard_trending_datasets_infos_by_builder_name_table = gr.DataFrame(pd.DataFrame({"Builder name": [], "Count": [], r"% of all datasets with infos": [], r"% of all public datasets": []}))
                 gr.Markdown("### Trending datasets coverage (is-valid)")
-                home_dashboard_trending_datasets_coverage_stats_table = gr.DataFrame(pd.DataFrame({"Num trending datasets": [], "HTTP Status": [], "Preview": [], "Viewer": [], "Search": []}))
-                home_dashboard_trending_datasets_coverage_table = gr.DataFrame(pd.DataFrame({"All trending datasets": [], "HTTP Status": [], "Preview": [], "Viewer": [], "Search": []}))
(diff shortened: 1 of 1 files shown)


lhoestq (approved):

severo:
The admin UI has the filter column now

severo:
I don't get why lmsys/lmsys-chat-1m has a red cross for filter if https://datasets-server.huggingface.co/is-valid?dataset=lmsys/lmsys-chat-1m

```json
{"preview":true,"viewer":true,"search":true,"filter":true}
```

severo:
I think it's due to some responses not having 4 fields in the response... and we don't hardcode the number of columns or the name of the fields, so... maybe we're filling another row.

severo:
Indeed, the second one: https://datasets-server.huggingface.co/is-valid?dataset=andersonbcdefg/synthetic_retrieval_tasks

has no filter field:

```json
{"preview":true,"viewer":true,"search":true}
```

severo:
I don't want to spend too much time on the admin UI. @lhoestq feel free to look at it if you want (to hardcode the expected fields, maybe). I'll be looking at why some cache entries still miss the 'filter' field

lhoestq:
ok !

lhoestq:
It's a green cross now:

image

severo:
yes because https://datasets-server.huggingface.co/is-valid?dataset=andersonbcdefg/synthetic_retrieval_tasks is fixed and has 4 fields now, not 3

lhoestq:
I mean for lmsys/lmsys-chat-1m ;)

severo:
> I mean for lmsys/lmsys-chat-1m ;)

I mean this too, and that's the bug: the red cross in lmsys/lmsys-chat-1m was a consequence of andersonbcdefg/synthetic_retrieval_tasks only having 3 fields. Because we fill the cells in order, not necessarily 4 cells per rows, we can introduce a shift (1 cell on one row, then 3 cells in the following one, which is completely incoherent)...

PR #2244, opened 2024-01-08
Refactor dataset updates but dont support private datasets

severo:
This PR contains the refactoring needed for supporting the private datasets (https://github.com/huggingface/datasets-server/pull/2224), saving the support for private datasets itself.

The most important part is that we now control a dataset only with:
- operations.update()
- operations.delete()
It's where we check if a dataset is supported (blocked, private, viewer is disabled, etc)

Once a job has been created, we consider the dataset as supported, and we don't check it anymore in the next steps.

Also:
- refactor the tests a bit to speed up the e2e tests, and to reduce complexity
- remove two unused (I think) admin endpoints: /backfill-plan and /backfill (and remove them from admin UI)
- delete the assets and cached_assets from a dataset when we call operations.delete_dataset, which removes the need to run a cron job to delete obsolete assets
- the backfill job is different: it only checks the current entries in the database to complete them if needed, or delete them if they are not supported anymore. It does not allow anymore to add a dataset that was not in the database. Anyway, if a dataset page is loaded, and the dataset is missing from the cache, a new job will be created immediately, so... not so important, I think.
- more tests


AndreaFrancis (approved):

libs/libcommon/src/libcommon/operations.py
```diff
@@ -2,41 +2,89 @@
...
+    return (
+        SupportStatus.UNSUPPORTED
+        if (
+            is_blocked(dataset_info.id, blocked_datasets)
+            # ^ TODO: should we check here?
+            # As we don't keep the track of unsupported datasets,
+            # we lose the ability to send a meaningful error message to the user, as previously done:
+            #   "This dataset has been disabled for now. Please open an issue in"
+            #   " https://github.com/huggingface/datasets-server if you want this dataset to be supported."
+            # A solution is to double-check if the dataset is blocked in the API routes.
```
lhoestq, line 57: sounds like a simple solution

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -647,253 +667,196 @@ class DatasetRemovalPlan(Plan):
...
+        context=f"dataset={dataset}",
+    ):
+        logging.info(f"Analyzing {dataset}")
         with StepProfiler(
-            method="DatasetOrchestrator.set_revision",
-            step="all",
-            context=f"dataset={self.dataset}",
+            method="set_revision",
+            step="plan",
+            context=f"dataset={dataset}",
```
lhoestq, line 747: Should we stop passing "context" ? It makes Grafana unusable since there are so many datasets

lhoestq (approved):
LGTM ! It makes me wonder if there is an opportunity to make the BackFillPlan run faster (couldn't check on Grafana how much time it takes though)

libs/libcommon/src/libcommon/operations.py
severo: I'll do it in another PR.

libs/libcommon/src/libcommon/orchestrator.py
severo: Makes sense, lets clean that in another PR. We can pass what we want, really it's up to you if it helps. Feel free to open an issue.

severo:
> It makes me wonder if there is an opportunity to make the BackFillPlan run faster (couldn't check on Grafana how much time it takes though)

Maybe... do you have an idea?

severo:
thanks both for the review!

libs/libcommon/src/libcommon/operations.py
severo: Done here: https://github.com/huggingface/datasets-server/pull/2246

PR #2246, opened 2024-01-08
return a specific error message when dataset is not supported

severo:
See https://github.com/huggingface/datasets-server/pull/2244#discussion_r1445259783

Instead of returning DatasetNotFound, return the reason why the dataset is not supported: disabled repo, disabled viewer, private repo, or blocked dataset. When we do the check, the user is already authenticated and authorized to access the dataset, so: it makes no sense to hide the reason as if the dataset would not exist (the initial idea was to avoid disclosing the existence of a private dataset).

Also: refactoring to "simplify": we raise if a dataset is not supported. No need for a "DatasetStatus" abstraction

Also (unrelated): delete the docker network when running make test, to avoid errors.

### DIFF
--- a/e2e/tests/test_11_api.py
+++ b/e2e/tests/test_11_api.py
@@ -231,7 +231,7 @@ def test_normal_user_blocked(csv_path: str) -> None:
         # ^ should be caught by COMMON_BLOCKED_DATASETS := "__DUMMY_DATASETS_SERVER_USER__/blocked-*"
     ) as dataset:
         poll_parquet_until_ready_and_assert(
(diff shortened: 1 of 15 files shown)


severo:
still an issue with the e2e tests

severo:
On this basis, it will be easier to implement https://github.com/huggingface/datasets-server/issues/1823#issuecomment-1881155186 (delete all the cache when we update a dataset from a webhook) @lhoestq @AndreaFrancis. I'll create a PR for that once this one is merged.

AndreaFrancis (approved):

lhoestq (approved):
lgtm :)

PR #2247, opened 2024-01-09
Change "retries" for "failed_runs" in cache collection

AndreaFrancis:
For aggregated steps, the "retries" field is increased each time a child is processed, which can prevent retrying jobs when there is an error.
See for example:
image

This cache has been updated 11 times because of its children, but it was a crash error from "dataset-size" last time. Since it already has 11 retries, backfill might not consider it to append a new job.
New logic (great suggestion by @severo):
- remove "retries" and instead use "failed_runs"
- increase "failed_runs" only when the cached result is different than 200 OK and is the same revision

I moved the increased logic to orchestrator instead of simple_cache because the responsibility of simple cache should only be a layer of access to the database and not to business logic.
This way, we should always have retries less than the limit = 3 https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/constants.py#L23


severo:
What if we set it to 0 when http_status is 200 OK? We're just interested in incrementing if we have an error. The field could reflect that it only counts errors: `failed_runs`?

AndreaFrancis:
> We're just interested in incrementing if we have an error. The field could reflect that it only counts errors: failed_runs?

It would be easier to understand. I can also change the field name and prepare a migration for it.

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -787,6 +790,21 @@ def finish_job(
         processing_step = processing_graph.get_processing_step_by_job_type(job_info["type"])
     except ProcessingStepDoesNotExist as e:
         raise ValueError(f"Processing step for job type {job_info['type']} does not exist") from e
+
+    try:
+        previous_response = get_response_metadata(
+            kind=processing_step.cache_kind, dataset=params["dataset"], config=params["config"], split=params["split"]
+        )
+        previous_failed_runs = 0 if previous_response is None else previous_response["failed_runs"]
```
severo, line 798: hmmm: it cannot be None, if I understand correctly

severo (approved):
excellent! it's a lot simpler now.

Just a detail about the types

libs/libcommon/src/libcommon/orchestrator.py
AndreaFrancis: Correct, I fixed it. Thanks!

PR #2248, opened 2024-01-09
recreate the dataset on webhook to avoid mixed content

### LINKED ISSUE #1823: Avoid providing "mixed" cache while refreshing a dataset

severo:
fixes #1823. Note that we now always require the user to be trusted (ie to pass a secret with the /webhook request) because all the actions are destructive.

Depends on merging https://github.com/huggingface/datasets-server/pull/2246 first

### DIFF
--- a/e2e/tests/test_11_api.py
+++ b/e2e/tests/test_11_api.py
@@ -231,7 +231,7 @@ def test_normal_user_blocked(csv_path: str) -> None:
         # ^ should be caught by COMMON_BLOCKED_DATASETS := "__DUMMY_DATASETS_SERVER_USER__/blocked-*"
     ) as dataset:
         poll_parquet_until_ready_and_assert(
-            dataset=dataset, expected_status_code=404, expected_error_code="ResponseNotFound"
+            dataset=dataset, expected_status_code=501, expected_error_code="DatasetInBlockListError"
         )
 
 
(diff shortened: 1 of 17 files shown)


lhoestq (approved):
great !

PR #2249, opened 2024-01-09
remove CACHE_MAX_DAYS feature

severo:
it simplifies the code, and it avoids periodic waste of resources to refresh entries that are already OK. If we want to refresh old entries, we can do it easily.


lhoestq (approved):
always nice to remove code

Issue #2253, opened 2024-01-10
Cache entries seem to be deleted again and again if dataset page is accessed when the jobs are running


severo:
I observed on https://huggingface.co/datasets/wisenut-nlp-team/FiD-Merge-Dataset_temp_addlong:

- erroneous contents: https://huggingface.co/datasets/wisenut-nlp-team/FiD-Merge-Dataset_temp_addlong/discussions/2
- I recreated the dataset through the admin UI
- some cache entries were created, the config-parquet-and-info job was still pending
- I accessed the dataset page
- then, the dataset status only showed no cache entries, and a pending job for dataset-config-names

It means that accessing one of the API endpoints triggered an update of the dataset (delete everything, and recreate it from the start).

It's an awful bug, related to https://github.com/huggingface/datasets-server/pull/2244. I'm on it.

severo:
Also note that I recreated the dataset with HIGH priority, but after the bug, the new jobs have NORMAL priority

severo:
Hmmm, I'm not sure it's triggered by accessing the dataset page (at least, it was not me). Maybe in "finish_job", some configuration in the results led to that issue.

severo:
Created https://huggingface.co/datasets/severo/FiD-Merge-Dataset_temp_addlong to reproduce

severo:
I reduced the number of files to make it easier to reproduce: only one parquet file at the root, the issue still appears.

severo:
It's the same with any dataset... https://huggingface.co/datasets/severo/doc-image-4 for example which used to work:

<img width="1035" alt="Capture d’écran 2024-01-10 à 10 57 00" src="https://github.com/huggingface/datasets-server/assets/1676121/76a14ec1-329b-4e77-8e09-c5f41f9f56fa">

One of the jobs deletes everything and recreates the first step job again, with priority normal. It's a very bad issue!

PR #2254, opened 2024-01-10
Dont delete dataset cache if webhook is update on another branch

lhoestq:
close https://github.com/huggingface/datasets-server/issues/2253


severo (approved):
Thanks for the fix @lhoestq !!!!

PR #2255, opened 2024-01-10
filter webhook scope

lhoestq:
it was updating datasets completely on event scopes like "discussion.comment"

### DIFF
--- a/services/api/src/api/routes/webhook.py
+++ b/services/api/src/api/routes/webhook.py
@@ -25,14 +25,14 @@ schema = {
             "properties": {
                 "headSha": {"type": "string"},
                 "name": {"type": "string"},
-                "type": {
-                    "type": "string",
-                    "enum": ["repo", "repo.content", "repo.config", "discussion", "discussion.comment"],
-                },
+                "type": {"type": "string", "enum": ["dataset", "model", "space"]},
             },
             "required": ["type", "name"],
         },
-        "scope": {"type": "string", "enum": ["add", "remove", "update", "move"]},
+        "scope": {
+            "type": "string",
+            "enum": ["repo", "repo.content", "repo.config", "discussion", "discussion.comment"],
+        },
     },
     "required": ["event", "repo", "scope"],
 }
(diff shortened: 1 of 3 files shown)


severo (approved):
Ok, just try to fix the styke

services/api/src/api/routes/webhook.py
```diff
@@ -25,12 +25,16 @@
             "properties": {
                 "headSha": {"type": "string"},
                 "name": {"type": "string"},
-                "type": {"type": "string", "enum": ["dataset", "model", "space"]},
+                "type": {
+                    "type": "string",
+                    "enum": ["repo", "repo.content", "repo.config", "discussion", "discussion.comment"],
```
severo: it will break if another scope is added, no?

severo: maybe just let it as "string"

services/api/src/api/routes/webhook.py
```diff
@@ -51,6 +55,7 @@ class MoonWebhookV2Payload(TypedDict):
     event: Literal["add", "remove", "update", "move"]
     movedTo: Optional[str]
     repo: MoonWebhookV2PayloadRepo
+    scope: Literal["repo", "repo.content", "repo.config", "discussion", "discussion.comment"]
```
severo, line 58: and here too: only str

services/api/src/api/routes/webhook.py
severo: alternative: only list the accepted values here: if another one, such as "discussion" appears, the JSON validation will fail, and the webhook will be discarded

severo: also: is "scope" always present (even in "delete")?

lhoestq: this one was a bad copy paste

services/api/src/api/routes/webhook.py
lhoestq: ok let me do the change

PR #2269, opened 2024-01-10
hardcode the error codes to retry

severo:
There is no need to have this as an environment variable, and it clutters the code.

Note that the two lists we had (default one and backfill one) were different, by the way


severo:
I still have to fix the tests

AndreaFrancis (approved):

PR #2281, opened 2024-01-11
increase difficulty based on failed runs

AndreaFrancis:
Some jobs fail many times with the same difficulty, but others run if they are processed by heavier workers.


severo (approved):
Sounds good.

PR #2286, opened 2024-01-12
remove check on disks

severo:
we needed this when we used NFS in production, but now that we use EFS and S3, no need for it.

Based on #2287.


AndreaFrancis (approved):
Thanks!

PR #2290, opened 2024-01-12
add retry for download hub file

AndreaFrancis:
Related to https://github.com/huggingface/datasets-server/issues/1443
For `split-duckdb-index` and `split-descriptive-statistics`, I found 11 records in the cache with UnexpectedErrors due to ReadTimeout error when downloading files from hub.
Screenshot from 2024-01-12 14-44-28
Screenshot from 2024-01-12 14-44-46

 When calling a force refresh, they worked, so instead of adding this error to the error_codes_to_retry list, it is better to retry downloads when job execution is in progress.


severo (approved):
OK, good idea

Issue #2297, opened 2024-01-15
/rows generates empty images


severo:
See https://huggingface.co/datasets/HuggingFaceM4/WebSight/viewer/default/train?p=1

<img width="1386" alt="Capture d’écran 2024-01-15 à 16 12 40" src="https://github.com/huggingface/datasets-server/assets/1676121/a20f3628-5ddd-400c-a067-17d47a6caedb">

Note that /first-rows images are OK.

<img width="1025" alt="Capture d’écran 2024-01-15 à 16 13 14" src="https://github.com/huggingface/datasets-server/assets/1676121/c34c00d5-e195-43ae-990e-dad431f6a6d0">

Reported by @VictorSanh on Slack (internal): https://huggingface.slack.com/archives/C02V51Q3800/p1705330620412079

severo:
No error is raised in the code:

![Uploading Capture d’écran 2024-01-15 à 17.31.52.png…]()

lhoestq:
Hmm for some reason the image files are 0 bytes, let me investigate

PR #2299, opened 2024-01-16
Don't create empty images in /rows

lhoestq:
Don't create empty images. Otherwise on a second call to /rows it will return the first empty image.
We were already doing something similar for audio.

Fix https://github.com/huggingface/datasets-server/issues/2297

TODO after merged:
- [ ] delete the cached assets for HuggingFaceM4/WebSight to remove the empty images


severo (approved):
OK, thanks for looking at the issue.

Feel free to merge once the conflicts are resolved

AndreaFrancis:
>  delete the cached assets for HuggingFaceM4/WebSight to remove the empty images

I removed cached assets, and it looks to be working fine.
image

lhoestq:
Cool thanks !

albertvillanova:
Could it be possible that other datasets were also impacted besides "HuggingFaceM4/WebSight"?

Below, "cc_news" and even for /first-rows:
Screenshot from 2024-01-17 08-33-43

If so, we would have to delete their cached assets for all of them...

severo:
maybe @AndreaFrancis knows of a way to look for all the /assets on s3 that have a size of 0 bytes?

AndreaFrancis:
> maybe @AndreaFrancis knows of a way to look for all the /assets on s3 that have a size of 0 bytes?

Let me check. I think it is possible to use aws cli I will try.

Issue #2053, opened 2023-11-03
Delete duckdb files when a split is deleted


severo:
See https://huggingface.co/datasets/HuggingFaceH4/ultrachat_200k/tree/refs%2Fconvert%2Fparquet/default/train

The `train` split existed, but has been deleted. Now, the datasets has `train_sft` and `train_gen`. But the duckdb index file still exists for `train` in `refs/convert/parquet`. Note that the parquet files have been deleted.

severo:
cc @albertvillanova or @AndreaFrancis

AndreaFrancis:
Fixed 
![image](https://github.com/huggingface/datasets-server/assets/5564745/744e470d-f24a-4409-8b67-2eab99c91894)

AndreaFrancis:
I will try to find a way to identify if other datasets have been affected, surely, yes, but it will automatically be fixed whenever the split-duckdb-index step is executed again for a given config.

severo:
> I will try to find a way to identify if other datasets have been affected, surely, yes, but it will automatically be fixed whenever the split-duckdb-index step is executed again for a given config.

OK, so: don't worry, we will have to compute them all when upgrading duckdb

PR #2308, opened 2024-01-18
Delete obsolete split duckdb files

AndreaFrancis:
Fix for https://github.com/huggingface/datasets-server/issues/2053
Delete obsolete files for splits that no longer exist including `index.duckdb`

### DIFF
--- a/services/worker/src/worker/job_runners/split/duckdb_index.py
+++ b/services/worker/src/worker/job_runners/split/duckdb_index.py
@@ -4,6 +4,7 @@
 import copy
 import logging
 import os
+import re
 from pathlib import Path
 from typing import Optional
 
@@ -44,6 +45,7 @@ from worker.utils import (
     HF_HUB_HTTP_ERROR_RETRY_SLEEPS,
     LOCK_GIT_BRANCH_RETRY_SLEEPS,
     create_branch,
+    get_split_names,
     hf_hub_url,
     retry,
 )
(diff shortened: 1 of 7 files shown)


services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -80,6 +82,29 @@ def check_indexable(feature: FeatureType) -> None:
...
+    # delete old files from non existent splits
+    files_to_ignore: set[str] = {
+        file
+        for file in only_config_files
+        # ignore existing files in all current splits
+        if pattern_only_config_splits_files.match(file)
+        # but not the old index in split
+        and file != index_file_location
+    }
+    files_to_delete = only_config_files - files_to_ignore
```
severo, line 104: to simplify the logic and avoid double negations, maybe set files_to_delete with the loop?

Something like
```suggestion
    files_to_delete: set[str] = {
        file
        for file in only_config_files
        # delete the obsolete index in the current split
        if file == index_file_location
        # and files from unknown splits
        or not pattern_only_config_splits_files.match(file)
    }
```

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -80,6 +82,28 @@ def check_indexable(feature: FeatureType) -> None:
...
+    pattern_only_config_splits_files = re.compile(
+        f"^({'|'.join(re.escape(f'{config}/{split_name}') for split_name in split_names)})/"
+    )
+
+    # delete old files from non existent splits
+    files_to_delete: set[str] = {
+        file
+        for file in only_config_files
+        # delete the obsolete index in the current split
+        if file == index_file_location
```
severo, line 100: btw: we don't need to remove this, since we will always create CommitOperationAdd with the new value. And huggingface_hub will thus ignore this CommitOperationDelete

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -82,26 +82,19 @@ def check_indexable(feature: FeatureType) -> None:
...
+        CommitOperationDelete(path_in_repo=file)
         for file in only_config_files
-        # delete the obsolete index in the current split
-        if file == index_file_location
-        # and files from unknown splits
-        or not pattern_only_config_splits_files.match(file)
-    }
-    return [CommitOperationDelete(path_in_repo=file) for file in files_to_delete]
+        if not pattern_only_config_splits_files.match(file)
+    ]
```
severo, line 97: nit: we could even do

```
    same_config = re.compile(f"^({re.escape(config)})/")
    existing_split = re.compile(
        f"^({'|'.join(re.escape(f'{config}/{split_name}') for split_name in split_names)})/"
    )
    return [
        CommitOperationDelete(path_in_repo=file)
        for file in all_repo_files
        if same_config.match(file)
        and not existing_split.match(file)
    ]
```

severo (approved):
see the comment... I think refactoring a bit more would help understand what the code does

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: thanks! I applied your suggestion

PR #2314, opened 2024-01-19
Refactor storage client

severo:


libs/libcommon/src/libcommon/storage_client.py
```diff
@@ -18,10 +18,11 @@ class StorageClient:
     A resource that represents a connection to a storage client.
 
     Args:
-        protocol (:obj:`str`): The fsspec protocol (supported s3 or file)
+        protocol (:obj:`str`): The fsspec protocol (supported: "file" or "s3")
```
AndreaFrancis, line 21: Maybe we can make the protocol a Literal or enum with fixed allowed values.

AndreaFrancis (approved):

libs/libcommon/src/libcommon/storage_client.py
severo: it's what I tried first, but then I realized that we directly put the environment variable there, and it's a string.

We could, but it would mean doing better validation when loading the environment variables (https://pypi.org/project/environs/ support env.enum, maybe try this).

Maybe in a dedicated issue where we improve the env vars validation?

PR #2315, opened 2024-01-19
Sign assets urls in first rows

severo:
When the API serves /first-rows, and if the CloudFront config has been passed through the environment variables, we now traverse the response content and sign every URL we find.

The response content contains unsigned URLs and comes from the database. We have to sign the URLs on the fly. Indeed, as they only last 1 hour, we cannot store them in the database.

<strike>
Note that:
> sign every URL we find
currently means:
> any string that starts with `https://datasets-server.huggingface.co/assets/`

This brings some false positives like:
1. an already signed URL
2. a string that is not a URL but starts with `https://datasets-server.huggingface.co/assets/`

Case 1. should not exist, but nothing prevents people from including case 2. in their data. As unsigned assets are easily forgeable, the only way would be to ensure it's from a feature like Image or Audio. But it means parsing more conscientiously. Do you think we should do it? (To me: it's good enough, and we can open an issue to work on it later if needed).
</strike>

-> we now only sign image and audio src we generate ourselves.

Follow-ups:
- setup the CloudFront config for the staging environment (in `/chart` and in infra's terraform configuration)
- setup the CloudFront config for the prod environment (in `/chart` and in infra's terraform configuration)


libs/libcommon/src/libcommon/storage_client.py
```diff
@@ -92,11 +86,26 @@ def get_unsigned_url(self, path: str) -> str:
...
 
     def sign_url_if_available(self, url: str) -> str:
-        if self.cloudfront:
-            url = self.cloudfront.sign_url(url=url)
+        if self.url_signer:
+            url = self.url_signer.sign_url(url=url)
             logging.debug(f"signed url: {url}")
         return url
 
+    def sign_urls_in_obj(self, obj: Any) -> Any:
```
lhoestq, line 94: maybe you can just check for dicts with a "src" field to have less false positives ?

lhoestq (commented):
Nice ! I think it requires some optimisations or it will be too slow in certain cases

lhoestq:
Another way to make this fast it to only consider the columns which have image and audio types.

```python
from datasets import Audio, Features, Image
from datasets.features.features import FeatureType, _visit

FEATURES_WITH_ASSETS = (Audio, Image)


def get_columns_with_assets(
    features: Features,
    features_with_assets: tuple = FEATURES_WITH_ASSETS,
) -> list[str]:
    columns_with_assets = []

    for column, feature in features.items():
        str_column = str(column)
        with_assets = False

        def classify(feature: FeatureType) -> None:
            nonlocal with_assets
            if isinstance(feature, features_with_assets):
                with_assets = True

        _visit(feature, classify)
        if with_assets:
            columns_with_assets.append(str_column)
    return columns_with_assets

```

libs/libcommon/src/libcommon/storage_client.py
AndreaFrancis, line 94: I think It is better to handle all this logic in the signer instead to avoid giving extra responsibilities to a Storage Client class.

libs/libcommon/tests/test_storage_client.py
```diff
@@ -0,0 +1,58 @@
...
+        (ARRAY_WITH_URL, 2),
+        (NESTED_OBJECT_WITH_URL, 3),
+        # below are limitations, we would ideally expect 0. Hopefully will not happen
+        (STRING_WITH_URL_AT_THE_START, 1),
+        (ALREADY_SIGNED_URL, 1),
+    ],
+)
+def test_sign_urls_in_obj(content: Any, num_replaced_urls: int) -> None:
+    url_signer = FakeUrlSigner()
+    storage_client = StorageClient(protocol="file", storage_root="/tmp", base_url=BASE_URL, url_signer=url_signer)
```
AndreaFrancis, line 56: > I think It is better to handle all this logic in the signer instead to avoid giving extra responsibilities to a Storage Client class.

also note that none of `protocol="file", storage_root="/tmp", base_url=BASE_URL` attributes are helpful to perform the  sign_urls_in_obj method so, it would have been invoked directly from the signer.

AndreaFrancis (approved):
LGTM

libs/libcommon/src/libcommon/storage_client.py
severo: yes, it makes sense, I'll change this.

libs/libcommon/tests/test_storage_client.py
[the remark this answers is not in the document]
severo: it's not used (as mentioned by Andrea). I'll move this to UrlSigner.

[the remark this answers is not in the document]
severo: Done

libs/libcommon/src/libcommon/storage_client.py
severo: done

severo:
> Another way to make this fast it to only consider the columns which have image and audio types.

Yes, but the issue is that we don't have the Features object as such, only a serialized version.

I was starting to find the Image and Audio columns but then reminded that we can have any level of nesting of these types (lists of images, etc.).

The solution might be to visit the whole `features` dict, store the JSON paths to assets URLs, and then sign all of them in the `rows` field. I'll try that.

---

Also, the first rows cells can be truncated, so: it's another point to take into account (ignore the truncated cells)

lhoestq:
The serialized version can be parsed using Features.from_dict btw

libs/libcommon/src/libcommon/storage_client.py
severo: done

libs/libcommon/src/libcommon/storage_client.py
```diff
@@ -92,11 +86,26 @@ def get_unsigned_url(self, path: str) -> str:
...
+        if not self.url_signer:
+            logging.debug("url_signer is not configured, object is left unchanged")
+            return obj
+        if isinstance(obj, dict):
+            return {k: self.sign_urls_in_obj(v) for k, v in obj.items()}
+        elif isinstance(obj, list):
+            return [self.sign_urls_in_obj(v) for v in obj]
```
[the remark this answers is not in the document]
severo: followed https://github.com/huggingface/datasets-server/pull/2315#issuecomment-1900633561

severo:
It should be good for review!

severo:
I forgot to handle the truncated rows... changing to draft again

severo:
Oh, in fact the Audio and Image cells are never truncated:

https://github.com/huggingface/datasets-server/blob/fbb0b22074444a24c1e3ae0eea3821baf0871085/libs/libcommon/src/libcommon/viewer_utils/rows.py#L222

But we still have to handle the possible case of nested objects that contain URLs to sign

severo:
> I forgot to handle the truncated rows... changing to draft again

Done!

libs/libcommon/src/libcommon/url_signer.py
```diff
@@ -0,0 +1,131 @@
...
+    return asset_url_paths
+
+
+class URLSigner(ABC):
+    @abstractmethod
+    def sign_url(self, url: str) -> str:
+        pass
+
+    def _sign_asset_url_path_in_place(self, cell: Any, asset_url_path: AssetUrlPath) -> Any:
+        if len(asset_url_path.path) == 0:
```
lhoestq, line 79: cells can be None

```suggestion
        if not cell:
            return cell
        elif len(asset_url_path.path) == 0:
```

libs/libcommon/src/libcommon/url_signer.py
```diff
@@ -0,0 +1,131 @@
...
+    feature: FeatureType, func: Callable[[FeatureType, VisitPath], Optional[FeatureType]], visit_path: VisitPath = []
+) -> FeatureType:
+    """Visit a (possibly nested) feature.
+
+    Args:
+        feature (FeatureType): the feature type to be checked
+    Returns:
+        visited feature (FeatureType)
+    """
+    if isinstance(feature, dict):
```
lhoestq, line 48: There are two possibilities in Sequence:

```
Sequence(Image())  # e.g. [image0, image1, ...]
```

or 

```
Sequence({"images": Image()})  # e.g. {"images": [image0, image1, ...]}
```

you can take this into account by adding these two lines:

```suggestion
    if isinstance(feature, Sequence) and isinstance(feature.feature, dict):
            feature = {k: [f] for k, f in feature.items()}
    if isinstance(feature, dict):
```

the original `_visit` in `datasets` doesn't need this since it doesn't keep track of the path

lhoestq (approved):
Nice job on the nested URL signer ! It's going to be pretty fast :)

libs/libcommon/src/libcommon/url_signer.py
```diff
@@ -0,0 +1,131 @@
...
+    path: VisitPath
+
+    def enter(self) -> "AssetUrlPath":
+        if len(self.path) == 0:
+            raise ValueError("Cannot enter an empty path")
+        return AssetUrlPath(feature_type=self.feature_type, path=self.path[1:])
+
+
+# invert to_features_list
+def to_features_dict(features: list[FeatureItem]) -> Features:
```
AndreaFrancis, line 34: Maybe this function should go to a more generic utils.py file?

libs/libcommon/src/libcommon/utils.py
```diff
@@ -49,16 +49,33 @@ def utf8_lead_byte(b: int) -> bool:
...
+    pass
+
+
+def serialize_and_truncate(obj: Any, max_bytes: int) -> str:
+    """
+    Serialize an object as JSON and truncate it if it's bigger than max_bytes
+
+    Args:
+        obj (`Any`): the Python object (can be a primitive type, a list, a dict, etc.)
+        max_bytes (int): the maximum number of bytes
```
AndreaFrancis, line 62: Not sure if this PR will be approved but then I would add '' to each data type https://github.com/huggingface/datasets-server/pull/2351

AndreaFrancis (approved):

libs/libcommon/src/libcommon/url_signer.py
severo: Done

libs/libcommon/src/libcommon/utils.py
severo: I think https://github.com/huggingface/datasets-server/pull/2351 will be merged first, I will adapt

severo: Done

libs/libcommon/src/libcommon/url_signer.py
severo: I sent https://github.com/huggingface/datasets-server/pull/2315/commits/2ca0df11e5fd89333ab49d38ce897aff7c175b59, but I think I didn't get it right, because the tests pass with or without these two lines at the start of `_visit`.

Could you have a look @lhoestq?

lhoestq: added a comment in your commit

severo: thanks! The issue was: that I hadn't enabled my test

AndreaFrancis (approved):
Great work!

lhoestq (approved):
🔥

severo:
thanks for the reviews! I rebased, nothing changed. Waiting for the CI and I merge

Issue #1914, opened 2023-10-03
upgrade duckdb to >=0.9 and pandas to >=2.1.1


severo:
See previous intent here: https://github.com/huggingface/datasets-server/pull/1827

If we apply it, we must recreate the duckdb files created with duckdb <= 0.8

AndreaFrancis:
I suggest the following plan to handle duckdb upgrades:
1. update poetry dependencies
2. increase job_runner version for `split-duckdb-index` so that, all entries are computed again with new version 
3. migration to add `duckdb_version` field in all cache entries with kind `split-duckdb-index` with a default value of 0.8 (previous version) 
4. change /search to validate that `split-duckdb-index` cache entry duckdb_version is equal to a new DUCKDB_VERSION_CURRENT constant; if not, raise an error
5. we can also invalidate `dataset-is-valid`, `config-is-valid`, and `split-is-valid` cache entries (set search:false and filter:false) so that we temporarily hide the functionality in the UI and avoid having errors when calling /search and /filter endpoints; meanwhile all cache entries are being refreshed

severo:
Good plan. Maybe we can simplify it a bit since the duckdb team swore 0.8 -> 0.9 was the last breaking change in the duckdb file format

1. update poetry dependencies and increase job_runner version for split-duckdb-index (even if it's not really needed with step 2)
2. in the migration job:
- (before launching the migration) stop the workers
- delete all the split-duckdb-index entries
- set search:false and filter:false in dataset-is-valid, config-is-valid, and split-is-valid cache entries so that we temporarily hide the functionality in the UI and avoid having errors when calling /search and /filter endpoints; meanwhile all cache entries are being refreshed

PR #2316, opened 2024-01-19
Update pandas and duckdb version

AndreaFrancis:
Fix for https://github.com/huggingface/datasets-server/issues/1914

Duckdb migration plan:

- [x] Update poetry dependencies
- [x] Increase job_runner version for split-duckdb-index
- [x] Migration to delete all the split-duckdb-index entries
- [x] Migration to set search:false and filter:false in dataset-is-valid, config-is-valid, and split-is-valid cache entries so that we temporarily hide the functionality in the UI and avoid having errors when calling /search and /filter endpoints

Deploy plan: stop all workers and run migration, then wait for backfill to compute again all split-duckdb-index entries (It could require an increase of the resources/workers).


AndreaFrancis:
It looks like the last pre-release of duckdb will break FTS for nested data; I opened https://github.com/duckdb/duckdb/issues/10281 and will proceed with the migration once it is fixed.

AndreaFrancis:
I will keep this PR on hold until there is an available Linux package for duckdb fix 0.9.3dev3058 https://pypi.org/project/duckdb/0.9.3.dev3058/#files

services/search/src/search/duckdb_connection.py
```diff
@@ -5,6 +5,5 @@
 
 def duckdb_connect(**kwargs: Any) -> duckdb.DuckDBPyConnection:
     con = duckdb.connect(read_only=True, **kwargs)
-    con.sql("SET enable_external_access=false;")
```
AndreaFrancis, line 8: With the new version it throws the following error:
```
    # duckdb.duckdb.Error: Extension Autoloading Error:
    # An error occurred while trying to automatically install the required extension

```

AndreaFrancis: Doc says https://duckdb.org/docs/sql/configuration.html:
`Allow the database to access external state (through, e.g., loading/installing modules, COPY TO/FROM, CSV readers, pandas replacement scans, etc)`

It looks like duckdb now tries to dynamically load FTS extension from the db, so probably it makes sense to remove it.

front/admin_ui/pyproject.toml
severo, line 1: also: change in https://github.com/huggingface/datasets-server/blob/66c1e089e204ab33195b957e1b99b0da6a4dd2de/front/admin_ui/requirements.txt#L6

severo (approved):
just a detail in admin_ui

services/search/src/search/duckdb_connection.py
AndreaFrancis: I found this [comment](https://github.com/huggingface/datasets-server/pull/1418#issuecomment-1742788076) cc. @albertvillanova @lhoestq

front/admin_ui/pyproject.toml
AndreaFrancis: Done, thanks!

services/search/src/search/duckdb_connection.py
lhoestq: As said in my comment, using `enable_external_access=false` should block HTTP requests (e.g. loading external data or extension) and avoid globbing local files to scan the local filesystem.

Is it possible to first load the FTS extension and then SET `enable_external_access=false` ?

lhoestq:
As you known updating duckdb will break filter and search for existing indexes. Indeed all our duckdb files are in the 0.8 format and 0.9 can NOT read them.

I would wait for duckdb to be backwards compatible (0.10 release coming mid-february) before migrating.

Also that means we might need to use two duckdb versions in the API pods to be able to query both old and new data while the jobs are being processed to update the duckdb files ?

This will also be a good opportunity to move the duckdb files to somewhere else.

lhoestq:
I just saw the discussion at https://github.com/huggingface/datasets-server/issues/1914 sorry

So the idea is to break search and filter while the jobs are processed ? It seems quite drastic since the migration will take time (days, maybe weeks ?)

severo:
> So the idea is to break search and filter while the jobs are processed ? It seems quite drastic since the migration will take time (days, maybe weeks ?)

indeed, it's surely too drastic. Let's propose another strategy?

severo:
> This will also be a good opportunity to move the duckdb files to somewhere else.

Yes.

severo:
Related discussion here: https://github.com/huggingface/datasets-server/issues/1694#issuecomment-1737010286 (store the duckdb files elsewhere)

lhoestq:
Having API workers that could read both duckdb 0.8 and 0.10 sounds like a bit of a pain actually.

Maybe it will be easier to add a new job to compute the new indexes, then after some time switch the API to use them once all the indexes are ready and then the old indexes ?

AndreaFrancis:
Closing this PR since it will be better handled in https://github.com/huggingface/datasets-server/issues/2413

severo:
0.10.0 has been released, we can work on https://github.com/huggingface/datasets-server/issues/2413

PR #2318, opened 2024-01-22
Post messages script - remove "sleep"

### LINKED ISSUE #2251: Finish the post-messages script earlier
The post-messages have been running for more than 15 hours now even if it does not take more than 2 minutes to complete.

https://cd.internal.huggingface.tech/applications/argo-cd/datasets-server-prod?view=pods&node=%2FPod%2Fdatasets-server%2Fprod-datasets-server-job-post-messages-28412650-w4x4c%2F0

Capture d’écran 2024-01-09 à 15 42 33

severo:
try to fix #2251

Also:
- remove the 1 second sleep after creating a discussion (not needed anymore). Maybe it was the source of the never finishing job
- add logs
- refactoring
- don't get all the discussions, only test if the bot has already created one (see https://github.com/huggingface/moon-landing/pull/8742 -> it's why we skip the tests for now)

### DIFF
--- a/jobs/cache_maintenance/src/cache_maintenance/discussions.py
+++ b/jobs/cache_maintenance/src/cache_maintenance/discussions.py
@@ -3,7 +3,6 @@
 
 import logging
 from dataclasses import dataclass
-from time import sleep
 from typing import Literal, Optional
 from urllib import parse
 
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
does this mean we should use another user name for the CI ?

severo:
> does this mean we should use another user name for the CI ?

yes, once https://github.com/huggingface/moon-landing/pull/8742 is deployed, I'll do a PR to change the user/org names in the tests, and re-enable the tests here.

severo:
deployed

PR #2336, opened 2024-01-23
Add extract_split_name_from_parquet_url function to parquet utils

egndz:
Fixes #1755 

This PR adds `extract_split_name_from_url` function to parquet utils and function is called when necessary.

Issue #1085, opened 2023-04-25
CI spawning check error: The provided API token is invalid.


albertvillanova:
See: https://github.com/huggingface/datasets-server/actions/runs/4796071992/jobs/8531405657?pr=1084#logs
```
______________________ test_real_check_spawning_response _______________________

app_config = AppConfig(assets=AssetsConfig(base_url='http://localhost/assets', storage_directory=None), cache=CacheConfig(mongo_dat...per_batch=1000, spawning_token='dummy_spawning_token', max_concurrent_requests_number=100, max_requests_per_second=50))

    @pytest.mark.asyncio
    async def test_real_check_spawning_response(app_config: AppConfig) -> None:
        semaphore = Semaphore(value=10)
        limiter = AsyncLimiter(10, time_period=1)
    
        headers = {"Authorization": f"API {CI_SPAWNING_TOKEN}"}
        async with ClientSession(headers=headers) as session:
            image_url = "http://testurl.test/test_image.jpg"
            image_urls = [image_url]
            spawning_url = app_config.urls_scan.spawning_url
            spawning_response = await check_spawning(image_urls, session, semaphore, limiter, spawning_url)
            assert spawning_response and isinstance(spawning_response, dict)
>           assert spawning_response["urls"] and isinstance(spawning_response["urls"], list)
E           KeyError: 'urls'

tests/job_runners/split/test_opt_in_out_urls_scan_from_streaming.py:309: KeyError
```
where `spawning_response` is:
```
{'detail': 'The provided API token is invalid.'}
```

AndreaFrancis:
I will try to run again, it shouldn't have failed since the token is configured in the secrets and is a valid one.

albertvillanova:
I think the root cause is because the PR is made from a fork.

Current implementation does not allow contributions from open source community members, who necessarily use forks.

albertvillanova:
Related to:
- #712

severo:
Related: we cannot run the test locally, and `cd services/worker; make test;` fails for that reason

severo:
keep open. cc @AndreaFrancis

severo:
keep open

severo:
keep open. cc @AndreaFrancis

PR #2337, opened 2024-01-24
Remove METRICS_MONGOENGINE_ALIAS and MetricsMongoResource

### LINKED ISSUE #2210: Deprecate/remove MetricsMongoResource and METRICS_MONGOENGINE_ALIAS from code
It's only used in the migrations, and nobody will ever run them again from an old version.

egndz:
Fixes #2210 

- This PR removes METRICS_MONGOENGINE_ALIAS and MetricsMongoResource along with its references

### DIFF
--- a/libs/libapi/src/libapi/duckdb.py
+++ b/libs/libapi/src/libapi/duckdb.py
@@ -12,6 +12,7 @@ import anyio
 from anyio import Path
 from huggingface_hub import hf_hub_download
 from libcommon.constants import DUCKDB_INDEX_DOWNLOADS_SUBDIRECTORY, SPLIT_DUCKDB_INDEX_KINDS
+from libcommon.parquet_utils import extract_split_name_from_parquet_url
 from libcommon.prometheus import StepProfiler
 from libcommon.simple_cache import CacheEntry
 from libcommon.storage import StrPath, init_dir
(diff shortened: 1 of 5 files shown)


severo:
@AndreaFrancis: I think it's expected that the e2e tests fail, see https://github.com/huggingface/datasets-server/pull/2336#issuecomment-1907797371.

See #1085

AndreaFrancis:
Let us remove `METRICS_MONGOENGINE_ALIAS` and `MetricsMongoResource`, but instead of deleting all the objects that use them, change to the new `ALIAS`, that is, metrics live now in different collections and databases: 
For cache metrics: https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/simple_cache.py#L179 instead of using `METRICS_MONGOENGINE_ALIAS`, we should use `CACHE_MONGOENGINE_ALIAS`
For queue metrics: https://github.com/huggingface/datasets-server/blob/main/libs/libcommon/src/libcommon/queue.py#L251 instead of using `METRICS_MONGOENGINE_ALIAS`, we should use `QUEUE_MONGOENGINE_ALIAS`.
Also, even if the code runs correctly in the test, it can break production migration history if we remove the migration order. So, instead of removing them, we can change and point to the new Alias according to the table.
This may include adding a new ALIAS here: https://github.com/huggingface/datasets-server/blob/main/jobs/mongodb_migration/src/mongodb_migration/migration.py#L75 (one by each collection) or creating different classes per collection.

Issue #2332, opened 2024-01-23
[Croissant] The API supports a parameter to produce either an abridged JSON-LD or a full-length JSON-LD


marcenacp:
**Issue**: Croissant JSON-LD are embedded in Hugging Face's HTML page. As JSON-LD can get lengthy, they are truncated ([see the code](https://github.com/huggingface/datasets-server/blob/1759d65fa668f35067175cfb3210666c1b7c3f47/services/api/src/api/routes/croissant.py#L99)).

**Proposed solution**: Add a parameter `?full=true/false` (with a default to false) to toggle between the abridged version or the full-length version of the JSON-LD. Hugging Face would use the abridged version to embed in their HTML page, while still allowing users to retrieve the full version to produce a valid Croissant file. Ideally the description of the abridged version would state that it's (or may be) abridged, and provide the URL of the full version. This is a temporary solution while waiting for <meta> tags to support URLs rather than in-lined JSON.

@lhoestq @severo What do you think? Also, what convention do you usually use for the URL parameters? Do you have examples of such routes? If I have your go, I could implement this. Thanks for your input!

severo:
Good idea. Feel free to open a PR!

> what convention do you usually use for the URL parameters?

We have no other boolean parameter, I think. @lhoestq should we pass:
- `?full=true/false`
- or `?max_columns=1000`, and if not present: return the full response?

If we opt for the integer parameter, you could take example on https://github.com/huggingface/datasets-server/blob/1759d65fa668f35067175cfb3210666c1b7c3f47/services/rows/src/rows/routes/rows.py#L72

else, we should do the same, but for a boolean value. If so, I think checking for exact value `true` is OK, and coherent with what we do on the Hub (https://huggingface.co/spaces/enzostvs/hub-api-playground)

---

> Ideally the description of the abridged version would state that it's (or may be) abridged, and provide the URL of the full version

OK, good idea. As we return a croissant content that we "directly" embed in the HTML, it would be better if it's part of the croissant specification, or at least not incompatible with it

PR #2340, opened 2024-01-25
Support a `full` parameter to produce either an abridged JSON-LD or a full-length JSON-LD.

marcenacp:
Fixes: https://github.com/huggingface/datasets-server/issues/2332

### DIFF
--- a/services/api/src/api/routes/croissant.py
+++ b/services/api/src/api/routes/croissant.py
@@ -64,7 +64,9 @@ def _escape_name(name: str, names: set[str]) -> str:
     return escaped_name
 
 
-def get_croissant_from_dataset_infos(dataset: str, infos: list[Mapping[str, Any]], partial: bool) -> Mapping[str, Any]:
+def get_croissant_from_dataset_infos(
+    dataset: str, infos: list[Mapping[str, Any]], partial: bool, full_jsonld: bool
+) -> Mapping[str, Any]:
     repo_name = "repo"
     names: set[str] = set(repo_name)
     distribution = [
@@ -96,7 +98,7 @@ def get_croissant_from_dataset_infos(dataset: str, infos: list[Mapping[str, Any]
         )
         skipped_columns = []
         for column, feature in features.items():
-            if len(fields) >= MAX_COLUMNS:
+            if len(fields) >= MAX_COLUMNS and not full_jsonld:
                 description_body += f"\n- {len(features) - MAX_COLUMNS} skipped column{'s' if len(features) - MAX_COLUMNS > 1 else ''} (max number of columns reached)"
                 break
             fields_names: set[str] = set()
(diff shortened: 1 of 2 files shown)


services/api/src/api/routes/croissant.py
```diff
@@ -199,6 +201,16 @@ def get_croissant_from_dataset_infos(dataset: str, infos: list[Mapping[str, Any]
...
 
 
+def _get_full_jsonld_parameter(request: Request) -> bool:
+    """Parameter to retrieve the full JSON-LD (full=True) or a truncated/abridged JSON-LD (full=False) with less features."""
+    full_jsonld = get_request_parameter(request, "full", default="false")
+    if full_jsonld.lower() == "true":
+        return True
+    elif full_jsonld.lower() == "false":
+        return False
+    raise MissingRequiredParameterError("Parameter 'full' must have value 'full=true' or 'full=false'")
```
severo, line 211: I think it's ok to just return False in any other case

```suggestion
    return full_jsonld.lower() == "true"
```

services/api/tests/routes/test_croissant.py
```diff
@@ -65,12 +69,23 @@ def test_get_croissant_from_dataset_infos() -> None:
     assert isinstance(croissant["recordSet"][0]["field"], list)
     assert isinstance(squad_info["features"], dict)
     assert len(croissant["recordSet"][0]["field"]) == len(squad_info["features"]) - 1
+    assert "1 skipped column: answers" in croissant["recordSet"][0]["description"]
```
severo, line 72: I don't get what is this `: answers"`. The string should contain

```
\n- 1 skipped column (max number of columns reached)
```

no?

severo:
The e2e tests cannot run on external branches, so no worry about it (see #1085)

services/api/tests/routes/test_croissant.py
marcenacp: The dataset contains 5 features. But I noticed there's a skipped column for this dataset (hence the number `4` a bit below in the file, rather than the expected `5`). So I added a test here.

The message is inserted here: https://github.com/huggingface/datasets-server/blob/4d4b2f1e1163257664f707564bcfef337b4f947e/services/api/src/api/routes/croissant.py#L147

marcenacp: `answers` corresponds to this feature: https://github.com/huggingface/datasets-server/blob/4d4b2f1e1163257664f707564bcfef337b4f947e/services/api/tests/routes/test_croissant.py#L15

services/api/src/api/routes/croissant.py
marcenacp: Done.

marcenacp:
@severo Thanks for the comments! Just took them into account.

services/api/tests/routes/test_croissant.py
severo: Oh OK, thanks! Sorry about that

severo (approved):

severo:
Thanks for the PR. Please tell us if it works as expected on datasets with a lot of columns, like https://huggingface.co/datasets/mstz/arcene.

severo:
https://datasets-server.huggingface.co/croissant?dataset=mstz/arcene -> 235 KB

https://datasets-server.huggingface.co/croissant?dataset=mstz/arcene?full=true -> `"Unexpected error."`

PR #2341, opened 2024-01-25
doc - Update minor details in the doc

AndreaFrancis:
Update minor details in the doc:
- partial field for some responses
- filter example
- rename `duorc` by `ibm/duorc` instead (even though the hub redirects to the correct organization, datasets-server API does not)

Note.- I haven't update yet `Server infrastructure` page, I will do it in another PR


AndreaFrancis:
By the way, I noticed that transferred datasets are getting `500 Unexpected errors` instead of `Not found`.
For instance:
- old dataset https://datasets-server.huggingface.co/splits?dataset=blimp -> 500 ERROR (It should be 404 since it not longer exists)
- new dataset https://datasets-server.huggingface.co/splits?dataset=nyu-mll/blimp

severo (approved):
good fixes, thanks

Issue #2344, opened 2024-01-29
The parquet files are deleted when the duckdb file is created


severo:
See https://huggingface.co/datasets/jslin09/wikisource_tw/commit/81a728d6be9f0334dd63fcb26750ff3b958057c3

<img width="1563" alt="Capture d’écran 2024-01-29 à 10 26 48" src="https://github.com/huggingface/datasets-server/assets/1676121/e607e060-8ed5-4955-90f9-19a52d148fe4">

Possibly related to https://github.com/huggingface/datasets-server/pull/2308.

severo:
Once fixed, we will have to look at the affected datasets, and recreate them

AndreaFrancis:
Fixed by https://github.com/huggingface/datasets-server/pull/2346

![image](https://github.com/huggingface/datasets-server/assets/5564745/1ae6f20a-1aa3-4b96-b044-3da984aa810b)
![image](https://github.com/huggingface/datasets-server/assets/5564745/1856b68d-1a4b-499c-a090-2154dddfda73)

I don't find an easy way to identify the datasets that have been affected because it was at hub level (commits to delete files) but I will force-refresh config-parquet-and-info step for the following list: 
[force-refresh-result.csv](https://github.com/huggingface/datasets-server/files/14085835/force-refresh-result.csv) (318 configs)

I used this query to get them `{kind:"split-duckdb-index", http_status:200, "content.partial": true, updated_at:{$gte : new ISODate("2024-01-18T00:00:00Z")}}` having in consideration that https://github.com/huggingface/datasets-server/pull/2308 has been deployed nearly 2024-01-19.

PR #2346, opened 2024-01-29
fix add pattern for partial splits when removing old duckb index files

AndreaFrancis:
fix for https://github.com/huggingface/datasets-server/issues/2344

### DIFF
--- a/services/worker/src/worker/job_runners/split/duckdb_index.py
+++ b/services/worker/src/worker/job_runners/split/duckdb_index.py
@@ -98,6 +98,7 @@ def get_delete_operations(all_repo_files: set[str], split_names: set[str], confi
         CommitOperationDelete(path_in_repo=file)
         for file in all_repo_files
         if same_config_pattern.match(file)
+        and file.endswith(".duckdb")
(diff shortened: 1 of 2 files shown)


services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -87,10 +87,20 @@ def get_delete_operations(all_repo_files: set[str], split_names: set[str], confi
...
+    existing_partial_split_pattern = re.compile(
+        f"^({'|'.join(re.escape(f'{config}/partial-{split_name}') for split_name in split_names)})/"
+    )
+
     return [
         CommitOperationDelete(path_in_repo=file)
         for file in all_repo_files
-        if same_config_pattern.match(file) and not existing_split_pattern.match(file)
+        if same_config_pattern.match(file)
+        and file.endswith(".duckdb")
```
severo, line 101: indeed: it seems safe that the duckdb job only deletes duckdb files, and not parquet files

severo (approved):

lhoestq (approved):
lgtm !

PR #2347, opened 2024-01-29
Fix filter no features in duckdb index cache

lhoestq:
fix https://huggingface.co/datasets/fka/awesome-chatgpt-prompts/viewer/default/train?f[act][min]=4&f[act][max]=8&f[act][transform]=length

image

This was caused by `RuntimeError: The indexing process did not store the features.`

### DIFF
--- a/services/search/src/search/routes/filter.py
+++ b/services/search/src/search/routes/filter.py
@@ -138,6 +138,10 @@ def create_filter_endpoint(
                         hf_token=hf_token,
                     )
                 with StepProfiler(method="filter_endpoint", step="get features"):
+                    # Features can be missing in old cache entries,
+                    # but in this case it's ok to get them from the Arrow schema.
+                    # Indded at one point we refreshed all the datasets that need the features
+                    # to be cached (i.e. image and audio datasets).
                     if "features" in duckdb_index_cache_entry["content"] and isinstance(
                         duckdb_index_cache_entry["content"]["features"], dict
                     ):
--- a/services/search/src/search/routes/filter.py
+++ b/services/search/src/search/routes/filter.py
@@ -4,7 +4,7 @@
 import logging
 import re
 from http import HTTPStatus
-from typing import Optional
+from typing import Literal, Optional, Union
 
 import anyio
 import duckdb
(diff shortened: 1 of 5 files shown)


services/search/src/search/routes/filter.py
```diff
@@ -138,20 +138,21 @@ async def filter_endpoint(request: Request) -> Response:
...
-                                for name, feature in duckdb_index_cache_entry["content"]["features"].items()
-                                if name != ROW_IDX_COLUMN
-                            }
-                        )
-                    except (KeyError, AttributeError):
-                        raise RuntimeError("The indexing process did not store the features.")
+                    if "features" in duckdb_index_cache_entry["content"] and isinstance(
+                        duckdb_index_cache_entry["content"]["features"], dict
+                    ):
+                        features = Features.from_dict(duckdb_index_cache_entry["content"]["features"])
```
albertvillanova, line 144: Note that you are now also including the `ROW_IDX_COLUMN`, contrary to what it was done before...

albertvillanova (commented):
Thanks for the fix.

Do you know why there might be no features? I remember having asked when creating the /filter endpoint and you told me that we could safely assume there were features (only very exceptionally we would not find and we could fail in that case)...

I found the conversation: https://github.com/huggingface/datasets-server/pull/1418#discussion_r1334517809
> You can raise an error if it's None imo.
>
> It can only happen in very rare cases (parquet dataset processed with old version of datasets for example)

services/search/src/search/routes/filter.py
lhoestq: Yes noted ! I added a test for create_response to take this into account

lhoestq:
> Do you know why there might be no features? I remember having asked when creating the /filter endpoint and I was told that we could safely assume there were features (only very exceptionally we would not find and we could fail in that case)...

Because the first version of the "config-parquet" job didn't store the features and now it does.

So old datasets may have features: None for subsequent processing jobs like split-duckdb-index

It's not a big deal since all the image and audio datasets (those that actually require the "features" info in the cache) were refreshed at that time. So for all the datasets with missing "features" we can infer the features from e.g. the pyarrow schema

services/search/src/search/routes/filter.py
albertvillanova: IMO ROW_IDX_COLUMN is not part of the dataset features (we just added that column in the DuckDB table for convenience).

lhoestq: The split-duckdb-index job does store ROW_IDX_COLUMN in the features

https://github.com/huggingface/datasets-server/blob/fc3e26e23d6d2aaa6dfa2f472a0276181370183d/services/worker/src/worker/job_runners/split/duckdb_index.py#L284-L285

Here I reuse the same features but since `use_row_idx_column=True` the ROW_IDX_COLUMN is not returned in `create_response`

severo:
maybe add a comment in the code to make it clear?

albertvillanova:
I found the conversation: https://github.com/huggingface/datasets-server/pull/1418#discussion_r1334517809
> You can raise an error if it's None imo.
>
> It can only happen in very rare cases (parquet dataset processed with old version of datasets for example)

services/search/src/search/routes/filter.py
lhoestq: If the "features" field exists in the split-duckdb-index cache content, it contains ROW_IDX_COLUMN

If it doesn't exist, we do `Features.from_arrow_schema(pa_table.schema),` which also contains ROW_IDX_COLUMN

In any case `to_rows_list` takes care of removing the column from the row in the end

https://github.com/huggingface/datasets-server/blob/fc3e26e23d6d2aaa6dfa2f472a0276181370183d/libs/libapi/src/libapi/utils.py#L213

PR #2348, opened 2024-01-29
[Croissant] JSON-LD requests are `full=true` by default.

marcenacp:


services/api/src/api/routes/croissant.py
```diff
@@ -203,7 +203,7 @@ def get_croissant_from_dataset_infos(
 
 def _get_full_jsonld_parameter(request: Request) -> bool:
     """Parameter to retrieve the full JSON-LD (full=True) or a truncated/abridged JSON-LD (full=False) with less features."""
-    full_jsonld = get_request_parameter(request, "full", default="false")
+    full_jsonld = get_request_parameter(request, "full", default="true")
     if full_jsonld.lower() == "true":
         return True
     return False
```
severo, line 209: maybe

```suggestion
    full_jsonld = get_request_parameter(request, "full", default="true")
    if full_jsonld.lower() == "false":
        return False
    return True
```

It would be full, unless we specify `full=false`

severo (approved):

severo:
waiting for https://github.com/huggingface/moon-landing/pull/8794 (internal) before merging and deploying

PR #2375, opened 2024-02-02
Fix backfill when dataset only has one cache entry

### LINKED ISSUE #2274: Only one cache entry, for the first step
See causalnlp/corr2cause

Capture d’écran 2024-01-11 à 09 49 42

The first step was successful, but no other step was computed.

reported here: https://huggingface.co/datasets/causalnlp/corr2cause/discussions/5

severo:
fixes #2274

### DIFF
--- a/jobs/cache_maintenance/src/cache_maintenance/backfill.py
+++ b/jobs/cache_maintenance/src/cache_maintenance/backfill.py
@@ -6,7 +6,7 @@ from typing import Optional
 
 from libcommon.dtos import Priority
 from libcommon.exceptions import NotSupportedError
-from libcommon.operations import update_dataset
+from libcommon.operations import backfill_dataset
 from libcommon.simple_cache import get_all_datasets
 from libcommon.storage_client import StorageClient
 
@@ -36,7 +36,7 @@ def backfill_cache(
     for dataset in datasets_in_database:
         analyzed_datasets += 1
         try:
-            update_dataset(
+            backfill_dataset(
                 dataset=dataset,
                 hf_endpoint=hf_endpoint,
                 blocked_datasets=blocked_datasets,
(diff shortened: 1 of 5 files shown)


severo:
Let's examine the list of datasets with the bug.

## `iNeil77/paired-single-code-2'`

Deleted (or private?) https://huggingface.co/datasets/iNeil77/paired-single-code-2
But it has not been deleted
Maybe a concurrency issue between:
- the job that was running to create the cache entry
- the webhook that deleted the cache entries and the job entries
When we remove a dataset, we delete, in order: the jobs, then the cache entries, then the assets and cached assets.
Note that we don't delete the job locks... should we? maybe it's related?

## Remaining

```
leyanpan/sat-solver
Leogrin/real-toxicity-prompts_first_5K
ayymen/Weblate-Translations
CyberHarem/miyu_edelfelt_fgo
Recag/Rp_CommonC_241
Recag/Rp_CommonC_151
CyberHarem/golden_hind_azurlane
cutterd/gelgen_tar_29
Recag/Rp_CommonC_539_1
lchakkei/OpenOrca-Traditional-Chinese-ChatML-Format-TwoPercent
jarvisx17/Medical-ASR-EN
Treza12/Cancer-Dataset
yn01/test_20240124_01
Recag/Rp_CommonC_412_1
x_stance
nekoknights/catchthesedeals
kenhktsui/open-toolformer-retrieval-multi-neg-result-new-kw
iNeil77/paired-single-code-2
```

severo:
I replicated https://huggingface.co/datasets/kenhktsui/open-toolformer-retrieval-multi-neg-result-new-kw, run `set_revision` on it, and effectively, the `DatasetBackfillPlan` contains 0 tasks, ie: it will not create the missing cache entries!

severo:
In fact, we have two bugs:
1. at some point, a cache entry is created, but all the jobs are deleted
2. in the backfill cron job, the dataset is "updated", but the jobs for the missing cache entries are not recreated as they should

The bug 2. is reproduced in https://github.com/huggingface/datasets-server/pull/2375/commits/8d979ae228d6f46e386492ea681128a7013ff973. Let's try to fix it.

severo:
OK. I think it's OK: we do a real backfill on all the steps, in the backfill cron job. Beware, it might be too long, maybe we could parallelize. Let's do it in another PR.

PR #2389, opened 2024-02-05
handle visibility (public/private) changes

### LINKED ISSUE #2379: Don't ignore webhooks when a dataset changes visibility
See:
- https://github.com/huggingface/moon-landing/issues/8779
- https://github.com/huggingface/moon-landing/pull/8825

A webhook is sent when a dataset is toggle between public and private.

Currently, we ignore them due to 

https://github.com/huggingface/datasets-server/blob/66c1e089e204ab33195b957e1b99b0da6a4dd2de/services/api/src/api/routes/webhook.py#L84-L89

Instead, I think we should delete+update the dataset

severo:
side-effect : any configuration change recreates everything in the datasets-server.
Possibly, the Hub will implement "scope": "repo.config.visibility", but for now, it's the way to react to visibility changes.

fixes #2379

### DIFF
--- a/services/api/src/api/routes/webhook.py
+++ b/services/api/src/api/routes/webhook.py
@@ -71,7 +71,7 @@ def process_payload(
     hf_timeout_seconds: Optional[float] = None,
     storage_clients: Optional[list[StorageClient]] = None,
 ) -> None:
-    if payload["repo"]["type"] != "dataset" or payload["scope"] not in ("repo", "repo.content"):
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Cool ! In general it will be from private/unsupported to public/supported so it's fine for now

coyotte508 (approved):
(you could store visibility & see if new visibility is different without relying on scope?)

AndreaFrancis (approved):

severo:
> (you could store visibility & see if new visibility is different without relying on scope?)

yes, we might do it indeed. Opening an issue: https://github.com/huggingface/datasets-server/issues/2390

PR #2386, opened 2024-02-02
Add loading tags, starting with croissant and webdataset

lhoestq:
Added two tags:
- "croissant" for datasets that have a valid info (= a valid croissant json)
- "webdataset" fro datasets with `info["builder_name"] == "webdataset"`

I implemented this using a new job "dataset-loading-tags" that aims at detecting tags related to dataset loading tools, and then adding the tags to "dataset-hub-cache".

Later we can enrich "dataset-loading-tags" with loading argument that can be used when showing code snippets in dataset pages.

TODO
- [x] add to hub-cache
- [x] tests

PR #2391, opened 2024-02-05
return list of tags in /hub-cache

severo:
Add a list of tags in /hub-cache used to refresh the Hub's datasets cache. For now, we add the "croissant" tag if `dataset-info` exists and is complete (all the configs).


severo:
let's add webdataset at the same time, see #2386

Issue #2387, opened 2024-02-05
Simple jobs have difficulty 100


severo:
After a backfill on 2024/02/03, we got a lot of new jobs, and in particular, a lot of them have a high difficulty, much higher than what they should reach.

<img width="643" alt="Capture d’écran 2024-02-05 à 10 41 58" src="https://github.com/huggingface/datasets-server/assets/1676121/35a31d46-451f-476f-9691-01b7d302df23">

severo:
fixed by #2403

PR #2403, opened 2024-02-06
delete only waiting jobs

AndreaFrancis:
Related to https://github.com/huggingface/datasets-server/issues/2387
As per [internal conversation](https://huggingface.slack.com/archives/C04L6P8KNQ5/p1707031392098359), only waiting jobs should be deleted on backfill to avoid started processes without a related Job record.


libs/libcommon/src/libcommon/queue.py
```diff
@@ -545,18 +545,18 @@ def create_jobs(self, job_infos: list[JobInfo]) -> int:
             return 0
 
     def delete_jobs_by_job_id(self, job_ids: list[str]) -> int:
```
severo, line 547: maybe change the name to `delete_waiting_jobs_by_job_id` to be more explicit

severo: done in [dbc2f25](https://github.com/huggingface/datasets-server/pull/2403/commits/dbc2f25763674ece0be3513f221d8d58b5767626)

PR #2406, opened 2024-02-06
Update cryptography dependency to 42.0 to fix vulnerability

albertvillanova:
This should fix 11 dependabot alerts.


severo (approved):
thanks!

PR #2415, opened 2024-02-06
catch ValueError before ImportError

### LINKED ISSUE #2385: Replace `DatasetModuleNotInstalledError` errors with `DatasetWithScriptNotSupportedError`
We should never have a `DatasetModuleNotInstalledError` error, because we should return a `DatasetWithScriptNotSupportedError` error before

See https://github.com/huggingface/datasets-server/issues/1067#issuecomment-1924305954

severo:
fixes #2385 

I didn't increment the job runner version, because it only happen in 235 datasets. We will update them manually.

```
db.cachedResponsesBlue.countDocuments({error_code: "DatasetWithScriptNotSupportedError", kind: "dataset-config-names"})
235
```

### DIFF
--- a/services/worker/src/worker/job_runners/dataset/config_names.py
+++ b/services/worker/src/worker/job_runners/dataset/config_names.py
@@ -75,17 +75,20 @@ def compute_config_names_response(
         ]
     except _EmptyDatasetError as err:
         raise EmptyDatasetError("The dataset is empty.", cause=err) from err
-    except ImportError as err:
-        raise DatasetModuleNotInstalledError(
-            "The dataset tries to import a module that is not installed.", cause=err
-        ) from err
-    except Exception as err:
-        if isinstance(err, ValueError) and "trust_remote_code" in str(err):
+    except ValueError as err:
+        if "trust_remote_code" in str(err):
             raise DatasetWithScriptNotSupportedError(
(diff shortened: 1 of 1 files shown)


services/worker/src/worker/job_runners/dataset/config_names.py
```diff
@@ -75,18 +75,21 @@ def compute_config_names_response(
...
                 "if you think this is an error and tag @lhoestq and @severo."
             ) from err
         raise ConfigNamesError("Cannot get the config names for the dataset.", cause=err) from err
+    except ImportError as err:
+        # this should only happen if the dataset is in the allow list, which should soon disappear
+        raise DatasetModuleNotInstalledError(
+            "The dataset tries to import a module that is not installed.", cause=err
+        ) from err
+    except Exception as err:
+        raise ConfigNamesError("Cannot get the config names for the dataset.", cause=err) from err
```
albertvillanova, line 92: I don't like having to repeat the raising of `ConfigNamesError` twice in the code, but I can't find a better way to do it...

albertvillanova (approved):
Thanks.

services/worker/src/worker/job_runners/dataset/config_names.py
severo: same

PR #2424, opened 2024-02-08
rename X-Revision to X-Repo-Commit

### LINKED ISSUE #2417: Rename X-Revision header to X-Repo-Commit
Align with the convention on the Hub: https://github.com/huggingface/moon-landing/issues/8169#issuecomment-1824620102

severo:
fixes #2417


lhoestq:
We use `X-Revision` in `datasets` to check for the Parquet export. If you rename it it will break features that rely on the Parquet export in `datasets` (e.g. use the Parquet export instead of a loading script)

PR #2432, opened 2024-02-09
Hardcode 100 rows max in /first-rows

### LINKED ISSUE #1921: Refactor FIRST_ROWS_MAX_NUMBER and MAX_NUM_ROWS_PER_PAGE
As reported by @severo (see: https://github.com/huggingface/datasets-server/pull/1418#discussion_r1338867209):
- Should we handle `MAX_NUM_ROWS_PER_PAGE` with `FIRST_ROWS_MAX_NUMBER`? Or hardcode `FIRST_ROWS_MAX_NUMBER` and use `MAX_NUM_ROWS_PER_PAGE` instead (I think so)
- but we set `FIRST_ROWS_MAX_NUMBER` to 4 in the tests...

Also note that `MAX_NUM_ROWS_PER_PAGE` was already used in "/rows" and "/search", besides "/filter" once #1418 merged.

severo:
fixes #1921


albertvillanova (approved):
Thanks. Much simpler now!

PR #2445, opened 2024-02-13
Update gradio

albertvillanova:
Update `gradio` to ^4.17.0.

Fix #2444.

PR #2447, opened 2024-02-13
requirements.txt is used by spaces, pyproject.yaml is ignored

severo:
follows #2445. cc @albertvillanova fyi. I already did the same error... We should find a way to have only one source of truth for the packages, instead of ... 3!


severo:
hmmm, does not work, the space stills builds with radio 4.16.0

severo:
it works now

Capture d’écran 2024-02-13 à 13 45 22

PR #2448, opened 2024-02-13
Log more details when worker crashes

severo:
Also, random updates:
- fix names: we don't "cancel" jobs anymore, we "delete" them
- remove obsolete catch of `DatasetInBlockListError` exception when running `finish_job()` -> it's not checked anymore

### DIFF
--- a/services/worker/src/worker/executor.py
+++ b/services/worker/src/worker/executor.py
@@ -201,7 +201,7 @@ class WorkerExecutor:
             explanation = f"exit code {err.exit_code}"
             if err.exit_code == -9:
                 explanation += " SIGKILL - surely an OOM"
-            error_msg = f"[Worker executor] Worker crashed ({explanation})"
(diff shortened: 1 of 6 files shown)


services/worker/src/worker/executor.py
```diff
@@ -198,13 +198,13 @@ def is_worker_alive(self, worker_loop_executor: OutputExecutor) -> bool:
         try:
             worker_loop_executor.stop()  # raises an error if the worker returned unexpected exit code
         except ProcessExitedWithError as err:
-            explanation = f"exit code f{err.exit_code}"
+            explanation = f"exit code {err.exit_code}"
```
severo, line 201: @lhoestq : ultra nit, but: was this `f` char wanted? or a typo?

lhoestq: it's a typo

PR #2452, opened 2024-02-14
Add croissant page

severo:
also add mention of support for the private datasets

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -28,7 +28,7 @@
       title: Get the number of rows and the bytes size
     - local: statistics
       title: Explore dataset statistics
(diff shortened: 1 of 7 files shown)


julien-c (approved):
looks good, but shouldn't it be in there:

image

and/or in there:

image

severo:
> looks good, but shouldn't it be in there:

Croissant is somewhat special because it relies on /croissant, while all the other ones rely on /parquet. But, I think it's a detail that is irrelevant for the users. Instead of focusing on libraries that can load Parquet files, we can showcase libraries that can load the data "in general".

julien-c:
yes agree

julien-c (approved):

Issue #1843, opened 2023-09-22
 Descriptive stats: unsupported operand type(s) for -: 'NoneType' and 'NoneType


AndreaFrancis:
There are 1516 cache records with the following error:
```

unsupported operand type(s) for -: 'NoneType' and 'NoneType'
unsupported operand type(s) for *: 'NoneType' and 'float'

```
See example https://datasets-server.huggingface.co/statistics?dataset=mozilla-foundation/common_voice_9_0&config=ha&split=validation&offset=0&limit=100
See for more details in db:
`{kind:"split-descriptive-statistics", http_status:500, "content.error":/unsupported operand/}`

Related to https://github.com/huggingface/datasets-server/issues/1443

AndreaFrancis:
Current state:
```
Atlas atlas-x5jgb3-shard-0 [primary] datasets_server_cache> db.cachedResponsesBlue.countDocuments( {kind:"split-descriptive-statistics", http_status:500, "content.error":/unsupported operand/})
2294

```

AndreaFrancis:
Currently we have:
```
Atlas atlas-x5jgb3-shard-0 [primary] datasets_server_cache> db.cachedResponsesBlue.countDocuments( {kind:"split-descriptive-statistics", http_status:500, "content.error":/unsupported operand/})
1166

```

Those should be fixed with a step refresh (Fixed by https://github.com/huggingface/dataset-viewer/pull/2453)

PR #2453, opened 2024-02-14
Statistics for list feature

polinaeterna:
Return distribution of lengths for lists (independently of what's inside).
_Note_: empty list is not counted as `na` value (so, not increment `na_count`), it is len=0.

I haven't done a proper testing of speed in an isolated environment (because i'm fighting with dev dockers locally, idk why), but a worker running locally without a container processes this dataset https://huggingface.co/datasets/polinaeterna/benchmark very fast (much more time was spend on download). So I assume it wouldn't be too much time to recompute.  

Also I tried to refactor the code for computation a bit, it can be better, feel free to suggest the way to improve it :)

~~Before merge I wanted to merge [this PR](https://github.com/huggingface/datasets-server/pull/2469) - to better track stats duration for bigger datasets, but I'm not sure I'm doing it right so would appreciate if someone who knows Prometheus take a look.~~ - merged

\+ this PR fixes this issue https://github.com/huggingface/datasets-server/issues/1843


severo:
> can also add other iterables (like dict, anything else?)

for lists, I imagine we want to show a histogram of the lists lengths?

for dicts, it would be a histogram of the number of keys? I feel like a dict column will nearly always have the same number of keys.

are there other iterables? (I imagine that Sequence is the same as list)

severo:
So... we review https://github.com/huggingface/datasets-server/pull/2469 before, right? is it ready?

polinaeterna:
> So... we review https://github.com/huggingface/datasets-server/pull/2469 before, right? is it ready?

@severo yes, I think it's not ready but I'm stuck and not sure if I'm going in the right direction. And if it's even needed. If not - you can ignore that pr and review this one

polinaeterna:
actually there are quite a lot of datasets with list/Sequence feature. I cannot get real numbers rn because I'm not able to connect to the cluster :/ but I tested on a subset of 1000 datasets through API and there are 300 of dataset with at least one config with a list feature. so an estimate of 30% from 100k would be 30k

libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -525,7 +525,7 @@ def parse_id(id: str) -> tuple[str, str, Optional[str], Optional[str], str]:
     },
     "split-descriptive-statistics": {
         "input_type": "split",
-        "triggered_by": "config-split-names",
+        "triggered_by": ["config-split-names", "config-parquet-metadata"],
```
polinaeterna, line 528: I added this because it requires parquet metadata, wdyt?

severo: OK! it explains why we have frequent `PreviousStepStillProcessingError`. This should fix it.

By the way, I think it's even:

```suggestion
        "triggered_by": "config-parquet-metadata",
```

because we don't require `config-split-names` to have been computed, we don't use it.

severo: surely test_processing_graph.py must be updated

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -577,47 +648,60 @@ def compute_descriptive_statistics_response(
...
+                    return StringColumn(
+                        feature_name=dataset_feature_name, n_samples=num_examples, n_bins=histogram_num_bins
+                    )
+
+                if dataset_feature.get("dtype") == "bool":
+                    return BoolColumn(feature_name=dataset_feature_name, n_samples=num_examples)
+        return None
+
+    all_stats = []
+    for feature_name, feature in features.items():
```
severo, line 687: do you think we could run them in parallel? would it help?

services/worker/tests/job_runners/split/test_descriptive_statistics.py
```diff
@@ -487,6 +521,8 @@ def test_string_statistics(
         assert expected == computed
 
 
+#
+#
```
severo, line 525:
```suggestion
```

severo (approved):
excellent!

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: Maybe but I wouldn't do this to avoid potential fights between Polars' threads and python processes because Polars is multithreaded when possible (I still have nightmares with multithreaded duckdb in a subprocess :D)

I assume it would be more efficient to rewrite code somehow to apply processing funcs to all columns (or maybe all columns of the same type) at once using native Polars code but I think this is a pretty massive change... 

for reference: https://docs.pola.rs/user-guide/misc/multiprocessing/

polinaeterna: > or maybe all columns of the same type

i think this is actually quite possible but maybe in another pr if we see that jobs stuck?

lhoestq:
🚀

Issue #2413, opened 2024-02-06
Update duckdb to 0.10.0


severo:
It will be the first version that will ensure backward-compatibility of the .duckdb format

severo:
https://github.com/huggingface/datasets-server/pull/2316#issuecomment-1930065020

> Having API workers that could read both duckdb 0.8 and 0.10 sounds like a bit of a pain actually.
> 
> Maybe it will be easier to add a new job to compute the new indexes, then after some time switch the API to use them once all the indexes are ready and then the old indexes ?

severo:
released: https://duckdb.org/2024/02/13/announcing-duckdb-0100

albertvillanova:
Happy to see the switch to a backward-compatible version (we could easily update in the future) and also to have the DuckDB file in a different place than the Parquet files.

AndreaFrancis:
From https://github.com/huggingface/datasets-server/pull/2316#issuecomment-1930065020:
> Having API workers that could read both duckdb 0.8 and 0.10 sounds like a bit of a pain actually.
> 
> Maybe it will be easier to add a new job to compute the new indexes, then after some time switch the API to use them once all the indexes are ready and then the old indexes ?

Do you mean to have two parallel job runners? One for the previous version and a new one for 0.10? 
If it is only one for the new one 0.10, search/filter will be available only for the already processed datasets.

wdyt @lhoestq @severo ?

severo:
I think it's two in parallel to make the transition. And we could take the opportunity to store the files elsewhere (not along with the parquet files)

AndreaFrancis:
> I think it's two in parallel to make the transition. And we could take the opportunity to store the files elsewhere (not along with the parquet files)

Let me check if it is possible to have two different versions of the same package in poetry (not sure yet).

lhoestq:
It's probably not possible directly :/

We might need a second python environment and run the indexing in a subprocess using the second python environment.

Or maybe it's possible to install the duckdb cli with the new version and run the indexing in a subprocess simply by calling the cli

AndreaFrancis:
> And we could take the opportunity to store the files elsewhere (not along with the parquet files)

Also related to https://github.com/huggingface/datasets-server/issues/1694#issuecomment-1737010286
and see internal discussion on Slack: https://huggingface.slack.com/archives/C067U56EQBB/p1701174059111469
Also: https://huggingface.slack.com/archives/C0311GZ7R6K/p1686665484050679
[...]

PR #2463, opened 2024-02-16
Update duckdb to 0.10.0

AndreaFrancis:
Will fix https://github.com/huggingface/datasets-server/issues/2413

This PR will include the following changes:
- Add a new job runner that will compute duckdb files with `0.10.0` -> Temporal cache kind `split-duckdb-index-010`
- Still have the old job runner that will compute duckdb files with `0.8.1` (Will use duckdb CLI to support two different duckdb versions, it was not possible with poetry) -> Cache kind is still split-duckdb-index to allow /search and /filter service working as usual while `split-duckdb-index-010` is being computed
- Will point the new job runner  `split-duckdb-index-010` to `/refs/convert/duckdb` instead or `refs/conver/parquet`
- Will add a new field in `split-duckdb-index-010` to specify the duckdb version used
- CLI [duckdb-081](https://github.com/duckdb/duckdb/releases/download/v0.8.1/duckdb_cli-linux-amd64.zip) is downloaded and unzipped in the docker container. 

Deploy plan - Once this PR is merged, deployed,  and all `split-duckdb-index-010` entries have been computed, send another PR to:
- Migration to delete all previous `split-duckdb-index` cache entries
- Migration to rename all cache entries with kind `split-duckdb-index-010` to `split-duckdb-index`
- Update the duckdb version in `/search` service
- Remove old code (I put TODOs to quickly identify which parts need to be removed from the old job runner)
- Refresh all entries for config-duckdb-index-size, split-is-valid, dataset-duckdb-index-size, config-is-valid, dataset-is-valid, dataset-hub-cache (Maybe increase job runner version and backfill?)
- Stop workers when deploying


chart/values.yaml
```diff
@@ -273,7 +273,7 @@ duckDBIndex:
   # the git commit message when the duckdb index file is uploaded to the Hub. Defaults to `Update duckdb index files`.
   commitMessage: "Update duckdb index files"
   # the git revision of the dataset where to store the duckdb index file. Defaults to `refs/convert/parquet`.
```
severo, line 275:
```suggestion
  # the git revision of the dataset where to store the duckdb index file. Defaults to `refs/convert/duckdb`.
```

services/worker/tools/duckdb
severo, line 1: best in services/worker/vendors/duckdb. It's already used to copy/paste other tools (a python library in the current case)

services/worker/src/worker/dtos.py
```diff
@@ -147,6 +147,11 @@ class SplitDuckdbIndex(SplitHubFile):
     num_bytes: Optional[int]
 
 
+# TODO: Rename to SplitDuckdbIndex after all split-duckdb-index-010 cache entries have been computed
+class SplitDuckdbIndexWithVersion(SplitDuckdbIndex):
+    version: str
```
severo, line 152: maybe `duckdb_version`?

severo (commented):
The plan seems good.

I don't know much about subprocesses. How would we handle a failure in the duckdb CLI? will it be exactly the same as now?

AndreaFrancis:
> I don't know much about subprocesses. How would we handle a failure in the duckdb CLI? will it be exactly the same as now?

It will raise a RuntimeError in the worker. I would instead not go deeper into the kind of exceptions this process can throw because I would like to remove the whole 0.8.1 job runner as soon as the 0.10.0 cache entries are computed.

services/worker/src/worker/job_runners/split/duckdb_index_0_8_1.py
```diff
@@ -0,0 +1,352 @@
...
+    retry,
+)
+
+# TODO: Remove file when all split-duckdb-index-010 entries have been computed
+
+
+def index_with_cli(
+    extensions_directory: Optional[str],
+    cli_path: Optional[str],
+    database_name: str,
```
lhoestq, line 72: rename the argument database_path ?

services/worker/tests/job_runners/split/test_duckdb_index.py
```diff
@@ -31,6 +31,7 @@
 from worker.job_runners.split.duckdb_index import (
```
lhoestq, line 31: should we duplicate these tests for the one based on the CLI as well ?

lhoestq:
> Is it ok to copy the executable [duckdb-081](https://github.com/duckdb/duckdb/releases/download/v0.8.1/duckdb_cli-linux-amd64.zip) in the Dockerfile? It could also be a download/unzip but since it will be removed, I thought any related logic should be as simple as possible.

Sounds good to me ! Better have it as a download/unzip to not make this git repo too big IMO

services/worker/tools/duckdb
AndreaFrancis: I changed it to download from the official site but not sure why it is still failing in the tests https://github.com/huggingface/datasets-server/pull/2463/files#diff-cd01935367648c9c9e517986ecce6ef485028972ecdc07408577d13e5eb73f2dR34 
:(

services/worker/tests/job_runners/split/test_duckdb_index.py
AndreaFrancis: I duplicated just the part of [test](https://github.com/huggingface/datasets-server/pull/2463/files#diff-4da26c1315609343b7f12e022b53fe14f12ea1ce66eac62ad21995bb2f1ff144R32) computing

services/worker/tools/duckdb
lhoestq: Not sure but on my side adding `bash` seemed to help:

```python
duckdb_0_8_1_command = ["bash", cli_path, database_path]
```

lhoestq: Ah actually it helped only because I failed splitting the rest of the command correctly, sorry.

Could it be a absolute/relative path issue ?

lhoestq:
For unit tests: maybe `duckdb` download is missing in `.github/workflows/_unit-tests-python.yml` ?

services/worker/tools/duckdb
AndreaFrancis: Maybe, I am passing it as an absolute path in /src/services/worker/duckdb

AndreaFrancis:
After ten tries, I was able to have it working for the unit test. It is ready for review.
I didn't find a way to download/unzip the CLI only for worker tests. I would really like to remove all the "parallel" code for 0.8.1 after 0.10.0 entries are computed.

severo (approved):

lhoestq (approved):
Awesome !

PR #2469, opened 2024-02-19
Change Prometheus histogram buckets for duration of `split-descriptive-statistics` job

polinaeterna:
I want to increase time range to track stats worker running time in Grafana but I'm not sure I'm doing it right.

And not sure it's much needed but I believe it would be useful for me to monitor


severo:
Good idea! We have the same issue on the charts in Grafana (https://grafana.huggingface.tech/d/i7gwsO5Vz/global-view) about the response time of some endpoints: as the max is 10s by default, we lose the info for longer durations. We might use the same technique for the starlette metrics

Capture d’écran 2024-02-28 à 19 04 16

severo:
Feel free to fix the tests, and we can merge and try with grafana to ensure it works as expected

polinaeterna:
@severo tests are green, shall we merge and see if that works? 
(i'll probably won't be able to check because i have problems with access to our services (something with my network) again :/// )

libs/libcommon/src/libcommon/constants.py
```diff
@@ -62,3 +62,21 @@
 
 CROISSANT_MAX_CONFIGS = 100
 MAX_NUM_ROWS_PER_PAGE = 100
+
+DESCRIPTIVE_STATISTICS_PROMETHEUS_HISTOGRAM_BUCKETS = (
```
severo, line 66: the buckets are a bit unbalanced. What if we do a generic list that could be used for any worker, with an exponential increase?

Such as:

```
0.01 / 0.02 / 0.05 / 0.1 / 0.2 / 0.5 / 1 / 2 / 5 / 10 / 30 / 60 / 60*2 / 60*5 / 60*10 / 60*20 / 60*60?
```

severo: and let's call it WORKER_PROMETHEUS_HISTOGRAM_BUCKETS?

severo: ok, too late ....

severo: Or, to keep closer to the [default values in prometheus](https://github.com/prometheus/client_python/blob/7a80f001237fe881d3607861947292abc85bf205/prometheus_client/metrics.py#L586):

```
WORKER_PROMETHEUS_HISTOGRAM_BUCKETS = (.005, .01, .025, .05, .075, .1, .25, .5, .75, 1.0, 2.5, 5.0, 7.5, 10.0, 30.0, 60.0, 2.5*60.0, 5*60.0, 7.5*60.0, 10*60.0, 30*60.0, 60*60.0, INF)
```

severo: ^ no need to tweak for minutes/hours, in fact... these boundaries will be used to compute statistics. Let's use the same space between steps, until reaching the max value we want (5000 = 1h23minutes is ok, as we stop the job runner after 40 minutes)

PR #2471, opened 2024-02-19
Remove codecov from CI

### LINKED ISSUE #2470: Remove codecov from CI
Remove `codecov` from CI.

See: https://github.com/huggingface/datasets-server/pull/2464#issuecomment-1952372916

> I personally don't really use it, and I surely didn't configure it well, because due to the monorepo structure of our repo (I think) the notifications are meaningless

albertvillanova:
Remove `codecov` from CI.

Fix #2470.


severo (commented):
Here we remove integration of codecov in the CI (ie. pushing the results of pytest-cov to an external service), but also the ability to compute the code coverage with `make coverage`.
I think it's OK, since 1. we don't really use it to be honest, 2. it's easy to put again if someday we want it again.
In this case, we surely want to also remove the `coverage` target in tools/PythonTest.mk

severo:
hmmm, I'm thinking: should we remove the `trec-car-tools` vendored dependency, now that we don't support python scripts anymore? I'm not sure which dataset required it (it's not https://huggingface.co/datasets/trec)

services/worker/vendors/trec-car-tools/python3/conf.py
```diff
@@ -27,7 +27,7 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    'sphinx.ext.coverage',
+    # 'sphinx.ext.coverage',
```
severo, line 30: I never changed anything in `services/worker/vendors/trec-car-tools`, it's an exact copy of the dependency... Not sure if we should change it.

Anyway, maybe we should delete this dependency anyway

severo (approved):

services/worker/vendors/trec-car-tools/python3/conf.py
severo: Also: it seems to break the worker tests

albertvillanova:
I am rebasing this PR onto master.

services/worker/vendors/trec-car-tools/python3/conf.py
albertvillanova: I was trying to have our CI green, but it didn't work anyway.

PR #2475, opened 2024-02-20
remove dependency 'trec-car-tools' from worker

### LINKED ISSUE #2474: Remove vendored dependency `trec-car-tools`
See https://github.com/huggingface/datasets-server/tree/main/services/worker/vendors/trec-car-tools

It's not required anymore since we don't allow datasets with script.

severo:
fixes #2474


albertvillanova (approved):
I agree it seems better to remove it if not necessary...

PR #2477, opened 2024-02-20
remove parallel steps

### LINKED ISSUE #2460: Order of the columns is not consistent between /first-rows and /rows
See /first-rows:

Capture d’écran 2024-02-16 à 11 52 25

/rows

Capture d’écran 2024-02-16 à 11 52 32

### LINKED ISSUE #2473: Remove the parallel steps?
I think it would be simpler to delete the concept of "parallel" steps.

Currently, we have two pairs of parallel steps:

1. `config-split-names-from-streaming` and `config-split-names-from-info`
2. `split-first-rows-from-streaming` and `split-first-rows-from-parquet`

It generates some complications:
- we don't want to store them both, in particular, the first-rows that have the heaviest content in the database
- the responses can be incoherent (see https://github.com/huggingface/datasets-server/issues/2460)
- it implies specific handling in the code (see `parallel_step_names` in the code)

The parallel steps are currently used to try to get a quick response from streaming the dataset, when it's faster than converting to parquet. But we should always trust the parquet ones more than the streaming ones. Also: computing from parquet (once we have parquet files) is super fast, while it can take some time from streaming.

So, I propose:

- to remove the concept of parallel steps (and raise an error if a processing graph has parallel steps)
- to have a conditional in the unique implementation of `config-split-names` and `split-first-rows`: use parquet/info preferentially, else use streaming
- (note: we have to ensure we don't have concurrency issue, and we always end up with the parquet version when possible, not the streaming one)

What do you think @huggingface/datasets-server?

severo:
With this PR, we remove the weird concept of parallel steps (which do not exist in DAG software/theory, I think).
 
Benefits:
- multiple parts of the code are now simpler
- fixes #2460

Drawbacks:
- we might end up computing from "streaming" twice when the Parquet has any error. Anyway, it should be 1. exceptional, 2. fast.

fixes #2473 

Details:
- [x] change array of kinds in api/endpoint to one kind
- [x] take only one step in get_cache_entry_from_steps, and move the corresponding tests (if some are still necessary) from service/api/tests/test_endpoint.py to libapi
- [x] replace get_best_response with get_response_or_missing_error
- [x] delete get_best_response
- [x] remove concept of BestResponse
- [x] remove variables with "best_response" in their name
- [x] take only one processing_step_name in try_backfill_dataset_then_raise
- [x] take only one kind in fetch_names
- [x] take only one step in get_previous_step_or_raise
- [x] has_any_successful_response -> is_successful_response
- [x] remove parallel mechanism
- [x] remove ResponseAlreadyComputedError
- [x] implement, test and creation migration for
  - [x] config-split-names (and fix tests_split_names)
  - [x] split-first-rows


libs/libcommon/src/libcommon/constants.py
```diff
@@ -48,20 +48,15 @@
...
-]
+CONFIG_HAS_VIEWER_KIND = "config-size"
+CONFIG_INFO_KIND = "config-info"
+CONFIG_PARQUET_METADATA_KIND = "config-parquet-metadata"
+CONFIG_SPLIT_NAMES_KIND = "config-split-names"
+DATASET_CONFIG_NAMES_KIND = "dataset-config-names"
+DATASET_INFO_KIND = "dataset-info"
+SPLIT_DUCKDB_INDEX_KIND = "split-duckdb-index"
+SPLIT_HAS_PREVIEW_KIND = "split-first-rows"
+SPLIT_DUCKDB_INDEX = "split-duckdb-index"
```
AndreaFrancis, line 59: Why not to use SPLIT_DUCKDB_INDEX_KIND instead?

services/worker/tests/test_job_manager.py
```diff
@@ -200,52 +198,52 @@ def test_job_runner_set_crashed(app_config: AppConfig) -> None:
...
-        job_info=job_info,
-        app_config=app_config,
-    )
-
-    job_manager = JobManager(job_info=job_info, app_config=app_config, job_runner=job_runner)
-    with pytest.raises(CustomError) as exc_info:
-        job_manager.raise_if_parallel_response_exists(parallel_step_name=stepA)
-    assert exc_info.value.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
-    assert exc_info.value.code == "ResponseAlreadyComputedError"
+# TODO: remove
```
AndreaFrancis, line 201: Is it going to be removed in this PR?

AndreaFrancis (approved):
Great! Much simpler and will reduce the number of generated jobs in backfill

libs/libcommon/src/libcommon/constants.py
severo: indeed. fixed

services/worker/tests/test_job_manager.py
severo: done, thanks

PR #2482, opened 2024-02-21
Switch to duckdb 010 in search/filter service

AndreaFrancis:
Part of https://github.com/huggingface/datasets-server/issues/1914
Following https://github.com/huggingface/datasets-server/pull/2463

**WARNING:** This PR should be merged only once all "split-duckdb-index-010" entries have been computed. And workers should be scaled down to 0 before deploying.
Changes done:
- Migration to delete all records with split-duckdb-index
- Migration to rename split-duckdb-index-010 to split-duckdb-index
- Update duckdb version in search and filter
- Change target_revision to refs/convert/duckdb in search and filter
- Change the downloading folder for indexes files to avoid cached files in search/filter (It will make easy to do a bulk delete in the future)
- Switch split-duckdb-index to split-duckdb-index-010 job runner
- Increase version of config-duckdb-index-size and split-is-valid to refresh new split-duckdb-index children steps


jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -330,4 +330,13 @@ def get_migrations(self) -> list[Migration]:
                 version="20240221160700",
                 description="merge 'split-first-rows-from-streaming' and 'split-first-rows-from-parquet' responses to 'split-first-rows'",
             ),
+            CacheDeletionMigration(
+                cache_kind="split-duckdb-index",
+                version="20240221134400",
```
severo, line 335: could you increase this value a bit, so that it's higher than the previous migration? 🙏  Our PRs were created concurrently

libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -585,17 +585,10 @@ def parse_id(id: str) -> tuple[str, str, Optional[str], Optional[str], str]:
...
-        "triggered_by": "config-parquet-metadata",
-        "job_runner_version": 1,
-        "difficulty": 70,
-        "bonus_difficulty_if_dataset_is_big": 20,
-    },
     "config-duckdb-index-size": {
         "input_type": "config",
-        "triggered_by": "split-duckdb-index",
-        "job_runner_version": 1,
+        "triggered_by": ["split-duckdb-index"],
```
severo, line 590:
```suggestion
        "triggered_by": "split-duckdb-index",
```

a string is also valid. I find it more readable when we have only one entry

services/search/src/search/routes/filter.py
```diff
@@ -21,7 +21,7 @@
     get_request_parameter_length,
     get_request_parameter_offset,
 )
-from libapi.response import ROW_IDX_COLUMN, create_response
```
severo, line 24: what was it used for? why do we remove it?

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -330,4 +330,13 @@ def get_migrations(self) -> list[Migration]:
...
+            ),
+            CacheRenamingMigration(
+                cache_kind="split-duckdb-index-010",
+                new_cache_kind="split-duckdb-index",
+                version="20240221134500",
+            ),
```
severo, line 341: note that we should also do the same with the jobs, if any, or the workers will break when they pick one.

severo (approved):

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: Thanks! I forgot it

services/search/src/search/routes/filter.py
AndreaFrancis: It was used in execute_filter_query (Line 188). Since all split-duckdb-index have been computed, all cache entries now have the features as part of the content, so, no need to add it to the list of columns.

jobs/mongodb_migration/src/mongodb_migration/collector.py
AndreaFrancis: Done

libs/libcommon/src/libcommon/processing_graph.py
AndreaFrancis: Done, thanks!

severo (approved):

albertvillanova (approved):
Thanks for taking care of this!

services/search/src/search/routes/filter.py
```diff
@@ -188,7 +188,7 @@ def execute_filter_query(
 ) -> tuple[int, pa.Table]:
     with duckdb_connect(database=index_file_location) as con:
         filter_query = FILTER_QUERY.format(
-            columns="*" if columns == "*" else ",".join([f'"{column}"' for column in [ROW_IDX_COLUMN] + columns]),
+            columns="*" if columns == "*" else ",".join([f'"{column}"' for column in columns]),
```
albertvillanova, line 191: Thanks for this fix.

I guess this bug was introduced by this PR:
- #2347
- See my comment: https://github.com/huggingface/datasets-server/pull/2347#discussion_r1469701184

IMHO, I find misleading to use the word "features" with 2 different meanings:
- the meaning I was supporting in my comment to that PR: "features" refers to the _original_ dataset features, those handled exclusively by the `datasets` library
- now we use "features" with a different meaning: all the columns stored in the DuckDB index file, which contain the original dataset features, plus any additional auxiliary columns required by datasets-server (like, "__hf_index_id", and any other additional column we may need in the future)

CC: @lhoestq for visibility

services/search/tests/routes/test_filter.py
```diff
@@ -81,7 +81,7 @@ def test_execute_filter_query(index_file_location: str, columns: Union[list[str]
     assert num_rows_total == 2
     expected = pa.Table.from_pydict({"__hf_index_id": [3], "name": ["Simone"], "gender": ["female"], "age": [30]})
     if isinstance(columns, list):
-        for column in set(expected.column_names) - set([ROW_IDX_COLUMN] + columns):
+        for column in set(expected.column_names) - set(columns):
             expected = expected.drop(column)
     assert pa_table == expected
```
lhoestq, line 86: I think the expected Table should have `__hf_index_id ` otherwise we can't format the filter output to have the right row_idx

severo (changes requested):
agree with @lhoestq about the missing column

AndreaFrancis:
@lhoestq @albertvillanova @severo  sorry, I should have explained better the problem regarding filter change.
Previously with duckdb version 0.8.1, we used to execute filter with these params:
Columns retrieved from "split-duckdb-index": `['image', 'text', 'rating', '__hf_index_id']`
Query generated at https://github.com/huggingface/datasets-server/blob/main/services/search/src/search/routes/filter.py#L198 
```
SELECT "__hf_index_id","image","text","rating","__hf_index_id"
    FROM data
    WHERE rating=3
    LIMIT 100
    OFFSET 0
```

See that `__hf_index_id` is used twice, in the 0.8.1 version there was a bug/feature that let us execute this kind of queries without problem, but now with version 0.10.0 it gives this error:
`pyarrow.lib.ArrowInvalid: Schema at index 0 was different
`

See this code snippet running 0.8.1 vs 0.10.0

```
import duckdb
con = duckdb.connect()
con.sql("CREATE TABLE t1 AS SELECT 42 AS i, 84 AS j;")
result = con.sql("SELECT i, i FROM t1;").arrow() # select twice the same column
print(result.to_pydict())

```

**0.8.1 result:**
`{'i': [42]}`

**0.10.0 result:**
```
    result = con.sql("SELECT i, i FROM t1;").arrow()
  File "pyarrow/table.pxi", line 4057, in pyarrow.lib.Table.from_batches
  File "pyarrow/error.pxi", line 154, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 91, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Schema at index 0 was different: 
i: int32
i: int32
vs
i: int32
i_1: int32

```

lhoestq:
Ah yes that makes sense ! Indeed `ROW_IDX_COLUMN` is already part of the feature from the `split-duckdb-index` job

lhoestq (approved):
LGTM then :)

severo (approved):
OK, thanks for the explanation

PR #2493, opened 2024-02-22
set updated_at field to the same value for all steps of a dataset

severo:
After the migrations:
- '_20240221103200_cache_merge_config_split_names'
- '_20240221160700_cache_merge_split_first_rows'

the updated_at field is not accurate anymore, leading to backfilling a lot of them (in the daily cronjob), leading to unnecessary jobs.

To fix that, we take the updated_at value for the root step, and set it to all the steps for the same dataset.

See https://huggingface.slack.com/archives/C04L6P8KNQ5/p1708640247358519?thread_ts=1708637658.767199&cid=C04L6P8KNQ5 (internal)

### DIFF
--- a/jobs/mongodb_migration/src/mongodb_migration/collector.py
+++ b/jobs/mongodb_migration/src/mongodb_migration/collector.py
@@ -72,6 +72,9 @@ from mongodb_migration.migrations._20240221103200_cache_merge_config_split_names
 from mongodb_migration.migrations._20240221160700_cache_merge_split_first_rows import (
     MigrationMergeSplitFirstRowsResponses,
 )
+from mongodb_migration.migrations._20240221160800_cache_set_updated_at_to_root_step import (
+    MigrationSetUpdatedAtToOldestStep,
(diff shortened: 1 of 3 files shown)


severo:
The comparison of dates, to decide if a step should be backfilled, is here:

https://github.com/huggingface/datasets-server/blob/2982e1ca2c42bd7f34a5bf7683e099e2cbfebbb7/libs/libcommon/src/libcommon/state.py#L95-L98

from:

https://github.com/huggingface/datasets-server/blob/2982e1ca2c42bd7f34a5bf7683e099e2cbfebbb7/libs/libcommon/src/libcommon/orchestrator.py#L629-L640

So, in case of equality, it's `False`, ie. it's not considered for backfilling.

AndreaFrancis (approved):
LGTM, the only risk I find is that it will be unclear to refresh specific steps in case we want to do it by any issue/bug at particular dates. But I think we can do it by the dataset-config-names step.

PR #2518, opened 2024-02-27
Tweak the production resources

severo:
See https://www.notion.so/huggingface2/Infrastructure-b4fd07f015e04a84a41ec6472c8a0ff5 and https://huggingface.slack.com/archives/C04L6P8KNQ5/p1702553107305559 (internal links)

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -266,11 +266,11 @@ storageAdmin:
   replicas: 1
   resources:
     requests:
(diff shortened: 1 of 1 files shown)


severo:
Finally, I reverted, because the infra will soon change, and they will have tools to auto-tweak these values and have alerts when the values are not optimal. Let's keep as is for now. Just removing obsolete code;

PR #2519, opened 2024-02-27
backfill retryable errors every 20 minutes

### LINKED ISSUE #2439: "Errors to retry" only retried once a day
It's strange: `JobManagerCrashedError` is listed in the errors to retry, but I witnessed a dataset where it was not retried (`failed_runs: 0`, no jobs).

See discussion opened here: https://huggingface.co/datasets/Samuel-Martineau/kryptik/discussions/1

Capture d’écran 2024-02-12 à 12 53 36

severo:
fixes #2439.

For example, if a Parquet file commit to the Hub fails, we want to retry quickly (but not immediately, in case the Hub has an issue). So... checking and retrying every 20 minutes seems a good compromise.

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -192,13 +192,30 @@ backfill:
     level: "debug"
   action: "backfill"
   schedule: "45 21 * * *"
-  # every four hours
+  # every day
   nodeSelector:
     role-datasets-server: "true"
   resources:
     requests:
       cpu: 1
-      memory: "4Gi"
+      memory: "1Gi"
+    limits:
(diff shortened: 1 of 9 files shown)


lhoestq (approved):
LGTM, though let's monitor the JobManagerCrashedError we retry and try to fix them. They can take a lot of time / resources

severo:
> LGTM, though let's monitor the JobManagerCrashedError we retry and try to fix them. They can take a lot of time / resources

Sure, it's the more crucial one (even if we want to reduce all the unexpected errors as well).

PR #2522, opened 2024-02-27
Fix backfill wrong difficulties, causing OOMs

lhoestq:
The backfill was creating duckdb indexing jobs for big datasets without adjusting the difficulty.

Because of that the medium workers were processing them and OOMing and every retry.

### DIFF
--- a/libs/libcommon/src/libcommon/orchestrator.py
+++ b/libs/libcommon/src/libcommon/orchestrator.py
@@ -4,6 +4,7 @@
 import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from functools import lru_cache
 from http import HTTPStatus
 from typing import Optional, Union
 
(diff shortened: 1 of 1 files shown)


libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -689,9 +690,21 @@ def _create_plan(self) -> None:
...
+                return False
+            else:
+                return num_bytes > self.processing_graph.min_bytes_for_bonus_difficulty
+
         for artifact_state in artifact_states:
             valid_pending_jobs_df = artifact_state.job_state.valid_pending_jobs_df
             if valid_pending_jobs_df.empty:
+                difficulty = artifact_state.processing_step.difficulty
+                if isinstance(artifact_state.config, str) and is_big(config=artifact_state.config):
+                    difficulty += artifact_state.processing_step.bonus_difficulty_if_dataset_is_big
```
severo, line 707: maybe limit to 100

severo:
```
            # increase difficulty according to number of failed runs
            difficulty = min(DEFAULT_DIFFICULTY_MAX, difficulty + self.failed_runs * DIFFICULTY_BONUS_BY_FAILED_RUNS)
```

^might be factorized too

lhoestq: I'm changing the logic in AfterJobPlan in https://github.com/huggingface/datasets-server/pull/2523 so no need to factorize imo

severo (approved):

PR #2524, opened 2024-02-27
Compute croissant in workers

### LINKED ISSUE #2521: Move /croissant to the cache
As it's called on every dataset page, we should cache it

severo:
The croissant metadata is computed by workers and stored in the cache

When served by the API service, and if `?full=false` is passed in the query, the columns are limited to 1,000 in the response.

fix #2521


severo:
wait for https://github.com/huggingface/datasets-server/issues/2624 (itself blocked by https://github.com/huggingface/moon-landing/pull/9106, internal)

services/worker/src/worker/dtos.py
```diff
@@ -251,6 +251,12 @@ class DatasetModalitiesResponse(TypedDict):
     modalities: list[DatasetModality]
 
 
+class DatasetCroissantCrumbsResponse(TypedDict):
+    croissant_crumbs: Mapping[str, Any]
+    truncated_configs: bool
+    partial: bool
```
lhoestq, line 257: The front-end expects the `croissant_crumbs` directly instead of this nested object no ?

lhoestq (commented):
Note that the `/croissant-crumbs` will fail until the `dataset-croissant-crumbs` jobs are finished, but it should be quite fast imo

Just one comment:

severo:
> Note that the /croissant-crumbs will fail until the dataset-croissant-crumbs jobs are finished, but it should be quite fast IMO

you're right, we could do it in two steps, but tbh, in the worst case, we would not include the croissant fields in the dataset pages for 24h, which is acceptable.

lhoestq:
yea I agree

services/worker/src/worker/dtos.py
severo: yes, you're right! I'll fix it

severo: should be fixed in https://github.com/huggingface/datasets-server/pull/2524/commits/6f33c4ae11cb334b0a844aa6427b664a633b6d7b

lhoestq (approved):
LGTM :)

Issue #2095, opened 2023-11-10
Scale automatically the number of pods for the workers


severo:
Ideally, scale the number of light, medium, and heavy workers, depending on the contents of the jobs queue.

At least, or in addition, have metrics on the occupation/idleness of the worker pods + alerts when we have too many, or too few of them.

severo:
To do so, we have to expose the three queue sizes (light, medium, heavy) on a prometheus custom exporter to pilot the kubernetes autoscaler.

severo:
Practically: add the number of waiting jobs for light (difficulty <=50), medium (difficulty in ]50, 80]) and heavy (difficulty > 80) to /metrics in services/admin.

severo:
Done!

severo:
See #2537 and #2538 in particular

PR #2526, opened 2024-02-28
add metrics on jobs count per worker type

severo:
See #2095.

Add metrics `worker_size_jobs_count{pid="10",worker_size="heavy"}: 12` to prometheus (value for worker_size: `heavy`, `medium` and `light`).

It will be used to autoscale the workers 

Also:
- difficulty_min is now excluded (and 0 is not a valid difficulty anymore)
- in prod, medium and heavy workers will now only process jobs for their range, not smaller ones


lhoestq:
> in prod, medium and heavy workers will now only process jobs for their range, not smaller ones

Why this change ? I feel like it's ok for heavy workers to do easier jobs instead of doing nothing if there are no hard jobs

severo:
> Why this change ? I feel like it's ok for heavy workers to do easier jobs instead of doing nothing if there are no hard jobs

the issue is that it's not true: they can do easier jobs, even if there are hard jobs.

lhoestq:
I see ! makes sense

libs/libcommon/src/libcommon/queue.py
```diff
@@ -600,7 +647,7 @@ def _get_next_waiting_job_for_priority(
         if job_types_only:
             filters["type__in"] = job_types_only
         if difficulty_min is not None and difficulty_min > DEFAULT_DIFFICULTY_MIN:
-            filters["difficulty__gte"] = difficulty_min
+            filters["difficulty__gt"] = difficulty_min
```
lhoestq, line 650:
```suggestion
            filters["difficulty__gte"] = difficulty_min
```

libs/libcommon/src/libcommon/queue.py
```diff
@@ -245,34 +248,78 @@ class JobTotalMetricDocument(Document):
...
+    jobs_count = IntField(required=True, default=0)
+    created_at = DateTimeField(default=get_datetime)
+
+    @staticmethod
+    def get_worker_size(difficulty: int) -> str:
+        if difficulty <= 40:
+            return WorkerSize.light
+        if difficulty <= 70:
+            return WorkerSize.medium
+        return WorkerSize.heavy
```
lhoestq, line 283: Fix typing by returning strings instead of type WorkerSize

```suggestion
    @staticmethod
    def get_worker_size(difficulty: int) -> str:
        if difficulty <= 40:
            return WorkerSize.light.value
        if difficulty <= 70:
            return WorkerSize.medium.value
        return WorkerSize.heavy.value
```

libs/libcommon/src/libcommon/queue.py
```diff
@@ -1010,6 +1060,19 @@ def get_jobs_total_by_type_and_status(self) -> JobsTotalByTypeAndStatus:
...
+        """Count the number of jobs by worker size.
+
+        Returns:
+            an object with the total of jobs by worker size.
+            Keys are worker_size, and values are the jobs count.
+        """
+        return {
+            WorkerSize.heavy.name: JobDocument.objects(difficulty__lte=100, difficulty__gt=70).count(),
+            WorkerSize.medium.name: JobDocument.objects(difficulty__lte=70, difficulty__gt=40).count(),
+            WorkerSize.light.name: JobDocument.objects(difficulty__lte=40, difficulty__gt=0).count(),
```
lhoestq, line 1073: just cor consistency

```suggestion
            WorkerSize.heavy.value: JobDocument.objects(difficulty__lte=100, difficulty__gt=70).count(),
            WorkerSize.medium.value: JobDocument.objects(difficulty__lte=70, difficulty__gt=40).count(),
            WorkerSize.light.value: JobDocument.objects(difficulty__lte=40, difficulty__gt=0).count(),
```

lhoestq (approved):
LGTM :)

PR #2527, opened 2024-02-28
Don't unquote path_in_repo

lhoestq:
Support paths containing `%`


lhoestq:
hmm maybe check that we don't get https://github.com/huggingface/datasets-server/issues/1433 back

PR #2531, opened 2024-02-29
Restore missing triggers between steps

### LINKED ISSUE #2530: Missing triggers between steps
possibly due to https://github.com/huggingface/datasets-server/pull/2489

We get errors like:
- https://huggingface.co/datasets/jmc255/aphantasia_drawing_dataset/discussions/1
- https://huggingface.co/datasets/yzeng58/CoBSAT/discussions/1

where the first step `dataset-config-names` gives an error, and no other entry is created, leading to the API service responding `404 - Error code:   ResponseNotFound`, while it should tell that the issue is we don't support script-based datasets anymore.

severo:
fixes #2530

it:
- reverts #2489
- adds the missing triggering step
- adds a check when creating the processing graph, to ensure all the config- and dataset-level steps have at least one triggering step of the same level, to handle the case where we couldn't get the configs or the splits.

### DIFF
--- a/libs/libcommon/src/libcommon/processing_graph.py
+++ b/libs/libcommon/src/libcommon/processing_graph.py
@@ -40,9 +40,10 @@ def guard_int(x: Any) -> int:
 
 
 def check_one_of_parents_is_same_or_higher_level(processing_graph: ProcessingGraph) -> None:
+    first_steps = processing_graph.get_first_processing_steps()
     for step in processing_graph.get_topologically_ordered_processing_steps():
         parents = processing_graph.get_parents(step.name)
-        if len(parents) > 0:
+        if parents:
             parent_input_types = [p.input_type for p in parents]
             if step.input_type == "dataset":
                 if "dataset" not in parent_input_types:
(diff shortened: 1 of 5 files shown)


severo:
After merging, the first full backfill should fix the issues.

libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -39,6 +39,24 @@ def guard_int(x: Any) -> int:
     raise ValueError(f"Invalid int: {x}")
 
 
+def check_one_of_parents_is_same_or_higher_level(processing_graph: ProcessingGraph) -> None:
+    for step in processing_graph.get_topologically_ordered_processing_steps():
+        parents = processing_graph.get_parents(step.name)
+        if len(parents) > 0:
```
AndreaFrancis, line 45: Is it worth validating that steps without parents are part of _first_processing_steps?

severo: well, it does not cost much, I'll add it

PR #2534, opened 2024-02-29
use a generic LongStepProfiler

severo:
follow-up to #2469 and #2533


services/worker/src/worker/loop.py
```diff
@@ -124,12 +123,7 @@ def process_next_job(self) -> bool:
                 logging.debug(e)
                 return False
 
-        buckets = (
-            DESCRIPTIVE_STATISTICS_PROMETHEUS_HISTOGRAM_BUCKETS
-            if job_info["type"] == "split-descriptive-statistics"
-            else None
-        )
-        with StepProfiler("loop", "run_job", job_info["type"], buckets=buckets):
+        with LongStepProfiler("loop", "run_job", job_info["type"]):
```
polinaeterna, line 126: so we do this for all workers now, right?

polinaeterna (approved):
thank you! now i see what i did wrong heh

services/worker/src/worker/loop.py
severo: yes

PR #2543, opened 2024-03-04
Truncate binary for major tom

lhoestq:
Fix viewer for https://huggingface.co/datasets/Major-TOM/Core-S2L2A

The dataset has one row per row group because each row contains multiple binary columns with a lot of data.

I fixed this by truncating binary data iteratively on row groups to not fill up the RAM

For now I only enable this logic on Major-TOM datasets to experiment

### DIFF
--- a/libs/libcommon/src/libcommon/parquet_utils.py
+++ b/libs/libcommon/src/libcommon/parquet_utils.py
@@ -303,12 +303,15 @@ class ParquetIndexWithMetadata:
 
         with StepProfiler(method="parquet_index_with_metadata.query_truncated_binary", step="read the row groups"):
             # This is a simple heuristic of how much we need to truncate binary data
-            max_binary_length = max(int(
-                (self.max_arrow_data_in_memory - in_memory_max_non_binary_size)
-                / (last_row_group_id + 1 - first_row_group_id)
-                / len(binary_columns)
-                / 2  # we divide more in case the row groups are not evenly distributed
-            ), 20)  # we use a minimum length to not end up with too empty cells
+            max_binary_length = max(
+                int(
+                    (self.max_arrow_data_in_memory - in_memory_max_non_binary_size)
+                    / (last_row_group_id + 1 - first_row_group_id)
+                    / len(binary_columns)
+                    / 2  # we divide more in case the row groups are not evenly distributed
+                ),
+                20,
+            )  # we use a minimum length to not end up with too empty cells
             try:
                 pa_tables: list[pa.Table] = []
                 truncated_columns: set[str] = set()
--- a/libs/libcommon/src/libcommon/parquet_utils.py
+++ b/libs/libcommon/src/libcommon/parquet_utils.py
@@ -1,9 +1,10 @@
 import asyncio
 import logging
 import os
+from collections.abc import Iterable
 from dataclasses import dataclass, field
 from functools import lru_cache
-from typing import Iterable, Literal, Optional, TypedDict, Union
+from typing import Literal, Optional, TypedDict, Union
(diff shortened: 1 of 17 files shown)


libs/libcommon/src/libcommon/parquet_utils.py
```diff
@@ -163,6 +185,142 @@ def __post_init__(self) -> None:
...
+        Returns:
+            `pa.Table`: The requested rows.
+            `list[strl]: List of truncated columns.
+        """
+        all_columns = set(self.features)
+        binary_columns = set(column for column, feature in self.features.items() if feature == Value("binary"))
+        if not binary_columns:
+            return self.query(offset=offset, length=length), []
+        with StepProfiler(
+            method="parquet_index_with_metadata.query", step="get the parquet files than contain the requested rows"
```
AndreaFrancis, line 215:
```suggestion
            method="parquet_index_with_metadata.query_truncated_binary", step="get the parquet files than contain the requested rows"
```

libs/libcommon/src/libcommon/parquet_utils.py
```diff
@@ -163,6 +185,142 @@ def __post_init__(self) -> None:
...
+            )
+            parquet_offset = (
+                offset - parquet_file_offsets[first_parquet_file_id - 1] if first_parquet_file_id > 0 else offset
+            )
+            urls = self.parquet_files_urls[first_parquet_file_id : last_parquet_file_id + 1]  # noqa: E203
+            metadata_paths = self.metadata_paths[first_parquet_file_id : last_parquet_file_id + 1]  # noqa: E203
+            num_bytes = self.num_bytes[first_parquet_file_id : last_parquet_file_id + 1]  # noqa: E203
+
+        with StepProfiler(
+            method="parquet_index_with_metadata.query", step="load the remote parquet files using metadata from disk"
```
AndreaFrancis, line 233:
```suggestion
            method="parquet_index_with_metadata.query_truncated_binary", step="load the remote parquet files using metadata from disk"
```

AndreaFrancis (approved):
LGTM

severo:
Super useful.

Three comments:
- we might factorize code between [`query`](https://github.com/huggingface/datasets-server/blob/87c224fa218420d69b20cd28d7befee0daf3c236/libs/libcommon/src/libcommon/parquet_utils.py#L330) and [`query_truncated_binary`](https://github.com/huggingface/datasets-server/blob/87c224fa218420d69b20cd28d7befee0daf3c236/libs/libcommon/src/libcommon/parquet_utils.py#L188) (basically, every code block surrounded by a "StepProfiler" could be a different method/function
- add unit tests for the new code
- instead of computing [`max_binary_length`](https://github.com/huggingface/datasets-server/blob/87c224fa218420d69b20cd28d7befee0daf3c236/libs/libcommon/src/libcommon/parquet_utils.py#L306), why not returning a very small number of bytes? I imagine that truncated binary is unusable anyway, so... maybe just returning the first 50 bytes is enough.

lhoestq:
Yes sounds good, and then we can enable this logic for any dataset with binary features

severo:
do you want to take care of it?

severo:
Note that https://huggingface.co/datasets/Major-TOM/Core-S2L1C/viewer/default/train?p=1 still does not work:

```
Rows from parquet row groups are too big to be read: 1.02 GiB (max=286.10 MiB)

Error code:   TooBigContentError
```

severo:
and https://huggingface.co/datasets/Major-TOM/Core-S2L2A/viewer/default/train?p=1 is suuuuper long

PR #2544, opened 2024-03-04
Set max workers for `retry_and_validate_get_parquet_file_and_size` in `config-parquet-and-info` to 4

polinaeterna:
I assume the fact that we don't restrict the max number of workers (which results in [32 workers](https://github.com/tqdm/tqdm/blob/master/tqdm/contrib/concurrent.py#L44)* from `tqdm.contrib.concurrent.thread_map` which we use at this step) might result in (likely) OOM for `bigcode/the-stack-v2`. Since there are about 1k parquet files in this dataset.

\+ add logs to see where exactly that happens to support this hypothesis.

*UPD: i'm not sure here because in theory Kubernetes should restrict the max number of accessible cpu's to 8 https://github.com/huggingface/datasets-server/blob/main/chart/env/prod.yaml#L452
But when I monitor in `htop` I see that all cpu's are loaded at this moment.

### DIFF
--- a/services/worker/src/worker/job_runners/config/parquet_and_info.py
+++ b/services/worker/src/worker/job_runners/config/parquet_and_info.py
@@ -656,6 +656,7 @@ def fill_builder_info(
                 data_files[split],
                 unit="pq",
                 disable=True,
+                max_workers=8,
             )
             parquet_files, sizes = zip(*parquet_files_and_sizes)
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
LGTM !

Reducing the number of workers can also make it a bit slower, but it should be fine

PR #2551, opened 2024-03-06
Update openapi specs according to changes in stats #2453

polinaeterna:
Some values of numerical results might be null now (rare cases when all values in column are null)

See https://github.com/huggingface/datasets-server/pull/2453

### DIFF
--- a/docs/source/openapi.json
+++ b/docs/source/openapi.json
@@ -1081,7 +1081,7 @@
       },
       "ColumnType": {
         "type": "string",
-        "enum": ["float", "int", "class_label", "string_label", "string_text", "bool"]
+        "enum": ["float", "int", "class_label", "string_label", "string_text", "bool", "list"]
       },
       "Histogram": {
         "type": "object",
@@ -1105,13 +1105,7 @@
         "type": "object",
         "required": [
           "nan_count",
-          "nan_proportion",
-          "min",
-          "max",
-          "mean",
-          "median",
-          "std",
-          "histogram"
+          "nan_proportion"
         ],
         "properties": {
           "nan_count": {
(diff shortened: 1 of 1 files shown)


docs/source/openapi.json
```diff
@@ -1105,13 +1105,7 @@
...
         "required": [
           "nan_count",
-          "nan_proportion",
-          "min",
-          "max",
-          "mean",
-          "median",
-          "std",
-          "histogram"
+          "nan_proportion"
```
polinaeterna, line 1108: actually not sure about this change, all keys should be in response, just some of them might be null. should i revert?

lhoestq: I just checked and if the fields are always in the response (regardless of whether they're null or not) then they can can be marked as "required"

So you can revert this change imo

lhoestq (approved):

docs/source/openapi.json
```diff
@@ -1121,23 +1121,21 @@
             "type": "number"
           },
           "min": {
-            "type": "number"
+            "type": { "anyOf": [{ "type": "number" }, { "type": "null" }] }
```
severo, line 1124: Can the stats by partially null (eg "min", "max"), and partially non-null (eg "mean" and "median")? Or is it always: all null, or all non-null?

polinaeterna: it's either all null (if all values in column are null) or all non-null

severo: OK. So... I think it would be easier to handle on the client's side if we change the type a bit: return `null` or an object with all the stats as "number" (ie: no individually optional number). Do you think it would make sense?

polinaeterna: you mean return `null` as a whole response instead of the object with only two filled fields (`nan_count`, `nan_proportion`)?

severo: yes. Because otherwise, the client has to assume that we can have min=null and max=10, which is not true.

polinaeterna: right... yes, that sounds very reasonable. but i'm not sure it is aligned with the server logic.
we have:
```
class NumericalStatisticsItem(TypedDict):
    nan_count: int
    nan_proportion: float
    min: Optional[float]  # might be None in very rare cases when the whole column is only None values
    max: Optional[float]
    mean: Optional[float]
    median: Optional[float]
    std: Optional[float]
    histogram: Optional[Histogram]


class CategoricalStatisticsItem(TypedDict):
    nan_count: int
    nan_proportion: float
    no_label_count: int
    no_label_proportion: float
    n_unique: int
    frequencies: dict[str, int]


class BoolStatisticsItem(TypedDict):
    nan_count: int
    nan_proportion: float
    frequencies: dict[str, int]


SupportedStatistics = Union[NumericalStatisticsItem, CategoricalStatisticsItem, BoolStatisticsItem]

class StatisticsPerColumnItem(TypedDict):
    column_name: str
    column_type: ColumnType
    column_statistics: SupportedStatistics
```
In response `column_statistics` is always non-null, does it make sense to make it nullable? it would mean then that it can be null only in a specific case - for numerical columns with all-None values. `CategoricalStatisticsItem` and `BoolStatisticsItem` don't do the same because they have all other values set to some non-null values.

btw i also realized now that front doesn't show anything for such numerical cases, while i think it should as it does for categories and bool:
https://huggingface.co/datasets/polinaeterna/only_null

severo: -> https://github.com/huggingface/moon-landing/issues/9216 (internal)

severo: hmmm, I think I was wrong. It's OK to have all the values as `Optional[float]` / `Optional[Histogram]` after all

Issue #1953, opened 2023-10-09
switch hardcoded `MAX_NUM_STRING_LABELS` from 30 to some relative number (n_unique / n_samples)


polinaeterna:

polinaeterna:
unstale

severo:
btw @polinaeterna: setting a priority (P0/P1/P2) disables the stale bot. It's what I did last week, since we want to implement this feature.

polinaeterna:
I just though that ideally we would have a way to configure the viewer somehow, for example in readme's yaml. 
I imagine we can have an additional `"viewer"` or smth key where users can set some configurable parameters. For example, if a column is string or category, which subset to show by default (not always we want to show the one first by alphabet).

Yes, for categories we can set `ClassLabel` column in `dataset_info` but this changes the underlying data type from string to int (@lhoestq am i right here?) which might be unwanted.

wdyt @huggingface/datasets-server ?

lhoestq:
Maybe we should show the default subset first ? It can be marked using `default: true` in YAML.

> Yes, for categories we can set ClassLabel column in dataset_info but this changes the underlying data type from string to int (@lhoestq am i right here?) which might be unwanted.

Yes correct, and it also requires to specify the types of all the columns which is not ideal. Not sure yet if such configuration would be used a lot though

polinaeterna:
> Maybe we should show the default subset first ? It can be marked using default: true in YAML.

also an option, we have an issue for that https://github.com/huggingface/datasets-server/issues/2541
should it be changed on the front-end or on our side in datasets-server? if the latter, i guess just the order of configs in `dataset-config-names` step should be changed?

severo:
Done.

PR #2553, opened 2024-03-07
Determine if a string column is string or category by proportion of unique values

polinaeterna:
...instead of by an absolute value. Will solve https://github.com/huggingface/datasets-server/issues/1953
Set to 0.2.

UPD: i also added the maximum value of categories to 1000 (for example, there are ~700 languages in the-stack-v2). Yes, it makes no sense for the viewer to display 1k labels but i think it should be fixed in the frontend - for example by displaying only top-N categories and the number of unique values. Because computing histograms for classes also doesn't make any sense, especially if column is language  code and they are all of the same length. 
Also if there are <= num_histogram_bins (currently 10) - treat as category independently of the proportion value because the viewer will display this well for sure.

Ideally this should be fixed by manual configuration (hope to start working on this soon!) and for now i think it's okay like this.


services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -495,7 +499,7 @@ def _compute_statistics(
         # TODO: count n_unique only on the first parquet file?
         nan_count, nan_proportion = nan_count_proportion(data, column_name, n_samples)
         n_unique = data[column_name].n_unique()
-        if n_unique <= MAX_NUM_STRING_LABELS:  # TODO: make relative
+        if n_unique / n_samples < MAX_PROPORTION_STRING_LABELS:
```
severo, line 502: nit: let's keep `<=` instead of `<`?

services/worker/tests/job_runners/split/test_descriptive_statistics.py
```diff
@@ -362,7 +362,7 @@ def count_expected_statistics_for_string_column(column: pd.Series) -> dict:  # t
     nan_count = column.isna().sum()
     value_counts = column.value_counts(dropna=True).to_dict()
     n_unique = len(value_counts)
-    if n_unique <= MAX_NUM_STRING_LABELS:
+    if n_unique / n_samples <= MAX_PROPORTION_STRING_LABELS:
```
severo, line 365: and note we use `<=` here

severo (approved):

severo:
should we recompute some datasets? I think it's not necessary, right?

polinaeterna:
@severo i think we can let this happen naturally, wdyt? :D maybe recompute manually some popular datasets that we know require this change

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: did this https://github.com/huggingface/datasets-server/pull/2553/commits/e7dc6f44cdd33027d94a0b03e60308d50f3aad3c
and also switched 0.3 to 0.15, i feel like it's more realistic

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -42,9 +42,13 @@
...
-# for example, 16 unique values in 100 samples -> strings
-# 16 unique values in 1000 samples -> category
-MAX_PROPORTION_STRING_LABELS = 0.15
+# for example, 21 unique values in 100 samples -> strings
+# 200 unique values in 1000 samples -> category
+# 201 unique values in 1000 samples -> string
+MAX_PROPORTION_STRING_LABELS = 0.2
+# if there are more than MAX_NUM_STRING_LABELS unique strings, consider column to be a string
+# even if proportion of unique values is lower than MAX_PROPORTION_STRING_LABELS
+MAX_NUM_STRING_LABELS = 1000
```
severo, line 51: OK, good idea

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -498,7 +502,11 @@ def _compute_statistics(
         logging.info(f"Compute statistics for string column {column_name} with polars. ")
         nan_count, nan_proportion = nan_count_proportion(data, column_name, n_samples)
         n_unique = data[column_name].n_unique()
-        if n_unique / n_samples <= MAX_PROPORTION_STRING_LABELS:
+        if (
+            n_unique / n_samples <= MAX_PROPORTION_STRING_LABELS
+            and n_unique <= MAX_NUM_STRING_LABELS
+            or n_unique <= n_bins
```
severo, line 508: maybe add parentheses to make it clearer as `and` and `or` are mixed

polinaeterna: [b4c1c46](https://github.com/huggingface/datasets-server/pull/2553/commits/b4c1c46a8aef1d692d27a29847cf4ff4268dd8ba)

Issue #2541, opened 2024-03-04
Display default config first in the viewer


polinaeterna:
Would be nice to have an option to choose which data to show first because sometimes dataset starts with ugly data lol. 
example: [bigcode/the-stack-v2](https://huggingface.co/datasets/bigcode/the-stack-v2). (yes, there is only one config with all the data for now but in theory it should be each folder inside `data/` is a config and the first folder should not appear as first in the viewer) 

The most straightforward thing to do is to display config marked as `default` first. I don't know if that require changes here on on the Hub's side?
cc @anton-l

Issue #2552, opened 2024-03-07
 `config-parquet-and-info` crashes with OOM at pq files validation, presumably when there are a lot of files


polinaeterna:
[bigcode/the-stack-v2](https://huggingface.co/datasets/bigcode/the-stack-v2) keeps failing at config-parquet-and-info step and seems that the problem is that there are too many parquet files to validate (there are about 1k of them). 

It [starts](https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/parquet_and_info.py#L645) [ `retry_and_validate_get_parquet_file_and_size`](https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/config/parquet_and_info.py#L651): 

![image](https://github.com/huggingface/datasets-server/assets/16348744/a88733c6-145a-4bae-b6b3-8980e33b5d00)

and then fails with OOM in about a minute:

![image](https://github.com/huggingface/datasets-server/assets/16348744/bc8d7221-fa0a-4ff1-84c1-513b5e1ebc6e)

See also https://github.com/huggingface/datasets-server/pull/2544 which didn't help.

Note that [bigcode/the-stack-v2-dedup](https://huggingface.co/datasets/bigcode/the-stack-v2-dedup) which is about 2-4 times smaller works at this step.

cc @anton-l

severo:
Could we refresh other datasets with the same issue (crash in config-parquet-and-info)?

Let me know if you want me to take care of it

polinaeterna:
@severo hm i don't know if we can easily find datasets with this specific issue because it happens when there are a lot of parquet files to read the metadata from (for example, the-stack-v2 has 1k parquet files) and it doesn't raise any specific error but `missing heartbeats`. and we don't have that many datasets like this. for comparison - [the-stack-v2-dedup](https://huggingface.co/datasets/bigcode/the-stack-v2-dedup) is ~3x smaller but still there are many files - assume 300 - and it worked fine.

but if we have resources we can try to refresh all `missing heartbeats` and see maybe something will work, why not

severo:
as of today:
<img width="310" alt="Capture d’écran 2024-03-12 à 15 33 06" src="https://github.com/huggingface/datasets-server/assets/1676121/9f18b269-2b52-4cc5-83f1-c017b4704817">

If it will fix 95% of them, I would be ok to refresh them. If it will fix 5% of them, I think it's not worth it. What do you think?

[...]

PR #2554, opened 2024-03-07
Default config first

lhoestq:
Show the default config first in the Viewer.

I did this by ordering the configs in `dataset-config-names` to have the default config in first position

fix https://github.com/huggingface/datasets-server/issues/2541

No need to recompute all the entries, it's fine to apply this to new / updated datasets imo

cc @HugoLaurencon for WebSight

### DIFF
--- a/services/worker/src/worker/job_runners/dataset/config_names.py
+++ b/services/worker/src/worker/job_runners/dataset/config_names.py
@@ -4,7 +4,7 @@
 import logging
 from typing import Optional
 
-from datasets import get_dataset_config_names
+from datasets import get_dataset_config_names, get_dataset_default_config_name
 from datasets.data_files import EmptyDatasetError as _EmptyDatasetError
 from libcommon.exceptions import (
     ConfigNamesError,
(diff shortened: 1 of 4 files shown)


services/worker/tests/job_runners/dataset/test_config_names.py
```diff
@@ -77,12 +77,13 @@ def test_compute_too_many_configs(
     )
 
     with patch("worker.job_runners.dataset.config_names.get_dataset_config_names", return_value=configs):
-        if error_code:
-            with pytest.raises(CustomError) as exc_info:
-                job_runner.compute()
-            assert exc_info.value.code == error_code
-        else:
-            assert job_runner.compute() is not None
+        with patch("worker.job_runners.dataset.config_names.get_dataset_default_config_name", return_value=None):
```
severo, line 80: Nice feature.

Maybe add a unit test here to be sure we set the default config first.

polinaeterna:
i love this but this also would break the [bigcode/the-stack-v2-dedup](https://huggingface.co/datasets/bigcode/the-stack-v2-dedup) viewer which we struggled for so long to make it work :crying_cat_face: 
the problem is - default config in bigcode/the-stack-v2-dedup is the one that includes all the data and it doesn't work, see https://github.com/huggingface/datasets-server/issues/2552. so @anton-l and I decided to split all the data into many small configs (one per language) so that the first small config is shown in the viewer. and it works now! 

so please don't deploy until we find a solution :pray:

polinaeterna:
this is why i was dreaming about having a dedicated field for the viewer :D https://github.com/huggingface/datasets-server/issues/1953#issuecomment-1983496422

services/worker/src/worker/job_runners/dataset/config_names.py
```diff
@@ -4,7 +4,7 @@
 import logging
 from typing import Optional
 
-from datasets import get_dataset_config_names
+from datasets import get_dataset_config_names, get_dataset_default_config_name
```
severo, line 7: it's not clear from https://github.com/huggingface/datasets/blob/f45bc6caa25115a04c41b278671a5a89457eb66c/src/datasets/inspect.py#L371 when the result i None. 

I was wondering when we should reorder, and when not to do it.

I imagine that we should do it only when `default: true` is present in the YAML file, no? Is it the case?

lhoestq: It's None when there are multiple configs and no default config.

A default config is a config named "default" or a config with `default: true`

severo: OK thanks! Opened: https://github.com/huggingface/datasets/pull/6722

severo (approved):

PR #2557, opened 2024-03-07
Add `/compatible-libraries` route

lhoestq:
- Renamed `dataset-loading-tags` -> `dataset-compatible-libraries` in code
- Added migration for queue and cache
- Added route


services/api/src/api/config.py
```diff
@@ -88,6 +88,7 @@ class EndpointConfig:
                 "split": "split-is-valid",
             },
             "/statistics": {"split": "split-descriptive-statistics"},
+            "/compatible-libraries": {"dataset": "dataset-compatible-libraries"},
```
severo, line 91: why do we need this endpoint?

jobs/mongodb_migration/src/mongodb_migration/collector.py
```diff
@@ -347,4 +347,14 @@ def get_migrations(self) -> list[Migration]:
...
+            QueueRenamingMigration(
+                version="20240307191001",
+                job_type="dataset-loading-tags",
+                new_job_type="dataset-compatible-libraries",
+            ),
+            CacheRenamingMigration(
+                cache_kind="dataset-loading-tags",
+                new_cache_kind="dataset-compatible-libraries",
+                version="20240307191000",
+            ),
```
severo, line 359: maybe the metrics collections too

services/api/src/api/config.py
lhoestq: For the front end to access the list of compatible libraries and the code snippets to show to the user.

Let me know if this endpoint is fine is if you have something else in mind

severo: Hmmm, OK, indeed, a lot of the data are not passed through sse-api (from `dataset-hub-cache`).

severo (approved):
see if you prefer to remove the metrics by hand, or through a migration, nit

lhoestq:
by hand is fine :)

PR #2559, opened 2024-03-08
Rollback delete duckdb indexes job

AndreaFrancis:
Once datasets-server-job-delete-duckdb-indexes job finishes in prod, we can merge this PR in order to avoid analyzing again the datasets to delete old duckdb indexes from refs/convert/parquet


severo:
The last log for the job gives:

```
INFO: 2024-03-08 03:13:28,308 - root - 89690 analyzed datasets (total: 114274 datasets): 10118 already ok (11.28%), 24584 raised an exception (27.41%). 104156 have been cleaned old duckdb files (116.13%). 
```

Not sure why we don't reach 100% of analyzed datasets. Should we run it again? Also: note the >100% value for the last metric, which surely only means that we deleted more than 1 duckdb file per dataset, which is expected (a lot of datasets have more than 1 split)

AndreaFrancis:
> Not sure why we don't reach 100% of analyzed datasets.

I think there was a bug in the counter, see https://github.com/huggingface/datasets-server/blob/main/jobs/cache_maintenance/src/cache_maintenance/duckdb_indexes.py#L42 , the counter was increased after calling the hf API, but there were some cases where a dataset does not have the refs/convert/parquet revision.
 
humm, and I think my code was incorrect regarding counters (notice that num_successful_datasets would increase even if there was an error with the dataset on num_error_datasets). But to be sure if we sum num_successful_datasets+num_untouched_datasets to get the total analyzed datasets of 10118 + 104156 = 114274.
I would say we don't need to run the job again.

PR #2569, opened 2024-03-08
remove temporary EFS

### LINKED ISSUE #2568: Use local storage (NVMe) instead of EFS volumes
The worker machines have NVMe, and we should use this local storage instead of EFS for:
- stats cache
- HF datasets cache
- duckdb index cache

We still need to use EFS for parquet metadata, since it's shared data.

Note that `services/admin` and `clean-...`  cron jobs access the same EFS volumes, but it's respectively to: 1. provide metrics on EFS storage, and 2. clean the old files.

Once fixed, `admin` will only mount and report on usage of the parquet-metadata volume, and the `clean-...` cron job will be deleted.

severo:
fixes #2568 

- [x] descriptive statistics
- [x] HF datasets
- [x] duckdb index (only for workers)


lhoestq (approved):
Looks all good !

Maybe make sure when deploying that the HF_DATASETS_CACHE (which is now the `datasets` default value at `~/.cache/huggingface/datasets`) is a place that have read/write permissions ?

AndreaFrancis (approved):
I think it will be necessary a manually delete of the job_runner folder from /storage/duckdb-index/job_runner after deleting the clean-duckdb-index-job-runner

severo:
> Maybe make sure when deploying that the HF_DATASETS_CACHE (which is now the datasets default value at ~/.cache/huggingface/datasets) is a place that have read/write permissions ?

It will be a subdirectory of `/tmp/hf-datasets-cache`

severo:
Merging

PR #2571, opened 2024-03-11
check openapi spec against spectral

severo:
also: fix action name
also: update openapi.json detail just to trigger the action


severo:
Perfect, it detects the errors in the spec (I saw them in VSCode, thanks to the spectral extension):

Capture d’écran 2024-03-11 à 23 26 48

https://github.com/huggingface/datasets-server/runs/22536128411

severo:
merging, even if the CI is red, because... it's the point of adding this check: the openapi spec is currently incorrect.

PR #2585, opened 2024-03-13
Update polars to fix PanicException

polinaeterna:
should fix the viewer for https://huggingface.co/datasets/teknium/OpenHermes-2.5 and presumably many others

There is and issue for this in polars: https://github.com/pola-rs/polars/issues/3942.

I didn't understand though why some cases with nested structs work while some like this doesn't


polinaeterna:
i wanted to add a testcase for this but i don't understand what's different comparing to [what we already have](https://github.com/huggingface/datasets-server/blob/main/services/worker/tests/fixtures/descriptive_statistics_dataset.py#L866)

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -193,7 +193,7 @@ def compute_histogram(
         hist_df_reverted = df.with_columns(pl.col(column_name).mul(-1).alias("reverse"))["reverse"].hist(
             bins=bins_edges_reverted
         )
-        hist_reverted = hist_df_reverted["reverse_count"].cast(int).to_list()
+        hist_reverted = hist_df_reverted["count"].cast(int).to_list()
```
severo, line 196: why this change? where is the "count" field computed?

severo (approved):
my review is somewhat superficial, because I don't know the code well, and we could not reproduce the issue with a unit test...
Could you test on a specific dataset and see if it fixed the issue?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: ok, thanks!

polinaeterna:
@severo i added a test with a real file [452fbcd](https://github.com/huggingface/datasets-server/pull/2585/commits/452fbcd63c461b7a1809f59f4efea4842377e102)

i checked on the CI with an old version of polars that it doesn't pass:
image
and now i passes

services/worker/tests/job_runners/split/test_descriptive_statistics.py
```diff
@@ -662,6 +664,34 @@ def test_list_statistics(
...
+        repo_type="dataset",
+        local_dir=dir_name,
+        token=CI_USER_TOKEN,
+    )
+    return str(dir_name / "test_polars_panic_error.parquet")
+
+
+def test_polars_struct_thread_panic_error(struct_thread_panic_error_parquet_file: str) -> None:
+    from polars import Float64, List, Struct
+    from polars import Utf8 as String  # string type in polars <0.20 is called Utf8, in 0.20 it's an alias to String
```
severo, line 685: Now that we use 0.20, we can just import String, right?

severo (approved):

services/worker/tests/job_runners/split/test_descriptive_statistics.py
severo: I don't know. I understand that it was for 0.19, but now that we use 0.20, it's not needed anymore, right?

Issue #2586, opened 2024-03-14
Cells are marked as truncated whereas they're not


severo:
See https://datasets-server.huggingface.co/first-rows?dataset=HuggingFaceH4/deita-6k-v0-sft&config=default&split=train_sft

<details><summary>JSON response</summary>
<pre>
{
  "dataset": "HuggingFaceH4/deita-6k-v0-sft",
  "config": "default",
  "split": "train_sft",
  "features": [
    {
      "feature_idx": 0,
      "name": "prompt",
      "type": { "dtype": "string", "_type": "Value" }
    },
    {
      "feature_idx": 1,
      "name": "prompt_id",
      "type": { "dtype": "string", "_type": "Value" }
    },
    {
      "feature_idx": 2,
      "name": "messages",
      "type": [
        {
          "content": { "dtype": "string", "_type": "Value" },
          "role": { "dtype": "string", "_type": "Value" }
        }
      ]
    }
  ],
  "rows": [
    {
      "row_idx": 0,
      "row": {
        "prompt": "I want to learn store Procedure, Can you teach me from the basic .",
        "prompt_id": "100cfc5f4d14b9c640cac30ebb28d0f91f2dc7662e9a578a544923a48fd5a59f",
        "messages": [
          {
            "content": "I want to learn store Procedure, Can you teach me from the basic .",
            "role": "user"
          },
          ...
    },
    {
      "row_idx": 9,
      "row": {
        "prompt": "\"Act as business consultant. Prepare a content for thirty slides about ageism for business training.",
        "prompt_id": "994c40cee21a1e60e11fb326a41999f05079bd568681c6066efe83660e91fd83",
        "messages": "[{\"content\":\"Act as business consultant. Prepare a content for thirty slides about ageism for busine"
      },
      "truncated_cells": ["prompt", "messages"]
    }
  ],
  "truncated": true
}
</pre></details>

The cells `prompt` and `messages` are always listed in `truncated_cells`, whereas they are not always truncated. For example, they are not truncated in cell 0.

```json
   {
      "row_idx": 0,
      "row": {
        "prompt": "I want to learn store Procedure, Can you teach me from the basic .",
        "prompt_id": "100cfc5f4d14b9c640cac30ebb28d0f91f2dc7662e9a578a544923a48fd5a59f",
        "messages": [
          {
            "content": "I want to learn store Procedure, Can you teach me from the basic .",
            "role": "user"
          },
          ...
        ],
[...]

PR #2595, opened 2024-03-14
Fix cell truncation

lhoestq:
We were using the same `list` for all the cells. So if a cell has truncated columns then all the cells end up with the same value for `truncated_cells`

Close https://github.com/huggingface/datasets-server/issues/2586


severo (approved):
😮

severo:
I guess that we will recompute the first rows entries on a case by case basis, because I don't think it affects too many datasets, and it's a long step.

lhoestq:
Yep, I can take care of HuggingFaceH4/deita-6k-v0-sft once it's deployed

severo:
feel free to deploy!

PR #2597, opened 2024-03-14
Add retry when loading dataset builder

AndreaFrancis:
Related to https://github.com/huggingface/datasets-server/issues/1443 for `config-parquet-and-info`, we have 1065 entries with HfHubHTTPError UnexpectedError, we should retry because sometimes it is not a final error but just an issue with Hub's connectivity.


severo (commented):
Good idea. And we should also catch the error to avoid UnexpectedError anyway, and have a proper error code

PR #2577, opened 2024-03-12
Add "statistics" field to is-valid workers

polinaeterna:
todo: 
- [x]  update openapi.json
- [x] update admin-ui

PR #2599, opened 2024-03-14
add missing dependency between steps

severo:
follows #2577

also: add two missing tests


severo:
cc @polinaeterna for viz, we forgot this change in #2577

polinaeterna:
argh again i forgot it, thank you!

PR #2602, opened 2024-03-15
Compute stats for Sequence feature only if it's underlying schema is List

polinaeterna:
Should fix https://huggingface.co/datasets/OpenAssistant/oasst1 (see "emojis" and "label" columns).
When sequence feature is dict of lists (not list of dicts), compute lengths over first subfield values.


services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -578,7 +583,25 @@ def _compute_statistics(
...
+        column_schema = data.schema[column_name]
+        if isinstance(column_schema, List):
+            lengths_df = df_without_na.with_columns(pl.col(column_name).list.len().alias(lengths_column_name))
+
+        elif isinstance(column_schema, Struct):
+            if not all(isinstance(field.dtype, List) for field in column_schema.fields):
+                raise PolarsSequenceSchemaError(
+                    f"Error parsing {column_name=} {column_schema=}. Subfields of Struct must be Lists. "
+                )
+            first_field_name = column_schema.fields[0].name  # assuming all fields have equal lengths within a sample
```
polinaeterna, line 595: to compute lengths in case when feature is dict of lists, i take values of the first field assuming that they all have different lengths. i don't know if it's possible that they actually have different lengths? cc @lhoestq

services/worker/tests/fixtures/descriptive_statistics_dataset.py
```diff
@@ -1109,18 +1154,27 @@ def nan_column() -> list[None]:
             "list__sequence_column": Sequence(Value("int64")),
             "list__sequence_nan_column": Sequence(Value("int64")),
             "list__sequence_all_nan_column": Sequence(Value("int64")),
-            "list__sequence_dict_column": [
+            "list__sequence_dict_column": Sequence(
```
polinaeterna, line 1157: that was incorrect before

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -578,7 +583,25 @@ def _compute_statistics(
...
+            lengths_df = df_without_na.with_columns(
+                pl.col(column_name).struct[first_field_name].list.len().alias(lengths_column_name)
+            )
+        else:
```
polinaeterna, line 599: idk if there might be something else (except for List and Struct), i assume that no but we can check after deployment if there are errors like this

services/worker/src/worker/job_runners/split/descriptive_statistics.py
lhoestq: They can have different lengths yes. For dicts I'm not sure what stats we should compute

polinaeterna: this specific case is for dicts of lists (it's just another option to represent list of structs) - so the same stats as for lists

lhoestq: maybe just ignore structs for this PR ?

polinaeterna: like here in "emojis" and "labels" https://huggingface.co/datasets/OpenAssistant/oasst1?row=0

polinaeterna: >  maybe just ignore structs for this PR ?

this is the fix itself :smiley:

lhoestq: I think the correct associated type is `struct` (not `sequence`) and for now we should not show anything for now. At least until we figure out a high level stat that makes sense for this type (maybe number of fields ?)

severo: OK: so, discriminating over the Sequence feature type is not a good idea because it can represent a list of things, or a dict of lists. We should show them as different types (here and/or on the Hub) to be sure to handle them differently, no?

polinaeterna: > in my opinion showing the histogram of length of the first list in the struct doesn't really make sense.

from what i see, this is just another way of storing the same data like this:
```python
[ {"name": "+1", "count": 10},  {"name": "_skip_reply", "count": 1},  {"name": "_skip_ranking", "count": 4}]
```
both ways in `datasets` are encoded as `Sequence({"name": Value("string"), "count": Value("int32")})` type; 
but in polars they are either `List` or `Struct(List)` depending on their raw values (if they are lists or dicts, as in example I provide here, or dicts of lists, as in your example above).

I can check if they have the same lengths, and return histogram only if it is true. But my point was that it is probably computationally inefficient while unnecessary, since we know that they are marked as `Sequence` so logically should be of the same lengths but technically there is no constraint for this. 
I can implement this, maybe at least compare if values of the first two keys have equal length in all rows.

>  the correct associated type is struct (not sequence) and for now we should not show anything for now

No, it is Sequence: https://huggingface.co/datasets/OpenAssistant/oasst1/blob/main/README.md?code=true#L58. I don't compute anything for struct for now indeed, only for sequences (you can check [_column_from_feature()](https://github.com/huggingface/datasets-server/pull/2602/files#diff-51b62bca3846b24d6554358a34bd39d0db4db7f24e378182fde2d549e2d1268bR756) function)

does this all make sense? :D i see it might me confusing

lhoestq: It's named sequence for legacy reasons with tfds. Since the lists may not have the same length you can't really say it's equivalent to

```
[ {"name": "+1", "count": 10},  {"name": "_skip_reply", "count": 1},  {"name": "_skip_ranking", "count": 4}]
```

severo: I never really understood the Sequence type, but I assumed it's like a list, although it can be encoded in two ways (list of things, dict of lists). If it's the case (a list-like type), we should show the same stats as lists. If we're not sure it's always the case, we should maybe only process the "list" one, and ignore the "dict" one.

polinaeterna: > Since` the lists may not have the same length

@lhoestq do you think it makes sense to compare lengths of lists in first two fields and return stats if they are all equal?

polinaeterna: @severo we chatted with @lhoestq and for now we decided to disable sequence-dict case until we decide on what to do with nested things in the viewer.

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -733,13 +735,14 @@ def compute_descriptive_statistics_response(
...
-        if isinstance(dataset_feature, list):
-            return ListColumn(feature_name=dataset_feature_name, n_samples=num_examples, n_bins=histogram_num_bins)
-
-        if isinstance(dataset_feature, dict):
-            if dataset_feature.get("_type") == "Sequence":
+        if isinstance(dataset_feature, list) or (
+            isinstance(dataset_feature, dict) and dataset_feature.get("_type") == "Sequence"
+        ):
+            schema = pl.scan_parquet(local_parquet_glob_path).schema[dataset_feature_name]
+            if isinstance(schema, List):  # compute only if it's internally a List! because it can also be Struct
```
lhoestq, line 742: add link to the docs

```suggestion
            # Compute only if it's internally a List! because it can also be Struct, see
            # https://huggingface.co/docs/datasets/v2.18.0/en/package_reference/main_classes#datasets.Features
            if isinstance(schema, List):
```

services/worker/tests/fixtures/descriptive_statistics_dataset.py
```diff
@@ -862,7 +862,553 @@ def nan_column() -> list[None]:
...
+            [[True, False, False, False, True], [None, True]],
+            [[True], [False], [False], [False, None], [True, None, True]],
+            [[True]],
+            None,
+            [[True, False, False, False], [None, True]],
+            [[True], [None, True, False]],
+            [[True], [True, False], [True, False]],
+        ],
+        "list__sequence_of_sequence_bool_all_nan_column": nan_column(),
+        "list__sequence_of_sequence_dict_column": [
```
polinaeterna, line 956: this case would work but seems to be always of length=1 because internally it's 
```
[
	[{'author': ['cat'], 'likes': [5]}], 
	[{'author': ['cat', 'cat'], 'likes': [5, 5]}], 
	[{'author': ['cat', 'cat'], 'likes': [5, 5]}], 
         ...
 	[{'author': [], 'likes': []}]
]
```
i don't expect to see this in reality though :D

polinaeterna:
Sequence feature is making me going mad a bit but i think it's ready for review now @severo @lhoestq

lhoestq (approved):
LGTM !

severo (approved):
thanks for fixing this (and adding 500 lines of tests :P)

PR #2605, opened 2024-03-18
Support exif images

mariosasko:
The EXIF orientation tag is currently ignored when displaying images (see https://huggingface.co/datasets/mariosasko/exif_image).

This PR should fix this.


severo (approved):
Thanks! I think it's OK without a test, as it's only applying a PIL method.

severo:
@mariosasko have you identified datasets where the images appear rotated in the dataset viewer?

lhoestq:
fix on the `datasets` side too: https://github.com/huggingface/datasets/pull/6739

PR #2606, opened 2024-03-18
Add Modalities tags

lhoestq:
Add "image", "audio" and "text". Later we can think about adding "tabular" (?) or more exotic ones like

TODO
- [x] add to hub-cache
- [x] tests


services/worker/src/worker/job_runners/dataset/modalities.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        modalities: set[DatasetModality] = set()
+
+        def classify_modality(feature: FeatureType) -> None:
+            nonlocal modalities
+            if isinstance(feature, Audio):
+                modalities.add("audio")
+            elif isinstance(feature, Image):
+                modalities.add("image")
+            elif isinstance(feature, Value) and feature.dtype == "string":
```
severo, line 54: maybe add large_string and the two translation types too?

severo: also: note that a numerical dataset that only has a string id column would be detected as a "text" dataset. Maybe it's a corner-case, but I have the intuition that nearly all the datasets will have a string column, which will defeat a bit the purpose of detecting a modality. Maybe it's not an important comment

services/worker/src/worker/job_runners/dataset/modalities.py
```diff
@@ -0,0 +1,78 @@
...
+                modalities.add("text")
+
+        for config_info in content["dataset_info"].values():
```
severo, line 57: as per the loop here, the processing step is fundamentally a config-level step. Not sure we need the modality per config though, so: it seems a good idea to just loop over all the configs.

services/worker/src/worker/job_runners/dataset/modalities.py
lhoestq: I guess we can improve later if needed

lhoestq:
I added the missing "text" types :)

severo (approved):

PR #2610, opened 2024-03-19
Fix hub cache libraries

lhoestq:
Follows the changes in https://github.com/huggingface/datasets-server/pull/2454

The `dataset-hub-cache` should have `libraries` as a list of strings, not a list of CompatibleLibrary objects

### DIFF
--- a/services/worker/src/worker/job_runners/dataset/hub_cache.py
+++ b/services/worker/src/worker/job_runners/dataset/hub_cache.py
@@ -6,7 +6,7 @@ import logging
 from libcommon.exceptions import PreviousStepFormatError
 from libcommon.simple_cache import CachedArtifactNotFoundError, get_previous_step_or_raise
 
-from worker.dtos import DatasetHubCacheResponse, DatasetLibrary, DatasetTag, JobResult
+from worker.dtos import CompatibleLibrary, DatasetHubCacheResponse, DatasetLibrary, DatasetTag, JobResult
 from worker.job_runners.dataset.dataset_job_runner import DatasetJobRunner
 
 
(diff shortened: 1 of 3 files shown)


AndreaFrancis (approved):

services/worker/src/worker/job_runners/dataset/hub_cache.py
```diff
@@ -70,7 +70,8 @@ def compute_hub_cache_response(dataset: str) -> tuple[DatasetHubCacheResponse, f
             kind="dataset-compatible-libraries", dataset=dataset
         )
         tags = compatible_libraries_response["content"]["tags"]
-        libraries = compatible_libraries_response["content"]["libraries"]
+        compatible_libraries: list[CompatibleLibrary] = compatible_libraries_response["content"]["libraries"]
```
severo, line 73: interesting. It means that we are passing the complete code samples, etc, to the Hub through SSE, and the Hub can cache them in its own DB, removing the need to call datasets-server on every dataset page creation.

Hmmm, I don't know at which point we want to do this though. We don't want to replicate all the cache-entries database to the Hub's one.

julien-c: yes we want to only store the minimum amount of data in the Hub DB

it's important DX-wise on the Hub side 🙏 
🙏 

calling datasets-server on dataset pages is fine

severo: OK. For now it's sent in the SSE, but not stored on the Hub side, so it's only some additional traffic. Once we finish the "Use in libraries" thing, I'll remove the bits we don't use from this endpoint.

severo: I opened an issue for this: https://github.com/huggingface/datasets-server/issues/2618

lhoestq: We are no longer passing the code samples btw, that's the point of this PR

severo: You're right, I misread

Issue #2608, opened 2024-03-19
Systematic ExternalServerError on HackerNoon/tech-company-news-data-dump


severo:
On step  split-opt-in-out-urls-scan, we always get:

```
{"error": "Error when trying to connect to https://opts-api.spawningaiapi.com/api/v2/query/urls"}
```

It seems to be the case only for this dataset (or some other ones).

cc @AndreaFrancis for viz

PR #2611, opened 2024-03-19
Fix  ExternalServerError for None URLs in split-opt-in-out-scan

AndreaFrancis:
Fix for https://github.com/huggingface/datasets-server/issues/2608
Many URLs are None and the job runner failed when doing the join operation at https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py#L38

### DIFF
--- a/services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
+++ b/services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
@@ -46,7 +46,7 @@ async def opt_in_out_task(
     try:
         spawning_response = await check_spawning(image_urls, session, semaphore, limiter, spawning_url)
     except Exception as err:
-        raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}") from err
+        raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}", cause=err) from err
     if "urls" not in spawning_response:
         raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}: '{spawning_response}'")
     opt_in_urls_indices = [i for i in range(len(image_urls)) if spawning_response["urls"][i]["optIn"]]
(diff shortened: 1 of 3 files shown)


services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -45,8 +45,8 @@ async def opt_in_out_task(
 ) -> tuple[list[Any], list[Any]]:
     try:
         spawning_response = await check_spawning(image_urls, session, semaphore, limiter, spawning_url)
-    except Exception:
-        raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}")
+    except Exception as err:
+        raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}") from err
```
severo, line 49: even:

```suggestion
        raise ExternalServerError(message=f"Error when trying to connect to {spawning_url}", cause=err) from err
```

services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py
```diff
@@ -224,7 +224,7 @@ def compute_opt_in_out_urls_scan_response(
 
     # get the urls
     num_scanned_rows = len(rows)
-    urls = [row[urls_column] for row in rows for urls_column in image_url_columns]
+    urls = [row[urls_column] if row[urls_column] else "" for row in rows for urls_column in image_url_columns]
```
severo, line 227: shouldn't we filter instead?

```suggestion
    urls = [row[urls_column] for row in rows for urls_column in image_url_columns if row[urls_column]]
```

AndreaFrancis: I thought the same, but it will affect when matching the opt indices at https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py#L52

AndreaFrancis: The problem is when sending the concatenated list of URLs at  https://github.com/huggingface/datasets-server/blob/main/services/worker/src/worker/job_runners/split/opt_in_out_urls_scan_from_streaming.py#L38 we have to keep the order or maybe think another logic to map the URL to the index.

severo (approved):

severo:
Once deployed, we should refresh the step for HackerNoon/tech-company-news-data-dump, and surely other ones

severo:
Note: it's not widespread

Capture d’écran 2024-03-20 à 16 03 18

Issue #1929, opened 2023-10-05
Add a "feature" or "column" level for better granularity


severo:
For example, if we support statistics for a new type of columns, or if we change the way we compute some stats, I think that we don't want to recompute the stats for all the columns, just for one of them.

It's a guess, because maybe it's more efficient to have one job that downloads the data and computes every possible stats, than having N jobs that download the same data and compute only one stat. To be evaluated

severo:
cc @polinaeterna

severo:
not needed for now

PR #2612, opened 2024-03-19
Stats for audio

polinaeterna:
~~i will maybe create a separate worker for media features (audio and image), haven't decided yet~~

In the end I didn't create a new worker / a new level (feature-type) because it's still not needed for our datasets sizes, see https://github.com/huggingface/datasets-server/pull/2612#issuecomment-2015482880

Failed e2e tests build seem to be unrelated to this PR (weird though).

**didn't i forget anything?**


severo:
clever. it could be a good idea to have one step per column kind: string, number, audio, images, etc. We had https://github.com/huggingface/datasets-server/issues/1929 (column-level jobs) which we discarded. Having one step per column kind would give some level of granularity; which would make it easier to only update the image stats, or the string stats, etc.

polinaeterna:
@severo yeah i also thought so but now i think maybe leave it as is again because: 
1. most of audio datasets have a single audio column (i haven't found any datasets on the hub with at least two audio columns), and we also have not that many audio datasets  
2. considering that we take only up to 5gb of raw data, processing seems to be fast as is 

so IMO adding column type step level would be beneficial if we extend parquet exports to more than 5gb which i'd love! but I don't know if we have plans for it

but on the other hand if (when) we also add image stats, that might be more recourse consuming and it would be reasonable to group audio and images. but we haven't decided what to show for images yet :D

severo:
for reference, asked here: https://huggingface.slack.com/archives/C039P47V1L5/p1701102125668309 (to ping them when it's available in prod)

severo:
I opened https://github.com/huggingface/datasets-server/issues/2636 for the CI error

services/worker/tests/job_runners/split/test_descriptive_statistics.py
```diff
@@ -471,6 +474,51 @@ def descriptive_statistics_string_text_partial_expected(datasets: Mapping[str, D
...
+        "column_statistics": audio_statistics,
+    }
+    nan_audio_lengths = [1.0, 3.0]  # take first and third audio file for this testcase
+    nan_audio_statistics = count_expected_statistics_for_numerical_column(
+        column=pd.Series(nan_audio_lengths), dtype=ColumnType.FLOAT
+    )
+    expected_nan_statistics = {
+        "column_name": "audio_nan",
+        "column_type": ColumnType.AUDIO,
+        "column_statistics": nan_audio_statistics,
```
severo, line 495: does it include `"nan_count": 2`? From the code it seems not. But the array contains two audio files and two NaN, right?

docs/source/openapi.json
```diff
@@ -1084,7 +1084,7 @@
       },
       "ColumnType": {
         "type": "string",
-        "enum": ["float", "int", "class_label", "string_label", "string_text", "bool", "list"]
+        "enum": ["float", "int", "class_label", "string_label", "string_text", "bool", "list", "audio"]
```
severo, line 1087: maybe also add an example in /statistics?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -594,15 +578,59 @@ def _compute_statistics(
...
+        parquet_directory: Path,
+        column_name: str,
+        n_samples: int,
+        n_bins: int,
+    ) -> NumericalStatisticsItem:
+        table = pq.read_table(parquet_directory, columns=[column_name]).drop_null()
+        if table.shape[0] == 0:
+            return all_nan_statistics_item(n_samples)
+
+        data = table.to_pydict()[column_name]
```
polinaeterna, line 611: actually as all data is in-memory here, it might take quite a lot of additional RAM.
It will be up to 5Gb of RAM for parquet + same size to copy as python dict.
Should I load file by file instead of full directory at once?

services/worker/tests/job_runners/split/test_descriptive_statistics.py
polinaeterna: oh that was a bug than you! fixed [4f7db83](https://github.com/huggingface/datasets-server/pull/2612/commits/4f7db83f772f750f422a44c73ae7e1da8e1fefc8)

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: changed to reading file by file [1e1bf60](https://github.com/huggingface/datasets-server/pull/2612/commits/1e1bf6086c1d63d16f542fcd85d104c134f76161)

docs/source/openapi.json
polinaeterna: it would be easier to add a 100%-sure working example when we have them in prod 🥺

polinaeterna: [feeb31c](https://github.com/huggingface/datasets-server/pull/2612/commits/feeb31cc3feff830642756a267bafc6e2732404f)
but i don't know, do we have an automated way to format openapi specs?

severo: Yes, if you use vscode (and the workspace defined here: https://github.com/huggingface/datasets-server/blob/main/.vscode/monorepo.code-workspace), install the recommended extensions (in particular `stoplight.spectral`). Please tell me if it works!

severo (approved):

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -594,15 +578,84 @@ def _compute_statistics(
...
+        parquet_directory: Path,
+        column_name: str,
+        n_samples: int,
+        n_bins: int,
+    ) -> NumericalStatisticsItem:
+        logging.info(f"Compute statistics for Audio column {column_name} with librosa and polars. ")
+        parquet_files = list(parquet_directory.glob("*.parquet"))
+        durations = []
+        for filename in parquet_files:
+            shard_audios = pq.read_table(filename, columns=[column_name]).drop_null().to_pydict()[column_name]
```
lhoestq, line 611: This can lead to quite some memory usage. For now it's maybe fine but it fails too often you can iterate on row groups instead of loading the full table.

lhoestq (approved):
Awesome !

services/worker/src/worker/job_runners/split/descriptive_statistics.py
severo: good idea. As even the row groups can be big

docs/source/openapi.json
polinaeterna: eh i use pycharm :upside_down_face:

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: > This can lead to quite some memory usage.

I think it would take up to ~1Gb peak usage at each iteration because max parquet file size is ~500Mb and when it contains audio bytes, the size is more or less equal to the uncompressed size. But maybe if audio files are big, loading with librosa in parallel might increase memory usage significantly, I just haven't encountered this case because usually audio datasets contain short segments of texts.  So yeah, we will see

Issue #2455, opened 2024-02-14
Add "modality" tags in /hub-cache


severo:
Could be: `image` and `audio`, to start with.

Related issues:
- https://github.com/huggingface/datasets-server/pull/2454 (libraries).
- https://github.com/huggingface/datasets-server/issues/2408 (tasks).
- internal: https://github.com/huggingface/moon-landing/issues/7557

Previous issue:
- https://github.com/huggingface/datasets-server/issues/561 (column types)

julien-c:
also maybe `format: {json, csv, parquet, imagefolders, etc}` (not sure if this is the same issue though)

PR #2617, opened 2024-03-20
Add formats tags

lhoestq:
Added `formats: list[str]` to "hub-cache". It is computed in "compatible-libraries" since compatibilities depend on the format.

Possibles values are
- csv
- json
- parquet
- text
- imagefolder
- audiofolder

Mentioned in https://github.com/huggingface/datasets-server/issues/2455

I chose to have it as a list for simplicity, but lmk if you think we should set it to `format: Optional[str]` instead

includes a minor change: updated the libraries order

### DIFF
--- a/services/worker/src/worker/dtos.py
+++ b/services/worker/src/worker/dtos.py
@@ -221,6 +221,7 @@ class IsValidResponse(TypedDict):
 
 DatasetTag = Literal["croissant"]
 DatasetLibrary = Literal["mlcroissant", "webdataset", "datasets", "pandas", "dask"]
+DatasetFormat = Literal["json", "csv", "parquet", "imagefolder", "audiofolder", "webdataset", "text"]
 ProgrammingLanguage = Literal["python"]
 
 
@@ -240,6 +241,7 @@ class CompatibleLibrary(TypedDict):
 class DatasetCompatibleLibrariesResponse(TypedDict):
     tags: list[DatasetTag]
     libraries: list[CompatibleLibrary]
+    formats: list[DatasetFormat]
 
 
 class DatasetHubCacheResponse(TypedDict):
@@ -249,6 +251,7 @@ class DatasetHubCacheResponse(TypedDict):
     num_rows: int
     tags: list[DatasetTag]
     libraries: list[DatasetLibrary]
+    formats: list[DatasetFormat]
 
 
 class DatasetParquetResponse(TypedDict):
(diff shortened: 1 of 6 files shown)


services/worker/src/worker/dtos.py
```diff
@@ -221,6 +221,7 @@ class IsValidResponse(TypedDict):
 
 DatasetTag = Literal["croissant"]
 DatasetLibrary = Literal["mlcroissant", "webdataset", "datasets", "pandas", "dask"]
+DatasetFormat = Literal["json", "csv", "parquet", "imagefolder", "audiofolder", "webdataset", "text"]
```
severo, line 224: we cannot discriminate between json and jsonl, right?

services/worker/src/worker/dtos.py
```diff
@@ -240,6 +241,7 @@ class CompatibleLibrary(TypedDict):
 class DatasetCompatibleLibrariesResponse(TypedDict):
     tags: list[DatasetTag]
     libraries: list[CompatibleLibrary]
+    formats: list[DatasetFormat]
```
severo, line 244: having two formats means that two configs are in different formats, right?

services/worker/src/worker/job_runners/dataset/compatible_libraries.py
```diff
@@ -607,20 +619,22 @@ def compute_compatible_libraries_response(
...
         if builder_name in get_compatible_library_for_builder:
             try:
                 compatible_library = get_compatible_library_for_builder[builder_name](dataset, hf_token)
                 libraries.append(compatible_library)
             except NotImplementedError:
                 pass
-    return DatasetCompatibleLibrariesResponse(tags=tags, libraries=libraries)
+        # mlcroissant library
+        libraries.append(get_mlcroissant_compatible_library(dataset, infos, partial=partial))
+        tags.append("croissant")
```
severo, line 636: good idea to move them to the end of the libraries list (as we will respect this order on the Hub)

severo (approved):

services/worker/src/worker/dtos.py
lhoestq: Currently having multiple formats is not supported (only one is detected)

services/worker/src/worker/dtos.py
lhoestq: We could differentiate but it requires to read some data. For now I'd keep it under "json"

services/worker/src/worker/dtos.py
severo: Ok. I was wondering why we return a list of formats

Issue #2620, opened 2024-03-21
Check `permissions` instead of `read` in JWT


severo:
Here:

https://github.com/huggingface/datasets-server/blob/dd2a81568b2334f69a005f2c1dfa224ad2024e9f/libs/libapi/src/libapi/jwt_token.py#L305-L307

Once https://github.com/huggingface/moon-landing/pull/9306 (internal) is deployed. The new field is:

```
permissions: {"repo.content.read": true}
```

Note that meanwhile, the `read` field will still be provided.

severo:
See https://github.com/huggingface/moon-landing/pull/9306/files#diff-9f01a5c8bad78651b5a7da78470337ed07ae5eadc9ad1f46becb6ff0cc457539R1306 for details of the new format

PR #2622, opened 2024-03-22
adapt JWT verification to new field

severo:
The JWT now contains `permissions: {"repo.content.read": true}` instead
of the (soon deprecated) ``read: true`.

fixes https://github.com/huggingface/datasets-server/issues/2620.

### DIFF
--- a/libs/libapi/src/libapi/jwt_token.py
+++ b/libs/libapi/src/libapi/jwt_token.py
@@ -268,7 +268,7 @@ def validate_jwt(
                 jwt=token,
                 key=public_key,
                 algorithms=[algorithm],
-                options={"require": ["exp", "sub", "read"], "verify_exp": verify_exp},
+                options={"require": ["exp", "sub", "permissions"], "verify_exp": verify_exp},
             )
             logging.debug(f"Decoded JWT is: '{public_key}'.")
             break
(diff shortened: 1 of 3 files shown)


coyotte508 (approved):
Note that technically those also work for us:

- `{"repo.content.write": true}`
- `{"repo.write": true}`
- `{"repo.read": true}`

We'll probably leave "repo.content.read" to be predictable, but if you want to 100% future-proof, you can check any of those is `true`

coyotte508:
> OK. Good to know. Is there a specific algorithm to use to check these different permissions? Or just check that any of them is `true` (which is what I did in [the last commit](https://github.com/huggingface/datasets-server/pull/2622/commits/53f8e87663e237fb8823f21e33e7f86c714bd25e))?

it's like webhooks, basically

write=true => gives access to read
higher scopes gives access to lower scope ("repo". gives access to "repo.content")

severo:
and should we raise here?

https://github.com/huggingface/datasets-server/pull/2622/files#diff-9fdf2c5e5d43723757c70f9c7143257d2990b894107400e957918d30b2692462R334

when the permissions are: `{"repo.write": True, "repo.content.read": False}`

coyotte508:
it works with an "allow" rather than "block" principle, so `{"repo.content.read:" false, "repo.content.write": true}` is valid

AndreaFrancis (approved):

Issue #2628, opened 2024-03-25
Increase nginx gateway timeout limit?


severo:
https://datasets-server.huggingface.co/search?dataset=wikimedia%2Fwikipedia&config=20231101.en&split=train&offset=0&length=100&query=deep+learning

gives

```
502 Bad Gateway
nginx/1.20.2
```

severo:
Fixed by @AndreaFrancis with #2638

PR #2633, opened 2024-03-25
Try to improve fts for big datasets by batches

AndreaFrancis:
Should fix https://github.com/huggingface/datasets-server/issues/2628 and help with search in big datasets.
When doing the 

```
FTS_BY_TABLE_COMMAND = (
    "SELECT * EXCLUDE (__hf_fts_score) FROM (SELECT *, fts_main_data.match_bm25(__hf_index_id, ?) AS __hf_fts_score"
    " FROM data) A WHERE __hf_fts_score IS NOT NULL ORDER BY __hf_fts_score DESC OFFSET {offset} LIMIT {length};"
)
```

FTS command performs the fts_main_data.match_bm25 macro over the whole "data" table in the index.duckdb file causing many OOM fails because it needs to load all the tables in memory that is for instance, for `wikimedia/wikipedia` en config, it tries to load nearly 9 GB of data (the current pod only has 6GB of memory).
I haven't found a better way to restrict and match to a better query plan than doing it by batches using __hf_index_id index. Hopefully, it will help a lot with big index files (I tried it, and worked super fast at least for the first pages), so I enabled it only for `wikimedia/wikipedia`, if it works, we can enable it for other datasets or use it for all.
**Note:**
Currently, it is like an "incremental" scanning, which can lead to a long query time if someone looks for page 1M for example, I was thinking of processing the batches in parallel, but it will be unnecessary for small datasets. Maybe in an upcoming PR, try to fit the better approach by the dataset size.
I also tried loading in two steps, first, only `SELECT __hf_index_id, fts_main_data.match_bm25(__hf_index_id, ?)` (which is super fast since it uses the __hf_index_id index) instead of `SELECT *, fts_main_data.match_bm25(__hf_index_id, ?)` and then join it with the data table but for some reason, it still gets OOM even if the first left table is reduced (Will continue investigating that approach).

### DIFF
--- a/services/search/src/search/routes/search.py
+++ b/services/search/src/search/routes/search.py
@@ -47,38 +47,40 @@ from starlette.responses import Response
 
 from search.duckdb_connection import duckdb_connect
 
-FTS_COMMAND_COUNT = "SELECT COUNT(*) FROM (SELECT __hf_index_id, fts_main_data.match_bm25(__hf_index_id, ?) AS __hf_fts_score FROM data) A WHERE __hf_fts_score IS NOT NULL;"
+FTS_COUNT_COMMAND = "SELECT COUNT(*) FROM (SELECT __hf_index_id, fts_main_data.match_bm25(__hf_index_id, ?) AS __hf_fts_score FROM data) A WHERE __hf_fts_score IS NOT NULL;"
 
 
-FTS_COMMAND = (
+FTS_BY_TABLE_COMMAND = (
     "SELECT * EXCLUDE (__hf_fts_score) FROM (SELECT *, fts_main_data.match_bm25(__hf_index_id, ?) AS __hf_fts_score"
     " FROM data) A WHERE __hf_fts_score IS NOT NULL ORDER BY __hf_fts_score DESC OFFSET {offset} LIMIT {length};"
 )
-FTS_BATCH_COMMAND = "SELECT * FROM (SELECT *, fts_main_data.match_bm25(__hf_index_id, ?) AS __hf_fts_score FROM data WHERE __hf_index_id BETWEEN {offset} and {length}) A WHERE __hf_fts_score IS NOT NULL;"
+
+FTS_BY_BATCHES_COMMAND = "SELECT * FROM (SELECT *, fts_main_data.match_bm25(__hf_index_id, ?) AS __hf_fts_score FROM data WHERE __hf_index_id BETWEEN {offset} and {length}) A WHERE __hf_fts_score IS NOT NULL;"
+
 FTS_BATCH_SIZE = 1000
 
 logger = logging.getLogger(__name__)
 
-DATASETS_WITH_FTS_BATCHES = ["wikimedia/wikipedia"]
+DATASETS_WITH_FTS_BY_BATCHES = ["wikimedia/wikipedia", "asoria/bolivian-recipes"]
 
 
-def _full_text_search_table(con: DuckDBPyConnection, query: str, offset: int, length: int) -> pa.Table:
+def _search_by_table(con: DuckDBPyConnection, query: str, offset: int, length: int) -> pa.Table:
     query_result = con.execute(
-        query=FTS_COMMAND.format(offset=offset, length=length),
+        query=FTS_BY_TABLE_COMMAND.format(offset=offset, length=length),
         parameters=[query],
     )
     return query_result.arrow()
 
 
-def _full_text_search_batches(
+def _search_by_batches(
     num_original_rows: int, con: DuckDBPyConnection, query: str, offset: int, length: int
 ) -> pa.Table:
     num_matched_rows = 0
     index = 0
     pa_tables = []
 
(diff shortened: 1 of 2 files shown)


AndreaFrancis:
The e2e error looks to be related to another issue in admin (not related to this PR I think), will check:
```
 #58 77.15 
#58 77.15   ChefBuildError
#58 77.15 
#58 77.15   Backend subprocess exited when trying to invoke build_wheel
#58 77.15   
#58 77.15   Will build with KenLM max_order set to 6
```

services/search/src/search/routes/search.py
```diff
@@ -46,33 +47,74 @@
...
+        fts_query = FTS_BY_BATCHES_COMMAND.format(offset=index, length=index + FTS_BATCH_SIZE)
+        logging.info(fts_query)
+        pa_table = con.execute(
+            query=fts_query,
+            parameters=[query],
+        ).arrow()
+        pa_tables.append(pa_table)
+        num_matched_rows += pa_table.num_rows
+        logging.info(f"{pa_table.num_rows} rows for batch {index} - {num_matched_rows} for table")
+        if num_matched_rows < (offset + length):
```
severo, line 92: isn't it the opposite? `>`?

are we testing this (offset, length) ?

services/search/src/search/routes/search.py
```diff
@@ -46,33 +47,74 @@
...
+            break
+    pa_result = pa.concat_tables(pa_tables)
+    pa_result = pa_result.sort_by([("__hf_fts_score", "descending")]).drop("__hf_fts_score")
```
severo, line 95: weren't we returning the results by row_idx, instead of FTS score?

AndreaFrancis: I don't quite remember the related issue, but It was done in this PR https://github.com/huggingface/datasets-server/pull/1790

services/search/src/search/routes/search.py
AndreaFrancis: Thanks, my mistake while moving from while loop to for instead. I will add more tests for the offset, length

AndreaFrancis:
> weren't we returning the results by row_idx, instead of FTS score?

I found the [internal thread](https://huggingface.slack.com/archives/C04L6P8KNQ5/p1694102592134799) for the discussion. 
However, there is an issue related to offset/length in this PR. The "incremental search" by batches won't work under the score sort criteria because there can be other records after the current batch that can have a higher similarity score and that changes the results.
On the other hand, if we change to `row_idx`, the current PR can work and help to compute fast the first 100 records for the search.
If we don't want to change the sort by  `row_idx`, I think changing this PR to run in parallel by batches can work, and at the end, sort the results by the score and get the page by the offset and length. (Which is heavy for big datasets because it will anyway query the whole table but it will probably avoid some OOM).
WDYT @severo @lhoestq @albertvillanova ?

polinaeterna:
i'd be in favor of sorting by relevance but i don't see an easy solution either - running in parallel would be so nice but we also should avoid mixing python parallelization and native duckdb multithreading (is it used in their search implementation by the way? i'd assume so but not sure), this might lead to hanging processes as it was with statistics :((

lhoestq:
> many OOM fails because it needs to load all the tables in memory

Do you know what it load exactly ? Does it load the full table, or only the rows that match the query, or something else ?

> I also tried loading in two steps, first, only SELECT __hf_index_id, fts_main_data.match_bm25(__hf_index_id, ?) (which is super fast since it uses the __hf_index_id index) instead of SELECT *, fts_main_data.match_bm25(__hf_index_id, ?) and then join it with the data table but for some reason, it still gets OOM even if the first left table is reduced (Will continue investigating that approach).

That's a great idea, weird that it still OOMs. Does it load everything in memory even if it's going to stash most of the data ? Is there a different method of joining that is smarter ?

> If we don't want to change the sort by row_idx, I think changing this PR to run in parallel by batches can work

It's worth trying, it could help for memory. Though maybe it can be slower because it could do random access to more samples ?

AndreaFrancis:
> Do you know what it load exactly ? Does it load the full table, or only the rows that match the query, or something else ?
It loads the full "data" table because it needs to execute the macro over all the rows to then filter which rows matches the score not null criteria

> Does it load everything in memory even if it's going to stash most of the data ? Is there a different method of joining that is smarter ?

I will go deeper with the duckdb query planner, it does not makes sense to have the full table loaded if using a INNER JOIN but will confirm and get back to you, hopefully this could be a better solution ;)

AndreaFrancis:
> I will go deeper with the duckdb query planner, it does not makes sense to have the full table loaded if using a INNER JOIN but will confirm and get back to you, hopefully this could be a better solution ;)

LOL I was assuming we had a primary key on __hf_index_id but "data" table was a "heap" table (without a clustered index) which is why it is too slow.  
image
I think everything makes sense now, I will see if adding it helps in our current process or better to change to the approach to first get the primary keys that match the search query and then get the other columns.

Issue #2619, opened 2024-03-20
Include authentication information in /compatible-libraries code


severo:
If a dataset is private or gated, we should include authentication in the loading code snippet.

Note that we don't store in datasets-server if a dataset is gated or private. We could, see https://github.com/huggingface/datasets-server/issues/2208. Or we could check the status from the job runner

severo:
cc @lhoestq

lhoestq:
Actually maybe we can have a link to the [huggingface-cli docs](https://huggingface.co/docs/huggingface_hub/guides/cli) for login ? Not sure if we can put it in the code (could be redundant in the case of multiple configs) or maybe in the text before the code snippet ?

severo:
Hmmm, it would work for the datasets library, but every library has potentially a different way to authenticate.

For datasets, pandas and dask, I understand that we rely on hfh, so the same would apply.
For webdatasets and croissant... not sure

lhoestq:
For webdataset we also rely on hfh. We haven't included authentication in
mlcroissant yet

Le jeu. 21 mars 2024, 17:39, Sylvain Lesage ***@***.***> a
écrit :

severo:
Some options:
1. don't do anything and let the user find in the docs how to access the dataset
2. in moon-landing, add the same sentence for all the libraries, shown only if the dataset requires authentication. Something like: `Authenticate with HF before running the code. Check the docs...` Or even show the code for login with hfh.
3. here: prepend a row about the potential need to authenticate, and include the token in the loading method.
4. here: check if the dataset requires a token, and if so: include the authentication in the code. Else leave the same code untouched.

My preferences: 4 > 3 > 2 > 1

lhoestq:
Let's do 4

PR #2635, opened 2024-03-26
Add login comment in code snippets

lhoestq:
Example:

```python
from datasets import load_dataset

# Login using e.g. `huggingface-cli login` to access this dataset
ds = load_dataset("imagenet-1k")
```

The comment is only show for datasets that require a token to be accessed

Close https://github.com/huggingface/datasets-server/issues/2619


severo (approved):
perfect

Issue #2593, opened 2024-03-14
Improve search functionality


AndreaFrancis:
Given that we have already implemented the search feature on the API and hub, let's discuss here what improvements can be made.
Some suggestions:

- [ ] Investigate if it is really used (Maybe it is needed a re design of the idea)
- [ ] Investigate execution time

lhoestq:
Maybe we can improve execution time once more (hf_transfer + more CPU), do some comms et check the usage evolution ?
I'll also ask about it to viewer users at HF

polinaeterna:
i also think it will be beneficial to improve execution time but i don't have any ideas on how to do this...
but i think it's worth working on this before adding any new improvements like other languages

AndreaFrancis:
I did a high-level but detailed investigation about our FTS feature [internal document](https://docs.google.com/document/d/1-6sbntvpoitg2Cn_aXUbzFun4pfvZSBUeR4pACB33hw/edit?usp=sharing)
Summary:
**Is the query not performant enough, and can it be improved?**
Using a new approach with a stage table reduced Memory usage by 2.44% and time execution by 7.9%. 
It might not be too impactful for a 9.1 GB dataset index file but could help for big datasets, so yes, we can use this approach.
I also tried adding a primary key on the __hf_index_id field to speed up the join with the stage table, but looking at the query plan, it always gets a full table scan.
Though I found the reason, using a primary key won't help, as mentioned on the official page https://duckdb.org/docs/guides/performance/schema#microbenchmark-the-effect-of-primary-keys:

> “In this case, primary keys will only have a (small) positive effect on highly selective queries such as when filtering on a single identifier. They do not have an effect on join and aggregation operators.”


> “For best bulk load performance, avoid defining primary key constraints if possible.”

**Are the resources enough, and can they be improved?**
Does the current machine need more Memory?
I don't think so. Running on an isolated m6a.4xlarge and c4.8xlarge with 64 GiB of RAM wasn't too different from the limit we have in the search pod (30 GiB).
Does the current machine need more CPU?
Running the experiments in a machine with a higher CPU **does not** help so, I would say no.
Does the current machine Storage impact?
**Yes**, the current machine with shared storage took too much time (**3 to 4** times more) than when running the same queries on a machine from the same AWS instance but with local storage.
Also see from: https://duckdb.org/docs/guides/performance/environment#disk

[...]

PR #2641, opened 2024-03-27
Apply faster fts by stage table for all datasets

AndreaFrancis:


lhoestq (approved):
Sounds good ! Before we merge, have you checked how much faster it is and/or if the memory usage is better ?

Let's monitor the frequency of OOMs just in case

AndreaFrancis:
I did a high-level analysis of how the solution proposed in this PR would help, it also includes some comparison among the current search pod and isolated m6a.4xlarge and c4.8xlarge machines to evaluate if CPU and other factors affect the performance but I will better explain at https://github.com/huggingface/datasets-server/issues/2593.

Is the old query performant enough, and can it be improved?
Using a new approach with a stage table, Memory usage was reduced by 2.44%, and time execution was reduced by 7.9%. 
See first section of internal doc https://docs.google.com/document/d/1-6sbntvpoitg2Cn_aXUbzFun4pfvZSBUeR4pACB33hw/edit?usp=sharing 
It might not be too impactful for a 9.1 GB dataset index file (which was used in the experiments)  but it could help for big datasets, so yes, we can try this approach.
I also tried adding a primary key on the `__hf_index_id` field to speed up the join with the stage table but looking at the query plan, it always gets a full table scan (instead of an index seek/scan as in other database engines). Though I found the reason, using a primary key won't help, as mentioned on the official page https://duckdb.org/docs/guides/performance/schema#microbenchmark-the-effect-of-primary-keys:

> In this case, primary keys will only have a (small) positive effect on highly selective queries such as when filtering on a single identifier. They do not have an effect on join and aggregation operators.

> For best bulk load performance, avoid defining primary key constraints if possible.

lhoestq:
thanks for the details !

PR #2642, opened 2024-03-27
Update Croissant URL to use the HF Hub `/api`

lhoestq:
I updated the URLs in the compatible-libraries jobs and also in the docs.

cc @severo

PR #2643, opened 2024-03-27
/croissant -> /croissant-crumbs (only specific fields)

### LINKED ISSUE #2624: remove fields from croissant once it's provided by the Hub API
Once the Hub provides an API to access the croissant metadata (https://github.com/huggingface/moon-landing/pull/9106, internal), we will be able to remove the following fields from the croissant response:
- `name`
- `description`
- `license`
- `url`

Also: we could provide `identifier` (DOI) from the Hub API, and remove it here.

Finally: it might be better to rename the datasets server endpoint /croissant-fragment, or something that indicates that it's a partial answer, not the full specification. The endpoint will only provide:
```json
{
  "@context": context,
  "@type": "sc:Dataset",
  "conformsTo": "http://mlcommons.org/croissant/1.0",
  "distribution": distribution,
  "recordSet": record_set
}
```

severo:
fixes #2624

### DIFF
--- a/docs/source/croissant.md
+++ b/docs/source/croissant.md
@@ -14,17 +14,16 @@ The [specification](https://github.com/mlcommons/croissant/blob/main/docs/croiss
 
 ## Get the metadata
 
-This guide shows you how to use Datasets Server's `/croissant` endpoint to retrieve the Croissant metadata associated to a dataset.
+This guide shows you how to use Hugging Face `/croissant` endpoint to retrieve the Croissant metadata associated to a dataset.
 
-
-The `/croissant` endpoint accepts the dataset name as its query parameter:
+The `/croissant` endpoint takes the dataset name in the URL, for example for the `ibm/duorc` dataset:
 
 <inferencesnippet>
 <python>
 ```python
 import requests
 headers = {"Authorization": f"Bearer {API_TOKEN}"}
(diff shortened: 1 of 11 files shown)


severo:
see also https://github.com/huggingface/datasets-server/pull/2642

severo:
fyi @marcenacp @ccl-core

lhoestq:
Maybe keep the bare minimum to still have a valid croissant file ?

severo:
> Maybe keep the bare minimum to still have a valid croissant file ?

Hmmm, I don't think so. I even wanted to rename it /croissant-fragment or /partial-croissant to be crystal clear that it's not a complete croissant file. We don't want people to use this endpoint, but the one on the Hub.

severo:
Or keep the required fields (I have to check which are), and in description, put clearly that the other endpoint should be used.

julien-c:
IMO it's simpler if it's not a valid croissant but only a (mostly internal) "parts of croissant" (croissant-crumbs?)

severo:
funny. Let's do that: /croissant-crumbs in parallel to /croissant, then I'll change in moon-landing, and remove /croissant here

severo:
I rebased on your PR https://github.com/huggingface/datasets-server/pull/2642 @lhoestq.

lhoestq (approved):

PR #2655, opened 2024-04-02
try if search works in big datasets

AndreaFrancis:
Before https://github.com/huggingface/datasets-server/pull/2641, I wanted to try changing the memory limit.
Maybe it is a dummy doubt, but I don't understand why we have many search replicas when there are a low number of requests (we have only 12 for rows, which is highly requested).
image
Isn't it better to have less replicas with a high memory limit so that we avoid OOM? (Not sure how it impacts in terms of infra costs) 
I might be wrong but I chose 7 replicas with a 30 GB memory based on:
- Our biggest index with FTS support is 30 GB size:
image

I am still investigating, but even with the JOIN with a small stage in the left table at https://github.com/huggingface/datasets-server/pull/2641, duckdb looks to be doing a SEQ_SCAN (full table scan) when retrieving all the columns. (I will ask the duckdb team and document here if possible). Doing a SEQ_SCAN implies almost loading the full table, so, if we have a 30 GB database, in the worst scenario, it will be needed a 30 GB of memory allocated.
 

I am not sure if it is ok to calculate this way, but I am assuming that before, we had 24 replicas with a memory limit of 9 GB, which gives  24*9 = 216 GB, and then 216/30 = 7.2 replicas

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -386,7 +386,7 @@ search:
   uvicornNumWorkers: "2"
   nodeSelector:
     role-datasets-server-search: "true"
-  replicas: 24
+  replicas: 7
   service:
     type: NodePort
   resources:
@@ -395,7 +395,7 @@ search:
       memory: "6500Mi"
     limits:
       cpu: 16
-      memory: "9500Mi"
+      memory: "28610Mi"
 
 sseApi:
   # Number of uvicorn workers for running the application
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -392,10 +392,10 @@ search:
   resources:
     requests:
       cpu: 1
-      memory: "6500Mi"
+      memory: "30Gi"
     limits:
       cpu: 16
-      memory: "28610Mi"
+      memory: "30Gi"
 
 sseApi:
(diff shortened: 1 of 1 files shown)


chart/env/prod.yaml
```diff
@@ -395,7 +395,7 @@ search:
       memory: "6500Mi"
     limits:
       cpu: 16
-      memory: "9500Mi"
+      memory: "28610Mi"
```
severo, line 398: Note that the machines (nodes) are defined here: 

https://github.com/huggingface/infra/blob/6ebd3fde4c220f8396de7cc7a884b649a0eb821d/projects/hub/00-aws/variables/hub-aws-prod-us-east-1.tfvars#L187-L196

Currently using m6a.4xlarge (https://aws.amazon.com/ec2/instance-types/m6a/); ie: each node only has 64GiB RAM, we might prefer bigger ones, even if it would work with two pods per node with this change.

severo: Also:

1. we should set the `requests.memory` and `limits.memory` parameters to the same value, because otherwise it will break quickly
2. with 64GiB and two pods per node, we can increase this value to 30Gi or 31Gi (we still have to let some RAM for the host machine)

PR #2657, opened 2024-04-04
Parallelize the backfill cron job

### LINKED ISSUE #2250: Run the backfill script in parallel?
The backfilll script is very slow (one dataset at a time) and also it breaks due to OOM (https://cd.internal.huggingface.tech/applications/argo-cd/datasets-server-prod?view=pods&node=%2FPod%2Fdatasets-server%2Fprod-datasets-server-job-backfill-28412480-2n9ps%2F0)

Maybe we could run the loop iterations in parallel, and also allocate more memory to the pod

severo:
fix #2250

### DIFF
--- a/jobs/cache_maintenance/src/cache_maintenance/backfill.py
+++ b/jobs/cache_maintenance/src/cache_maintenance/backfill.py
@@ -1,8 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
 
+import concurrent.futures
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Optional
 
 from libcommon.dtos import Priority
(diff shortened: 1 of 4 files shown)


AndreaFrancis (approved):

jobs/cache_maintenance/src/cache_maintenance/backfill.py
```diff
@@ -69,39 +73,67 @@ def backfill_retryable_errors(
...
-                        blocked_datasets=blocked_datasets,
-                        hf_token=hf_token,
-                        priority=Priority.LOW,
-                        hf_timeout_seconds=None,
-                        storage_clients=storage_clients,
-                    ),
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_BACKFILL_WORKERS) as executor:
+        futures_to_dataset = {
+            executor.submit(
```
lhoestq, line 115: It looks like the `future` can be a dictionary key, but note that weird objects in python may not be hashable

lhoestq (approved):
lgtm !

jobs/cache_maintenance/src/cache_maintenance/backfill.py
severo: this structure comes from the official docs pages, so I felt free to use it in our code too

severo: anyway, we could remove it, it's just used for a warning log

PR #2674, opened 2024-04-08
Define POETRY variable in docs Makefile

albertvillanova:
Define POETRY variable in docs Makefile so that the right Poetry version is used when parallel versions are installed.

This PR is a follow-up of:
- #2411
  - Commit: 42fe53a3e120de804e72d8d94ff475024152e31a

Note that `--no-root` argument is passed with current Poetry 1.7.1 to avoid the warning:
```
Installing the current project: dataset-viewer-doc (0.1.0)
The current project could not be installed: No file/folder found for package dataset-viewer-doc
If you do not want to install the current project use --no-root
```


severo:
merging main to fix the CI

PR #2676, opened 2024-04-08
Fix and re-enable pip-audit

albertvillanova:
Fix and re-enable pip-audit.


severo (approved):
oh! I had forgotten this

PR #2671, opened 2024-04-08
Fix fsspec and s3fs dependencies

albertvillanova:
This PR:
- Adds explicit dependency on fsspec by using the tilde requirement because fsspec does not use semantic versioning.
- ~~Removes dependency on s3fs because this package is no longer used.~~
- ~~Sets s3fs as dev dependency (also using the tilde requirement) because 2 tests use it (see [comment](https://github.com/huggingface/dataset-viewer/pull/2671#issuecomment-2042277624) below)~~
- Sets the indirect dependence on s3fs through the extras of fsspec: `fsspec[s3]` (see [comment](https://github.com/huggingface/dataset-viewer/pull/2671#issuecomment-2043126529) and [comment](https://github.com/huggingface/dataset-viewer/pull/2671#issuecomment-2043149269) below)

PR #2683, opened 2024-04-09
Make s3fs dependency explicit

albertvillanova:
Make s3fs dependency explicit in libcommon.

Related to:
- #2671

### DIFF
--- a/libs/libcommon/src/libcommon/storage_client.py
+++ b/libs/libcommon/src/libcommon/storage_client.py
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2023 The HuggingFace Authors.
 import logging
-from typing import Any, Optional
+from typing import Literal, Optional, Union, cast
 
 import fsspec
+from fsspec.implementations.local import LocalFileSystem
+from s3fs import S3FileSystem  # type: ignore
 
 from libcommon.config import S3Config
 from libcommon.url_signer import URLSigner
@@ -26,8 +28,8 @@ class StorageClient:
         url_signer (`URLSigner`, *optional*): The url signer to use for signing urls
     """
 
-    _fs: Any
-    protocol: str
+    _fs: Union[LocalFileSystem, S3FileSystem]
+    protocol: Literal["file", "s3"]
     storage_root: str
     base_url: str
     overwrite: bool
@@ -44,7 +46,7 @@ class StorageClient:
     ) -> None:
         logging.info(f"trying to initialize storage client with {protocol=} {storage_root=} {base_url=} {overwrite=}")
         self.storage_root = storage_root
-        self.protocol = protocol
+        self.protocol = cast(Literal["file", "s3"], protocol)
         self.base_url = base_url
         self.overwrite = overwrite
         self.url_signer = url_signer
--- a/libs/libcommon/src/libcommon/storage_client.py
+++ b/libs/libcommon/src/libcommon/storage_client.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2023 The HuggingFace Authors.
 import logging
-from typing import Literal, Optional, Union, cast
+from typing import Optional, Union
 
 import fsspec
 from fsspec.implementations.local import LocalFileSystem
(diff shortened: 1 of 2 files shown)


libs/libcommon/src/libcommon/storage_client.py
```diff
@@ -44,7 +46,7 @@ def __init__(
     ) -> None:
         logging.info(f"trying to initialize storage client with {protocol=} {storage_root=} {base_url=} {overwrite=}")
         self.storage_root = storage_root
-        self.protocol = protocol
+        self.protocol = cast(Literal["file", "s3"], protocol)
```
severo, line 49: maybe even set `protocol: Literal["file", "s3"]` in the constructor, and handle the parsing when we read the environment variables instead.

libs/libcommon/src/libcommon/storage_client.py
```diff
@@ -26,8 +28,8 @@ class StorageClient:
         url_signer (`URLSigner`, *optional*): The url signer to use for signing urls
     """
 
-    _fs: Any
-    protocol: str
+    _fs: Union[LocalFileSystem, S3FileSystem]
```
severo, line 31: much better!

libs/libcommon/src/libcommon/storage_client.py
```diff
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2023 The HuggingFace Authors.
 import logging
-from typing import Any, Optional
+from typing import Literal, Optional, Union, cast
 
 import fsspec
+from fsspec.implementations.local import LocalFileSystem
+from s3fs import S3FileSystem  # type: ignore
```
severo, line 8: it does not support types?

libs/libcommon/src/libcommon/storage_client.py
albertvillanova: The problem with that is that any service instantiating the `StorageClient` should have defined the passing `protocol` variable as `Literal` as well, but most of them define it as `str`. Therefore, mypy would complain:
```
error: Incompatible types in assignment (expression has type "str", variable has type "Literal['file', 's3']")  [assignment]
```

This way, any service can define `protocol` argument as `str` and we handle the cats within the `StorageClient` itself.

libs/libcommon/src/libcommon/storage_client.py
albertvillanova: No, at least the version we are using.

We should check again once we update fsspec/s3fs.

libs/libcommon/src/libcommon/storage_client.py
albertvillanova: I am refactoring the code...

albertvillanova: I have defined the StorageProtocol type and used it in AssetsConfig, CachedAssetsConfig and as parameter in StorageClient.

severo: Yes, I think it's better now. Extra thing (can be in another PR, even if it could enter here too): I think we should validate in the config. Maybe with `env.enum`: https://pypi.org/project/environs/, or the example they show below with marshmallow:

```
# using marshmallow validators
env.str(
    "NODE_ENV",
    validate=OneOf(
        ["production", "development"], error="NODE_ENV must be one of: {choices}"
    ),
)
```

wdyt?

albertvillanova: Done.

severo: Surely more nerd-sniping than a useful thing, but we could try to not repeat the list of valid strings, with something like https://stackoverflow.com/a/78297120/7351594 (only valid in 3.11 though). Alternative: declare the list of values along with `StorageProtocol` to be sure to update both if we support another protocol

```
StorageProtocol = Literal["file", "s3"]
STORAGE_PROTOCOL_VALUES: Tuple[StorageProtocol] = ["file", "s3"]
```

Feel free to ignore, though

severo (approved):

libs/libcommon/src/libcommon/storage_client.py
albertvillanova: Done, although with a slightly different logic.

PR #2686, opened 2024-04-09
Fix `Arrow datatype Extension(...) not supported by Polars` error

polinaeterna:
Some datasets use extension arrow datatypes ([`ArrayXD`](https://github.com/huggingface/datasets/blob/2.18.0/src/datasets/features/features.py#L524)) which are not supported by polars. Using `pl.read_parquet()` or `pl.scan_parquet()` throws an error:
```
"Arrow datatype {dt:?} not supported by Polars. You probably need to activate that data-type feature."
```
even when passing only columns of supported datatypes with `column=[name]` argument.

This PR uses pyarrow to read data directly instead of reading with polars and then pass supported columns to polars.
This PR just ignores such extension columns since it doesn't make much sense to compute anything basic over fixed size arrays, and also this feature is rarely used. This can be extended in a future PR to return for example arrays shape.

will fix https://github.com/huggingface/dataset-viewer/issues/2679


services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -776,20 +778,19 @@ def compute_descriptive_statistics_response(
...
         if isinstance(dataset_feature, list) or (
             isinstance(dataset_feature, dict) and dataset_feature.get("_type") == "Sequence"
         ):
-            schema = pl.scan_parquet(local_parquet_split_glob).schema[dataset_feature_name]
+            first_parquet_file = local_parquet_split_directory / split_parquet_files[0]["filename"]
+            feature_arrow_type = pq.read_schema(first_parquet_file).field(dataset_feature_name).type
             # Compute only if it's internally a List! because it can also be Struct, see
             # https://huggingface.co/docs/datasets/v2.18.0/en/package_reference/main_classes#datasets.Features
-            if isinstance(schema, List):
+            if isinstance(feature_arrow_type, ListType) or isinstance(feature_arrow_type, LargeListType):
```
polinaeterna, line 792: https://github.com/pola-rs/polars/blob/b91dedbae8a204579144a508124c72246473921a/crates/polars-core/src/datatypes/field.rs#L138C15-L138C16

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -847,7 +848,8 @@ def _column_from_feature(
             column_stats = column.compute_and_prepare_response(local_parquet_split_directory)
         else:
             try:
-                data = pl.read_parquet(local_parquet_split_glob, columns=[column.name])
+                table = pq.read_table(local_parquet_split_directory, columns=[column.name])
+                data: pl.DataFrame = pl.DataFrame._from_arrow(table)
```
polinaeterna, line 851: i should have used `pl.from_arrow()` instead of "accessing a protected member" in `pl.DataFrame._from_arrow()` but the former one returns `pl.DataFrame | pl.Series` instead of only `pl.DataFrame` which causes type check fail

AndreaFrancis (approved):
Great!

mariosasko (commented):
A few comments from me :)

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: @severo actually seems that reading with arrow is 1.5 slower - absolute numbers are tenths of a second though. memory consumption is also higher idk why 
image

but i changed the code anyway to use native polars reader (as before) when possible [c190265](https://github.com/huggingface/dataset-viewer/pull/2686/commits/c1902656dc3e6759c728446386e2d23c6d21eb6d)

mariosasko: Yes, they are often used outside of tests, too (e.g. you can find examples in `transformers`, `datasets`, `huggingface_hub`, etc.)

lhoestq: it's also ok to do if not isinstance(...): raise TypeError to calm mypy instead of assert ...

mariosasko: Yes, either this or explicitly tell `bandit` to ignore `assert` on this particular line:
```python
assert isinstance(data, pl.DataFrame)  # type narrowing  # nosec: B101
```

PR #2691, opened 2024-04-10
Allow sorting by column

### LINKED ISSUE #2555: Sort the rows along a column
See https://github.com/huggingface/moon-landing/issues/9201 (internal)

albertvillanova:
Allow sorting by column.

I am proposing a minimal implementation that adds the functionality, by adding a new query parameter to the corresponding endpoints. On this basis we can discuss the approach and suitability for our use case. 

Close #2555.


albertvillanova:
For the corresponding endpoints:
- /filter: the implementation only adds the `ORDER BY` clause to the corresponding SELECT query
  - The pagination works as expected in this case: first all the results are ordered, then sliced in to pages with offset and limit
- /search: does it make sense to sort in this case? Please note that this endpoint already sorts the results according to the search-score. Therefore, reordering the results will make useless the search-score.
- /rows: Current implementation uses the Parquet metadata to select the row groups. I should think about using the "min" and "max" metadata to implement sorting?

albertvillanova:
As discussed last week, let's implement this just on /filter and /first-rows for the moment.

CC: @severo @lhoestq

lhoestq:
Cool ! why /first-rows again ? (sorry I don't remember)

severo:
OK. Implementing sorting on the full dataset (/rows) would be difficult, so we want to allow sorting on the first 5GB at least. We should allow sorting with or without a filter because requiring a filter would not make sense from a user's point of view, IMO. I agree that it would not make sense to sort the search results.

Re /first-rows: the idea of sorting /first-rows was to account for a default sort order in the YAML configuration (e.g., sort by default by `likes` in https://huggingface.co/datasets/severo/trending-repos). We discussed it internally at https://github.com/huggingface/moon-landing/issues/9201#issuecomment-2048277255 -> I think we should not propose a default sorting order for now.

So, maybe we need to:
- use /filter without a "where" parameter but with a "sort" parameter when we want to sort the rows. This way, we use the 5GB duckdb file. So: "where" and "sort" would be optional on /filter. 
- make it explicit in the UI that the sort is done only on the first 5GB (if the full dataset is bigger). Related to https://github.com/huggingface/moon-landing/issues/9429 (internal): we don't always show that statistics, filter, search are done on the first 5GB.

albertvillanova:
Thanks for the explanation, @severo. Much clearer now.

albertvillanova:
I think this PR is ready for review @huggingface/dataset-viewer.

severo (approved):
wow, excellent!

albertvillanova:
Maybe we could merge this PR today... @huggingface/dataset-viewer.

severo:
Sure, feel free to merge when there is one approval (or none if you're confident)

lhoestq (approved):
Awesome !

Issue #2690, opened 2024-04-10
Increase in timeouts from the Hub?


severo:
It seems like today a lot of datasets show `The dataset viewer is taking too long to fetch the data`, and only work after some refreshes.

<img width="1016" alt="Capture d’écran 2024-04-10 à 10 15 38" src="https://github.com/huggingface/dataset-viewer/assets/1676121/d9e7738b-48bf-4817-b8a1-b16216332323">

See the video below. Every refresh can have a different result : no stats, no rows, or everything can be working as expected.

https://github.com/huggingface/dataset-viewer/assets/1676121/206f4a45-46ba-49d7-b5a4-135dda484f64

severo:
I guess it's an issue with the RAM on the API service (already reported here: https://github.com/huggingface/dataset-viewer/issues/2395).

<img width="1379" alt="Capture d’écran 2024-04-10 à 10 25 04" src="https://github.com/huggingface/dataset-viewer/assets/1676121/f0e24529-7f9b-46e1-8dd7-0735b0eb1bfa">

For now, let's just "restart" them.

severo:
Happening again.

@XciD @rtrompier could we increase the resources for the API pods? (`prod-datasets-server-api-...`) And maybe isolate in their own "role" instead of sharing with the other pods, like we did for the workers, search and filter pods?

severo:
re increase of resources:
- https://github.com/huggingface/infra/pull/940 (internal)
- https://github.com/huggingface/dataset-viewer/pull/2693

severo:
closing as a duplicate of #2395

PR #2693, opened 2024-04-10
increase resources for API service

severo:
Related to #2690

See https://github.com/huggingface/infra/pull/940 (internal) which provides more resources, in a specific new "role" (role-datasets-server-api)

It provides up to 6 m6a.4xlarge nodes, each of which has 16 vCPUs and 64GB RAM. So: each node should get 4 pods, we should use 3 nodes, and have room to scale up twice

---

blocked by https://github.com/huggingface/infra/pull/940

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -347,17 +347,17 @@ api:
   # https://docs.gunicorn.org/en/stable/design.html#how-many-workers
   uvicornNumWorkers: "9"
   nodeSelector:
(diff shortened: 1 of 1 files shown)


rtrompier (approved):

rtrompier:
Nodes are ready for production ;)

severo:
Merging, as https://github.com/huggingface/infra/pull/940 is deployed in production

PR #2697, opened 2024-04-11
Update duckdb from 0.10.0 to 0.10.1

### LINKED ISSUE #2604: Upgrade to duckdb 0.10.1
https://github.com/duckdb/duckdb/releases/tag/v0.10.1

albertvillanova:
Update duckdb from 0.10.0 to 0.10.1.

Fix #2604.


AndreaFrancis (changes requested):
I added the version to the downloads folder to make it easier to delete a whole version when recomputing all the indexes.
The current code uses [DUCKDB_VERSION](https://github.com/huggingface/dataset-viewer/blob/main/libs/libcommon/src/libcommon/constants.py#L12) to create the "downloading" folder for search, but it is currently set to 0.10.0. Changing to 0.10.1 does not make sense for all the indexes since some may have been created using the old version (at least we recompute all indexes again).
Maybe use the duckdb_version from [here](https://github.com/huggingface/dataset-viewer/blob/main/services/worker/src/worker/dtos.py#L148) and send it to [here](https://github.com/huggingface/dataset-viewer/blob/main/libs/libapi/src/libapi/duckdb.py#L67).

severo:
> The current code uses [DUCKDB_VERSION](https://github.com/huggingface/dataset-viewer/blob/main/libs/libcommon/src/libcommon/constants.py#L12) to create the "downloading" folder for search, but it is currently set to 0.10.0.

But anyway, these directories are temporary, and can be deleted without any impact, right? Maybe we should remove the duckdb version from the path. Or did I miss something?

severo:
OK, it was because during the migration we had both versions at the same time. Now that the duckdb format is backward-compatible, maybe it's better to remove the version from the path.

albertvillanova:
Thanks a lot for your comments, @AndreaFrancis and @severo.

I am addressing them.

albertvillanova:
Please, let me know if I addressed them properly, @AndreaFrancis and @severo.

services/worker/src/worker/dtos.py
```diff
@@ -145,7 +145,6 @@ class SplitDuckdbIndex(SplitHubFile):
     partial: Optional[bool]
     num_rows: Optional[int]
     num_bytes: Optional[int]
-    duckdb_version: str
```
severo, line 148: hmmm good question: do we want to remove this field? I think we can keep it for now

albertvillanova: We can keep it even if it is no longer used, for the moment...

severo: yes. The thing is all the entries in the db have it, so... let's not change that, I think, and it can be useful for stats purposes

albertvillanova: Done.

AndreaFrancis (approved):
Thanks!
After deploying, the service won't be able to find the indexes under /0.10.0/ folder, we might leave downloading everything is needed again and delete the old folder. Not sure if there is a fast way to move from /storage/duckdb-index/downloads/0.10.0/* to /storage/duckdb-index/downloads/*

severo (approved):

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -314,7 +311,7 @@ def compute_split_duckdb_index_response(
         partial=partial,
         num_rows=num_rows,
         num_bytes=num_bytes,
-        duckdb_version=DUCKDB_VERSION,
+        duckdb_version=duckdb.__version__,
```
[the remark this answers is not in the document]
albertvillanova: Yes, I checked: `"0.10.1"`

severo:
>  After deploying, the service won't be able to find the indexes under /0.10.0/ folder, we might leave downloading everything is needed again and delete the old folder. Not sure if there is a fast way to move from /storage/duckdb-index/downloads/0.10.0/* to /storage/duckdb-index/downloads/*

Let's just manually delete `downloads/` and `cache/`, no? And let the service download the duckdb files it needs.

PR #2699, opened 2024-04-11
Update numba from 0.56.4 to 0.59.1

### LINKED ISSUE #2698: Update numba
Numba dependency was introduced to the exact 0.56.4 version in: https://github.com/huggingface/dataset-viewer/pull/1026/files#diff-e127bcf65702533dbd7e316815e4b116552c3adef5d4d50db54147d430d868edR24
- #1026

If there is not a strong requirement to use that specific version, we should update it.

albertvillanova:
Update numba from 0.56.4 to 0.59.1.

Fix #2698.

### DIFF
--- a/front/admin_ui/poetry.lock
+++ b/front/admin_ui/poetry.lock
@@ -1445,7 +1445,6 @@ huggingface-hub = {version = "^0.21.4", extras = ["hf-transfer"]}
 mongo-types = "0.15.1"
 mongoengine = "^0.27.0"
 networkx = "^3.0"
-numba = "0.56.4"
 orjson = "^3.9.15"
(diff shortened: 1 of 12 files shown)


albertvillanova:
I have removed the explicit dependence on numba-0.56.4 from libcommon/pyproject.toml, as we do not use it in our code.
- Note the dependence is now indirect via: datasets > librosa > numba

Before merging, I would suggest waiting for @lhoestq reply on: https://github.com/huggingface/dataset-viewer/issues/2698#issuecomment-2049108864

albertvillanova:
It seems this PR could be merged without any further impact, @huggingface/dataset-viewer.

severo (approved):

PR #2701, opened 2024-04-11
Search service with local storage

AndreaFrancis:
Part of https://github.com/huggingface/dataset-viewer/issues/2593
When migrating to local storage, from dataset-viewer side, it is just needs to:
- [x] Point to the local storage
- [x] Change replicas and uvicornNumWorkers according to AWS instance type provisioned by the infra team (r6id.8xlarge)
- [x] Unmount all references to volumeDuckDBIndex
- [x] Move cleanDuckdbIndexCache cron job to run inside search service pod and reduce time interval from 3 days to something shorter
- [x] In search service, return  and error if storage is out of space

From infra side I think it is required to:
- [x] Change AWS instance type (r6id.8xlarge)
- [x] Mechanism to redirect requests to the same pod -> https://github.com/huggingface/dataset-viewer/pull/2708/files

image

Does it sound like a plan? @rtrompier @severo @lhoestq


lhoestq:
Sounds like a good plan !

services/search/src/search/routes/filter.py
```diff
@@ -153,6 +156,10 @@ async def filter_endpoint(request: Request) -> Response:
                         offset,
                         extensions_directory,
                     )
+                with StepProfiler(method="search_endpoint", step="clean old indexes"):
+                    # no need to do it every time
+                    if random.random() < clean_cache_proba:  # nosec
+                        clean_dir(str(duckdb_index_file_directory), expiredTimeIntervalSeconds)
```
severo, line 162: OK. It was the logic in /rows if I remember well.
We're profiling, so, we can assess if it impacts a lot the response time or not. But, if it does, maybe we should change the logic a bit, and run a regular job in parallel on the search pod (as is done in services/workers: the zombie killers, the long jobs killer, etc)

chart/values.yaml
```diff
@@ -337,42 +335,6 @@ backfillRetryableErrors:
...
-  nodeSelector: {}
-  resources:
-    requests:
-      cpu: 0
-    limits:
-      cpu: 0
-  tolerations: []
-  # the time interval at which cached downloaded data files will be considered as expired and will be deleted
-  expiredTimeIntervalSeconds: 600
-  subfolderPattern: "downloads/*"
```
severo, line 374: currently, are we still using both cache/ and downloads/? Only for the search service?

AndreaFrancis: Yes, it is only for search service using the shared storage.

services/search/src/search/routes/filter.py
AndreaFrancis: Yes, it's a simple logic initially. I am estimating to retain only 12 hours of data initially, which should take at most 1 minute. Agree, in case of taking too much time, to change to a parallel process.

libs/libapi/src/libapi/duckdb.py
```diff
@@ -67,25 +72,39 @@ def get_download_folder(root_directory: StrPath, dataset: str, revision: str, co
...
-        repo_type=REPO_TYPE,
-        revision=target_revision,
-        repo_id=dataset,
-        filename=repo_file_location,
-        local_dir=index_folder,
-        hf_token=hf_token,
-        cache_dir=cache_folder,
-    )
+    if not available_disk_space(cache_folder, size_bytes):
+        raise DownloadIndexError("Unable to download index file")  # Should we show the remaining disk space?
```
AndreaFrancis, line 93: I was unsure what exactly we would like to show the user. Getting a lack of storage should be something to alert in our internal process.

libs/libapi/src/libapi/duckdb.py
```diff
@@ -67,25 +72,39 @@ def get_download_folder(root_directory: StrPath, dataset: str, revision: str, co
...
+            revision=target_revision,
+            repo_id=dataset,
+            filename=repo_file_location,
+            local_dir=index_folder,
+            hf_token=hf_token,
+            cache_dir=cache_folder,
+        )
+    except OSError as err:
+        if err.errno == errno.ENOSPC:
+            raise DownloadIndexError("Unable to download index file", err)
```
AndreaFrancis, line 107: Same here. Even if the process had enough space before downloading, other workers could download files, which might lead to a lack of space. Again, I am unsure about what to say to the user.

services/search/src/search/routes/filter.py
severo: > which should take at most 1 minute

it means that a user who is launching a search will have to wait 1 minute for the response, because of our maintenance operation.

If the duration were like 2 seconds, then the strategy would be OK. But if we expect regularly 1-minute durations, I think we should change the implementation, and run the maintenance operations in parallel.

libs/libapi/src/libapi/duckdb.py
severo: It returns a `HTTPStatus.INTERNAL_SERVER_ERROR` error (500), which is OK. We also have the `DownloadIndexError` error code, so, if a user is reporting this, we will know what occurs

severo: But indeed, we could improve the text a bit, to make it clearer for the user (here and in the other comment). As it refers to the internal, maybe catch this specific exception in the /search service (and in the other ones in a similar way) and return another exception with: `Cannot perform the search due to a lack of disk space on the server. Please report the issue.`

severo: And I imagine there is a way to see with the infra to get alerts before this occurs

services/search/src/search/routes/filter.py
AndreaFrancis: Yeah, even currently, we have requests that only take 2 minutes to answer in the index processing. Considering the slow demand for the requests, initially, it will not have a high impact on the first version of local storage. Could we handle it in another PR?

libs/libapi/src/libapi/duckdb.py
AndreaFrancis: Done; I changed it to `Cannot perform the operation due to a lack of disk space on the server. Please report the issue.` because it it used for filter and search.

AndreaFrancis:
Not sure why the e2e is failed, but I was able to run the failed ones:
image

services/search/src/search/routes/filter.py
severo: yes, sure

severo:
Yes, e2e often fails, and works on a second intent. We surely should look more in details why

libs/libapi/src/libapi/duckdb.py
```diff
@@ -60,13 +64,27 @@ async def get_index_file_location_and_download_if_missing(
...
     subdirectory = "".join([c if re.match(r"[\w-]", c) else "-" for c in f"{dataset}-{hash_suffix}"])
     return f"{root_directory}/{DUCKDB_INDEX_DOWNLOADS_SUBDIRECTORY}/{DUCKDB_VERSION}/{subdirectory}"
 
 
+def available_disk_space(path: StrPath, required_space: int) -> bool:
+    disk_stat = os.statvfs(path)
+    # Calculate free space in bytes
+    free_space = disk_stat.f_bavail * disk_stat.f_frsize
+    logging.debug(f"{free_space} available space, needed {required_space}")
+    return free_space >= required_space
```
severo, line 85: nit (feel free to ignore): we could also rename to `check_available_disk_space` and raise DownloadIndexError here, instead of returning a boolean. It would simplify the code a bit

severo (approved):
Note that the EFS currently contains 14TB! Are we sure the files were deleted correctly? It seems a lot. Let's keep an eye on the size of the search pods and the new `DownloadIndexError`!

Capture d’écran 2024-04-12 à 18 13 29

services/search/src/search/routes/filter.py
severo: (maybe let's open an issue to keep track of it)

severo:
👏

Capture d’écran 2024-04-12 à 18 26 44

AndreaFrancis:
Note that the EFS currently contains 14TB! Are we sure the files were deleted correctly?

It looks weird for me, would like to monitor the process, will force temporary to run each 10 minutes: https://github.com/huggingface/dataset-viewer/pull/2716 It was previously scheduled at   schedule: "0 0 * * *"

services/search/src/search/routes/filter.py
AndreaFrancis: Done https://github.com/huggingface/dataset-viewer/issues/2718

libs/libapi/src/libapi/duckdb.py
AndreaFrancis: Done. Indeed, it is simpler, thanks!

AndreaFrancis:
> Note that the EFS currently contains 14TB! Are we sure the files were deleted correctly? It seems a lot. Let's keep an eye on the size of the search pods and the new DownloadIndexError!

It now is only 40 GB
Screenshot from 2024-04-12 13-48-14

severo:
It's more than 1TB again

Capture d’écran 2024-04-15 à 12 35 46

lhoestq:
Note that there may be two directories storing the indexes: the download directory where we read them from, and also maybe the cache directory for `huggingface_hub` where it stores the cached git blobs. Could it be the `huggingface_hub` cache contains the indexes files too and is not cleaned ?

AndreaFrancis:
cache uses 261 M at this time (Maybe there was an ongoing download)
image

For the downloads folder, it has been since Friday because there isn't any folder with last_access_datetime more than three days ago. We can reduce this interval to something short, like 2 or 1 day if needed.
Note that:
- If we reduce it to 1 day, it will delete 1186 folders
- If we reduce it to 2 days, it will delete 638 folder

Issue #468, opened 2022-07-25
Remove reverse-proxy, and use ALB/ingress instead?


severo:
It would remove a single point of failure.

Also: we could have one healthcheck per service (API, admin). Currently the `ELB-HealthChecker/2.0` healthcheck is responded by the API (/healthcheck), and it will not alert if the admin service is down (even if we can still monitor it with betteruptime on /admin/healthcheck).

severo:
>  Well we can't as you use cache and ALB don't have this feature.

this "cache" does not exist anymore in our nginx reverse proxy, so we can reopen

severo:
Replacing nginx with ALB can help routing the same user to the same pod, for example. It could be useful for the search pods. cc @rtrompier @AndreaFrancis (see internal discussion: https://huggingface.slack.com/archives/C04L6P8KNQ5/p1712821819546829?thread_ts=1712662138.160929&cid=C04L6P8KNQ5)

severo:
Done.

PR #2708, opened 2024-04-12
feat(chart): search api exposed directly with ALB

rtrompier:
As discussed in #468, let's start to remove nginx proxy.
In this PR, we begin with the search API.

### DIFF
--- a/chart/nginx-templates/default.conf.template
+++ b/chart/nginx-templates/default.conf.template
@@ -64,6 +64,15 @@ server {
     proxy_http_version 1.1;
   }
 
+  location /search {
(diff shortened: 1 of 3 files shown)


severo (commented):
Thanks for working on this.

I don't know much about k8s ingress, so... I cannot really review the ingress.yaml file.

But what is sure is that we have to include some type of conditional in the nginx template, because we rely on it in the e2e tests.

Another way might be to have a specific nginx template just for the e2e tests.

What do you think?

rtrompier:
Ok, so let's keep the proxy for now.
It will be bypassed by the ALB once deployed.

severo (approved):

PR #2710, opened 2024-04-12
fix(chart): search api exposed directly with ALB

rtrompier:


chart/env/prod.yaml
```diff
@@ -389,6 +390,12 @@ search:
   replicas: 7
   service:
     type: NodePort
+  ingress:
+    enabled: true
+    annotations:
+      alb.ingress.kubernetes.io/group.order: "1"
+      alb.ingress.kubernetes.io/target-node-labels: role-datasets-server-search=true
+      alb.ingress.kubernetes.io/target-group-attributes: stickiness.enabled=true,stickiness.lb_cookie.duration_seconds=300
```
severo, line 398: ok, but I don't think we pass cookies anyway

chart/env/prod.yaml
```diff
@@ -389,6 +390,12 @@ search:
   replicas: 7
   service:
     type: NodePort
+  ingress:
+    enabled: true
+    annotations:
+      alb.ingress.kubernetes.io/group.order: "1"
```
severo, line 396: OK! this was the missing part, right?

severo (approved):
OK!

Tbh, I didn't understand the manipulation on `datasetsServer.instance.ingress.annotations`

AndreaFrancis (approved):
Sorry, I can't give helpful feedback but it looks good overall.

chart/env/prod.yaml
rtrompier: Yep

chart/env/prod.yaml
rtrompier: let's try ;) 
The LB should create a cookie during the first call to remember the instance. During the 2nd call, the ALB will read the cookie, and redirect to the same instance.

severo: Ah ok!

PR #2712, opened 2024-04-12
Stats for images

polinaeterna:
Compute histograms over images widths


severo (approved):
Nice! + good refactoring

PR #2713, opened 2024-04-12
feat(chart): expose all services through ALB

rtrompier:
- [x] disable nginx proxy
- [x] expose `/` on `api` instances
- [x] expose `/admin` on `admin` instances
- [x] expose `/rows` on `rows` instances
- [x] expose `/search` on `search` instances
- [x] expose `/filter` on `search` instances
- [x] expose `/sse` on `sseApi` instances
- [x] expose `/openapi.json` on github
- [ ] customise 404.html error page ?

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -336,7 +336,6 @@ admin:
     annotations:
       alb.ingress.kubernetes.io/group.order: "1"
       alb.ingress.kubernetes.io/target-node-labels: role-datasets-server=true
-      alb.ingress.kubernetes.io/actions.openapi-redirect: '{"Type":"redirect","RedirectConfig":{"Host":"raw.githubusercontent.com","Path":"/huggingface/dataset-viewer/main/docs/source/openapi.json","Port":"443","Protocol":"HTTPS","Query":"#{query}","StatusCode":"HTTP_307"}}'
   resources:
     requests:
       cpu: 1
@@ -361,8 +360,8 @@ api:
   ingress:
     enabled: true
     annotations:
-      alb.ingress.kubernetes.io/group.order: "2"
       alb.ingress.kubernetes.io/target-node-labels: role-datasets-server-api=true
+      alb.ingress.kubernetes.io/actions.openapi-redirect: '{"Type":"redirect","RedirectConfig":{"Host":"raw.githubusercontent.com","Path":"/huggingface/dataset-viewer/main/docs/source/openapi.json","Port":"443","Protocol":"HTTPS","Query":"#{query}","StatusCode":"HTTP_307"}}'
   resources:
     requests:
       cpu: 4
@@ -385,7 +384,7 @@ rows:
   ingress:
     enabled: true
     annotations:
-      alb.ingress.kubernetes.io/group.order: "3"
+      alb.ingress.kubernetes.io/group.order: "2"
       alb.ingress.kubernetes.io/target-node-labels: role-datasets-server-rows=true
   resources:
     requests:
@@ -409,7 +408,7 @@ search:
   ingress:
     enabled: true
     annotations:
-      alb.ingress.kubernetes.io/group.order: "4"
+      alb.ingress.kubernetes.io/group.order: "3"
       alb.ingress.kubernetes.io/target-node-labels: role-datasets-server-search=true
       alb.ingress.kubernetes.io/target-group-attributes: stickiness.enabled=true,stickiness.lb_cookie.duration_seconds=300
       alb.ingress.kubernetes.io/target-type: ip
@@ -434,7 +433,7 @@ sseApi:
   ingress:
     enabled: true
     annotations:
-      alb.ingress.kubernetes.io/group.order: "5"
+      alb.ingress.kubernetes.io/group.order: "4"
       alb.ingress.kubernetes.io/target-node-labels: role-datasets-server=true
   resources:
     requests:
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -331,6 +331,11 @@ admin:
   replicas: 2
   service:
     type: NodePort
+  ingress:
+    enabled: true
+    annotations:
+      alb.ingress.kubernetes.io/group.order: "3"
+      alb.ingress.kubernetes.io/target-node-labels: role-datasets-server=true
   resources:
     requests:
       cpu: 1
@@ -352,6 +357,11 @@ api:
   replicas: 12
   service:
     type: NodePort
+  ingress:
+    enabled: true
+    annotations:
+      alb.ingress.kubernetes.io/group.order: "3"
+      alb.ingress.kubernetes.io/target-node-labels: role-datasets-server-api=true
   resources:
     requests:
       cpu: 4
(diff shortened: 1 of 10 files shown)


chart/templates/services/sse-api/ingress.yaml
```diff
@@ -0,0 +1,33 @@
...
+  rules:
+    - host: {{ include "datasetsServer.ingress.hostname" . }}
+      http:
+        paths:
+          - backend:
+              service:
+                name: "{{ include "name" . }}-sse-api"
+                port:
+                  name: http
+            path: /sse
```
AndreaFrancis, line 20:
```suggestion
            path: /hub-cache
```

AndreaFrancis:
>  expose assets and cached-assets ? Should not be served by API anymore right ?

Yes, it should be served by S3+CloudFront if I am not wrong

rtrompier:
@severo can you please tell me if `openapi.json` endpoint is still necessary ? 
And if we have to customize the 404 error page ?

severo:
Yes, we need it somewhere. Not necessarily at that URL though, if it's an issue. re 404: not important at all

severo:
re: openapi.json. Ideally, it would be served from the docs. See the previous discussions here: https://github.com/huggingface/dataset-viewer/commit/cf8042da9a2325a1d5706dbecded9b187d5cc8db and here: https://github.com/huggingface/doc-builder/issues/312#issuecomment-1675099444. So: not a problem at all if we move its URL, we would just have to update it in the docs.

chart/env/prod.yaml
```diff
@@ -331,6 +331,11 @@ admin:
   replicas: 2
   service:
     type: NodePort
+  ingress:
+    enabled: true
+    annotations:
```
severo, line 336: do we need the same annotations in the staging.yaml env?

severo: if so, we could delete the reverse-proxy from Helm altogether

chart/env/prod.yaml
```diff
@@ -331,6 +331,11 @@ admin:
...
+      alb.ingress.kubernetes.io/group.order: "3"
```
severo, line 337: is it possible to have several times the same group order (3)?

chart/templates/services/search/ingress.yaml
```diff
@@ -19,6 +19,16 @@ spec:
                   name: http
             path: /search
             pathType: Prefix
+    - host: {{ include "datasetsServer.ingress.hostname" . }}
```
severo, line 22: do we need to create a new "rule"? Isn't it only a new "path" in the previous rule?

severo (commented):
I'm wondering: would it make sense to have all the paths in the same ingress config file? or is it better to have one file per service?

rtrompier:
> re: openapi.json. Ideally, it would be served from the docs. See the previous discussions here: [cf8042d](https://github.com/huggingface/dataset-viewer/commit/cf8042da9a2325a1d5706dbecded9b187d5cc8db) and here: [huggingface/doc-builder#312 (comment)](https://github.com/huggingface/doc-builder/issues/312#issuecomment-1675099444). So: not a problem at all if we move its URL, we would just have to update it in the docs.

We can do the redirect with a custom annotation. So let me know if you want to move it or not ;)

rtrompier:
> I'm wondering: would it make sense to have all the paths in the same ingress config file? or is it better to have one file per service?

Yes we can. Do you think it will be easier for you? 
IMO it's better to keep it separate, in case you want to have a dedicated URL (ex: api.datasets.hf.co)

chart/env/prod.yaml
rtrompier: Yes done.
I kept the reverse proxy for now in case we need to re-enable it. But once deployed, yes we should be able to remove it.

chart/templates/services/admin/ingress.yaml
```diff
@@ -0,0 +1,33 @@
...
+      http:
+        paths:
+          - backend:
+              service:
+                name: "{{ include "name" . }}-admin"
+                port:
+                  name: http
+            path: /admin
+            pathType: Prefix
+  {{- if include "hf.common.ingress.certManagerRequest" ( dict "annotations" $annotations ) }}
```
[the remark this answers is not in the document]
rtrompier: we can merge all the rule in a same file, and have a rule by path.
For example for moon-landing, we have one ingress by component as well

chart/templates/services/search/ingress.yaml
severo: OK!

severo:
> We can do the redirect with a custom annotation. So let me know if you want to move it or not ;)

Please let's keep it unchanged if it's possible, and we can move it later in a dedicated PR

severo:
We should finish this PR. Happy to give you a hand @rtrompier

rtrompier:
> We should finish this PR. Happy to give you a hand @rtrompier

Thx for the reminder. I plan to finish it in the week.

chart/templates/services/admin/ingress.yaml
```diff
@@ -0,0 +1,31 @@
...
+        paths:
+          - path: /admin
+            backend:
+              service:
+                name: "{{ include "name" . }}-admin"
+                port:
+                  name: http
+
+            pathType: Prefix
+          {{ if hasKey $annotations "alb.ingress.kubernetes.io/actions.openapi-redirect" -}}
```
severo, line 23: I think it's not very important, but to avoid problems later (eg if we protect /admin in some way), can we move the openapi redirect to api/ingress.yaml instead?

chart/env/prod.yaml
```diff
@@ -352,6 +358,11 @@ api:
   replicas: 12
   service:
     type: NodePort
+  ingress:
+    enabled: true
+    annotations:
+      alb.ingress.kubernetes.io/group.order: "2"
```
severo, line 364: as api ingress will take everything at the root, we should put it as the last group, so: 5, and move the other ones.

same for staging

severo: an alternative is to explicitly list all the api endpoint, but there are many

chart/templates/services/api/ingress.yaml
```diff
@@ -0,0 +1,23 @@
...
+        paths:
+          - backend:
+              service:
+                name: "{{ include "name" . }}-api"
+                port:
+                  name: http
+            path: /
+            pathType: Prefix
+{{- include "ingress.tls" (merge (dict "annotations" $annotations) $ ) | indent 2}}
+{{- end }}
```
severo, line 23: we have to forbid /metrics here, I think

chart/templates/services/admin/ingress.yaml
rtrompier: done

chart/env/prod.yaml
rtrompier: Removed the annotation. Will use the default group.order (100)

chart/templates/services/api/ingress.yaml
rtrompier: The best way to do that is too expose metrics on another port and not expose it.
But let's try to block the path at the ALB level.

severo: hmmm ok. https://github.com/huggingface/dataset-viewer/issues/2812

severo (approved):
I think it's good now, right?

rtrompier:
> I think it's good now, right?

I think so yes :)

PR #2716, opened 2024-04-12
Force clean downloads folder each 10 minutes to monitor

AndreaFrancis:


severo (approved):

AndreaFrancis:
It looks like since adding duckdb 0.10.0 to the storage path, it has not deleted the old files because the folder inside /storage/duckdb-index/downloads/* 0.10.0 has always had a recent last accessed time LOL.
Hopefully, https://github.com/huggingface/dataset-viewer/pull/2697 fixed it by downloading the individual split folders at the downloads level.

PR #2731, opened 2024-04-16
Set non-package mode

albertvillanova:
Once that we have updated Poetry to 1.8.2, we can set non-package mode in the components we are not packaging:
- docs
- admin-ui
- e2e

Note that currently a warning was raised when installing (poetry@1.8.2 install) the docs and the admin-ui components:
```
Warning: The current project could not be installed: No file/folder found for package dataset-viewer-doc
If you do not want to install the current project use --no-root.
If you want to use Poetry only for dependency management but not for packaging, you can disable package mode by setting package-mode = false in your pyproject.toml file.
In a future version of Poetry this warning will become an error!
```

### DIFF
--- a/.github/workflows/_quality-python.yml
+++ b/.github/workflows/_quality-python.yml
@@ -40,15 +40,19 @@ jobs:
         run: |
           poetry env use "${{ env.python-version }}"
           poetry install
-      - name: Run ruff linter on src
-        run: poetry run ruff check src
-      - name: Run ruff linter on tests
-        run: poetry run ruff check tests --ignore=ARG
+      - name: Run ruff linter
+        run: |
+          if [ -d src ]; then poetry run ruff check src; fi
+          if [ -d tests ]; then poetry run ruff check tests --ignore=ARG; fi
       - name: Run ruff formatter
-        run: poetry run ruff format --check src tests
+        run: |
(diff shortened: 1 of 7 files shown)


severo (approved):
Good idea. I imagine that we can now delete e2e/src? I created it to remove the warning

severo:
e2e CI: we can't run ruff on src/ since we deleted it... I don't know if it's worth the pain handling this case in the github actions, maybe better to revert https://github.com/huggingface/dataset-viewer/pull/2731/commits/1b0a9879f3df7759b9f93c6130b172d294ebd75b

.github/workflows/_quality-python.yml
```diff
@@ -40,15 +40,19 @@ jobs:
         run: |
           poetry env use "${{ env.python-version }}"
           poetry install
-      - name: Run ruff linter on src
-        run: poetry run ruff check src
-      - name: Run ruff linter on tests
-        run: poetry run ruff check tests --ignore=ARG
+      - name: Run ruff linter
+        run: |
+          if [ -d src ]; then poetry run ruff check src; fi
```
severo, line 45: excellent!

severo (approved):

albertvillanova:
@severo I am merging this PR, after the fix in the Makefiles and the addition of py.typed to cache_maintenance.

PR #2734, opened 2024-04-17
Fix search storage cleaning

AndreaFrancis:
Before moving storage cleaning to a parallel process in https://github.com/huggingface/dataset-viewer/issues/2718, there are some missing details and issues with the current process to fix:
- I removed `clean-directory` action from cache maintenance since we don't use this task anymore (we can add it later if needed)
- I found that clean_dir wasn't deleting the folder with a file that was last accessed time before X seconds ago. Index files change the last accessed time when being read by /search or /filter, but its parent folder didn't, so there wasn't a real cleaning when using the glob pattern; I also modified the test to cover this scenario.


services/search/src/search/routes/search.py
```diff
@@ -211,9 +211,13 @@ async def search_endpoint(request: Request) -> Response:
                         extensions_directory,
                     )
                     # no need to do it every time
+                    # TODO: Will be moved to another process in parallel
```
severo, line 214: I'm not a fan of putting TODOs in the code (I also do it too often, but I think it's a bad idea because it promises something that might never occur). We already have an open issue for it.

services/search/src/search/routes/filter.py
```diff
@@ -159,9 +159,13 @@ async def filter_endpoint(request: Request) -> Response:
                         extensions_directory,
                     )
                     # no need to do it every time
+                    # TODO: Will be moved to another process in parallel
```
severo, line 162: same comment as below

severo (approved):

PR #2735, opened 2024-04-17
Change search resources

AndreaFrancis:
The current instance type for the search service is [r6id.16xlarge](https://github.com/huggingface/infra/blob/1288ec190493ef0da24cbbc5422ea94ed84924ea/projects/hub/00-aws/variables/hub-aws-prod-us-east-1.tfvars#L192C24-L192C37)
image
I think we can adjust the new values for CPU and memory and also reduce the number of replicas (we still have a low number of requests)

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -377,7 +377,7 @@ search:
   uvicornNumWorkers: "2"
   nodeSelector:
     role-datasets-server-search: "true"
-  replicas: 4
+  replicas: 3
   service:
     type: NodePort
   ingress:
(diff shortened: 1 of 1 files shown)


rtrompier (approved):

chart/env/prod.yaml
```diff
@@ -377,7 +377,7 @@ search:
   uvicornNumWorkers: "2"
   nodeSelector:
     role-datasets-server-search: "true"
-  replicas: 4
+  replicas: 3
```
severo, line 380: nit: as 250Gi means two pods per node, I imagine we can have an even number of replicas: 2 (1 node) or 4 (2 nodes)

severo (approved):

chart/env/prod.yaml
severo: well nevermind, OK to reduce to 3, let's see.

PR #2737, opened 2024-04-18
Store transformed values in duckdb index file

polinaeterna:
Add the following columns:
* `{column_name}.length` - for lists and strings (**all strings**, both string-strings and class-strings)
* `{column_name}.duration` - for audios
* `{column_name}.width` and `{column_name}.height` - for images


severo:
If we're going to recompute the duckdb files, I wonder if we could look again at the following possibility:
- storing the indexes and additional columns into the .duckdb file,
- while using the original parquet files to store the data

I don't know if it could work. If so, it could also make the requests slower or more complex. But it might reduce the cold-start latency of /filter and /search, as well as the local storage needs.

AndreaFrancis:
> storing the indexes and additional columns into the .duckdb file,

I think that could be possible if we create the data table and then the index and finally remove the data table.
But should reduce disk usage in theory, but I have seen that duckdb has an interesting way ot managing "drops" actions. 
See [Limitations on Reclaiming Disk Space](https://duckdb.org/docs/sql/statements/drop.html#limitations-on-reclaiming-disk-space):

> Running DROP TABLE should free the memory used by the table, but not always disk space. Even if disk space does not decrease, the free blocks will be marked as “free”. For example, if we have a 2 GB file and we drop a 1 GB table, the file might still be 2 GB, but it should have 1 GB of free blocks in it. To check this, use the following PRAGMA and check the number of free_blocks in the output:
> 
> PRAGMA database_size;
>  

For that, I run a simple experiment and looks weird for me but expected by duckdb:
```
andrea@asoria:~/Documents/tools$ ./duckdb010/duckdb test.duckdb
v0.10.0 20b1486d11
Enter ".help" for usage hints.
D CREATE TABLE data AS FROM 'https://huggingface.co/datasets/allenai/c4/resolve/refs%2Fconvert%2Fparquet/en/partial-validation/0001.parquet';
 99% ▕███████████████████████████████████████████████████████████▍▏ 
100% ▕████████████████████████████████████████████████████████████▏ 
D 
D select count(*) from data;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│       141509 │
└──────────────┘               
D PRAGMA database_size;
┌───────────────┬───────────────┬────────────┬──────────────┬─────────────┬─────────────┬──────────┬──────────────┬──────────────┐
│ database_name │ database_size │ block_size │ total_blocks │ used_blocks │ free_blocks │ wal_size │ memory_usage │ memory_limit │
│    varchar    │    varchar    │   int64    │    int64     │    int64    │    int64    │ varchar  │   varchar    │   varchar    │
├───────────────┼───────────────┼────────────┼──────────────┼─────────────┼─────────────┼──────────┼──────────────┼──────────────┤
│ test          │ 363.2 MiB     │     262144 │         1453 │        1453 │           0 │ 0 bytes  │ 363.2 MiB    │ 12.3 GiB     │
└───────────────┴───────────────┴────────────┴──────────────┴─────────────┴─────────────┴──────────┴──────────────┴──────────────┘
D .exit
andrea@asoria:~/Documents/tools$ ls -la
total 744052
...
-rw-rw-r--  1 andrea andrea 380907520 abr 19 11:23 test.duckdb
andrea@asoria:~/Documents/tools$ ./duckdb010/duckdb test.duckdb
v0.10.0 20b1486d11
Enter ".help" for usage hints.
D drop table data;
D PRAGMA database_size;
┌───────────────┬───────────────┬────────────┬──────────────┬─────────────┬─────────────┬──────────┬──────────────┬──────────────┐
│ database_name │ database_size │ block_size │ total_blocks │ used_blocks │ free_blocks │ wal_size │ memory_usage │ memory_limit │
│    varchar    │    varchar    │   int64    │    int64     │    int64    │    int64    │ varchar  │   varchar    │   varchar    │
├───────────────┼───────────────┼────────────┼──────────────┼─────────────┼─────────────┼──────────┼──────────────┼──────────────┤
│ test          │ 363.2 MiB     │     262144 │         1453 │        1453 │           0 │ 64 bytes │ 256.0 KiB    │ 12.3 GiB     │
└───────────────┴───────────────┴────────────┴──────────────┴─────────────┴─────────────┴──────────┴──────────────┴──────────────┘
D .exit
andrea@asoria:~/Documents/tools$ ls -la
total 744052
...
-rw-rw-r--  1 andrea andrea 380907520 abr 19 11:24 test.duckdb
andrea@asoria:~/Documents/tools$ ./duckdb010/duckdb test.duckdb
v0.10.0 20b1486d11
Enter ".help" for usage hints.
D PRAGMA database_size;
┌───────────────┬───────────────┬────────────┬──────────────┬─────────────┬─────────────┬──────────┬──────────────┬──────────────┐
│ database_name │ database_size │ block_size │ total_blocks │ used_blocks │ free_blocks │ wal_size │ memory_usage │ memory_limit │
│    varchar    │    varchar    │   int64    │    int64     │    int64    │    int64    │ varchar  │   varchar    │   varchar    │
├───────────────┼───────────────┼────────────┼──────────────┼─────────────┼─────────────┼──────────┼──────────────┼──────────────┤
│ test          │ 363.2 MiB     │     262144 │         1453 │           1 │        1452 │ 0 bytes  │ 256.0 KiB    │ 12.3 GiB     │
└───────────────┴───────────────┴────────────┴──────────────┴─────────────┴─────────────┴──────────┴──────────────┴──────────────┘
D .exit
andrea@asoria:~/Documents/tools$ la -la
total 744052
...
-rw-rw-r--  1 andrea andrea 380907520 abr 19 11:24 test.duckdb
```

See that:
- First, file size was 363Mib (After table creation)
-  After the drop, the size remains the same (I would have expected it to reduce)
- What changed in duckdb was the free_blocks, as mentioned in the docs.

So, I am not sure if this can help us :/

polinaeterna:
Should be ready for review now @huggingface/dataset-viewer 

@severo regarding storing index+transformed values and data separately - it probably would be possible to avoid problem that @AndreaFrancis described above by creating [new connection ](https://duckdb.org/2024/01/26/multi-database-support-in-duckdb.html#mix-and-match) and new database and copy only what's needed there (I'm not sure, haven't tested yet). But anyway I don't completely understand how it would make search/filter more efficient since we will have to download parquet files with data too anyway, no? I only see that download could be performed in parallel in this case.

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -205,9 +307,12 @@ def compute_split_duckdb_index_response(
             create_index_sql = CREATE_INDEX_COMMAND.format(columns=indexable_columns)
             logging.info(create_index_sql)
             con.sql(create_index_sql)
+
     finally:
         con.close()
 
+    features["__hf_index_id"] = {"dtype": "int64", "_type": "Value"}
```
polinaeterna, line 314: I don't update features with newly created transformed columns to avoid including them in search/filter response

lhoestq:
Should we name the columns like `{column_name}.length` ? It would feel more natural later when running queries in duckdb sql on the hub.

And in case of collision we can ignore I guess

lhoestq:
> while using the original parquet files to store the data

it would make things too slow or complicated imo, especially in downstream usage of the duckdb files from users

polinaeterna:
> especially in downstream usage of the duckdb files from users

I agree, if we ever want to advertise using `duckdb.index` file, it wouldn't be useful without data

severo:
Yes, it was an open question. The benefit would have been to reduce the size of the file we have to download. But no worry if it's not feasible or too complex.

services/worker/src/worker/job_runners/split/duckdb_index.py
AndreaFrancis: It looks to be duplicated here https://github.com/huggingface/dataset-viewer/pull/2737/files#diff-d9a2c828d7feca3b7f9e332e040ef861e842a16d18276b356461d2aa34396a8aR398

AndreaFrancis (approved):
Great! I tried it, and it works amazing!
Just one detail (I can also fix it in another PR) related to /search:
In this [command] (https://github.com/huggingface/dataset-viewer/blob/main/services/search/src/search/routes/search.py#L54), I used `data.*`. If we have a dataset with many string/audio/image columns, it would end up loading them and increasing memory usage while they are not needed. Filter, on the other hand, uses only the columns retrieved from "features," which is correct and will not load the transformed columns.
Search (We should avoid loading the transformed columns even if they are ignored when creating the response):
Screenshot from 2024-05-03 18-45-42
Filter (OK):
Screenshot from 2024-05-03 18-46-24

Do you mind applying the same logic as in filter but for /search? As I mentioned, it would be only heavy for datasets with many transformed columns.

polinaeterna:
@AndreaFrancis thank you for pointing out to this issue with loading all columns, hopefully it's fixed now

polinaeterna:
> And in case of collision we can ignore I guess

@lhoestq I realized it might cause problems at the stage of filtering. Let's say a dataset itself contains string column `text` and `text.length` but `text.length` is not an integer column corresponding to lengths of texts but something different, it might be even of different dtype. For example, string or class label with values "long" and "short". But when we filter, we assume that `{column_name}.length` contains integers corresponding to lengths. 

I don't know if examples like this exist though lol

polinaeterna:
okay i checked text, audio and image datasets and found one example with column names collision. i didn't check lists but i guess we are fine anyway :D

services/search/src/search/routes/search.py
```diff
@@ -219,16 +223,6 @@ async def search_endpoint(request: Request) -> Response:
                                 expiredTimeIntervalSeconds,
                             )
                 with StepProfiler(method="search_endpoint", step="create response"):
-                    # Features can be missing in old cache entries,
-                    # but in this case it's ok to get them from the Arrow schema.
-                    # Indded at one point we refreshed all the datasets that need the features
-                    # to be cached (i.e. image and audio datasets).
-                    if "features" in duckdb_index_cache_entry["content"] and isinstance(
```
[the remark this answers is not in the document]
AndreaFrancis: I think it is ok. Looks like all records have the "features" field:
image

AndreaFrancis (approved):

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -78,6 +94,76 @@ def check_indexable(feature: FeatureType) -> None:
...
+        return shapes_df
+    target_df.insert_column(target_df.shape[1], shapes_df[width_column_name])
+    target_df.insert_column(target_df.shape[1], shapes_df[height_column_name])
+    return target_df
+
+
+def compute_transformed_data(parquet_directory: Path, features: dict[str, Any]) -> Optional[pl.DataFrame]:
+    transformed_df = None
+    for feature_name, feature in features.items():
+        if isinstance(feature, list) or (isinstance(feature, dict) and feature.get("_type") == "Sequence"):
```
lhoestq, line 149: if `feature` is `Sequence` and `feature.feature` is a dict, then `feature` is a pa.struct()

Checking this should also avoid using `is_list_pa_type` (you already have all the necessary info in `features`)

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -656,28 +693,41 @@ class AudioColumn(MediaColumn):
...
+    def get_width(example: Optional[dict[str, Any]]) -> Optional[int]:
         """Get image widths."""
+        if example is None:
+            return None
         with io.BytesIO(example["bytes"]) as f:
             image = Image.open(f)
             return image.size[0]
 
+    @staticmethod
+    def get_shape(example: Optional[dict[str, Any]]) -> Union[tuple[None, None], tuple[int, int]]:
```
lhoestq, line 721: you might want to take into account the case where example is `bytes` ("bytes" only)

severo (approved):
As there are a lot of changes, and the code was already a bit complex, I'm not sure I really understand everything for my review (which is why I propose exporting part of the code to its proper module - I let you decide to do it or not, it's not blocking).

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -591,38 +626,40 @@ def transform(cls, example: dict[str, Any]) -> Any:
...
 
-        if not transformed_values:
+    @classmethod
+    def _compute_statistics(
+        cls,
+        parquet_directory: Path,
+        column_name: str,
+        n_samples: int,
+    ) -> SupportedStatistics:
+        transformed_values = cls.compute_transformed_data(parquet_directory, column_name, cls.transform)
```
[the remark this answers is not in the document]
polinaeterna: > we're computing the transformed data twice, once to create the duckdb file, and once to compute the stats, right? Is it an issue, or is it a detail that does not affect the computation time too much?

ah you're right i haven't though about that the same job is done twice honestly. i'm not sure if that's an issue as these transformations are quite fast, except for maybe audio/image, which are also fast considering that the data size is small but not as fast as polars' native measures.
It's possible to make the stats step depended on `duckdb-index` to avoid this double compute but in this case the error with fts index for example would cause stats to fail too. I'd personally avoid it. wdyt @lhoestq @AndreaFrancis ?

> If I remember well, before, we depended on the duckdb index file to compute the stats, then we changed to rely on polars instead, is this right?

We did not depend on the duckdb index file, we queried parquet files directly with duckdb. Now we query them with polars.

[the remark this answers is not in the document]
lhoestq: I think it's fine as it is for now

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -39,6 +41,13 @@
 
 from worker.config import AppConfig, DuckDbIndexConfig
 from worker.dtos import CompleteJobResult, SplitDuckdbIndex
+from worker.job_runners.split.descriptive_statistics import (
```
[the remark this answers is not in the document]
polinaeterna: I tried to move them to libcommon but that lead me to importing test data from libcommon to worker to ensure integrity which was meh so I created `statistics_utils.py` file in worker instead. not sure it looks good either but probably okay?..

services/worker/src/worker/job_runners/split/duckdb_index.py
```diff
@@ -80,6 +96,76 @@ def check_indexable(feature: FeatureType) -> None:
...
+    duration_column_name = f"{column_name}.duration"
+    durations = AudioColumn.compute_transformed_data(parquet_directory, column_name, AudioColumn.get_duration)
+    duration_df = pl.from_dict({duration_column_name: durations})
+    if target_df is None:
+        return duration_df
+    target_df.insert_column(target_df.shape[1], duration_df[duration_column_name])
+    return target_df
+
+
+def compute_image_width_length_column(
```
severo, line 132: I imagine the idea was:

```suggestion
def compute_image_width_height_column(
```

severo (approved):
It will be super useful!

polinaeterna:
I haven't thought before - maybe we should also update duckdb version since we are recomputing everything? But I don't know if it's needed.

AndreaFrancis:
Not needed for now IMO

polinaeterna:
sorry, one more thing i forgot :see_no_evil: shouldn't i update the job runner's version?

severo:
it would recompute all the stats. I think it's what we want, right?

PR #2743, opened 2024-04-23
Hardcode number of bins in histogram instead of env var

polinaeterna:
...to simplify things. I think I've seen some discussion and related PRs from @severo recently with the idea that the less configuration options the better. But I cannot find it...
(Honestly, I have no idea why I decided that it would be a good idea to make it an environmental variable.)


lhoestq (approved):

severo:
yes here: https://huggingface.slack.com/archives/C067U56EQBB/p1713435248465759 (internal)

PR #2747, opened 2024-04-25
Update DatasetWithScriptNotSupportedError message suggesting CLI

### LINKED ISSUE #2742: Improve the message for DatasetWithScriptNotSupportedError
Now we have a CLI tool to convert a script dataset to a data-only one (parquet). See https://github.com/huggingface/datasets/releases/tag/2.19.0
Capture d’écran 2024-04-23 à 11 41 10

We should reflect that in all the error messages for DatasetWithScriptNotSupportedError: https://github.com/search?q=repo%3Ahuggingface%2Fdataset-viewer+%22raise+DatasetWithScriptNotSupportedError%22&type=code, and propose the user to use the CLI to convert the dataset. Ideally with a link to a doc that shows how to do it (I think the doc page should lie in https://github.com/huggingface/datasets/tree/main/docs, see https://github.com/huggingface/datasets/issues/6830).

albertvillanova:
Suggest using the `datasets` convert_to_parquet CLI instead of opening a Discussion.

Additionally, this PR sets a default error message, instead of repeating the same everywhere.

Fix #2742.

### DIFF
--- a/services/worker/src/worker/job_runners/config/parquet_and_info.py
+++ b/services/worker/src/worker/job_runners/config/parquet_and_info.py
@@ -1241,11 +1241,7 @@ def compute_config_parquet_and_info_response(
         raise EmptyDatasetError(f"{dataset=} is empty.", cause=err) from err
     except ValueError as err:
         if "trust_remote_code" in str(err):
-            raise DatasetWithScriptNotSupportedError(
-                "The dataset viewer doesn't support this dataset because it runs "
-                "arbitrary python code. Please open a discussion in the discussion tab "
-                "if you think this is an error and tag @lhoestq and @severo."
-            ) from err
+            raise DatasetWithScriptNotSupportedError from err
         raise
     except FileNotFoundError as err:
         raise DatasetNotFoundError("The dataset, or the revision, does not exist on the Hub.") from err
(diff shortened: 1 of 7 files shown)


services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -1241,11 +1241,7 @@ def compute_config_parquet_and_info_response(
         raise EmptyDatasetError(f"{dataset=} is empty.", cause=err) from err
     except ValueError as err:
         if "trust_remote_code" in str(err):
-            raise DatasetWithScriptNotSupportedError(
-                "The dataset viewer doesn't support this dataset because it runs "
-                "arbitrary python code. Please open a discussion in the discussion tab "
-                "if you think this is an error and tag @lhoestq and @severo."
-            ) from err
+            raise DatasetWithScriptNotSupportedError from err
```
severo, line 1244: btw, we should have passed the cause error as a parameter:

```suggestion
            raise DatasetWithScriptNotSupportedError(cause=err) from err
```

It's unrelated to the subject of this PR though, and we surely could be smarted and get the value automatically when we use `from err`

severo (approved):

services/worker/src/worker/job_runners/config/parquet_and_info.py
albertvillanova: Yes, we can get the value automatically from the `__cause__` attribute of the raised Exception with a `from` clause.

severo: https://github.com/huggingface/dataset-viewer/issues/2753

PR #2748, opened 2024-04-25
Group sleep logs in retry()

lhoestq:
they're spamming too much, e.g. 20k logs just for the parquet-and-info jobs of fineweb

### DIFF
--- a/libs/libcommon/src/libcommon/utils.py
+++ b/libs/libcommon/src/libcommon/utils.py
@@ -165,9 +165,47 @@ RETRY_ON: tuple[type[Exception]] = (Exception,)
 class retry:
     """retries with an increasing sleep before every attempt"""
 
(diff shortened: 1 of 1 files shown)


libs/libcommon/src/libcommon/utils.py
```diff
@@ -178,13 +216,14 @@ def decorator(*args: Any, **kwargs: Any) -> Any:
                 try:
                     """always sleep before calling the function. It will prevent rate limiting in the first place"""
                     duration = self.sleeps[attempt]
-                    logging.info(f"Sleep during {duration} seconds to preventively mitigate rate limiting.")
+                    self.grouped_log(f"Sleep during {duration} seconds to preventively mitigate rate limiting.")
```
severo, line 219: Don't we prefer to just remove this log?

PR #2755, opened 2024-04-29
Adding large_string to duckdb-index worker and tests

AndreaFrancis:
Fix for https://github.com/huggingface/dataset-viewer/issues/2613

### DIFF
--- a/services/worker/src/worker/job_runners/split/duckdb_index.py
+++ b/services/worker/src/worker/job_runners/split/duckdb_index.py
@@ -67,7 +67,7 @@ def get_indexable_columns(features: Features) -> list[str]:
 
         def check_indexable(feature: FeatureType) -> None:
             nonlocal indexable
-            if isinstance(feature, Value) and feature.dtype == "string":
+            if isinstance(feature, Value) and feature.dtype in ("string", "large_string"):
                 indexable = True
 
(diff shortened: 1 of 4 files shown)


severo (approved):

severo:
The idea is to recompute the datasets manually, right? I imagine that only few datasets use this type.

AndreaFrancis:
> The idea is to recompute the datasets manually, right? I imagine that only few datasets use this type.

Yes, I will put here the datasets that were refreshed when it is ready.

severo:
(the number is enough... better not to give the names, as some might be private)

AndreaFrancis:
I will refresh 3364 splits; the logic is:  Get all datasets with "text" modality and intersect with the list of splits with "has_fts": false.

severo:
OK. Maybe some datasets have both string and large_string: we will not refresh them, right? Is it a way to find them too? I think the answer is no, without scanning every "/info" response

AndreaFrancis:
> OK. Maybe some datasets have both string and large_string: we will not refresh them, right? Is it a way to find them too? I think the answer is no, without scanning every "/info" response

In case they had both (string and large_string), they should already be "has_fts": true, so there should be no refresh for them.

PR #2758, opened 2024-04-30
Set number of retries per error code

severo:
See https://huggingface.slack.com/archives/C04L6P8KNQ5/p1714480880576699?thread_ts=1714033939.429809&cid=C04L6P8KNQ5 (internal)

Errors like `CreateCommitError` should always be retried, because they correspond to the Hub being down, which should always be a temporary situation. I set the limit to 30 in that case (instead of 3). I set 30 to other error codes as well: `HfHubError`, `LockedDatasetTimeoutError`, `PreviousStepStillProcessingError`


lhoestq (approved):

severo:
CI error is unrelated, see https://github.com/huggingface/dataset-viewer/issues/2759

severo:
It's working as expected: for the first backfill on retryable errors, a lot of new jobs

Capture d’écran 2024-04-30 à 16 08 09

severo:
And the number of `CreateCommitError`s is decreasing, soon zero

Capture d’écran 2024-04-30 à 17 16 34

severo:
Is there an issue with the metrics? We never get back to zero:

Capture d’écran 2024-05-02 à 13 37 52
Capture d’écran 2024-05-02 à 13 38 35
Capture d’écran 2024-05-02 à 13 38 09

severo:
For `PreviousStepStillProcessingError`, it seems it's due to the previous step being `JobManagerCrashedError`. When we detect that a job has crashed, maybe we're not creating jobs for its children?

I opened an issue: https://github.com/huggingface/dataset-viewer/issues/2765

severo:
Also: the metrics include `details.copied_from_artifact: false | true`, ie: we currently have only 4 datasets with the `CreateCommitError`, but 40 cache entries for all the following steps, and the metrics above show 40.

For the 4 datasets, we reached `failed_runs: 30`, so we must have a bug here. One of these is https://huggingface.co/datasets/venetis/VMMRdb_make_model_test. I'm not sure what happens. Opening an issue: https://github.com/huggingface/dataset-viewer/issues/2766

PR #2760, opened 2024-04-30
Fix size in tests

### LINKED ISSUE #2759: services/worker tests are failing
See https://github.com/huggingface/dataset-viewer/actions/runs/8895306050/job/24425363938

We get 55 instead of 50 (not sure for which field, we don't have enough logs as of now).

severo:
fixes #2759

I'm not sure why we need to add 5 bytes to the size though...


albertvillanova (approved):
Thanks for the fast fix.

severo:
@lhoestq do you have an idea of why we need to add 5 bytes? Is it insignificant, or should we dig a bit to understand what is happening?

lhoestq:
it's insignificant

lhoestq (approved):

PR #2764, opened 2024-04-30
Index Translations and TranslationVariableLanguages data type for search

AndreaFrancis:
We have 2171 splits with translation or TranslationVariableLanguages data types that need to be indexed while duckdb supports it.
See these examples:

- https://huggingface.co/datasets/vgaraujov/americasnlp-mt-21
- https://huggingface.co/datasets/ted_hrlr


severo (approved):
OK. I'm wondering how it works, as it's not a string but an object. Maybe it's worth adding a test to ensure the search works (we can retrieve the contents, in whichever language). Also: if we search for a language string ("en"), will it be retrieved?

AndreaFrancis:
> OK. I'm wondering how it works, as it's not a string but an object

It treats the translation as a struct like this:
image

I added a [test](https://github.com/huggingface/dataset-viewer/pull/2764/files#diff-0dda92a634f5813ca87e9119b3ae9a187b300a06415ab3fc323850a864417892R495) to cover the scenario

AndreaFrancis:
> if we search for a language string ("en"), will it be retrieved?

Interesting, it will treat the key as part of the content to look for:
image
See the second select looking for "aym", it returns the whole table

severo:
excellent! Thanks for the examples

PR #2770, opened 2024-05-03
support more than 10k parquet files in config-parquet-metadata

### LINKED ISSUE #2768: FineWeb: Unexpected end of stream: Page was smaller (1862094) than expected (2055611)
The config-parquet-metadata job succeeds but the split-first-rows job fails when using `compute_first_rows_from_parquet_response`.

In the meantime I set the error code in the config-parquet-metadata response as `CachedResponseNotFound` to make the split-first-rows succeed

severo:
fixes #2768


albertvillanova (approved):
Thanks for the investigation and the neat fix.

PR #2776, opened 2024-05-06
Set max number of splits

lhoestq:
To avoid cases with 300 splits = 300 load_dataset() * 300 data files resolutions = 90k calls to the Hub API to list files

### DIFF
--- a/services/worker/src/worker/config.py
+++ b/services/worker/src/worker/config.py
@@ -259,6 +259,22 @@ class ConfigNamesConfig:
             )
 
 
+SPLIT_NAMES_MAX_NUMBER = 30
+
+
+@dataclass(frozen=True)
+class SplitNamesConfig:
+    max_number: int = SPLIT_NAMES_MAX_NUMBER
+
+    @classmethod
+    def from_env(cls) -> "SplitNamesConfig":
+        env = Env(expand_vars=True)
+        with env.prefixed("SPLIT_NAMES_"):
+            return cls(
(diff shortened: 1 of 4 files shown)


services/worker/src/worker/job_runners/config/split_names.py
```diff
@@ -83,10 +87,17 @@ def compute_split_names_from_streaming_response(
             f"Cannot get the split names for the config '{config}' of the dataset.",
             cause=err,
         ) from err
+    if len(split_name_items) > max_number:
+        split_examples = ", ".join([split_name_item["split"] for split_name_item in split_name_items[:5]])
+        raise DatasetWithTooManySplitsError(
+            f"The {config} config contains {len(split_name_items)} while it should generally contain 3 splits maximum (train/validation/test). "
+            f"If the splits {split_examples}... are not used to differentiate between training and evaluation, please consider defining configs of this dataset instead. "
```
severo, line 94: excellent!

services/worker/src/worker/config.py
```diff
@@ -259,6 +259,22 @@ def from_env(cls) -> "ConfigNamesConfig":
             )
 
 
+SPLIT_NAMES_MAX_NUMBER = 30
```
severo, line 262: frankly, I would hardcode it

lhoestq: I like that it's at the same place as other constants.

But yea having a config is overkill, we can change that later

albertvillanova (commented):
You can merge the main branch once the PR with the vulnerability fix has been merged:
- #2774

PR #2788, opened 2024-05-07
Smart dataset update: don't recompute everything all the time

lhoestq:
Don't recompute a viewer if the commit is an update in the main readme content or in yaml tags that don't influence the viewer.

Only available for datasets in `datasets-maintainers` org for now.

It works this way

1. take the old_sha and new_sha of the webhook
2. check that the cache revision is old_sha
3. check the updated files and fields in the YAML readme file
4. if there is no need to recomptue anything, update the revisions of the cache entries and storage directories

PS1: it doesn't update the revision of the entries in the queue because:
- running jobs might be writing to the old storage directory: they should fail and be re-run on the new revision
- waiting jobs are not relevant since we need to re-run their parent running jobs anyway

PS2: I added a very simple out-of-order commit handling by retrying in case the old_sha doesn't match

Fix this issue https://github.com/huggingface/dataset-viewer/issues/2757 once available for all datasets


severo:
hmmm

Capture d’écran 2024-05-13 à 22 05 11

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -211,14 +224,44 @@ def run(self) -> TasksStatistics:
...
+
 @dataclass
 class DeleteDatasetStorageTask(Task):
     dataset: str
     storage_client: StorageClient
 
     def __post_init__(self) -> None:
         # for debug and testing
-        self.id = f"DeleteDatasetStorageTask,{self.dataset},{self.storage_client}"
+        self.id = f"DeleteDatasetStorageTask,{self.dataset},1"
```
severo, line 264: I'm not sure: I think we want to show the details of the storage client here (see it `__repr__()` method)

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -815,6 +1032,46 @@ def set_revision(
...
+    hf_token: Optional[str] = None,
+) -> TasksStatistics:
+    """
+    Set the current revision of the dataset in a smarter way.
+
+    /!\ This logic is WIP and hsould only be used on a subset of datasets for now.
+
+    If the revision is already set to the same value, this is a no-op.
+    Else: one job is created for every step that need to be recomputed,
+    and for the other step it updates its revision without recomputing.
```
severo, line 1051:
```suggestion
    Else, if only .gitignore, .gitattributes, or a non-significant (for the
      dataset viewer) part of README.md has been modified in the last
      commit, update the cache entries and assets to the last revision
      without recomputing.
    Else: raise. 
```

libs/libcommon/tests/test_smart_update.py
severo, line 1: I think it's the first test file which is not directly mapped to a source file. Why not, because in this case, test_orchestrator.py is already long. Maybe rename to `test_orchestrator_smart_update.py` to make it a bit clearer what it is testing?

libs/libcommon/tests/test_smart_update.py
```diff
@@ -0,0 +1,283 @@
...
+        assert_smart_dataset_update_plan(
+            plan,
+            files_impacted_by_commit=[],
+            tasks=["UpdateRevisionOfDatasetCacheEntriesTask,1"],
+        )
+    with pytest.raises(RequestException):  # if diff doesn't exist
+        get_smart_dataset_update_plan(processing_graph=PROCESSING_GRAPH_TWO_STEPS)
+
+
+def test_add_initial_readme_commit() -> None:
```
severo, line 148: another interesting test would be if the commit creates the README file, with the `configs` field in YAML for example -> it should raise `SmartUpdateImpossibleBecauseOfUpdatedYAMLField` iirc

libs/libcommon/tests/test_smart_update.py
```diff
@@ -0,0 +1,283 @@
...
+    put_cache(step=STEP_DA, dataset=DATASET_NAME, revision=OTHER_REVISION_NAME)
+    with put_diff(ADD_CONFIG_DIFF), put_readme(ONE_CONFIG_README):
+        with pytest.raises(SmartUpdateImpossibleBecauseOfUpdatedYAMLField):
+            get_smart_dataset_update_plan(processing_graph=PROCESSING_GRAPH_TWO_STEPS)
+
+
+def test_add_second_config_commit() -> None:
+    # Add a second config commit: raise
+    put_cache(step=STEP_DA, dataset=DATASET_NAME, revision=OTHER_REVISION_NAME)
+    with put_diff(ADD_SECOND_CONFIG_DIFF), put_readme(ONE_CONFIG_README, revision=OTHER_REVISION_NAME), put_readme(
```
severo, line 180: just to be sure I understand: when you call `put_readme` twice, the second one will mock the already mocked `hf_hub_download` again?

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -736,6 +812,147 @@ def is_big(config: str) -> bool:
...
+    diff: str = field(init=False)
+    files_impacted_by_commit: set[str] = field(init=False)
+    updated_yaml_fields_in_dataset_card: list[str] = field(init=False)
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        cache_kinds = [
+            processing_step.cache_kind for processing_step in self.processing_graph.get_first_processing_steps()
+        ]
+        for retry in range(3):
```
severo, line 851:
```suggestion
        # Try to be robust to a burst of webhooks or out-of-order webhooks
        # by waiting up to 2 seconds for a coherent state 
        for retry in range(3):
```

severo (commented):
It will be super useful!

For now, the main impact will be to avoid recomputing anything when the README contents (below YAML, or unrelated fields like the tags) of a big dataset are updated. It does not occur all the time, but when it does, we will be happy to avoid the load + the downtime for its viewer.

libs/libcommon/tests/test_smart_update.py
lhoestq: Yes, they wrap each other

libs/libcommon/src/libcommon/storage_client.py
```diff
@@ -113,5 +115,32 @@ def delete_dataset_directory(self, dataset: str) -> int:
...
+            new_revision (`str`): the new revision
+
+        Returns:
+            int: The number of directories updated (0 or 1)
+        """
+        # same pattern as in asset.py
+        old_dataset_revision_key = self.get_full_path(f"{parse.quote(dataset)}/{DATASET_SEPARATOR}/{old_revision}")
+        new_dataset_revision_key = self.get_full_path(f"{parse.quote(dataset)}/{DATASET_SEPARATOR}/{new_revision}")
+        try:
+            self._fs.mv(old_dataset_revision_key, new_dataset_revision_key, recursive=True)
```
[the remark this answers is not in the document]
lhoestq: Yeah using shutil it can be faster but only for local files (we have to keep this logic for S3)

services/api/src/api/routes/webhook.py
```diff
@@ -70,13 +72,13 @@ def process_payload(
     hf_token: Optional[str] = None,
     hf_timeout_seconds: Optional[float] = None,
     storage_clients: Optional[list[StorageClient]] = None,
-) -> None:
+) -> Optional[TasksStatistics]:
```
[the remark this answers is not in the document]
lhoestq: I added these for testing but I ended up not using all of them - will remove the unnecessary changes

services/api/src/api/routes/webhook.py
```diff
@@ -90,11 +92,34 @@ def process_payload(
...
+        revision = payload["repo"].get("headSha")
+        old_revision: Optional[str] = None
+        for updated_ref in payload.get("updatedRefs") or []:
+            ref = updated_ref.get("ref")
+            ref_new_sha = updated_ref.get("newSha")
+            ref_old_sha = updated_ref.get("oldSha")
+            if ref == "refs/heads/main" and isinstance(ref_new_sha, str) and isinstance(ref_old_sha, str):
+                revision = ref_new_sha
+                old_revision = ref_old_sha
         delete_dataset(dataset=dataset, storage_clients=storage_clients)
```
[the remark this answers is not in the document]
lhoestq: oops good catch !

severo:
I think we still want this PR, right @lhoestq (no hurry though)

libs/libcommon/src/libcommon/operations.py
```diff
@@ -209,7 +216,7 @@ def update_dataset(
     hf_timeout_seconds: Optional[float] = None,
     priority: Priority = Priority.LOW,
     storage_clients: Optional[list[StorageClient]] = None,
-) -> None:
+) -> TasksStatistics:
```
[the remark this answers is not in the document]
lhoestq: I'm reverting to return None, the output is not used anywhere anyway

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -736,6 +812,147 @@ def is_big(config: str) -> bool:
...
+class SmartDatasetUpdatePlan(Plan):
+    dataset: str
+    revision: str
+    hf_endpoint: str
+    old_revision: Optional[str] = None
+    processing_graph: ProcessingGraph = field(default=processing_graph)
+    storage_clients: Optional[list[StorageClient]] = None
+    hf_token: Optional[str] = None
+
+    cached_revision: str = field(init=False)
```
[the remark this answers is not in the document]
lhoestq: I'll use it in tests

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -736,6 +812,147 @@ def is_big(config: str) -> bool:
...
+                        old_revision=self.old_revision,
+                        new_revision=self.revision,
+                        storage_client=storage_client,
+                    )
+                )
+
+    def get_diff(self) -> str:
+        headers = build_hf_headers(token=self.hf_token, library_name="dataset-viewer")
+        resp = requests.get(
+            self.hf_endpoint + f"/datasets/{self.dataset}/commit/{self.revision}.diff", timeout=10, headers=headers
```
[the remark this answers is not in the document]
lhoestq: yes, we raise `SmartUpdateImpossibleBecauseCachedRevisionIsNotParentOfNewRevision` otherwise

libs/libcommon/src/libcommon/orchestrator.py
lhoestq: I switched to {self.storage_client.protocol}://{self.storage_client.storage_root}

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -36,6 +29,7 @@ class AudioSource(TypedDict):
 def generate_object_key(
     dataset: str, revision: str, config: str, split: str, row_idx: int, column: str, filename: str
 ) -> str:
+    # same pattern as in storage_client.py
```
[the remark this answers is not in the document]
lhoestq: I moved the function to StorageClient to have all the logic there

lhoestq:
This is ready for review :)

This mechanism will only be available to `datasets-maintainers` repos for testing for now

services/webhook/src/webhook/routes/webhook.py
```diff
@@ -90,10 +97,38 @@ def process_payload(
...
+            return None
+        revision = payload["repo"].get("headSha")
+        old_revision: Optional[str] = None
+        for updated_ref in payload.get("updatedRefs") or []:
+            ref = updated_ref.get("ref")
+            ref_new_sha = updated_ref.get("newSha")
+            ref_old_sha = updated_ref.get("oldSha")
+            if ref == "refs/heads/main" and isinstance(ref_new_sha, str) and isinstance(ref_old_sha, str):
+                if revision != ref_new_sha:
+                    logging.info(f"headSha {revision} is different from newSha {ref_new_sha}")
```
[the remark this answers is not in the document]
severo: also: should we take it into account if it occurs? (maybe it's not important, since we should expect the payload to be always correct)

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -2,27 +2,18 @@
...
 
-from libcommon.constants import DATASET_SEPARATOR
-from libcommon.storage import StrPath, remove_dir
 from libcommon.storage_client import StorageClient
 
 SUPPORTED_AUDIO_EXTENSION_TO_MEDIA_TYPE = {".wav": "audio/wav", ".mp3": "audio/mpeg", ".opus": "audio/opus"}
 SUPPORTED_AUDIO_EXTENSIONS = SUPPORTED_AUDIO_EXTENSION_TO_MEDIA_TYPE.keys()
 
 
-def delete_asset_dir(dataset: str, directory: StrPath) -> None:
```
severo, line 21: unused. OK!

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -728,6 +801,153 @@ def is_big(config: str) -> bool:
...
+        # Try to be robust to a burst of webhooks or out-of-order webhooks
+        # by waiting up to 2 seconds for a coherent state
+        for retry in range(3):
+            cache_entries_df = get_cache_entries_df(
+                dataset=self.dataset,
+                cache_kinds=cache_kinds,
+            )
+            if len(cache_entries_df) == 0:
+                raise SmartUpdateImpossibleBecauseCacheIsEmpty(f"Failed to smart update to {self.revision[:7]}")
+            self.cached_revision = cache_entries_df.sort_values("updated_at").iloc[-1]["dataset_git_revision"]
```
severo, line 849: shouldn't we check that all the entries have the same value?

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -728,6 +801,153 @@ def is_big(config: str) -> bool:
...
+            for storage_client in self.storage_clients:
+                self.add_task(
+                    UpdateRevisionOfDatasetStorageTask(
+                        dataset=self.dataset,
+                        old_revision=self.old_revision,
+                        new_revision=self.revision,
+                        storage_client=storage_client,
+                    )
+                )
+
```
severo, line 895: maybe explain here why we don't touch the pending jobs?

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -728,6 +801,153 @@ def is_big(config: str) -> bool:
...
+                    self.dataset,
+                    "README.md",
+                    repo_type="dataset",
+                    token=self.hf_token,
+                    revision=self.revision,
+                    endpoint=self.hf_endpoint,
+                )
+            ).open(mode="r", newline="", encoding="utf-8") as f:
+                dataset_card_data_dict = DatasetCard(f.read()).data.to_dict()
+        except Exception:
```
severo, line 928: why do we hide the exceptions?

libs/libcommon/src/libcommon/orchestrator.py
```diff
@@ -728,6 +801,153 @@ def is_big(config: str) -> bool:
...
+                    self.dataset,
+                    "README.md",
+                    repo_type="dataset",
+                    token=self.hf_token,
+                    revision=self.old_revision,
+                    endpoint=self.hf_endpoint,
+                )
+            ).open(mode="r", newline="", encoding="utf-8") as f:
+                old_dataset_card_data_dict = DatasetCard(f.read()).data.to_dict()
+        except Exception:
```
severo, line 942: same

severo (approved):
some comments, but overall it's super nice! let's try it soon!

services/webhook/src/webhook/routes/webhook.py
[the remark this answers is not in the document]
lhoestq: I'm switching this to warning and modifying the text a bit.

It shouldn't occur but it's nice to have a warning in case we ever need to debug something related to that

libs/libcommon/src/libcommon/orchestrator.py
lhoestq: yup good idea, I'm adding an error in that case

libs/libcommon/src/libcommon/orchestrator.py
severo: sure, but I think we should let them raise, or raise any other error, instead of considering that the README YAML has not changed in a manner that could change the dataset viewer results. It would fallback to "normal" update, which is what I would expect.

lhoestq: Well if there is an error, the readme is considered empty, which does fall back to normal update if it used to be valid and have some content

lhoestq: I see your point though: if there was no metadata in the readme and the user adds malformed metadata we should raise an error instead of considering it to be an empty readme.

I'll raise an error then

PR #2790, opened 2024-05-09
Add lock-no-update command to Makefile

albertvillanova:
Add lock-no-update command to Makefile.

I find this command useful when there are issues in the local Python environment. See: https://github.com/huggingface/dataset-viewer/pull/2783#issuecomment-2102015730


severo (approved):
OK!

PR #2792, opened 2024-05-10
Install libsndfile 1.2.2 from source to support 0.12 OPUS format

### LINKED ISSUE #2584: opus decoding error
see https://huggingface.co/datasets/stable-speech/mls_eng_10k/discussions/1#65ef6e9d440a5fc3d94a40ad

To fix this maybe we should pin `soundfile` library to `>=1.0.31` (first version that supported opus) like [we do in `datasets` library](https://github.com/huggingface/datasets/blob/main/src/datasets/config.py#L144).

albertvillanova:
Install libsndfile 1.2.2 from source in Docker image.

Fix #2584.

### DIFF
--- a/services/worker/Dockerfile
+++ b/services/worker/Dockerfile
@@ -19,6 +19,21 @@ RUN apt-get update \
     && apt-get install -y unzip wget procps htop ffmpeg libavcodec-extra libsndfile1 \
     && rm -rf /var/lib/apt/lists/*
 
+# Install libsndfile 1.2.2 from source because the version distributed by the package manager at the moment is 1.2.0 (see details here: https://github.com/libsndfile/libsndfile)
+# See issue with opus fomat: https://github.com/huggingface/dataset-viewer/issues/2584
+RUN apt-get update \
+    && apt-get install -y autoconf autogen automake build-essential libasound2-dev libflac-dev libogg-dev libtool libvorbis-dev libopus-dev libmp3lame-dev libmpg123-dev pkg-config git;
+WORKDIR /tmp
+RUN git clone --depth=1 --branch=1.2.2 https://github.com/libsndfile/libsndfile.git;
+WORKDIR /tmp/libsndfile
+RUN autoreconf -vif;
+RUN /tmp/libsndfile/configure --enable-werror;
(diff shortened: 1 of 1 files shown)


severo (commented):
Is it needed only in worker? I imagine we need this also in:
- rows
- search
- the CI (we should run tests to reproduce the issue and check it's fixed with the upgrade)

albertvillanova:
Thanks for the feedback: honestly, I took this issue without having a clear idea what to do... :sweat_smile:

albertvillanova:
@severo, while trying to implement a test, I discovered that I am not able to reproduce the issue with libsndfile-1.0.31 (our minimum required version).
I have checked our CI, and our tests also use libsndfile-1.0.31. See: https://github.com/huggingface/dataset-viewer/actions/runs/9079619818/job/24949210823
> Setting up libsndfile1:amd64 (1.0.31-2ubuntu0.1) ...

Do you know why our prod has libsndfile-1.2.0?

albertvillanova:
Maybe I could try to pin the version of libsndfile1 in all the Dockerfiles, e.g. https://github.com/huggingface/dataset-viewer/blob/af80a2486ffed3090ec08bd4f27f938848bec942/services/worker/Dockerfile#L19

What do you think?

severo:
Do you mean that your local system package is `libsndfile-1.0.31`, not `libsndfile-1.2.0`?

severo:
Yes, I think we first should confirm which sets of system + pip packages work wit opus, and which ones don't. Then ensure we use this set in our CI and prod (+ local setup for development)

albertvillanova:
Yes, of course, @severo 

What I wanted to pint out is that currently I are not be able to test in our CI what will happen in our prod, because the libsndfile1 versions are misaligned:
- CI uses 1.0.31
- prod uses 1.2.0

We should first align both. Do you agree?

severo:
Ah OK. It's because the tests in the CI don't use the docker file. The system (ubuntu, apt) is set in the GitHub action. Yes, we should align, you're right!

albertvillanova:
Superseded by:
- #2811

PR #2793, opened 2024-05-10
Fix image stats when image column type is bytes, not struct

polinaeterna:
should fix cases like this https://huggingface.co/datasets/lhoestq/test-cmcv

### DIFF
--- a/services/worker/src/worker/job_runners/split/descriptive_statistics.py
+++ b/services/worker/src/worker/job_runners/split/descriptive_statistics.py
@@ -583,12 +583,25 @@ class MediaColumn(Column):
     transform_column: type[Column]
 
     @classmethod
-    def transform(cls, example: dict[str, Any]) -> Any:
+    def transform(cls, example: dict[str, Any], is_struct: bool = True) -> Any:
         """
         Function to use to transform the original values to further pass these transformed values to statistics
         computation. Used inside ._compute_statistics() method.
         """
         raise NotImplementedError
+
+    @classmethod
+    def is_struct_type(cls, filename: Path, column_name: str) -> bool:
+        column_type = pq.read_schema(filename).field(column_name).type
+        if pa.types.is_struct(column_type):
+            return True
+        elif pa.types.is_binary(column_type):
+            return False
+        else:
+            raise ValueError(
+                f"Incorrect {cls.__name__} {column_name=} schema: {column_type=} "
+                f"but can be either bytes or struct. "
+            )
 
     @classmethod
     def _compute_statistics(
@@ -598,6 +611,7 @@ class MediaColumn(Column):
         n_samples: int,
     ) -> SupportedStatistics:
         parquet_files = list(parquet_directory.glob("*.parquet"))
+        is_struct = cls.is_struct_type(parquet_files[0], column_name)
         transformed_values = []
         for filename in parquet_files:
             shard_items = pq.read_table(filename, columns=[column_name]).drop_null().to_pydict()[column_name]
@@ -605,6 +619,7 @@ class MediaColumn(Column):
                 thread_map(
                     cls.transform,
                     shard_items,
+                    is_struct,
                     desc=f"Transforming values of {cls.__name__} {column_name} for {filename.name}",
                     leave=False,
                 )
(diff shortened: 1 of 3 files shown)


services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -583,13 +584,26 @@ class MediaColumn(Column):
...
 
+    @classmethod
+    def is_struct_type(cls, filename: Path, column_name: str) -> bool:
+        column_type = pq.read_schema(filename).field(column_name).type
+        if pa.types.is_struct(column_type):
+            return True
+        elif pa.types.is_binary(column_type) or pa.types.is_null(column_type):
+            return False
+        else:
+            raise ValueError(
```
polinaeterna, line 602: idk if i should do a type check and raise here actually

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -656,29 +672,35 @@ class AudioColumn(MediaColumn):
...
     transform_column = IntColumn
 
     @staticmethod
-    def get_width(example: dict[str, Any]) -> int:
+    def get_width(example: Optional[Union[bytes, dict[str, Any]]], is_struct: bool = True) -> Optional[int]:
         """Get image widths."""
-        with io.BytesIO(example["bytes"]) as f:
+        if example is None:
+            return None
+        example_bytes = example["bytes"] if is_struct else example  # type: ignore
```
polinaeterna, line 696: `type: ignore` here because example can be both bytes and dict and mypy is not happy with bytes being indexed with string. 
any better solution to this?

services/worker/tests/fixtures/descriptive_statistics_dataset.py
```diff
@@ -1640,9 +1640,9 @@ def nan_column() -> list[None]:
     },
     features=Features(
         {
-            "audio": Audio(sampling_rate=1600),
-            "audio_nan": Audio(sampling_rate=1600),
-            "audio_all_nan": Audio(sampling_rate=1600),
+            "audio": Audio(sampling_rate=1600, decode=False),
+            "audio_nan": Audio(sampling_rate=1600, decode=False),
+            "audio_all_nan": Audio(sampling_rate=1600, decode=False),
```
lhoestq, line 1645: why this change ?

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -656,29 +672,35 @@ class AudioColumn(MediaColumn):
...
 
     @staticmethod
-    def get_duration(example: dict[str, Any]) -> float:
+    def get_duration(example: Optional[Union[bytes, dict[str, Any]]], is_struct: bool = True) -> Optional[float]:
         """Get audio durations"""
-        with io.BytesIO(example["bytes"]) as f:
+        if example is None:
+            return None
+        example_bytes = example["bytes"] if is_struct else example  # type: ignore
+        with io.BytesIO(example_bytes) as f:
```
lhoestq, line 680: I guess you can just use isinstance ?
```suggestion
        if example is None:
            return None
        elif isinstance(example, bytes):
            example_bytes = example
        else:
            example_bytes = example["bytes"]
        with io.BytesIO(example_bytes) as f:
```

services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -656,29 +672,35 @@ class AudioColumn(MediaColumn):
     transform_column = FloatColumn
 
     @staticmethod
-    def get_duration(example: dict[str, Any]) -> float:
+    def get_duration(example: Optional[Union[bytes, dict[str, Any]]], is_struct: bool = True) -> Optional[float]:
```
lhoestq, line 675: and therefore you don't need `is_struct `
```suggestion
    def get_duration(example: Optional[Union[bytes, dict[str, Any]]]) -> Optional[float]:
```

services/worker/src/worker/job_runners/split/descriptive_statistics.py
lhoestq: if you use isinstance on the examples you might not need this function

lhoestq (commented):
Cool ! Just a few comments:

services/worker/src/worker/job_runners/split/descriptive_statistics.py
polinaeterna: yeah I first did like this but then I thought if I know in advance for all samples, why check every time. but if you think it's fine, i'd do like this

services/worker/tests/fixtures/descriptive_statistics_dataset.py
polinaeterna: to make `path` accessible here https://github.com/huggingface/dataset-viewer/pull/2793/files#diff-6cac2dec70bb036137022bcfb8bbe859ecc3d25b990558ef7efcb27bd75f5466R806

PR #2795, opened 2024-05-13
Enclose column names in double quotes in filter docs and tests

### LINKED ISSUE #2794: Improve robustness of column names containing non-alphanumeric characters
Currently, if a column name contains non-alphanumeric characters, some functionalities are broken, like filter.

We should improve the robustness of the viewer in this regard.

See discussion: https://huggingface.co/datasets/m-a-p/Matrix/discussions/2 

CC: @lhoestq

albertvillanova:
Suggest enclosing column names in double quotes in filter docs for robustness with non-alphanumeric characters.

Note that the viewer uses DuckDB: https://duckdb.org/docs/sql/introduction
> DuckDB’s SQL dialect follows the conventions of the PostgreSQL dialect.

And in PostgreSQL SQL Syntax: https://www.postgresql.org/docs/11/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
> Quoted identifiers can contain any character, except the character with code zero. (To include a double quote, write two double quotes.) This allows constructing table or column names that would otherwise not be possible, such as ones containing spaces or ampersands.

Fix #2794.

CC: @lhoestq

### DIFF
--- a/docs/source/filter.md
+++ b/docs/source/filter.md
@@ -21,27 +21,28 @@ The `/filter` endpoint accepts the following query parameters:
 - `length`: the length of the slice, for example `10` (maximum: `100`)
 
 The `where` parameter must be expressed as a comparison predicate, which can be:
-- a simple predicate composed of a column name, a comparison operator, and a value
+- a simple predicate composed of a column name in double quotes, a comparison operator, and a value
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
Great !

PR #2796, opened 2024-05-13
Catch DatasetGenerationCastError and show details

AndreaFrancis:
Will fix https://github.com/huggingface/dataset-viewer/issues/2601 and https://github.com/huggingface/dataset-viewer/issues/2723
Related to https://github.com/huggingface/dataset-viewer/issues/1443

Current number of records with this issue (UnexpectedError):
```
{ _id: { error: 'DatasetGenerationError' }, count: 5504 }
  { _id: { error: 'DatasetGenerationCastError' }, count: 1697 }
```

### DIFF
--- a/libs/libcommon/src/libcommon/exceptions.py
+++ b/libs/libcommon/src/libcommon/exceptions.py
@@ -78,6 +78,7 @@ CacheableErrorCode = Literal[
     "ConfigNamesError",
     "ConfigNotFoundError",
     "CreateCommitError",
+    "DatasetGenerationCastError",
     "DatasetInBlockListError",
     "DatasetManualDownloadError",
     "DatasetModuleNotInstalledError",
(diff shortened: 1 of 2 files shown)


severo (approved):
Nice!

Just for completeness: I was wondering if other steps could create this error. I checked in the db, and all the errors indeed come from this step.

AndreaFrancis:
> I was wondering if other steps could create this error

I saw the errors come from calling `builder.download_and_prepare` and `builder._prepare_split`, which look to be invoked only in this step so, I think we are ok for now.

severo:
The full trace now shows in the dataset viewer, which should help the users fix their dataset (it's generally a user issue, not a bug)

For DatasetGenerationError: https://huggingface.co/datasets/Recag/Rp_CommonC_185_1

Capture d’écran 2024-05-14 à 08 54 44

For DatasetGenerationCastError: https://huggingface.co/datasets/RealTimeData/wikitext_alltime_backup

Capture d’écran 2024-05-14 à 08 56 31

PR #2797, opened 2024-05-13
Fix stats for data with NaN (not a number) values

polinaeterna:
While the viewer shows both `null` (=python `None`) and `NaN` (=python `float("nan")`) values equally (as nulls, missing data), they are internally not equal. Parquet statistics do not include `NaN` values in `null` count; `polars` returns `NaN` as a mean/median/std value for column with `NaN` values (so, with at least one value), and histogram compute is currently broken.

I propose to treat `NaN`  values as `null` values. I don't see much benefit in introducing another category for them and also we don't differentiate between `NaN` and `null` in the viewer itself. Let me know what you think about 

I didn't think of this difference before :( so actually most of the changes in this PR is renaming `nan` to `null` in tests lol. would be also nice to rename in response schema but this should be avoided in general and in this case not that crucial.

---

^ example of such data: https://huggingface.co/datasets/lhoestq/test-cmcv (columns `longitude` and `latitude`)


severo:
It's so specific and, at the same time, so difficult...

NaN does not exist in JSON, for example, but is supported by `JSON.parse/serialize` in Javascript and by `json.dumps/parse` in Python.

NaN must not be used for min/max statistics in Parquet (https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L978).

I don't really know what we should do. Here, we're replacing the `NaN` values with `null` before computing stats, but we still send NaN values in JSON in /first-rows or /rows (even if it's outside the JSON spec).

severo:
interesting. Indeed, we use `orjson`, not `json`, and it serializes `NaN` to `null`:

https://github.com/ijl/orjson

> has strict JSON conformance in not supporting Nan/Infinity/-Infinity

> orjson.dumps() serializes Nan, Infinity, and -Infinity, which are not compliant JSON, as null

severo:
I'm wondering what occurs with the conversion to duckdb and the /filter endpoint. Are the nan elements conserved in the duckdb file? is it possible to filter on them? I'm not sure NaN is a thing in SQL at all.

Should we take a global decision for dataset viewer to consider (and convert) all the NaN values to null?

polinaeterna:
there are both `nan` and `null` in duckdb format:
image

so, the result of https://huggingface.co/datasets/lhoestq/test-cmcv/viewer/default/train?f[longitude]=null is empty

severo:
And https://huggingface.co/datasets/lhoestq/test-cmcv/viewer/default/train?f[longitude][value]=%27NaN%27 gives us the "NaN" values

polinaeterna:
thanks to discussion with @anton-l I realized that there actually might be a practical reason to differentiate between `NaN` and `null` in the viewer - to let people know what methods they should use when processing data. For example, `polars` has `.fill_null()` and `.fill_nan()`. so maybe we actually should display them differently and to count the number of `NaN` values as a separate category for floats.

severo:
yes, it adds complexity, but seems to be more respectful of the users' datasets

lhoestq:
yes I'd be in favor of differentiating them as well

lhoestq (approved):
LGTM ! feel free to mention this behavior in the docs btw

severo (approved):

PR #2816, opened 2024-05-15
Remove nginx reverse proxy from Helm chart

severo:
following #2713, ALB serves the routes, and the nginx reverse proxy is unused. We remove it from Helm.


rtrompier (approved):

services/reverse-proxy/README.md
```diff
@@ -1,10 +1,10 @@
 # Dataset viewer - reverse proxy
 
-> Reverse-proxy in front of the API
+> Reverse-proxy in front of the API. Only used in the docker compose (CI), not in the Helm chart (staging/prod)
 
 See [docker-compose-dataset-viewer.yml](../../tools/docker-compose-dataset-viewer.yml) for usage.
 
-Note that the template configuration is located in [chart/nginx-templates/](../../chart/nginx-templates/) in order to be reachable by the Helm chart to deploy on Kubernetes.
+Note that the template configuration is located in [tools/nginx-templates/](../../tools/nginx-templates/) in order to be reachable by the Helm chart to deploy on Kubernetes.
```
AndreaFrancis, line 7: I am confused. Will it be used by the Helm chart? (I thought no because of the comment in line 3)

AndreaFrancis (approved):
Thanks for removing also the old duckdb volume (I forgot to do it).

services/reverse-proxy/README.md
severo: you're right! I removed this sentence (and moved the templates to the same directory)

PR #2821, opened 2024-05-16
Fix admin again

severo:
fix #2820


albertvillanova (commented):
Thanks for the fix. It is weird that we did not notice the CI issue fixed by:
- https://github.com/huggingface/dataset-viewer/pull/2825#pullrequestreview-2062762116

PR #2822, opened 2024-05-16
disable dataset viewer for 'nfaa' tag

severo:
See https://huggingface.slack.com/archives/C042H6N5JR0/p1714771691945989 (internal)

On the first backfill cron job after this PR is merged, all the datasets with the "nfaa" tag will be deleted from the dataset viewer's cache.


albertvillanova (approved):
Thanks.

libs/libcommon/src/libcommon/constants.py
```diff
@@ -33,6 +33,7 @@
 PROCESSING_STEP_CONFIG_PARQUET_AND_INFO_ROW_GROUP_SIZE_FOR_BINARY_DATASETS = 100
 PARQUET_REVISION = "refs/convert/parquet"
 
+UNSUPPORTED_TAG_NFAA = "not-for-all-audiences"
```
Pierrci, line 36: we have other tags in addition to the official one for which we want to apply the same policy if possible, see https://github.com/huggingface/moon-landing/blob/ec6c0f42e48d3479a09b85a1a3dc81915cb88920/server/repo_types/Tags.ts#L2 (I think it's ok to just copy/paste, we'll try to ping you if/when we add more)

Pierrci:
cc @allendorf @No-m4d for vis on this

libs/libcommon/src/libcommon/constants.py
severo: OK, done in https://github.com/huggingface/dataset-viewer/pull/2823

PR #2825, opened 2024-05-16
fix e2e test

### LINKED ISSUE #2824: an e2e test is broken since #2821

severo:
fix #2824


albertvillanova (commented):
Thanks for the fix. It is weird that we did not notice it in our CI in the https://github.com/huggingface/dataset-viewer/pull/2821 PR.

PR #2829, opened 2024-05-17
fix: Empty split list 

AndreaFrancis:
Fix for https://github.com/huggingface/dataset-viewer/issues/2803

### DIFF
--- a/services/worker/src/worker/job_runners/config/split_names.py
+++ b/services/worker/src/worker/job_runners/config/split_names.py
@@ -15,7 +15,6 @@ from libcommon.exceptions import (
     EmptyDatasetError,
     PreviousStepFormatError,
     SplitNamesFromStreamingError,
-    SplitsEmptyError,
 )
 from libcommon.simple_cache import CachedArtifactError, CachedArtifactNotFoundError, get_previous_step_or_raise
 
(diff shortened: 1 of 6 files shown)


services/worker/src/worker/job_runners/config/split_names.py
```diff
@@ -87,6 +94,10 @@ def compute_split_names_from_streaming_response(
             f"Cannot get the split names for the config '{config}' of the dataset.",
             cause=err,
         ) from err
+
+    if not split_name_items:
+        raise SplitsEmptyError("The config has no splits.")
```
severo, line 99: What is the reason for a config to have an empty list of splits?

It would help enhance the error message to help the user fix the issue. Currently ("The config has no splits.") it's hard to take action based on it.

severo: > the README file does not include a split definition

but if no split is defined, we should have one split, named "train", by default, no?

AndreaFrancis: I am not a fan of assigning default values for user responses; I honestly like to fail on the first try (from-info) and leave the second try to get the correct split (I am unsure if it will always be 'train').

severo: sure, I meant: why can config-info have an empty list of splits?

AndreaFrancis: Thanks for helping me clarify; I think I found the real issue:
Some files have 0 rows, like https://huggingface.co/datasets/Vinay573/aarkoocustomdataset/tree/main/data?show_file_info=data%2Ftrain-00000-of-00001.parquet, and when trying to fill builder info https://github.com/huggingface/dataset-viewer/blob/main/services/worker/src/worker/job_runners/config/parquet_and_info.py#L692, it cleans all the previous content (even if it was possible to get the split before). Then, since it is 0 rows, no new split is added here: https://github.com/huggingface/dataset-viewer/blob/main/services/worker/src/worker/job_runners/config/parquet_and_info.py#L747.
I think the fix should be to add the split anyway but set the values to zero (as it is by default from the library):
image

severo (approved):
Well spotted!

PR #2831, opened 2024-05-17
Fix webdatset multipart ext and npz

lhoestq:
for a new dataset being released that has a field "url.txt" that is not read correctly as text; and another field "npz" that had an error

thsoe have been fixed in https://github.com/huggingface/datasets/pull/6904 and included in branch `datasets-2.19.1-hotfix`


lhoestq:
taking the liberty to merge for a dataset being released

see [internal](https://huggingface.slack.com/archives/C06T0REVAH1/p1714048738307959)

PR #2833, opened 2024-05-17
Don't truncate urls

lhoestq:
..in first-rows, otherwise we can't show images in the viewer

related to #1416 (issue for iamge and audio type that were also truncated at one point)

### DIFF
--- a/libs/libcommon/src/libcommon/viewer_utils/rows.py
+++ b/libs/libcommon/src/libcommon/viewer_utils/rows.py
@@ -17,8 +17,8 @@ from libcommon.utils import get_json_size
 from libcommon.viewer_utils.features import get_cell_value, to_features_list
 from libcommon.viewer_utils.truncate_rows import create_truncated_row_items
 
-
 URL_COLUMN_RATIO = 0.3
+
 
 def transform_rows(
     dataset: str,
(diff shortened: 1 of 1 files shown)


lhoestq:
taking the liberty to merge, same reason as https://github.com/huggingface/dataset-viewer/pull/2831

libs/libcommon/src/libcommon/viewer_utils/rows.py
```diff
@@ -155,7 +157,27 @@ def create_first_rows_response(
...
+        or (  # column of URLs
+            isinstance(feature, Value)
+            and feature.dtype == "string"
+            and len(transformed_rows) > 0
+            and sum(
+                (
+                    col in row
+                    and isinstance(row[col], str)
+                    and row[col].startswith("http://")
+                    or row[col].startswith("https://")
```
severo, line 173: hmmm is the precedence correct? I would have enclosed the two last lines with parentheses

lhoestq: good catch, I'll fix this

lhoestq:
I opened https://github.com/huggingface/dataset-viewer/pull/2834 with a fix and a test

PR #2834, opened 2024-05-17
Fix truncation for urls

lhoestq:


severo (approved):
thanks for the quick fix!

PR #2835, opened 2024-05-17
allow dataset viewer on private NFAA datasets for Pro and Enterprise Hub

severo:
cc @Pierrci 

We enable the viewer for *private* NFAA datasets, owned by a Pro user or Enterprise Hub org.

The viewer is still disallowed on all public NFAA datasets.


polinaeterna:
I wonder why we want to enable this feature to Pro/Enterprise? what's the difference with regular users in this context?

julien-c (approved):
why not

Pierrci:
(private takes precedence over tags)

PR #2832, opened 2024-05-17
added .opus as supported audio extension

Airop:
added .opus as supported audio extension

Issue #2818, opened 2024-05-16
Do not convert Opus audio files to WAV


albertvillanova:
As suggested by @severo, we could avoid the conversion of Opus audio files to WAV: https://github.com/huggingface/dataset-viewer/pull/2811#issuecomment-2113015400

Indeed, according to https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Audio_codecs, it's a well-supported codec among browsers (see "Browser compatibility" in the Opus table).

Beware:
> Safari supports Opus in the `<audio>` element only when packaged in a CAF file, and only on macOS High Sierra (10.13) or iOS 11.

Overall:
> Opus is a good all-around audio codec for use in your web applications, and can be used for any audio tasks you have in mind.

Airop:
Hey, I would like to work on that!

PR #2841, opened 2024-05-21
Fix the no-conversion of Opus audio files

albertvillanova:
This PR fixes the no-conversion of Opus audio file:
- #2818

because the following PR does not avoid the conversion of Opus audio files to WAV:
- #2832

CC: @Airop


severo (approved):
wow, thanks!!!

Issue #2843, opened 2024-05-21
The admin-ui space cannot build


severo:
https://huggingface.co/spaces/datasets-maintainers/dataset-viewer-admin-ui

<img width="736" alt="Capture d’écran 2024-05-21 à 14 28 30" src="https://github.com/huggingface/dataset-viewer/assets/1676121/56f36a4d-e598-4166-82d3-9f47c32c56ec">

```
ERROR: Wheel 'botocore' located at /tmp/pip-unpack-6pji9o0y/botocore-1.12.198-py2.py3-none-any.whl is invalid.

[notice] A new release of pip available: 22.3.1 -> 24.0
[notice] To update, run: pip install --upgrade pip

--> ERROR: process "/bin/sh -c pip install --no-cache-dir -r /tmp/requirements.txt" did not complete successfully: exit code: 1

```

severo:
fixed by #2850

PR #2850, opened 2024-05-22
fix admin ui deps

lhoestq:
had to pin a few s3 related deps (from poetry.lock) or pip couldn't resolve them


severo:
fixes #2843

PR #2855, opened 2024-05-27
Give more ram to backfill cron jobs

severo:
The backfill jobs can restart 2, 3, even 6 times... due to OOM. And we end up with successive cron jobs occurring at the same time.

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -174,21 +174,6 @@ mongodbMigration:
     limits:
       cpu: 1
 
-# --- jobs (post-install/upgrade hooks) ---
-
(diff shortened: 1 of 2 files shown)


severo:
8GB RAM was not enough for `prod-datasets-server-backfill-retryable-errors`!

> The container last terminated with exit code 137 because of OOMKilled.

https://cd.internal.huggingface.tech/applications/argo-cd/datasets-server-prod?view=pods&node=%2FPod%2Fdatasets-server%2Fprod-datasets-server-job-backfill-retryable-errors-28613490-olx0w&tab=summary

PR #2870, opened 2024-05-29
More webdataset fixes

lhoestq:
...by pinning a more recent version of `datasets@datasets-2.19.1-hotfix`

The webdataset fixes are needed for a dataset release, see [internal](https://huggingface.slack.com/archives/C06T0REVAH1/p1716998706658099?thread_ts=1714048738.307959&cid=C06T0REVAH1)

<s>I also included your fix for the columns order and Json improvement @albertvillanova </s>

### DIFF
--- a/front/admin_ui/poetry.lock
+++ b/front/admin_ui/poetry.lock
@@ -663,8 +663,8 @@ vision = ["Pillow (>=6.2.1)"]
 [package.source]
 type = "git"
 url = "https://github.com/huggingface/datasets.git"
-reference = "11a27e9a7484a17c6d08d6a838364f9913f7bfd3"
-resolved_reference = "11a27e9a7484a17c6d08d6a838364f9913f7bfd3"
+reference = "08ae5373f3bbcae9d7390476f358cd5fc7d76fcd"
+resolved_reference = "08ae5373f3bbcae9d7390476f358cd5fc7d76fcd"
 
 [[package]]
 name = "decorator"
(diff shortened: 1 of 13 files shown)


severo (approved):

severo:
if I read correctly, we now support webdatasets with compressed files (.gz for example). Should we recompute datasets after that? What is the current error?

For the column order fix: I imagine that we will not do anything, because it would require to recompute everything, right?

lhoestq:
Preciously compressed files would be a binary type, instead of their correct type (string, lists, whatever)

Yeah no need to recompute anything for webdataset, and no need for the column order either (except the reported one maybe)

lhoestq:
One of the e2e test is failing though, not sure if it's related to your change @albertvillanova ? 

```
tests/test_14_statistics.py::test_statistics_endpoint
```

albertvillanova (commented):
Maybe... I am investigating it.

albertvillanova (commented):
Yes, it is confirmed: pandas.read_json casts the float column to int if all the column values are indeed integers... Not sure how to prevent this behavior...

albertvillanova:
I have reported the issue to pandas:
- https://github.com/pandas-dev/pandas/issues/58866

albertvillanova (commented):
You are implementing 2 fixes here: the one for webdatasets and the one for JSON column order.

Is the fix for column order as urgent as the one for webdatasets?
If not, I would suggest including only the fix for webdatasets and wait for the JSON downcasting issue to be fixed.

Otherwise, including the fix for JSON column order will introduce the bug on JSON downcasting from float to int. It is true that this is a very edge case: only appears in JSON files (not in JSON-lines) and when ALL the column values can be cast fro float to int. But maybe this could raise issues for some users.

lhoestq:
I removed the json change

PR #2874, opened 2024-05-30
Re-add torch dependency

lhoestq:
needed for webdataset with .pth files in them


lhoestq:
arf it's failing, I'll try to install the cpu only version

lhoestq:
added to worker only actually, this should keep the other docker images lighter

severo (approved):

PR #2879, opened 2024-06-03
Use pymongoarrow to get dataset results as dataframe

AndreaFrancis:
Fix for https://github.com/huggingface/dataset-viewer/issues/2868

### DIFF
--- a/libs/libcommon/src/libcommon/queue.py
+++ b/libs/libcommon/src/libcommon/queue.py
@@ -17,14 +17,13 @@ from typing import Any, Generic, Literal, Optional, TypedDict, TypeVar
 from uuid import uuid4
 
 import pandas as pd
-import pyarrow as pa
+import pymongoarrow as pma
 import pytz
 from bson import ObjectId
 from mongoengine import Document
 from mongoengine.errors import DoesNotExist, NotUniqueError
 from mongoengine.fields import DateTimeField, EnumField, IntField, ObjectIdField, StringField
 from mongoengine.queryset.queryset import QuerySet
-from pymongoarrow.api import find_arrow_all
 
 from libcommon.constants import (
     DEFAULT_DIFFICULTY_MAX,
@@ -190,16 +189,15 @@ class JobDocument(Document):
     objects = QuerySetManager["JobDocument"]()
 
     @classmethod
-    def fetch_arrow_table(
+    def fetch_as_df(
         cls, query: Optional[Mapping[str, Any]] = None, projection: Optional[Mapping[str, Any]] = None
-    ) -> pa.Table:
+    ) -> pd.DataFrame:
         """
-        Fetch documents matching the query as an Apache Arrow Table.
+        Fetch documents matching the query as a Pandas Dataframe.
         """
         query = query if query is not None else {}
         collection = cls._get_collection()
-        arrow_table = find_arrow_all(collection, query, projection=projection)
-        return arrow_table
+        return pma.api.find_pandas_all(collection, query, projection=projection)  # type: ignore
 
     def info(self) -> JobInfo:
         return JobInfo(
@@ -501,8 +499,7 @@ def release_lock(key: str) -> None:
     )
 
 
-PA_PROJECTION = {
-    "job_id": 1,
+PENDING_JOBS_PROJECTION = {
     "type": 1,
     "priority": 1,
     "status": 1,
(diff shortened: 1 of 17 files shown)


AndreaFrancis:
I put two tests per _get_df function to show the execution difference between the previous mapping and pymongoarrow. If there is an improvement, I should remove the old code.

libs/libcommon/src/libcommon/queue.py
```diff
@@ -116,6 +119,21 @@ class JobQueryFilters(TypedDict, total=False):
     difficulty__lte: int
 
 
+PA_SCHEMA = Schema(
+    {
+        "_id": bson.ObjectId,
```
AndreaFrancis, line 124: https://www.mongodb.com/docs/languages/python/pymongo-arrow-driver/current/data-types/ for more reference

libs/libcommon/src/libcommon/queue.py
```diff
@@ -116,6 +119,21 @@ class JobQueryFilters(TypedDict, total=False):
...
+        "type": pa.string(),
+        "dataset": pa.string(),
+        "revision": pa.string(),
+        "config": pa.string(),
+        "split": pa.string(),
+        "priority": pa.string(),
+        "status": pa.string(),
+        "created_at": pa.timestamp("ms"),
```
AndreaFrancis, line 132: I tried to align with the old mapping "ns," but Pymongoarrow throws an error:
`TypeError: PyMongoArrow only supports millisecond temporal resolution compatible with MongoDB's UTC datetime type.`
I verified, and also, we support up to milliseconds in the Mongo Collection, so I think it is not a big deal.

libs/libcommon/tests/test_state.py
```diff
@@ -47,6 +48,42 @@ def cache_mongo_resource_autouse(cache_mongo_resource: CacheMongoResource) -> Ca
...
+
+
+@pytest.mark.limit_memory("2 MB")  # Success, it uses ~1.5 MB
+def test_get_cache_entries_df() -> None:
+    cache_kinds, expected_entries = populate_cache()
+    entries = get_cache_entries_df(dataset=DATASET_NAME, cache_kinds=cache_kinds)
+    assert entries.shape[0] == expected_entries
+
+
+@pytest.mark.limit_memory("2 MB")  # Will fail, it uses ~2.8 MB
```
AndreaFrancis, line 80: While getting cache df records, the percentage reduction is ~46%, so we are good by switching to pymongoarrow.

libs/libcommon/tests/test_orchestrator.py
```diff
@@ -194,6 +202,33 @@ def test_finish_job(
...
+
+
+@pytest.mark.limit_memory("1.4 MB")  # Success, it uses ~1.4 MB
+def test_get_pending_jobs_df() -> None:
+    populate_queue()
+    pending_jobs_df = Queue().get_pending_jobs_df(dataset=DATASET_NAME)
+    assert pending_jobs_df.shape == (250, 9)
+
+
+@pytest.mark.limit_memory("1.6 MB")  # Will fail, it uses ~1.6 MB
```
AndreaFrancis, line 225: For `get_pending_jobs_df,` the percentage reduction is ~12%; I will leave the team to decide if this is a good indicator to switch to pymongoarrow.
The most impactful is that after getting the data frame, it is needed to perform other transformations (Like renaming _id -> job_id and making status and priority a Categorical value to be used later in a sort statement in [here](https://github.com/huggingface/dataset-viewer/blob/main/libs/libcommon/src/libcommon/state.py#L53).
There surely is another way to handle the state logic, but I would like to make a few changes initially and improve progressively once we are sure that nothing breaks with the switch to pymongoarrow.

libs/libcommon/src/libcommon/queue.py
```diff
@@ -1019,6 +1046,27 @@ def _get_df(self, jobs: list[FlatJobInfo]) -> pd.DataFrame:
         # ^ does not seem optimal at all, but I get the types right
 
     def get_pending_jobs_df(self, dataset: str, job_types: Optional[list[str]] = None) -> pd.DataFrame:
+        filters = {"dataset": dataset}
+        if job_types:
+            filters["type"] = {"$in": job_types}  # type: ignore
+        df = JobDocument.fetch_as_df(query=filters)
+        df.rename(columns={"_id": "job_id"}, inplace=True)
```
AndreaFrancis, line 1053: I didn't find a way to load the data frame with a 'job_id' alias or a projection. Do you have any ideas?

libs/libcommon/src/libcommon/queue.py
```diff
@@ -1019,6 +1046,27 @@ def _get_df(self, jobs: list[FlatJobInfo]) -> pd.DataFrame:
...
+        df["priority"] = pd.Categorical(
```
AndreaFrancis, line 1054: It is needed for https://github.com/huggingface/dataset-viewer/blob/main/libs/libcommon/src/libcommon/state.py#L53
same for "priority".

severo (approved):

libs/libcommon/src/libcommon/simple_cache.py
```diff
@@ -834,6 +862,13 @@ def _get_df(entries: list[CacheEntryFullMetadata]) -> pd.DataFrame:
 
 
 def get_cache_entries_df(dataset: str, cache_kinds: Optional[list[str]] = None) -> pd.DataFrame:
+    filters = {"dataset": dataset}
+    if cache_kinds:
+        filters["kind"] = {"$in": cache_kinds}  # type: ignore
```
lhoestq, line 867: the syntax is not the same as `filters["kind__in"] = cache_kinds` ?

lhoestq (approved):
sounds good, feel free to iterate on this !

libs/libcommon/src/libcommon/queue.py
lhoestq: it's ok imo

libs/libcommon/src/libcommon/simple_cache.py
AndreaFrancis: No, I think kind__in is more for `mongoengine` but in this case it was needed a `pymongo` style

PR #2880, opened 2024-06-04
Allow mnist and fashion mnist + remove canonical dataset logic

lhoestq:


severo (approved):
cool!

severo:
about "remove canonical dataset logic": we possibly have other places in the code where we handle canonical repos. Maybe in another PR?

PR #2884, opened 2024-06-05
Add retry mechanism to get_parquet_file in parquet metadata step

polinaeterna:
...to see if it helps with [FineWeb](https://huggingface.co/datasets/HuggingFaceFW/fineweb) `config-parquet-metadata` issue. 

Currently the error says just `Server disconnected` which seems to be a `aiohttp.ServerDisconnectedError` error.

If that works, a more fundamental solution would be to completely switch to `HfFyleSystem` instead of `HTTPFileSystem` and remove retries.

### DIFF
--- a/services/worker/src/worker/job_runners/config/parquet_metadata.py
+++ b/services/worker/src/worker/job_runners/config/parquet_metadata.py
@@ -15,7 +15,7 @@ from libcommon.exceptions import (
 )
 from libcommon.simple_cache import get_previous_step_or_raise
 from libcommon.storage import StrPath
-from libcommon.utils import HF_HUB_HTTP_ERROR_RETRY_SLEEPS, retry
+from libcommon.utils import retry
 from libcommon.viewer_utils.parquet_metadata import create_parquet_metadata_file
 from tqdm.contrib.concurrent import thread_map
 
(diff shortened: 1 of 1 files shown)


services/worker/src/worker/job_runners/config/parquet_metadata.py
```diff
@@ -31,7 +33,10 @@ def create_parquet_metadata_file_from_remote_parquet(
     parquet_file_item: SplitHubFile, fs: HTTPFileSystem, hf_token: Optional[str], parquet_metadata_directory: StrPath
 ) -> ParquetFileMetadataItem:
     try:
-        parquet_file = get_parquet_file(url=parquet_file_item["url"], fs=fs, hf_token=hf_token)
+        retry_get_parquet_file = retry(on=[aiohttp.ServerConnectionError], sleeps=HF_HUB_HTTP_ERROR_RETRY_SLEEPS)(
```
polinaeterna, line 36: or should it be specifically `ServerDisconnectedError`?

lhoestq: if ServerConnectionError is a parent class of ServerDisconnectedError it's ok imo

lhoestq: btw `retry` will do a `sleep` before each first call to `get_parquet_file`, which can make the parquet-metadata job a little bit slower (it's unlikely to be longer that the maximum allowed 40min but better keep this in mind when re-running it for datasets like fineweb)

lhoestq (approved):
lgtm !

services/worker/src/worker/job_runners/config/parquet_metadata.py
polinaeterna: I changed the first value of sleeps from 1 to 0.2

polinaeterna:
seems that it helped! but it was slow indeed :( almost half an hour
image
image

lhoestq:
Great ! Given the number of files it's not absurd ^^

PR #2888, opened 2024-06-06
Fix get_shape in statistics when argument is bytes, not dict 

polinaeterna:
Will fix duckdb-index step for [common-canvas/commoncatalog-cc-by](https://huggingface.co/datasets/common-canvas/commoncatalog-cc-by)


lhoestq (approved):
good catch !

PR #2889, opened 2024-06-06
No auto backfill on most nfaa datasets

lhoestq:
they correlate with memory spikes, I'm just hardcoding this to see the impact on memory and further investigate

### DIFF
--- a/libs/libapi/src/libapi/utils.py
+++ b/libs/libapi/src/libapi/utils.py
@@ -2,6 +2,7 @@
 # Copyright 2022 The HuggingFace Authors.
 
 import logging
+import random
 from collections.abc import Callable, Coroutine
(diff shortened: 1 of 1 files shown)


libs/libapi/src/libapi/utils.py
```diff
@@ -2,6 +2,7 @@
 # Copyright 2022 The HuggingFace Authors.
 
 import logging
+import random
```
severo, line 5:
```suggestion
```

severo (approved):

AndreaFrancis (approved):

libs/libapi/src/libapi/utils.py
lhoestq: oops !

PR #2890, opened 2024-06-06
No mongo cache in DatasetRemovalPlan

lhoestq:
don't keep the full query result in memory (as mongoengine does by default)

this should reduce the frequency of memory spikes and could have an effect of the memory issues we're having


lhoestq:
taking the liberty to merge to try it out since sylvain is off

PR #2891, opened 2024-06-07
Make StorageClient not warn when deleting a non-existing directory

albertvillanova:
Make StorageClient not warn when trying to delete a non-existing directory.

I think we should only log a warning if the directory exists and could not be deleted.


severo (approved):
yes, it makes sense.

ideally, if `.rm()` raises a dedicated error when the directory does not exist, we could catch it, instead of calling two methods successively. But if it's not the case, the current test is good too

albertvillanova:
@severo I have checked that both `LocalFileSystem` and `S3FileSystem` raise `FileNotFoundError` if the directory does not exist.

PR #2896, opened 2024-06-11
Remove Prometheus context label

### LINKED ISSUE #2895: Too high label cardinality metrics in Prometheus
As discussed privately (CC: @McPatate), we may have too high label cardinality metrics in Prometheus. As stated in Prometheus docs: https://prometheus.io/docs/practices/naming/#labels
> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

Currently, our `StepProfiler` uses 3 labels: `method`, `step` and `context`.

Recently, we fixed an issue with object memory addresses as values in the `context` label:
- #2893

Note that sometimes we pass the dataset name in the `context` label, for example in `DatasetBackfillPlan`: https://github.com/huggingface/dataset-viewer/blob/19c049499af1b37cf4d9dfa93b58657d66b75337/libs/libcommon/src/libcommon/orchestrator.py#L490

Other times we pass the dataset and the config names in the `context` label, for example in `ConfigState`: https://github.com/huggingface/dataset-viewer/blob/19c049499af1b37cf4d9dfa93b58657d66b75337/libs/libcommon/src/libcommon/state.py#L202

For the (`method`, `step`, `context`) labels, this creates (number_of_method_values * number_of_step_values * **number_of_datasets**) or (number_of_method_values * number_of_step_values * **number_of_dataset_configs**) unique label values! This might be a too large cardinality.

One possible solution to reduce cardinality would be to suppress the `context` label. What do you think? Other proposals?
CC: @huggingface/dataset-viewer

albertvillanova:
Remove Prometheus context label.

Fix #2895.

### DIFF
--- a/libs/libapi/src/libapi/authentication.py
+++ b/libs/libapi/src/libapi/authentication.py
@@ -96,7 +96,6 @@ async def auth_check(
         with StepProfiler(
             method="auth_check",
             step="requests.get",
-            context=f"external_auth_url={external_auth_url} timeout={hf_timeout_seconds}",
         ):
             try:
                 logging.debug(
--- a/libs/libcommon/src/libcommon/orchestrator.py
+++ b/libs/libcommon/src/libcommon/orchestrator.py
@@ -125,7 +125,6 @@ class CreateJobsTask(Task):
         with StepProfiler(
             method="CreateJobsTask.run",
             step="all",
-            context=f"num_jobs_to_create={len(self.job_infos)}",
         ):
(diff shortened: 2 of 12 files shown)


severo (approved):
Perfect. So sorry about this, this info should have gone to logs, if any, but not in metrics

lhoestq:
nice ! thanks @albertvillanova !

PR #2897, opened 2024-06-11
Use `HfFileSystem` in config-parquet-metadata step instead of `HttpFileSystem`

polinaeterna:
`config-parquet-metadata` step is failing again for [FineWeb](https://huggingface.co/datasets/HuggingFaceFW/fineweb) with errors like 
```
"Could not read the parquet files: 504, message='Gateway Time-out', url=URL('https://huggingface.co/datasets/HuggingFaceFW/fineweb/resolve/refs%2Fconvert%2Fparquet/default/train-part1/4089.parquet')"
```
Maybe this would help (the same is used in `config-parquet-and-info` step which works).

Previous fix was https://github.com/huggingface/dataset-viewer/pull/2884

### DIFF
--- a/services/worker/src/worker/job_runners/config/parquet_metadata.py
+++ b/services/worker/src/worker/job_runners/config/parquet_metadata.py
@@ -28,7 +28,7 @@ from worker.dtos import (
     ParquetFileMetadataItem,
 )
 from worker.job_runners.config.config_job_runner import ConfigJobRunner
-from worker.utils import hf_hub_open_file, hf_hub_parquet_path
+from worker.utils import hf_hub_open_file, hffs_parquet_url
 
 SLEEPS = [0.2, 1, 1, 10, 10, 10]
 
@@ -45,7 +45,7 @@ def create_parquet_metadata_file_from_remote_parquet(
     parquet_file_item: SplitHubFile, hf_endpoint: str, hf_token: Optional[str], parquet_metadata_directory: StrPath
 ) -> ParquetFileMetadataItem:
     split_directory = extract_split_name_from_parquet_url(parquet_file_item["url"])
-    hfh_parquet_file_path = hf_hub_parquet_path(
+    hfh_parquet_file_path = hffs_parquet_url(
         repo_id=parquet_file_item["dataset"],
         config=parquet_file_item["config"],
         split_directory=split_directory,
--- a/services/worker/src/worker/job_runners/config/parquet_metadata.py
+++ b/services/worker/src/worker/job_runners/config/parquet_metadata.py
@@ -6,17 +6,19 @@ import logging
 from typing import Optional
 
 import aiohttp
-from fsspec.implementations.http import HTTPFileSystem
+from huggingface_hub import HfFileSystemFile
 from libcommon.dtos import JobInfo, SplitHubFile
 from libcommon.exceptions import (
     FileSystemError,
     ParquetResponseEmptyError,
     PreviousStepFormatError,
 )
+from libcommon.parquet_utils import extract_split_name_from_parquet_url
 from libcommon.simple_cache import get_previous_step_or_raise
 from libcommon.storage import StrPath
 from libcommon.utils import retry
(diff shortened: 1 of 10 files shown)


services/worker/src/worker/job_runners/config/parquet_metadata.py
```diff
@@ -26,17 +28,34 @@
...
 )
 from worker.job_runners.config.config_job_runner import ConfigJobRunner
-from worker.utils import get_parquet_file
+from worker.utils import hf_hub_open_file, hffs_parquet_url
 
 SLEEPS = [0.2, 1, 1, 10, 10, 10]
 
 
+# TODO: on which error?
+@retry(sleeps=SLEEPS)
```
polinaeterna, line 37: not sure which error I should retry on, in `config-parquet-and-info`, which basically does the same thing, it's `pa.ArrowInvalid`: 
https://github.com/huggingface/dataset-viewer/blob/e7e674ad8dcd0f8fcd1e2ede0591fd8c0afb4ee3/services/worker/src/worker/job_runners/config/parquet_and_info.py#L595C26-L595C38
But I'm not sure where this error might come from.

services/worker/src/worker/job_runners/config/parquet_metadata.py
```diff
@@ -26,17 +28,34 @@
...
+        repo_id=parquet_file_item["dataset"],
+        config=parquet_file_item["config"],
+        split_directory=split_directory,
+        filename=parquet_file_item["filename"],
+    )
     try:
-        retry_get_parquet_file = retry(on=[aiohttp.ServerConnectionError], sleeps=SLEEPS)(get_parquet_file)
-        parquet_file = retry_get_parquet_file(url=parquet_file_item["url"], fs=fs, hf_token=hf_token)
+        f = retry_open_file(
+            file_url=hfh_parquet_file_path, hf_endpoint=hf_endpoint, hf_token=hf_token, revision=PARQUET_REVISION
```
polinaeterna, line 56: maybe revision should be passed as a worker's config parameter?

services/worker/src/worker/job_runners/config/parquet_metadata.py
lhoestq: Not sure either, I had this error from time to time in prod.

I added this retry in https://github.com/huggingface/dataset-viewer/pull/1404

Maybe corrupted queries ?

services/worker/src/worker/job_runners/config/parquet_metadata.py
lhoestq: I think it's fine, if there is a parquet-and-info job writing parquet files it will retrigger parquet-metadata later anyway

services/worker/src/worker/job_runners/config/parquet_metadata.py
lhoestq: I guess you can use pa.ArrowInvalid as well

services/worker/src/worker/job_runners/config/parquet_metadata.py
```diff
@@ -26,17 +28,34 @@
...
+    file_url: str, hf_endpoint: str, hf_token: Optional[str], revision: Optional[str] = None
+) -> HfFileSystemFile:
+    return hf_hub_open_file(file_url=file_url, hf_endpoint=hf_endpoint, hf_token=hf_token, revision=revision)
+
+
 def create_parquet_metadata_file_from_remote_parquet(
-    parquet_file_item: SplitHubFile, fs: HTTPFileSystem, hf_token: Optional[str], parquet_metadata_directory: StrPath
+    parquet_file_item: SplitHubFile, hf_endpoint: str, hf_token: Optional[str], parquet_metadata_directory: StrPath
 ) -> ParquetFileMetadataItem:
+    split_directory = extract_split_name_from_parquet_url(parquet_file_item["url"])
```
lhoestq, line 47: maybe we can rename `extract_split_name_from_parquet_url` -> `extract_split_directory_from_parquet_url ` since the split directory is different from the split name: it can be e.g. `train` or `partial-train` or `train-part0`

PR #2900, opened 2024-06-11
2754 partial instead of error

### LINKED ISSUE #2754: Return partial dataset-hub-cache instead of error?
`dataset-hub-cache` depends on multiple previous steps, and any error in one of them makes it fail. It provokes things like https://github.com/huggingface/moon-landing/issues/9799 (internal): in the datasets list, a dataset is not marked as "supporting the dataset viewer", whereas the only issue is that we didn't manage to list the compatible libraries, to create the tags.

https://github.com/huggingface/dataset-viewer/blob/main/services/worker/src/worker/job_runners/dataset/hub_cache.py

In this case, we could return a partial response, or maybe return an empty list of libraries or modalities if we have an error.

What do you think @lhoestq?

severo:
fix #2754

### DIFF
--- a/services/worker/src/worker/dtos.py
+++ b/services/worker/src/worker/dtos.py
@@ -344,7 +344,7 @@ class DatasetHubCacheResponse(TypedDict):
     preview: bool
     viewer: bool
     partial: bool
-    num_rows: Union[int, None]
+    num_rows: Optional[int]
     tags: list[DatasetTag]
     libraries: list[DatasetLibrary]
     modalities: list[DatasetModality]
--- a/services/worker/src/worker/dtos.py
+++ b/services/worker/src/worker/dtos.py
@@ -344,7 +344,7 @@ class DatasetHubCacheResponse(TypedDict):
     preview: bool
     viewer: bool
     partial: bool
-    num_rows: int
+    num_rows: Union[int, None]
     tags: list[DatasetTag]
     libraries: list[DatasetLibrary]
     modalities: list[DatasetModality]
--- a/services/worker/src/worker/job_runners/dataset/hub_cache.py
+++ b/services/worker/src/worker/job_runners/dataset/hub_cache.py
@@ -72,7 +72,7 @@ def compute_hub_cache_response(dataset: str) -> tuple[DatasetHubCacheResponse, f
         raise
     except Exception:
         partial = False
-        num_rows = 0
+        num_rows = None
         size_progress = 0.0
 
     progress = min((p for p in [is_valid_progress, size_progress] if p is not None), default=0.0)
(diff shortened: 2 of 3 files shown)


severo:
I don't increment the version, let's wait for https://github.com/huggingface/dataset-viewer/issues/2898, to save some resources.

services/worker/src/worker/job_runners/dataset/hub_cache.py
```diff
@@ -51,23 +51,29 @@ def compute_hub_cache_response(dataset: str) -> tuple[DatasetHubCacheResponse, f
...
+                "Previous step 'dataset-size' did not return the expected content: 'partial' or 'size.dataset.num_rows'."
+            )
+        partial = content["partial"]
+        num_rows = content["size"]["dataset"]["num_rows"]
+        size_progress = size_response["progress"]
+    except (CachedArtifactNotFoundError, PreviousStepFormatError):
+        raise
+    except Exception:
+        partial = False
+        num_rows = 0
```
lhoestq, line 75: the hub won't show 0 rows on the dataset page right ?

lhoestq (approved):
LGTM !

services/worker/src/worker/job_runners/dataset/hub_cache.py
severo: hmmm maybe it should be optional... Looking at it

severo: Possible changes soon to this response:
- make num_rows an optional field?
- add num_rows_estimated (see https://github.com/huggingface/dataset-viewer/pull/2906)
- remove "tags"?
- replace "viewer / preview / partial" with one field that would simplify the logic on the Hub side (currently: `!hub_cache.viewer ? "preview" : hub_cache.partial ? "viewer-partial" : "viewer"`) -> `viewer: "viewer" | "viewer-partial" | "preview" | "none"`?

We might as well work all this in a new step (`dataset-moonlanding-cache` for example). And once we've finished this batch of changes, prepare the Hub to consume the new step, and remove `dataset-hub-cache` here.

wdyt @lhoestq?

severo: It's currently not used on moonlanding btw. So, we can change it. I agree that `num_rows = 0` is wrong.

Two options:
- `num_rows = None`
- don't set the `num_rows` field

Let's set it to `None`

services/worker/src/worker/dtos.py
```diff
@@ -344,7 +344,7 @@ class DatasetHubCacheResponse(TypedDict):
     preview: bool
     viewer: bool
     partial: bool
-    num_rows: int
+    num_rows: Union[int, None]
```
lhoestq, line 347: Optional in python means Union[int, None]

```suggestion
    num_rows: Optional[int]
```

lhoestq (approved):

services/worker/src/worker/dtos.py
severo: sure

severo:
Once deployed, I'll relaunch the datasets with error

PR #2902, opened 2024-06-12
Minor fix id with length of str dataset name

albertvillanova:
This is a minor fix of some Tasks ids containing the length of of the string dataset name.

I discovered this while investigating the memory leak issue.


severo (approved):

severo (commented):
hmmm, maybe set `"DeleteDatasetJobs,1"` and "DeleteDatasetCacheEntries,1" instead?
To be honest, I don't know if this is still important, it was more for debugging iirc

PR #2903, opened 2024-06-12
Pass copies of DataFrames instead of views

albertvillanova:
As the memory leak may be caused by improperly de-referenced objects, better pass copies of DataFrames instead of views.

In a subsequent PR I could try to optimize memory usage by not storing unnecessary data.


severo (commented):
could you add stress tests that loop on these methods, and use the memray decorators to show that we reduce the leaked (or used) memory?

or is it unrelated?

albertvillanova:
Superseded by:
- #2921

PR #2906, opened 2024-06-12
[Config-parquet-and-info] Compute estimated dataset info

lhoestq:
This will be useful to show the estimate number of rows of datasets that are partially converted to Parquet

I added `estimated_dataset_info` to the` parquet-and-info` response.
It contains estimations of:
- download_size
- num_bytes
- num_examples

Then we'll be able to propagate this info to the `size` jobs and then to `hub-cache`.

I'll run it on some datasets to check if it works fine and when it's ok I'll re-run the jobs for datasets for which we need to estimate the number of rows.

TODO:
- [x] compute sizes of non-read files
- [x] test with compressed files
- [x] test with zip archives
- [x] add track_read() test (will add before merging)


lhoestq:
Actually I still have to compute the sizes or un-read files

lhoestq:
It was a bit of a rabbit hole to do this in a clean way, but it's all good now ! Let me know if you have comments :)

I'll enable it only for allenai/c4 for now probably to try it out

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -840,6 +858,144 @@ def __exit__(
...
+                    f.readline = functools.partial(tracker.track_read, urlpath, f.readline)
+                if hasattr(f, "readlines"):
+                    f.readlines = functools.partial(tracker.track_read, urlpath, f.readlines)
+                tracker.files[urlpath] = {"read": 0, "size": int(f.size)}
+            return f
+
+        # track files reads from local, http and hf file-systems
+        self.exit_stack.enter_context(
+            patch.object(LocalFileSystem, "open", autospec=True)
+        ).side_effect = functools.partial(wrapped, fs_open=LocalFileSystem.open)
```
severo, line 979: wokay I'm totally lost. I trust you

severo (approved):
I read everything, but I understood only a part... I cannot do more than trust you and the tests.

I'm a bit worried by the size and complexity of parquet_and_info.py though, because it becomes harder to maintain with every addition. Also: few intermediate functions are tested, the tests are done on the results, and maybe don't cover all the cases (I don't know). Not necessarily to be addressed in this PR.

services/worker/src/worker/job_runners/config/parquet_and_info.py
lhoestq: I'll add some comments to make it clearer what it does ! (basically it replaces LocalFileSystem.open by a wrapper temporarily - the replacement is undone once the track_reads context exits)

lhoestq:
I added some comments, enabled on allenai/c4 and datasets-maintainers org and added a migration to add estimated_dataset_info to existing cache entries

severo:
Nice, thanks. Feel free to merge and deploy!

PR #2907, opened 2024-06-12
Move secrets to Infisical

rtrompier:


AndreaFrancis (approved):

severo (approved):
TBH, I don't have any idea if this is correct or not. Let's synchronize between the infra and datasets team when we deploy it, in order to fix things quickly if it breaks

PR #2908, opened 2024-06-13
Add `started_at` field to cached response documents

polinaeterna:
Will close https://github.com/huggingface/dataset-viewer/issues/2892

To pass `started_at` field from Job to CachedResponse, I updated `JobInfo` class so that it stores `started_at` info too. It is None by default (and set to actual time by `Queue._start_newest_job_and_delete_others()`). Maybe there is better way to pass this information? 
I think alternatively I can get a new timestamp inside worker's `Loop.process_next_job` and pass it to `finish` to avoid changing dtos, but I don't know if it's better.


libs/libcommon/tests/test_queue.py
```diff
@@ -541,7 +541,7 @@ def test_queue_get_zombies() -> None:
         difficulty=test_difficulty,
     )
     queue.start_job(job_types_only=[job_type])
-    assert queue.get_zombies(max_seconds_without_heartbeat=10) == [zombie.info()]
+    assert queue.get_zombies(max_seconds_without_heartbeat=10) == [zombie.reload().info()]
```
polinaeterna, line 544: reload here since `started_at` field has been updated after `.start_job()`

severo (commented):
I'm wondering: will we ever need the datetime? Or are we always interested in the job duration? If so, maybe it's better to store the duration instead because it will be easier to consume it and run queries.

Regarding "None" (https://github.com/huggingface/dataset-viewer/pull/2908#pullrequestreview-2125746016), I'm not sure it's an issue. If we don't have the information about the job duration, it is better to set None than 0. We can remove the Optional part later if all the entries have been updated.

polinaeterna:
close in favor of https://github.com/huggingface/dataset-viewer/pull/2967

polinaeterna:
closing in favor of https://github.com/huggingface/dataset-viewer/pull/2967

PR #2912, opened 2024-06-14
Additional modalities detection

severo:
- [x] detect tabular
- [x] detect timeseries

### DIFF
--- a/services/worker/src/worker/dtos.py
+++ b/services/worker/src/worker/dtos.py
@@ -333,7 +333,7 @@ class DatasetCompatibleLibrariesResponse(TypedDict):
     formats: list[DatasetFormat]
 
 
-DatasetModality = Literal["image", "audio", "text", "video", "geospatial", "3d"]
+DatasetModality = Literal["image", "audio", "text", "video", "geospatial", "3d", "tabular", "timeseries"]
 
 
 class DatasetModalitiesResponse(TypedDict):
--- a/services/worker/src/worker/job_runners/dataset/modalities.py
+++ b/services/worker/src/worker/job_runners/dataset/modalities.py
@@ -68,6 +68,8 @@ def detect_features_modalities(features: Features) -> set[DatasetModality]:
         for column_name, feature in features.items()
     ):
         modalities.add("timeseries")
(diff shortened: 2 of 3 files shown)


services/worker/src/worker/dtos.py
```diff
@@ -333,7 +333,7 @@ class DatasetCompatibleLibrariesResponse(TypedDict):
     formats: list[DatasetFormat]
 
 
-DatasetModality = Literal["image", "audio", "text", "video", "geospatial", "3d"]
+DatasetModality = Literal["image", "audio", "text", "video", "geospatial", "3d", "tabular", "timeseries"]
```
lhoestq, line 336: maybe name it "time-series" ?

severo: we use `timeseries` in moon-landing, so... better to stick with this, unless it's really an issue for you :P

severo:
thanks for the addition on timeseries @lhoestq

severo:
(feel free to add a test, it should be straightforward as we have similar cases already!)

lhoestq (approved):

zhangzw16:
Hi, thank you so much for your contribution!  @severo @lhoestq 

I’m new to the Hugging Face Dataset Viewer. I noticed that this PR adds support for handling time-series data. I’d like to ask if there is any documentation related to this feature. If the modality is detected as time-series, what happens next? How does it differ from other modalities?

Also, I want to visualize time-series-related columns in the Dataset Viewer. How can I achieve this? If it’s not currently supported, I’d like to contribute. Where should I start?

lhoestq:
The modality is useful to filter by modality in HF, e.g.

image

> Also, I want to visualize time-series-related columns in the Dataset Viewer. How can I achieve this? If it’s not currently supported, I’d like to contribute. Where should I start?

The best way to start would be with a demo app on HF Spaces (you can use Docker, gradio etc.). This is useful to let people share feedbacks and contribute :) No need for the demo to be a full Dataset Viewer but just focused on time series

For the anecdote the Dataset Viewer itself started as a demo on Spaces [here](https://huggingface.co/spaces/lhoestq/datasets-explorer)

PR #2922, opened 2024-06-18
admin-ui: Do not mark gated datasets as error

AndreaFrancis:
In the admin-ui, some datasets have all the features working (is-valid) but are shown as errors.
This is due to gated datasets; I just added another sign to identify those and let us know that the issue is unrelated to jobs not being correctly processed.
Another question is if we should consider this dataset as part of the coverage metric since we need to know the value.
Or maybe there is another way to automatically subscribe to the admin-UI user token for those datasets?


severo (approved):
Good fix! We can see in another PR how to get the data, but it's already a good step forward

PR #2926, opened 2024-06-19
only raise error in config-is-valid if format is bad

severo:
Should remove the remaining PreviousStepFormatError: https://github.com/huggingface/dataset-viewer/issues/2433#issuecomment-2179014134


severo:
OK... merging

PR #2929, opened 2024-06-20
Use current priority for children jobs

severo:
When we change the priority of a job manually after it has started, we want the children jobs to use the same priority


severo:
OK, merging

PR #2931, opened 2024-06-20
Create pastJobs collection

severo:
It will be used to apply rate limiting on the jobs, based on the total duration in a window (see https://github.com/huggingface/dataset-viewer/issues/2279#issuecomment-2178655627).


libs/libcommon/src/libcommon/queue/past_jobs.py
```diff
@@ -0,0 +1,88 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2022 The HuggingFace Authors.
```
AndreaFrancis, line 2:
```suggestion
# Copyright 2024 The HuggingFace Authors.
```

AndreaFrancis (approved):
LGTM. Just a question, should we delete all PastJobs entries for a dataset when it is being deleted? Maybe as part of https://github.com/huggingface/dataset-viewer/blob/main/libs/libcommon/src/libcommon/orchestrator.py#L748

severo:
> Just a question, should we delete all PastJobs entries for a dataset when it is being deleted

No, we have to keep it, because the budget for the dataset is still the same, even if the dataset is deleted for some time. It would be complex to remove it on deletion, additionally, because we delete the dataset before re-creating it when we update.

PR #2932, opened 2024-06-20
Fix estimate info for zip datasets

lhoestq:
I simply had to track metadata reads only once to fix the estimator.

(otherwise every file opened in a zip archive triggers an additional read of the metadata with the central directory of the zip file that prevents the estimator from converging)

ex: locally and on only 100MB of parquet conversion (prod is 5GB), it estimates 47871 examples on https://huggingface.co/datasets/datasets-maintainers/test_many_zip_with_images_MSCOCO_zip (true value is exactly 50k)

Unrelared, but I noticed that parquet conversion of zip files is super slooooooowww, we'll have to improve that at one point because it can surely take more than 40min to run. When tested in prod, an error happens before that though (JobManagerCrashedError) which suggests a termination by kubernetes.

### DIFF
--- a/services/worker/src/worker/job_runners/config/parquet_and_info.py
+++ b/services/worker/src/worker/job_runners/config/parquet_and_info.py
@@ -879,15 +879,22 @@ def get_urlpaths_in_gen_kwargs(gen_kwargs: dict[str, Any]) -> list[str]:
     shards = max(lists, key=len)
     urlpaths: set[str] = set()
     for shard in shards:
+        # Standard list of shards [data0.json, data1.json, ...]
         if isinstance(shard, str):
             urlpaths.add(shard.split("::")[-1])
-        if shard and isinstance(shard, tuple):
+        # Each item can also be an iterator
+        # (typically used in builders that support iterating on extracted zip files)
+        elif isinstance(shard, FilesIterable):
+            urlpaths.update(item.split("::")[-1] for item in shard)
+        # ImageFolder / AudioFolder list of shards
+        # Each item is a tuple like (optional original file, downloaded file)
+        elif shard and isinstance(shard, tuple):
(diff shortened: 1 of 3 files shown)


services/worker/tests/job_runners/config/test_parquet_and_info.py
```diff
@@ -1045,6 +1045,8 @@ def test_track_reads_zip_file(text_file: str, zip_file: str) -> None:
             open(text_file, "rb") as uncompressed_f,
         ):
             expected_output_size = len(uncompressed_f.read())
+            tracker.files["file://" + zip_file]["read"] > 0  # open does read metadata
```
severo, line 1048:
```suggestion
            assert tracker.files["file://" + zip_file]["read"] > 0  # open does read metadata
```

?

services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -880,6 +881,11 @@ def get_urlpaths_in_gen_kwargs(gen_kwargs: dict[str, Any]) -> list[str]:
     for shard in shards:
         if isinstance(shard, str):
             urlpaths.add(shard.split("::")[-1])
+        if shard and isinstance(shard, tuple):
+            if isinstance(shard[-1], FilesIterable):
```
severo, line 885: maybe a comment at least (or a test, at most) to give an example for each case

severo (approved):
again, very hard to review for me...

services/worker/src/worker/job_runners/config/parquet_and_info.py
lhoestq: I added comments as well as a test for every dataset builder and included normal data and zipped data

PR #2937, opened 2024-06-21
Enable estimate info (size) on all datasets

lhoestq:
TODO before merging:

- [x] merge https://github.com/huggingface/dataset-viewer/pull/2932
- [x] test json in prod (allenai/c4 en.noblocklist: correct relative error of only 0.02%)
- [x] test webdataset in prod (datasets-maintainers/small-publaynet-wds-10x: correct with relative error of only 0.07%)


severo:
we won't recompute for all the datasets, right? it will thus be applied progressively, when a dataset is updated.

lhoestq:
Yes correct. If needed we can manually recompute for some important datasets like C4

PR #2948, opened 2024-06-25
Exclude blocked datasets from Job metrics

AndreaFrancis:
Fix for https://github.com/huggingface/dataset-viewer/issues/2945


AndreaFrancis:
Wrong approach; it should be excluded from WorkerSizeJobsCountDocument instead, closing in favor of a new PR.

PR #2966, opened 2024-07-03
Use placeholder revision in urls in cached responses

lhoestq:
close https://github.com/huggingface/dataset-viewer/issues/2965

### DIFF
--- a/libs/libcommon/src/libcommon/viewer_utils/asset.py
+++ b/libs/libcommon/src/libcommon/viewer_utils/asset.py
@@ -3,15 +3,18 @@
 
 from io import BytesIO
 from tempfile import NamedTemporaryFile
-from typing import Optional, TypedDict
+from typing import TYPE_CHECKING, Optional, TypedDict
 
 from PIL import Image, ImageOps
 from pydub import AudioSegment  # type:ignore
 
-from libcommon.storage_client import StorageClient
+if TYPE_CHECKING:
+    from libcommon.storage_client import StorageClient
+
 
 SUPPORTED_AUDIO_EXTENSION_TO_MEDIA_TYPE = {".wav": "audio/wav", ".mp3": "audio/mpeg", ".opus": "audio/opus"}
 SUPPORTED_AUDIO_EXTENSIONS = SUPPORTED_AUDIO_EXTENSION_TO_MEDIA_TYPE.keys()
+DATASET_GIT_REVISION_PLACEHOLDER = "{dataset_git_revision}"
 
 
 class ImageSource(TypedDict):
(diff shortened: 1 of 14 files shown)


libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -35,25 +38,32 @@ def create_image_file(
...
+    storage_client: "StorageClient",
 ) -> ImageSource:
+    # We use a placeholder revision that will be replaced later by the
+    # dataset_git_revision of cache responses when the data will be accessed.
+    # This is useful to allow moving files to a newer revision without having
+    # to modify the cached rows content.
     object_key = storage_client.generate_object_key(
         dataset=dataset,
-        revision=revision,
+        revision=DATASET_GIT_REVISION_PLACEHOLDER,
```
lhoestq, line 49: main change is here

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -35,25 +38,32 @@ def create_image_file(
...
         buffer = BytesIO()
         image.save(fp=buffer, format=format)
         buffer.seek(0)
-        with storage_client._fs.open(storage_client.get_full_path(object_key), "wb") as f:
+        with storage_client._fs.open(storage_client.get_full_path(path), "wb") as f:
             f.write(buffer.read())
-    return ImageSource(src=storage_client.get_url(object_key), height=image.height, width=image.width)
+    return ImageSource(
+        src=storage_client.get_url(object_key, revision=revision), height=image.height, width=image.width
+    )
```
lhoestq, line 66: A few lines later, `get_url` applies the right revision if there is an URL preparator associated to the storage client.

In practive:
- /rows, /search and /filter have an URL preparator that inserts the revision and signs the urls from the parquet/duckdb data
- split-first-rows worker has no URL preparator: URLs are stored in the cached as unprepared (i.e. with the revision placeholder and unsigned)
- /first-rows has an URL preparator that inserts the revision and signs the urls that come form the cache

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -35,25 +38,32 @@ def create_image_file(
     filename: str,
     image: Image.Image,
     format: str,
-    storage_client: StorageClient,
+    storage_client: "StorageClient",
 ) -> ImageSource:
+    # We use a placeholder revision that will be replaced later by the
```
severo, line 43:
```suggestion
    # We use a placeholder revision in the JSON stored in the database,
    # while the path of the file stored on the disk/s3 contains the revision.
    # The placeholder will be replaced later by the
```

severo: or something like that, to help understand where the placeholder is used

severo (approved):

PR #2967, opened 2024-07-03
Add duration to cached steps

polinaeterna:
Will close https://github.com/huggingface/dataset-viewer/issues/2892

As suggested in https://github.com/huggingface/dataset-viewer/pull/2908#pullrequestreview-2126425919, this adds duration field to cached responses. Duration is computed using `started_at` field of a corresponding job.

Note that this PR adds new fields to major libcommon dtos `JobInfo` and `JobResult`.


libs/libcommon/tests/test_utils.py
```diff
@@ -147,3 +149,7 @@ def test_serialize_and_truncate_does_not_raise(obj: Any, max_bytes: int, expecte
 def test_serialize_and_truncate_raises(obj: Any, max_bytes: int) -> None:
     with pytest.raises(SmallerThanMaxBytesError):
         serialize_and_truncate(obj=obj, max_bytes=max_bytes)
+
+
+def test_get_duration() -> None:
```
polinaeterna, line 154: idk if that even needed...

services/worker/src/worker/job_manager.py
```diff
@@ -107,6 +107,7 @@ def run_job(self) -> JobResult:
                 "job_runner_version": self.job_runner_version,
                 "is_success": False,
                 "output": None,
+                "duration": get_duration_or_none(self.job_info["started_at"]),
```
polinaeterna, line 110: this is dumb but otherwise mypy warns that `Optional[datetime]` is being passed because job_info["started_at"] might be None if a job is not started yet.

services/worker/tests/test_job_manager.py
```diff
@@ -115,6 +117,8 @@ def test_backfill(priority: Priority, app_config: AppConfig) -> None:
     assert job_result["is_success"]
     assert job_result["output"] is not None
     assert job_result["output"]["content"] == {"key": "value"}
+    assert job_result["duration"] is not None
```
polinaeterna, line 120: this is the only test that checks that duration is set as expected for successful jobs. should I test it anywhere else?

PR #2971, opened 2024-07-04
Fix dataset name when decreasing metrics

AndreaFrancis:


libs/libcommon/src/libcommon/queue/jobs.py
```diff
@@ -368,7 +368,7 @@ def delete_waiting_jobs_by_job_id(self, job_ids: list[str]) -> int:
         try:
             existing = JobDocument.objects(pk__in=job_ids, status=Status.WAITING)
             for job in existing.all():
-                decrease_metric(dataset=job.type, job_type=job.type, status=job.status, difficulty=job.difficulty)
+                decrease_metric(dataset=job.dataset, job_type=job.type, status=job.status, difficulty=job.difficulty)
```
severo, line 371: oh! well spotted!!!

PR #2976, opened 2024-07-08
Update statistics.md: Add the length calculation statement.

Siki-cloud:
specify that the "string_text " length is calculated by its character number.


polinaeterna (approved):
awesome, good catch, thank you!

PR #2978, opened 2024-07-09
remove the 'tags' field from the Hub cache response

severo:
Note that I don't migrate to remove the field from the cache (we could do it, but not sure it's worth the pain)

Better wait for https://github.com/huggingface-internal/moon-landing/pull/10591 (internal) to be deployed on the Hub before merging.


severo:
See https://github.com/huggingface-internal/moon-landing/pull/10586#issuecomment-2216957260

lhoestq (approved):

AndreaFrancis (approved):

severo:
OK, it has been deployed in prod.

PR #2982, opened 2024-07-09
Docs: mention partial parquet conversion for big row groups

lhoestq:
asked at https://huggingface.slack.com/archives/C0709MK0PUN/p1720433431019569?thread_ts=1720210655.175099&cid=C0709MK0PUN

### DIFF
--- a/docs/source/parquet.md
+++ b/docs/source/parquet.md
@@ -199,9 +199,17 @@ To read and query the Parquet files, take a look at the [Query datasets from the
 
 ## Partially converted datasets
 
-The Parquet version can be partial if the dataset is not already in Parquet format or if it is bigger than 5GB.
+The Parquet version can be partial in two cases:
+- if the dataset is already in Parquet format but it contains row groups bigger than the recommended size (100-300MB uncompressed)
+- if the dataset is not already in Parquet format or if it is bigger than 5GB.
 
 In that case the Parquet files are generated up to 5GB and placed in a split directory prefixed with "partial", e.g. "partial-train" instead of "train".
+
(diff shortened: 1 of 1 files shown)


julien-c:
note that this should ideally be displayed directly on the hub in the cases when it happens, imo

severo (approved):

lhoestq:
> note that this should ideally be displayed directly on the hub in the cases when it happens, imo

maybe in the (i) ?

image

Something like "The table contains 100 rows per page, up to 5GB. It is loaded from converted Parquet files, since original files have row group sizes bigger than the recommended value (100-300MB uncompressed)"

severo:
I imagine that we can detect the case, if:
- the viewer is partial
- the dataset has the "parquet" file format

We might want to show the warning in the right column (where we show the size tags)

Capture d’écran 2024-07-10 à 10 47 52

text:

> ⚠️ The original Parquet files have row group sizes bigger than the recommended value (100-300MB uncompressed). The dataset viewer falls back to auto-converted Parquet files on the first 5GB of raw data.

lhoestq:
Yes you can indeed detect the case like that.

Not a big fan of the warning which can get a lot of the user's attention. Here is my suggestion for the text: 

> ⓘ The Viewer and auto-converted Parquet files contain data up to 5GB, since original files have row groups that are too big to be displayed. The recommended row group size is 100-300MB in-memory.

severo:
I created https://github.com/huggingface-internal/moon-landing/pull/10644. Not sure if it's the best way to do. Let's discuss there

PR #2987, opened 2024-07-12
fix text and arrow format

lhoestq:
fix "0 dataset in text format" reported at https://x.com/jedmaczan/status/1809280782588158452

image

I'll re-launch the job for text datasets

### DIFF
--- a/services/worker/src/worker/job_runners/dataset/compatible_libraries.py
+++ b/services/worker/src/worker/job_runners/dataset/compatible_libraries.py
@@ -634,6 +634,8 @@ get_format_for_builder: dict[str, DatasetFormat] = {
     "parquet": "parquet",
     "imagefolder": "imagefolder",
     "audiofolder": "audiofolder",
+    "text": "text",
+    "arrow": "arrow",
 }
 
 
--- a/services/worker/src/worker/dtos.py
+++ b/services/worker/src/worker/dtos.py
@@ -313,7 +313,7 @@ class IsValidResponse(TypedDict):
 
 
 DatasetLibrary = Literal["mlcroissant", "webdataset", "datasets", "pandas", "dask"]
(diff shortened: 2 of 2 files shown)


severo:
OK, we have 7 datasets with "text" format. I imagine we have to re-run over all the datasets if we want to have more...

re arrow: so, we have to add this format to moonlanding, otherwise it will be ignored.

severo:
> add this format to moonlanding

https://github.com/huggingface-internal/moon-landing/pull/10664

severo:
No, it should be immediate

lhoestq:
hmmm

The hub cache is updated for https://huggingface.co/datasets/sam-chami/asturian doesn't show the "text" format even though it's updated:

```json
{
  "dataset": "sam-chami/asturian",
  "kind": "dataset-hub-cache",
  "content": {
    "formats": [
      "text"
    ],
    ...
  },
  ...
}
```

severo:
hmmm. What if you create a job that recomputes the step?

lhoestq:
that fixed it ! 2.5k datasets in text format now

services/worker/src/worker/job_runners/dataset/compatible_libraries.py
```diff
@@ -634,6 +634,8 @@ def get_compatible_libraries_for_webdataset(
     "parquet": "parquet",
     "imagefolder": "imagefolder",
     "audiofolder": "audiofolder",
+    "text": "text",
+    "arrow": "arrow",
```
julien-c, line 638: seems a bit error-prone to have to list them here

severo: you mean: extracting this to our constants.py file, for example?

PR #2995, opened 2024-07-23
move the docs page

severo:
This change should move https://huggingface.co/docs/datasets-server to https://huggingface.co/docs/dataset-viewer.

cc @mishig25: we also need to redirect https://huggingface.co/docs/datasets-server to https://huggingface.co/docs/dataset-viewer. Where do we do it?

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -16,7 +16,7 @@ jobs:
     uses: huggingface/doc-builder/.github/workflows/build_main_documentation.yml@main
     with:
       commit_sha: ${{ github.sha }}
-      package: datasets-server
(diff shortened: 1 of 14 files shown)


mishig25:
this file need to be changed too https://github.com/huggingface/dataset-viewer/blob/main/.github/workflows/doc-pr-build.yml

mishig25:
to changes these lines as well https://github.com/huggingface-internal/moon-landing/blob/94daef6d9d6c77c70b23de6a0612dd1d93ea0978/server/views/pages/DocsPage/consts.ts#L256-L258

julien-c (approved):
lgtm

severo:
OK, let's go!

PR #2997, opened 2024-07-24
Add Polars loading code

nameexhaustion:
Hello from the Polars team! We've recently added support for scanning Hugging Face datasets, and to make it easier for users we're hoping that Polars code snippets could be added under the "Use this dataset" section on the dataset viewer webpage, next to the other libraries.

Specifically, here's where we'd like to add Polars as an option:
image

### DIFF
--- a/services/worker/src/worker/dtos.py
+++ b/services/worker/src/worker/dtos.py
@@ -312,7 +312,7 @@ class IsValidResponse(TypedDict):
     statistics: bool
 
 
-DatasetLibrary = Literal["mlcroissant", "webdataset", "datasets", "pandas", "dask"]
+DatasetLibrary = Literal["mlcroissant", "webdataset", "datasets", "pandas", "dask", "polars"]
 DatasetFormat = Literal["json", "csv", "parquet", "imagefolder", "audiofolder", "webdataset", "text", "arrow"]
 ProgrammingLanguage = Literal["python"]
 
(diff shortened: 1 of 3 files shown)


lhoestq (commented):
Cool thanks ! Excited to see Polars as an option on HF :)

When login is required, we show this message for Pandas and Dask, and I think we can show the same message for Polars no (if Polars supports reading the user token from disk) ?

```python
# Login using e.g. `huggingface-cli login` to access this dataset
```

(the token is stored in `~/.cache/huggingface/token` where `~/.cache/huggingface` is the default `HF_HOME`)

There will be a link in the UI anyway to the Polars docs on HF where we can explain other authentication, methods like `storage_options`.

If this authentication method is not implemented yet in Polars I'm happy to leave it as is :)

nameexhaustion:
> (the token is stored in ~/.cache/huggingface/token where ~/.cache/huggingface is the default HF_HOME)

Ah, I haven't seen this one before. Thanks for the explanation - I'll see if I can get some support in for this.

nameexhaustion:
Updated to show the `LOGIN_COMMENT` message

lhoestq (approved):
LGTM ! cc @severo for viz

severo:
Nice, thanks!

I opened https://github.com/huggingface/huggingface.js/pull/822. And then, I'll update the Hub's code

PR #3001, opened 2024-07-25
otherwise we have dataset_name: 'parquet'

severo:
Not sure what to with the existing entries (we don't want to recompute all the entries).

Alternative: remove the field?


lhoestq (approved):
no need to no anything imo, it's not used anywhere anyway

PR #3007, opened 2024-07-31
Stats for datetimes

polinaeterna:


polinaeterna:
I have no idea how `datetime.strptime("2024-01-01 00:00:00CET", "%Y-%m-%d %H:%M:%S%Z")` might work locally but fail on CI :sob:

services/worker/src/worker/statistics_utils.py
```diff
@@ -456,6 +477,24 @@ def is_class(n_unique: int, n_samples: int) -> bool:
             n_unique / n_samples <= MAX_PROPORTION_STRING_LABELS and n_unique <= MAX_NUM_STRING_LABELS
         ) or n_unique <= NUM_BINS
 
+    @staticmethod
+    def is_datetime(data: pl.DataFrame, column_name: str) -> tuple[bool, Optional[str]]:
+        """Check if first 1000 non-null samples in a column match datetime format. If true, also return datetime format"""
+
+        values = data.filter(pl.col(column_name).is_not_null()).head(1000)[column_name].to_list()
```
lhoestq, line 484: you can lower to first 100 to reduce the risk of OOM imo

```suggestion
        """Check if first 100 non-null samples in a column match datetime format. If true, also return datetime format"""

        values = data.filter(pl.col(column_name).is_not_null()).head(100)[column_name].to_list()
```

lhoestq (commented):
cool ! does it need any change on the front end side ? cc @severo for viz

Also this change requires implementing its associated filtering for /filter. I think DuckDB can filter after casting the strings to timestamps

severo:
Super useful. For the frontend, first: we will need the openapi spec to be updated, so that the API is clear. But I think it should work directly. One detail though: the edges are long strings, and THIS might break the design, so, surely something to add in moonlanding.

services/worker/src/worker/statistics_utils.py
polinaeterna: do you think there might be OOM with 1000 texts?

lhoestq: some datasets have 1 row = 1 full book so it can oom in rare cases like that

services/worker/src/worker/statistics_utils.py
```diff
@@ -699,3 +746,111 @@ def get_shape(example: Optional[Union[bytes, dict[str, Any]]]) -> Union[tuple[No
...
+                mean=None,
+                median=None,
+                std=None,
+                histogram=None,
+            )
+        original_timezone = None
+        if isinstance(data[column_name].dtype, pl.String):
+            # let polars identify format itself. provide manually in case of error
+            try:
+                original_timezone = get_timezone(data[column_name][0])
```
polinaeterna, line 806: I track original timezone here because polars converts datetimes to UTC when format is not provided and we need to switch it back later

lhoestq (approved):
Will do some testing in prod with @cfahlgren1, I enabled the string -> datetime stats only  for caleb and I's datasets

PR #3017, opened 2024-08-11
Re-enable Polars as a supported library for jsonl with glob paths

nameexhaustion:
Ref https://github.com/huggingface/dataset-viewer/pull/3006
Issue fixed by https://github.com/pola-rs/polars/pull/17958


services/worker/src/worker/job_runners/dataset/compatible_libraries.py
```diff
@@ -697,15 +697,13 @@ def fmt_code(
...
-            # The pattern 'datasets/[dataset]/**/*' is not supported yet
-            return None
         is_json_lines = ".jsonl" in first_file or HfFileSystem(token=hf_token).open(first_file, "r").read(1) != "["
 
         if is_json_lines:
             read_func = "read_ndjson"
         else:
-            read_func = "read_json"
+            # JSON not yet supported by Polars
+            return None
```
nameexhaustion, line 706: drive-by - removed this as polars doesn't have support for `read_json` yet

PR #3018, opened 2024-08-12
Update aiohttp 3.10.2 min version to fix vulnerability

albertvillanova:
Update aiohttp 3.10.2 min version to fix vulnerability.

Fix 12 dependabot alerts.


albertvillanova:
The error in the CI is unrelated to this PR.

See:
- #3019

PR #3020, opened 2024-08-15
doc: Read parquet files with PySpark 

AndreaFrancis:
Adding primary doc for reading parquet files with PySpark

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -46,8 +46,8 @@
           title: Polars
         - local: mlcroissant
           title: mlcroissant
-        - local: PySpark
-          title: pyspark
+        - local: pyspark
+          title: PySpark
 - title: Conceptual Guides
   sections:
     - local: configs_and_splits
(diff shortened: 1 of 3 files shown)


lhoestq (commented):
Cool ! Maybe put links to files in the repo rather than from the API which is more advanced usage ? This way people can read parquet, jsonl and more

AndreaFrancis:
> Cool ! Maybe put links to files in the repo rather than from the API which is more advanced usage ? This way people can read parquet, jsonl and more

Agree, but I wanted to follow the same pattern as the other sections (ClickHouse, cuDF, DuckDB, Pandas, and Polars), which point to refs/convert/parquet.

Additionally, it will be in the `Query datasets from Dataset Viewer API` section specifically.

lhoestq:
Ah yes sorry, this is the viewer's docs so yea it should show the viewer's parquet files

lhoestq (approved):
LGTM ! Would be cool to have spark docs in the hub docs as well !

PR #3031, opened 2024-08-21
Update to new dataset ID in example

davanstrien:
At the moment, the examples fail because rotten_tomatoes moved to https://huggingface.co/datasets/cornell-movie-review-data/rotten_tomatoes


severo (approved):
Thanks! I think we have a lot more occurrences of canonical datasets, in the docs or in the openapi file. I can take care of reviewing them

severo:
merging because I want to do the same with other datasets. Better to base on main

PR #3034, opened 2024-08-21
import from the correct file

severo:
see
https://github.com/huggingface/huggingface_hub/pull/2474#issuecomment-2302132315


Wauplin (approved):
Thanks!

PR #3036, opened 2024-08-22
More explicit test and comments about offset-naive datetimes read from mongo

### LINKED ISSUE #862: Ensure the dates stored in mongo (jobs, cache) are localized
See https://github.com/huggingface/datasets-server/pull/850#discussion_r1118518118.

severo:
I think it's enough to close #862


lhoestq (approved):
sounds good

PR #3037, opened 2024-08-22
Update datasets to 2.21.0

albertvillanova:
Update datasets to 2.21.0.

Fix #3024.

PR #3040, opened 2024-08-22
Update datasets to 2.20.0

albertvillanova:
Update datasets to 2.20.0.

This PR is intended to address the CI errors raised by this update, as a first step before:
- #3037

Fixes after the update of `datasets`:
- Pass `trust_remote_code=True` for script dataset: 227f2c4
- Use JSON-Lines (instead of JSON) dataset in `test_statistics_endpoint` to avoid pandas bug that downcasts float to int column: a86d040

### DIFF
--- a/e2e/tests/conftest.py
+++ b/e2e/tests/conftest.py
@@ -38,6 +38,15 @@ def json_path(tmp_path_factory: TempPathFactory) -> str:
 
 
 @pytest.fixture(scope="session")
+def jsonl_path(tmp_path_factory: TempPathFactory) -> str:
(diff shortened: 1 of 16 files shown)


albertvillanova:
Note I have commented an assertion is test_statistics_endpoint: [c437c9f](https://github.com/huggingface/dataset-viewer/pull/3040/commits/c437c9f09432be2949c4fd18e85c46b66595cea1)
- pandas downcasts a float column to int if all the numbers can be downcast to integers, e.g. [1.0, 2.0, 3.0]
- I don't think this is a big issue

severo (approved):

Issue #3010, opened 2024-08-01
Replace ConfigNamesError with the underlying error


severo:
100K cache entries (1%) have the `ConfigNamesError`. It would be better to show the underlying error, and help the user debug their data files.

severo:
Current state:

```
db.cachedResponsesBlue.aggregate([
  {
    $match: {
      error_code: "ConfigNamesError",
      "details.copied_from_artifact": { $exists: false },
    },
  },
  {
    $group: {
      _id: "$kind",
      count: {
        $sum: 1,
      },
    },
  },
  {
    $sort: {
      count: -1,
    },
  },
]);
{ _id: 'dataset-config-names', count: 10608 }
db.cachedResponsesBlue.aggregate([
  {
    $match: {
      kind: "dataset-config-names",
      error_code: "ConfigNamesError",
      "details.copied_from_artifact": { $exists: false },
    },
  },
  {
    $group: {
      _id: "$details.cause_exception",
      count: {
        $sum: 1,
      },
    },
  },
  {
    $sort: {
      count: -1,
    },
  },
]);
{ _id: 'FileNotFoundError', count: 4902 }
{ _id: 'DataFilesNotFoundError', count: 4575 }
{ _id: 'ValueError', count: 561 }
{ _id: 'DatasetWithScriptNotSupportedError', count: 228 }
{ _id: 'BadZipFile', count: 113 }
{ _id: 'TypeError', count: 41 }
{ _id: 'HfHubHTTPError', count: 30 }
{ _id: 'BrokenPipeError', count: 27 }
{ _id: 'DatasetNotFoundError', count: 23 }
{ _id: 'AttributeError', count: 22 }
{ _id: 'PermissionError', count: 12 }
{ _id: 'GatedRepoError', count: 12 }
{ _id: 'RuntimeError', count: 12 }
{ _id: 'InvalidConfigName', count: 11 }
{ _id: 'ConnectionError', count: 9 }
{ _id: 'KeyError', count: 8 }
{ _id: 'ReadTimeout', count: 4 }
{ _id: 'SyntaxError', count: 3 }
{ _id: 'UnicodeDecodeError', count: 3 }
{ _id: 'IndexError', count: 2 }
{ _id: 'JSONDecodeError', count: 2 }
{ _id: 'ScannerError', count: 2 }
{ _id: 'HTTPError', count: 1 }
{ _id: 'IsADirectoryError', count: 1 }
[...]

PR #3041, opened 2024-08-22
Raise specific errors instead of ConfigNamesError when appropriate

severo:
It will help show them better in moonlanding

see #3010

### DIFF
--- a/libs/libcommon/src/libcommon/constants.py
+++ b/libs/libcommon/src/libcommon/constants.py
@@ -43,6 +43,7 @@ DEFAULT_MAX_FAILED_RUNS = 3
 LARGE_MAX_FAILED_RUNS = 30  # for errors that should not be permanent
 MAX_FAILED_RUNS_PER_ERROR_CODE = {
     # default
+    "ConfigNamesError": DEFAULT_MAX_FAILED_RUNS,  # <- 20240822: to recompute all of them in the next backfill
     "RetryableConfigNamesError": DEFAULT_MAX_FAILED_RUNS,
     "ConnectionError": DEFAULT_MAX_FAILED_RUNS,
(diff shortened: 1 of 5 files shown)


libs/libcommon/src/libcommon/exceptions.py
```diff
@@ -99,6 +100,7 @@ def as_response(self) -> ErrorResponse:
     "ExternalServerError",
     "FeaturesError",
     "FeaturesResponseEmptyError",
+    "FileFormatMismatchBetweenSplits",
```
albertvillanova, line 103: I would suggest adding the word "Error" at the end, by convention.

albertvillanova (approved):
Thanks.

libs/libcommon/src/libcommon/exceptions.py
severo: sure!

severo: done

libs/libcommon/src/libcommon/constants.py
```diff
@@ -43,6 +43,8 @@
 LARGE_MAX_FAILED_RUNS = 30  # for errors that should not be permanent
 MAX_FAILED_RUNS_PER_ERROR_CODE = {
     # default
+    "ConfigNamesError": DEFAULT_MAX_FAILED_RUNS,  # <- 20240822: to recompute all of them in the next backfill
```
lhoestq, line 46: so you should remove this after the next backfill ?

lhoestq (approved):
cool !

libs/libcommon/src/libcommon/constants.py
severo: yes. Anyway, it can wait for some days, it will retry 3 times at most, and we don't have that many cases.

PR #3051, opened 2024-08-28
Fix typo "build > built"

david4096:


severo (approved):
Neat! Thanks for the fix

PR #3059, opened 2024-09-09
Add relaion2b dataset opt-out results

lhoestq:
previously we had the laion2b results, but laion re-released the dataset with an improved filtering based on safety


julien-c (approved):
nice!

you can update the Spawning peeps too

PR #3067, opened 2024-09-23
fine grained token for admin

lhoestq:
require fine grained token with repo.write permission to access the admin endpoints

### DIFF
--- a/services/admin/README.md
+++ b/services/admin/README.md
@@ -10,7 +10,7 @@ The service can be configured using environment variables. They are grouped by s
 
 Set environment variables to configure the application (`ADMIN_` prefix):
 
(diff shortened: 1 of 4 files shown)


coyotte508:
This isn't very future-proof (digging into the internals of the tokens definitions)

but it's ok, at worst admin is temporarily disabled on dataset-servers.

A more future-proof way would be to call `/api/:repoType(models|spaces|datasets)/datasets-maintainers/:repo/auth-check/write` 

And implement the endpoint on moon's side, one line in `server.ts`:

```
app.get("/api/:repoType(models|spaces|datasets)/:namespace?/:repo/auth-check/write", middlewares.authOptional(REPO_AUTH), middlewares.repoGuard([PermissionName.WriteRepoContent]), (req, res) => res.sendStatus(200));
```

But if doing a http call everytime is not practical, I understand

lhoestq:
It's easy to fix in the future if needed anyway.
And it's not a critical endpoint, it's only for manual endpoint

PR #3085, opened 2024-10-17
fix(infra): new token name

rtrompier:


severo:
I don't see the token in this repo secrets. I imagine we have to wait for it before merging, right?

rtrompier:
Removed and added again seem to fix the issue. Probably a cache issue on Github side ^^

severo (approved):

PR #3089, opened 2024-10-25
Add video support

lhoestq:


severo:
I remember you mentioned at one point that some video dataset have one video, and each row refers to a range [start, end], which would be accessed as "media fragments" in HTML: https://developer.mozilla.org/en-US/docs/Web/Media/Audio_and_video_delivery#specifying_playback_range. Is it still a thing, or do we just have one row = one video?

lhoestq:
one row = one video (we'll see later if we need to support other cases)

PR #3109, opened 2024-12-06
AudioCell - set audio type to string

cfahlgren1:
We are adding support for more generic types for the SQL Console in the AudioCell [Internal PR](https://github.com/huggingface-internal/moon-landing/pull/11843#discussion_r1871635683).

### DIFF
--- a/docs/source/openapi.json
+++ b/docs/source/openapi.json
@@ -773,7 +773,7 @@
             "type": "integer"
           }
         },
-        "required": ["src", "height", "width"]
+        "required": ["src"]
       },
       "NullableImagesListCell": {
         "type": "array",
--- a/docs/source/openapi.json
+++ b/docs/source/openapi.json
@@ -773,7 +773,7 @@
             "type": "integer"
           }
         },
-        "required": ["src"]
+        "required": ["src", "height", "width"]
       },
       "NullableImagesListCell": {
         "type": "array",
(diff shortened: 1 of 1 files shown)


cfahlgren1:
@severo went ahead and made the width and height optional while we are at it

docs/source/openapi.json
```diff
@@ -774,7 +773,7 @@
             "type": "integer"
           }
         },
-        "required": ["src", "height", "width"]
+        "required": ["src"]
```
severo, line 776: Hmmm, note that it comes from

https://github.com/huggingface/dataset-viewer/blob/539b43775760f68f48d6df252ebef6e14c94edb6/libs/libcommon/src/libcommon/viewer_utils/asset.py#L20

It means that data coming from the API will always contain height and width. If you want to set them as optional on the Hub, maybe do so outside of the openapi.d.ts file.

cfahlgren1: ahhh, makes sense. will revert this.

cfahlgren1: wouldn't this be the same for audio type? 

We would expect that the API would only return this?

```python
              "enum": ["audio/wav", "audio/mpeg", "audio/ogg"]
```

and we need to make new type on moonlanding?

severo: no, I checked in the code, it can be any extension after all

severo (approved):
Ok! Better now by relaxing the constraint, even if the API will always return height and width for now

PR #3115, opened 2024-12-08
don't stop when afterjobplan is running

lhoestq:


services/worker/src/worker/executor.py
```diff
@@ -149,8 +149,9 @@ def heartbeat(self) -> None:
             try:
                 Queue().heartbeat(job_id=job_id)
             except Exception as error:
-                logging.warning(f"Heartbeat failed for job {job_id}: {error}")
-                self.stop()
+                logging.warning(f"Heartbeat failed for job {job_id} (AfterJobPlan might be running): {error}")
+                # Don't stop since the AfterJobPlan may be running
```
severo, line 153: Should we keep sending heart beats during the "AfterJobPlan", and then uncomment the next line? Otherwise, it could be killed as a zombie anyway, right?

lhoestq: the job is marked as finished before the afterjobplan starts :) so no risk of having a zombie

severo: OK, thanks for clarifying

PR #3123, opened 2024-12-19
doc: Update inaccessible datasets

AndreaFrancis:
Replacing `barilan/blog_authorship_corpus` with `tasksource/blog_authorship_corpus` since the former is no longer accessible due to being a script-based dataset, and replacing `ETDataset/ett` with `GEM/SciDuet` for the same reason. Additionally, making minor fixes.


lhoestq (approved):
nice !

PR #3126, opened 2025-01-07
chore: update markdown formatting inside <Tip>

ruijdacd:
Markdown formatting isn't working well inside the `<Tip>` element.

This might be affecting other pages as well, but this is the one I came across.

### DIFF
--- a/docs/source/postgresql.md
+++ b/docs/source/postgresql.md
@@ -46,7 +46,9 @@ select * from squad limit 10;
 ```
 
 <Tip>
+  
 Full documentation for the `ai.load_dataset` function can be found [here](https://github.com/timescale/pgai/blob/main/docs/load_dataset_from_huggingface.md).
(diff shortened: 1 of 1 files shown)


lhoestq:
it doesn't seem to work well either with your change https://moon-ci-docs.huggingface.co/docs/dataset-viewer/pr_3126/en/postgresql

Maybe simply remove the <Tip> and keep it as plain text

ruijdacd:
That's a shame, I'll make that change now

lhoestq (approved):
thanks !

ruijdacd:
I've just thought that we could've just replaced the markdown elements with HTML and the formatting would be correct inside of the Tip component.

Would you like me to submit that change?

lhoestq:
I haven't tried that, not sure if it works. No strong opinion though

PR #3138, opened 2025-02-24
Support large list

lhoestq:
to avoid RowsPostProcessingError from `TypeError: could not determine the type of the data cell` errors in open R1 project


libs/libcommon/src/libcommon/viewer_utils/features.py
```diff
@@ -342,6 +343,25 @@ def get_cell_value(
             )
             for (idx, subCell) in enumerate(cell)
         ]
+    elif isinstance(fieldType, LargeList):
```
lhoestq, line 346: main change is here

PR #3144, opened 2025-03-03
When given, use the sequence len for array_shape

ccl-core:
I see that some datasets have the Sequence length defined, e.g.: https://datasets-server.huggingface.co/info?dataset=pollen-robotics/apple_storage

We should probably use this info.


libs/libcommon/src/libcommon/croissant_utils.py
```diff
@@ -126,16 +126,24 @@ def feature_to_croissant_field(
...
         }
     elif isinstance(feature, (LargeList, list, Sequence)):
+        array_shape = []
         if isinstance(feature, list):
             if len(feature) != 1:
                 return None
             sub_feature = feature[0]
+            array_shape.append("-1")
         else:
+            if shape := feature.length:
```
marcenacp, line 136: nit: For readability, collect `int | None` instead of str, and clean afterwards when building the actual arrayShape.

```python
array_shape.append(-1)
...
array_shape.append(feature.length)
...
array_shape.append(sub_feature.length)
...

str_array_shape = [str(shape) if shape else '-1' for shape in array_shape]
field["arrayShape"] = ",".join(str_array_shape)
```

marcenacp (approved):

libs/libcommon/src/libcommon/croissant_utils.py
ccl-core: Good point, done.

ccl-core:
@lhoestq gentle ping :)

lhoestq (approved):

PR #3150, opened 2025-03-18
Add function to convert a feature name to a valid jsonpath.

ccl-core:
This is used to set the correct Transform operations.

This was needed in particular for this dataset: https://huggingface.co/datasets/allenai/WildChat-1M as some of its feature names contain slashes. This creates a problem when the slashes are included non-escaped as part of the json_path attribute of a Transform operation.

An example of the corrected croissant def, as it will be now generated by HF, is here: https://github.com/mlcommons/croissant/pull/829

### DIFF
--- a/libs/libcommon/tests/test_croissant_utils.py
+++ b/libs/libcommon/tests/test_croissant_utils.py
@@ -42,6 +42,7 @@ def test_truncate_features_from_croissant_crumbs_response(num_columns: int) -> N
         ("feature'with'quote", r"['feature\'with\'quote']"),
         ("feature[with]brackets", r"['feature\[with\]brackets']"),
         ("feature[with/slash]'and'quote", r"['feature\[with/slash\]\'and\'quote']"),
+        (r"feature\'already\'escaped", r"['feature\'already\'escaped']"),
     ],
 )
 def test_escape_jsonpath_key(feature_name, expected_output):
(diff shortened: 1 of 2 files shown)


libs/libcommon/tests/test_croissant_utils.py
```diff
@@ -30,6 +34,21 @@ def test_truncate_features_from_croissant_crumbs_response(num_columns: int) -> N
...
 
 
+@pytest.mark.parametrize(
+    "feature_name, expected_output",
+    [
+        ("simple_feature", "simple_feature"),
+        ("feature/with/slash", "['feature/with/slash']"),
+        ("feature'with'quote", r"['feature\'with\'quote']"),
+        ("feature[with]brackets", r"['feature\[with\]brackets']"),
+        ("feature[with/slash]'and'quote", r"['feature\[with/slash\]\'and\'quote']"),
```
marcenacp, line 44: Can you add a test case when the feature_name already contains an escaped character? (To see how it reacts)

marcenacp (approved):

libs/libcommon/tests/test_croissant_utils.py
ccl-core: Done, thanks!

marcenacp (approved):

lhoestq (approved):
lgtm !

PR #3159, opened 2025-03-31
Delete ref branches

lhoestq:
I added two tasks that run when we call `remove_dataset()`: one to <s>empty</s> delete the refs/convert/parquet branch and one to empty the refs/convert/duckdb branch (and super_squash_commit)

Maybe there is a simpler way ? Like calling and endpoint that deletes the refs branches completely instead of emptying them ? EDIT: ended up using `HfApi.delete_branch()` :)

TODO:
- [x] add the COMMITTER_HF_TOKEN secret in the charts

cc @severo 

close https://github.com/huggingface/dataset-viewer/issues/2384


severo:
re:

> Maybe there is a simpler way ? Like calling and endpoint that deletes the refs branches completely instead of emptying them ?

I guess we can call this, no?

https://huggingface.co/docs/huggingface_hub/v0.30.0.rc3/en/package_reference/hf_api#huggingface_hub.HfApi.delete_branch

lhoestq:
Good point ! I just tried in prod and deleting the branch works fine with the committer_hf_token :) I'll do the change

lhoestq:
will deploy later today

lhoestq:
it works well, though it has the side effect to delete the branches every time there some data files are updated since `delete_dataset` is called at that time. It's ok imo

PR #3160, opened 2025-04-01
Use new cr:UInt* data types.

ccl-core:
We recently introduced support for cr:UInt* data types https://github.com/mlcommons/croissant/pull/836


ccl-core:
@lhoestq hi Quentin, can you PTAL (not urgent)? Thanks :)

PR #3166, opened 2025-04-03
Don't rely on `refs/convert/duckdb` anymore

lhoestq:
Create the duckdb index on the fly when /filter or /search is accessed and have it as usual in the duckdb indexes directory cache.

This can enable the removal of the `refs/convert/duckdb` branch entirely.

For now I'll only enable it for datasets with names that contain "NoDuckdbRef" for testing, and explore the performance, and expand progressively.

Note: the on-the-fly indexing is currently is exact same as the one done in the `split-duckdb-index` job: it ingests up to 5GB of Parquet files, applies transforms for extra indexing (e.g. extract audio length, image size), and computes the full text search index.

PS: I had to bump polars and fix the `.hist()` usage which got some breaking changes

TODO:
- [x] /filter
- [x] /search


severo:
good idea

PR #3178, opened 2025-04-22
Parallelize tokenization to enable fast on-the-fly indexing

lhoestq:
it was taking way too much time to index datasets for full text search, and it's pretty important now that we want to get rid of `refs/convert/duckdb` and index only when necessary

follows #3166 

todo:
- [x] revert debugging line

### DIFF
--- a/services/admin/src/admin/authentication.py
+++ b/services/admin/src/admin/authentication.py
@@ -16,6 +16,7 @@ async def auth_check(
     organization: Optional[str] = None,
     hf_timeout_seconds: Optional[float] = None,
     require_fine_grained_permissions: Sequence[str] = ("repo.write",),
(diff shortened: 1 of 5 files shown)


services/admin/src/admin/authentication.py
```diff
@@ -46,11 +49,27 @@ async def auth_check(
...
+                        "role" in json["auth"]["accessToken"]
+                        and any(
+                            json["auth"]["accessToken"]["role"] == require_token_and_org_permission[0]
+                            and any(
+                                org["name"] == organization and org["roleInOrg"] == require_token_and_org_permission[1]
+                                for org in json["orgs"]
+                            )
+                            for require_token_and_org_permission in require_token_and_org_permissions
+                        )
+                    )
```
lhoestq, line 72: this simply allows for write tokens, which is stronger than fine-grained. write tokens are used in the ci so this fixes ci debugging

PR #3180, opened 2025-04-25
Update e2e.yml

Love241982:
@pytest.fixture
def setup_dataset(client):
    dataset_id = "tmp-dataset-123"
    response = client.post(f"/datasets/{dataset_id}", json={"config": "default"})
    assert response.status_code == 200, "Failed to create dataset"
    yield dataset_id
    client.delete(f"/datasets/{dataset_id}")
def test_enterprise_org_private(setup_dataset, client):
    dataset_id = setup_dataset
    response = client.get(f"/parquet?dataset={dataset_id}&config=default")
    assert response.status_code == 200, f"Dataset not found: {response.json()}"

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,23 +1,36 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 The HuggingFace Authors.
 
-name: e2e
-on:
-  workflow_dispatch:
-  push:
-    branches:
-      - main
-    paths:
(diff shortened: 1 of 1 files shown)


Love241982 (commented):
def test_enterprise_org_private(setup_dataset, client):
    dataset_id = setup_dataset
    response = client.get(f"/parquet?dataset={dataset_id}&config=default")
    assert response.status_code == 200, f"Dataset not found: {response.json()}"

PR #3193, opened 2025-05-13
feat: Thumbnail and PDF for new column type

AndreaFrancis:
Part of https://github.com/huggingface/dataset-viewer/issues/2991
Given that https://github.com/huggingface/datasets/pull/7325 has been implemented, we can now process PDF columns in dataset-viewer.
TODOs:
- [X] Implement PDF processing in [get_cell_value](https://github.com/huggingface/dataset-viewer/blob/main/libs/libcommon/src/libcommon/viewer_utils/features.py#L273) for /first-rows and /rows
- [x] Add the "document" modality using the same logic as image/audio
- [x] Fix dependencies (poetry.lock)
- [x] Add unit tests in libcommon
- [x] Add unit tests for modalities in worker
- [x] Add e2e for new type

Note:
Doc will go in an upcoming PR once this one is deployed and confirm it works as expected and ready to share with the community.


AndreaFrancis:
Hi @lhoestq @severo,
This is a very early draft of how the PDF implementation could work.

I'd like to confirm a few details regarding UI integration. For that purpose, I’ve drafted a new class:

```python
class PDFSource(TypedDict):
    src: str  # path to the PDF file
    thumbnail: str  # path to a preview image of the PDF (first page) in PNG format
```

The idea is that `src` would point to the actual PDF file, while `thumbnail` would be used to display a preview in the UI.

Do you think this structure would work on the UI side?

severo:
Also, I wonder if we could return the size of the PDF in bytes? It might be a useful information to show in the dataset viewer

libs/libcommon/src/libcommon/viewer_utils/features.py
```diff
@@ -270,6 +277,65 @@ def get_video_file_extension(value: Any) -> str:
...
+        pdf_object = value
+
+    if not isinstance(pdf_object, pdfplumber.pdf.PDF):
+        raise TypeError(
+            "PDF cell must be a pdfplumber.pdf.PDF object or an encoded dict of a PDF, "
+            f"but got {str(value)[:300]}{'...' if len(str(value)) > 300 else ''}"
+        )
+
+    # Default thumbnail widht and height will result in a 150 PPI image
+    thumbnail_image = pdf_object.pages[0].to_image(width=DEFAULT_THUMBNAIL_WIDTH, height=DEFAULT_THUMBNAIL_HEIGHT)
```
lhoestq, line 320: does it respect the page image ratio ?

lhoestq (commented):
nice ! yes +1 on thumbnail width and height and it should be all good

libs/libcommon/src/libcommon/viewer_utils/features.py
AndreaFrancis: Actually no, I will fix it

lhoestq (commented):
Nice ! Looks mostly good to me :)

Feel free to also edit `transform_rows()` in libapi to enable multithreading when preparing the rows containing PDFs (right now multithreading is only enabled for audio and image) and we can merge !

I also added a comment on filtering

libs/libcommon/src/libcommon/url_preparator.py
```diff
@@ -113,10 +115,12 @@ def _prepare_asset_url_path_in_place(self, cell: Any, asset_url_path: AssetUrlPa
         elif len(asset_url_path.path) == 0:
             if not isinstance(cell, dict):
                 raise InvalidFirstRowsError("Expected the cell to be a dict")
-            src = cell.get("src")
-            if not isinstance(src, str):
-                raise InvalidFirstRowsError('Expected cell["src"] to be a string')
-            cell["src"] = self.prepare_url(src, revision=revision)
+            for key in cell.keys():
+                if "src" in key:
```
severo, line 119: OK, it's a way to get both `"str"` and `"thumbnail_src"` transformed. I think it's OK, but maybe add a comment to make it clear for future modifications or additions of fields: the ones that contain the substring `"src"` will be modified.

libs/libcommon/src/libcommon/viewer_utils/asset.py
```diff
@@ -32,6 +33,14 @@ class VideoSource(TypedDict):
     src: str
 
 
+class PDFSource(TypedDict):
+    src: str
+    size_bytes: int
+    thumbnail_src: str
```
severo, line 39: an alternative would be `thumbnail: ImageSource`. It adds one level of nesting, but reuses a type and avoids prefixes... Not sure

severo (approved):
Cooool!

I saw that you want to update the doc in another PR, but I think that the OpenAPI spec should be updated here, to reflect the current state.

As a follow-up: should we support lists of PDFs files as we do for images and audio?

libs/libcommon/pyproject.toml
```diff
@@ -9,7 +9,7 @@ license = "Apache-2.0"
 python = "3.9.18"
 appdirs = "^1.4.4"
 cryptography = "^43.0.1"
-datasets = {version = "3.6.0", extras = ["audio", "vision"]}
+datasets = {git = "https://github.com/huggingface/datasets.git", rev = "da1db8a5b89fc0badaa0f571b36e122e52ae8c61", extras = ["audio", "vision"]}
```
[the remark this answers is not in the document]
lhoestq: yup, but it will be released soon (tm)

e2e/tests/test_53_filter.py
```diff
@@ -162,3 +162,34 @@ def test_filter_audios_endpoint(normal_user_audios_public_dataset: str) -> None:
     # ensure the URL is valid
     response = poll(url, url="")
     assert response.status_code == 200, response
+
+
+def test_filter_pdfs_endpoint(normal_user_pdfs_public_dataset: str) -> None:
+    dataset = normal_user_pdfs_public_dataset
+    config, split = get_default_config_split()
+    where = "size_bytes=1319"
```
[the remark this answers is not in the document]
AndreaFrancis: I changed the column name to avoid confusion

libs/libcommon/src/libcommon/url_preparator.py
AndreaFrancis: I changed it to use ImageSource instead

libs/libcommon/src/libcommon/viewer_utils/asset.py
AndreaFrancis: Good idea! I changed it to `ImageSource` instead

lhoestq (approved):
LGTM !

> With these changes, is it safe to proceed with merging, or could this risk breaking something in the Hub?

let's see :p

AndreaFrancis:
@cfahlgren1 Not sure if any other configuration is needed from moonlanding/dataset-viewer infra as it already consumes the sources from https://datasets-server.huggingface.co/assets/* same as for images and audios. Maybe @severo?

Issue #3063, opened 2024-09-18
Simplify test code where a dataset is set as gated


severo:
[huggingface_hub@0.25.0](https://github.com/huggingface/huggingface_hub/releases/tag/v0.25.0) provides an API to set a repository as gated.

We had included a custom version of `update_repo_settings` because it lacked a `gated` parameter. Now we can switch back to the `huggingface_hub` method

https://github.com/huggingface/dataset-viewer/blob/4859100ef282dcf73257dfb60e6b5a20d5955c68/jobs/cache_maintenance/tests/utils.py#L41
https://github.com/huggingface/dataset-viewer/blob/4859100ef282dcf73257dfb60e6b5a20d5955c68/services/admin/tests/fixtures/hub.py#L24
https://github.com/huggingface/dataset-viewer/blob/4859100ef282dcf73257dfb60e6b5a20d5955c68/services/worker/tests/fixtures/hub.py#L35

severo:
depends on https://github.com/huggingface/dataset-viewer/issues/3064

severo:
good idea

ArjunJagdale:
Hi! any update about this issue? i have done a PR

PR #3206, opened 2025-07-03
refactor(tests): use HfApi.update_repo_settings to simplify gated dataset test setup

ArjunJagdale:
Now that `huggingface_hub>=0.25.0` supports the `gated` parameter in `HfApi.update_repo_settings`, this PR removes the custom implementations of `update_repo_settings()` from internal test utilities.

### Changes made:
- Removed redundant `update_repo_settings()` function from:
  - `jobs/cache_maintenance/tests/utils.py`
  - `services/admin/tests/fixtures/hub.py`
  - `services/worker/tests/fixtures/hub.py`
- Replaced each usage with the official `HfApi.update_repo_settings(...)` method
- Cleaned up related unused imports (`hf_raise_for_status`, `REPO_TYPES`, `REPO_TYPES_URL_PREFIXES`)

This is a pure cleanup refactor — it introduces no functional or behavioral changes.

Closes: #3063

### DIFF
--- a/jobs/cache_maintenance/tests/utils.py
+++ b/jobs/cache_maintenance/tests/utils.py
@@ -16,7 +16,6 @@ from huggingface_hub.constants import (
     REPO_TYPES_URL_PREFIXES,
 )
 from huggingface_hub.hf_api import HfApi
-from huggingface_hub.utils import hf_raise_for_status
 from libcommon.resources import Resource
 
 from .constants import (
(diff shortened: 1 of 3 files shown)


ArjunJagdale:
Hi! One test is currently failing:
```
FAILED tests/job_runners/config/test_parquet_and_info.py::test_parquet_file_path_in_repo[config-split-0-100001-False-None]
→ Failed: DID NOT RAISE <class 'ValueError'>
```

It seems the test expects a ValueError, but the current logic no longer raises one for this case.

Would it be better to reinstate the ValueError if it's still the correct behavior?

severo (approved):
Great contribution.

Image

This is the kind of change we love :)

PR #3207, opened 2025-07-03
Upgrade to `datasets` 4 (for torchcodec + List)

lhoestq:


services/worker/src/worker/job_runners/split/descriptive_statistics.py
```diff
@@ -214,7 +212,7 @@ def _column_from_feature(
         dataset_feature_name: str, dataset_feature: Union[dict[str, Any], list[Any]]
     ) -> Optional[SupportedColumns]:
         if isinstance(dataset_feature, list) or (
-            isinstance(dataset_feature, dict) and dataset_feature.get("_type") in ("LargeList", "Sequence")
+            isinstance(dataset_feature, dict) and dataset_feature.get("_type") in ("LargeList", "List", "Sequence")
```
severo, line 215: why do we keep testing "Sequence" here?

libs/libcommon/src/libcommon/duckdb_utils.py
```diff
@@ -168,7 +168,7 @@ def compute_transformed_data(parquet_paths: list[Path], features: dict[str, Any]
     transformed_df = None
     for feature_name, feature in features.items():
         if isinstance(feature, list) or (
-            isinstance(feature, dict) and feature.get("_type") in ("LargeList", "Sequence")
+            isinstance(feature, dict) and feature.get("_type") in ("LargeList", "List", "Sequence")
```
severo, line 171: why do we keep testing "Sequence" here?

severo (approved):

services/worker/src/worker/job_runners/split/descriptive_statistics.py
lhoestq: because of existing datasets: in the dataset-viewer cache they are Sequence

libs/libcommon/src/libcommon/duckdb_utils.py
lhoestq: (same)

PR #3214, opened 2025-07-09
Backward compat list

lhoestq:
`datasets<4` uses /info to get the exported dataset info, but it doesn't have the List type. this adds backward compatibility for `datasets<4`


services/worker/src/worker/job_runners/config/parquet_and_info.py
```diff
@@ -1257,6 +1257,23 @@ def commit_parquet_conversion(
         raise e
 
 
+def backward_compat_features(
+    features_dict: Union[dict[str, Any], str, list[Any]],
+) -> Union[dict[str, Any], str, list[Any]]:
+    """`datasets<4` use exported dataset_info and doesn't have the List type"""
```
severo, line 1263: datasets>=4 don't expect List in this endpoint?

lhoestq: it has backward compatibility :)

Issue #1522, opened 2023-07-17
Use `CONSTANT_LIST.copy` in list config fieds


severo:
See https://github.com/huggingface/datasets-server/pull/1508#discussion_r1265658458

In particular `get_empty_str_list` should not be used anymore. Same with `default_factory=list`

ArjunJagdale:
Can i work on this issue?

ArjunJagdale:
HI @severo! you can check PR #3219

PR #3219, opened 2025-07-10
refactor(config): replace get_empty_str_list with CONSTANT.copy in ParquetAndInfoConfig (#1522)

ArjunJagdale:
This PR addresses part of issue #1522 by removing the use of the helper function `get_empty_str_list` in `services/worker/src/worker/config.py`.

### ✅ Changes made:
- Removed the unused `get_empty_str_list()` helper.
- Introduced the constant `PARQUET_AND_INFO_FULLY_CONVERTED_DATASETS: list[str] = []`.
- Updated `ParquetAndInfoConfig` to use `default_factory=PARQUET_AND_INFO_FULLY_CONVERTED_DATASETS.copy`, following the pattern established in PR #1508.

This avoids shared mutable default values in dataclass fields and makes the configuration behavior more explicit.

---
cc @severo

### DIFF
--- a/services/worker/src/worker/config.py
+++ b/services/worker/src/worker/config.py
@@ -53,8 +53,8 @@ WORKER_SLEEP_SECONDS = 15
 WORKER_STATE_FILE_PATH = None
 
 
-def get_empty_str_list() -> list[str]:
-    return []
+# def get_empty_str_list() -> list[str]:
+#     return []
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


ArjunJagdale:
This PR only covers `config.py`.  I will be looking to refactor similar patterns in other config files!

services/worker/src/worker/config.py
```diff
@@ -53,8 +53,8 @@ def from_env(cls) -> "UvicornConfig":
 WORKER_STATE_FILE_PATH = None
 
 
-def get_empty_str_list() -> list[str]:
-    return []
+# def get_empty_str_list() -> list[str]:
+#     return []
```
severo, line 57: Let's delete it

```suggestion
```

severo (changes requested):

ArjunJagdale:
Done!

severo:
OK, I tried locally, and the file (orchestrator.py) is reformatted when I launch `make style` in the libs/libcommon directory.

Possibly the issue is the version of poetry you have locally? You might want to have a look at https://github.com/huggingface/dataset-viewer/blob/main/DEVELOPER_GUIDE.md to ensure your environment is set up as expected.

ArjunJagdale:
yeah i tried but the project requires python 3.9.18 i guess, and that is not available as `exe` even on official website. Editing toml will not be good i think. But I will try using `ruff format` in the `libs/libcommon`, that should follow the `pyproject.toml`.

ArjunJagdale:
Maybe this time it'll work

ArjunJagdale:
Kudos, Thanks!

PR #3220, opened 2025-07-11
Escape subfields.

ccl-core:
Some croissants for HF datasets have spaces in their IDs, e.g. https://huggingface.co/api/datasets/microsoft/WildFeedback/croissant with

```
{
              "@type": "cr:Field",
              "@id": "user_preference/history/turn id",
              "dataType": "sc:Text",
              "source": {
                "fileSet": {
                  "@id": "parquet-files-for-config-user_preference"
                },
                "extract": {
                  "column": "history"
                },
                "transform": {
                  "jsonPath": "turn id"
                }
              }
            }
```

This is invalid and will triggers a [ValidationError](https://github.com/mlcommons/croissant/blob/f3b751a8d3720ac09cb53ec33e6baa0fd97ddf6c/python/mlcroissant/mlcroissant/_src/core/json_ld.py#L199).

The `jsonpath` value also needs to be updated to make sure we always obtain a valid jsonpath.

Correct output should look like:

```
{
              "@type": "cr:Field",
              "@id": "user_preference/history/turn_id",  <-- No spaces.
              "dataType": "sc:Text",
              "source": {
                "fileSet": {
                  "@id": "parquet-files-for-config-user_preference"
                },
                "extract": {
                  "column": "history"
                },
                "transform": {
                  "jsonPath": "['turn id']"  <-- Valid jsonpath.
                }
              }
            }
```

This PR also adds two tests.

### DIFF
--- a/jobs/cache_maintenance/tests/utils.py
+++ b/jobs/cache_maintenance/tests/utils.py
@@ -5,17 +5,14 @@ import time
 from contextlib import suppress
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Optional, Union
+from typing import Literal, Optional, cast
 
 import requests
+from huggingface_hub import HfApi
 from huggingface_hub.community import DiscussionComment, DiscussionWithDetails
 from huggingface_hub.constants import (
     REPO_TYPE_DATASET,
-    REPO_TYPES,
-    REPO_TYPES_URL_PREFIXES,
 )
-from huggingface_hub.hf_api import HfApi
-from huggingface_hub.utils import hf_raise_for_status
 from libcommon.resources import Resource
 
 from .constants import (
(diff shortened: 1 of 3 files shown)


libs/libcommon/src/libcommon/croissant_utils.py
```diff
@@ -7,6 +7,28 @@
...
+
+    Reasons:
+    - `/` are used in the syntax as delimiters. So we replace them.
+    - Two FileObject/FileSet/RecordSet/Fields cannot have the same ID. So we append a postfix in case it happens.
+
+    Args:
+        id_to_escape: The initial non-escaped ID.
+        ids: The set of already existing IDs.
+    Returns:
+        `str`: The escaped name.
```
marcenacp, line 24: Maybe mention that the returned ID is unique

marcenacp (approved):
Thanks!

ccl-core:
@severo I thought my latest commits would fix the formatting errors and the tests (at least they do in my local env), do you have a clue as to why they still fail? Are the actions running on the latest commit?

severo (approved):
Great!

ccl-core:
Hi @lhoestq @severo , thank you for merging this!
Do you think it could be possible to refresh this croissant? https://huggingface.co/datasets/microsoft/WildFeedback

In particular, this PR should fix the id with space: `"@id": "user_preference/history/turn id"` and the wrong jsonpath: `"jsonPath": "turn id"`.

PR #3230, opened 2025-09-17
build: pin poetry version to 2.1.4

kszucs:


lhoestq (approved):
all green ! except e2e but it can only run on the main repo (while you're on a fork)

merging and let's see if e2e passes

PR #3348, opened 2026-05-23
fix: URL encode path segments when converting hf:// to https://

### LINKED ISSUE #3323: Dataset viewer fails to display images when subset name contains spaces (missing URL encoding)
When a dataset defines subset config names containing spaces in the YAML metadata (e.g., 'config_name: Test name' instead of 'config_name: Test_name'), the dataset viewer fails to render images. Instead, a white warning triangle is displayed in place of each image.

The config name is interpolated directly into the asset URL without URL-encoding spaces. This produces an invalid URL.

qubeena07:
Fixes #3323

When a dataset config or split name contains spaces, the asset URLs returned by the viewer are broken. This happens because the code that converts `hf://` paths to `https://` URLs does a plain string replace with no URL encoding, so spaces end up as literal spaces in the final URL.

For example, a config named `Test name` would produce:
```
https://huggingface.co/datasets/user/repo/resolve/sha/Test name/train/image.jpg
```
instead of:
```
https://huggingface.co/datasets/user/repo/resolve/sha/Test%20name/train/image.jpg
```

The broken URL makes images, audio, and video fail to load in the viewer.

## Changes

Added a `hf_to_https_url()` helper in `url_preparator.py` that splits the `hf://` URL at the `@` separator, then percent-encodes each segment of the file path using `urllib.parse.quote` before assembling the `https://` URL.

Replaced all four places that did the raw `.replace()` conversion:
- `url_preparator.py` `prepare_url()` (video files going through the URL preparator)
- `features.py` `image()` (images stored as HF URL references)
- `features.py` `audio()` (audio stored as HF URL references)  
- `features.py` `video()` (video stored as HF URL references)

Added tests covering paths with spaces in config names and split names.


lhoestq (commented):
lgtm !

PR #3352, opened 2026-05-28
fix(deps): bump starlette to >=1.0.1 monorepo-wide + migrate to ASGI lifespan

jagwar:
## Summary

Address **CVE-2026-48710 / GHSA-86qp-5c8j-p5mr** (Starlette Host header poisoning of `request.url.path`) by bumping `starlette` to `>=1.0.1,<2.0` in the shared `libs/libapi` and coordinating the unavoidable Starlette 1.0 API migration in all 6 services that pass `on_startup`/`on_shutdown` to `Starlette(...)`.

## CVE context

Starlette `<= 1.0.0` reconstructs `request.url` from `http://{Host}{path}` without validating the `Host` header. An attacker sending `Host: example.com/abc?bar=` makes `request.url.path` return `/abc` while the router still dispatches to the real route. Middleware or handlers that gate on `request.url.path` for authorization can be bypassed. Fixed in Starlette 1.0.1.

## Exposure assessment for dataset-viewer

Audit of this repo found **no** code path in any service that uses `request.url.path` for an authorization decision (CORS, auth middleware, JWT verification all rely on headers and the routed handler, not on `request.url.path`). This upgrade is **defense-in-depth**: it closes the upstream framework gap and protects future code paths from this class of bug.

## Why monorepo-wide instead of just one service

`starlette` is pinned in `libs/libapi`, which is a `path-develop` dependency of all 6 API services in this monorepo. Bumping libapi to a Starlette 1.0+ range forces a coordinated migration because Starlette 1.0 **also removed** the `on_startup` / `on_shutdown` constructor kwargs in favor of the ASGI `lifespan` protocol. 6 services use those kwargs today and would break at startup without the migration. So this PR has to land atomically.

- **5 services with sync-only `on_shutdown`** (`resource.release`): `admin`, `api`, `rows`, `search`, `webhook` — straightforward `@asynccontextmanager` wrapping a `try/yield/finally` that calls `resource.release()` for each resource.
- **1 service with mixed sync/async** (`sse-api`): `hub_cache_watcher.start_watching` is sync (`def start_watching(self) -> None`), `hub_cache_watcher.stop_watching` is async (`async def stop_watching(self) -> None`). The lifespan calls `start_watching()` directly and `await stop_watching()` on the shutdown side.

`starlette-prometheus 0.9.0` declares `starlette = ">=0.12.2"` (no upper bound) and stays compatible — no change needed there. A safety grep for other Starlette 1.0-removed patterns (`add_event_handler`, `on_event`, `iscoroutinefunction_or_partial`, etc.) returned no matches outside `on_startup` / `on_shutdown`.

## Why three commits

For easy review:
1. `fix(libapi): pin starlette>=1.0.1 ...` — the manifest bump (1 line + comment).
2. `refactor(services): migrate on_shutdown to ASGI lifespan ...` — the 5 trivial sync-only migrations.
3. `refactor(sse-api): migrate on_startup + on_shutdown ...` — the mixed-mode case, isolated so reviewers can pay extra attention.

## Lockfiles

Lockfile refresh was attempted locally but is **not included** in this PR: the repo pins `python = "3.12.11"` strictly in each `pyproject.toml`, and the local Poetry available could not regenerate locks with the canonical Poetry 2.2.1 format (resulted in noise from a downgrade to 1.8.5 — stripped `groups = ["main"]` markers, platform conditional rewrites, etc., not real dependency changes). The **manifest pin is the source of truth** — CI will regenerate lockfiles with the canonical Python/Poetry. Reviewers: please confirm CI's resolver picks up `starlette 1.0.1+` for libapi and all 6 services.

## Test plan

- [ ] CI green on libapi + all 6 services (admin, api, rows, search, sse-api, webhook)
- [ ] `make install` succeeds in libapi + the 6 services after lockfile regeneration
- [ ] Manual smoke per service: instantiate the app + hit `/healthcheck` — confirm clean startup AND clean shutdown (no leaked Mongo connections)
- [ ] sse-api specifically: confirm `hub_cache_watcher.start_watching()` runs at startup and `await hub_cache_watcher.stop_watching()` runs at shutdown under the new lifespan path

## Cross-reference

Tracking org-wide CVE-2026-48710 audit. Reference:
- huggingface/hf-inference-sdk#70 (merged — manifest pin only, no lifespan needed)
- huggingface/huggingface-inference-toolkit#125 (merged — validated the `@asynccontextmanager` lifespan pattern via remote + local CPU/GPU integration tests)

Follow-ups not in scope here: `huggingface/inferencer` (Pydantic v1->v2 + FastAPI 0.61->0.110+, larger refactor).


lhoestq (approved):
thx !

PR #3356, opened 2026-06-05
add LeRobot library

NikodemBartnik:
Add LeRobot library to the dataset filters.
Connected to 2206 in huggingface.js: https://github.com/huggingface/huggingface.js/pull/2206

### DIFF
--- a/services/worker/tests/job_runners/dataset/test_compatible_libraries.py
+++ b/services/worker/tests/job_runners/dataset/test_compatible_libraries.py
@@ -532,11 +532,7 @@ def test_compute_lerobot(
             {
                 "config_name": "default",
                 "arguments": {},
-                "code": (
-                    "from lerobot.datasets import LeRobotDataset\n"
-                    "\n"
-                    'dataset = LeRobotDataset("lerobot-dataset")'
-                ),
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
lgtm !

services/worker/tests/job_runners/dataset/test_compatible_libraries.py
```diff
@@ -382,6 +392,21 @@ def mock_hffs(tmp_path_factory: TempPathFactory) -> Iterator[fsspec.AbstractFile
...
 
+    (hf / "datasets" / LEROBOT_DATASET / "data" / "chunk-000").mkdir(parents=True)
+    ds.to_parquet(hf / "datasets" / LEROBOT_DATASET / "data" / "chunk-000" / "episode_000000.parquet")
+    (hf / "datasets" / LEROBOT_DATASET / "README.md").write_text(
+        "---\n"
+        "license: apache-2.0\n"
+        "task_categories:\n"
+        "- robotics\n"
+        "tags:\n"
+        "- LeRobot\n"
```
julien-c, line 403: if possible keep this case-sensitive because tag matching is case-sensitive on the Hub

lhoestq:
we can also add an heuristic to make it work even if the lerobot tag is not present btw @NikodemBartnik

PR #3366, opened 2026-06-18
feat: Preserve row-level agent traces when first rows fit

armand0e:
## Summary

- let `split-first-rows` keep the full `trace` cell for agent-trace datasets when the response still fits the configured byte budget
- fall back to the existing truncation behavior when keeping `trace` would exceed the response budget
- add regression coverage for both the preserved-trace and fallback-truncation paths

## Why

Hermes exports can store multiple sessions in one `sessions.jsonl` file. The dataset rows already contain a parsed row-level `trace`, but the shared `file_path` points every session at the same multi-session source file. Keeping the row-level trace available when it fits gives the Hub trace UI enough data to render a session detail without requiring one repo file per Hermes session.

## Validation

- `uvx ruff format libs/libcommon/src/libcommon/viewer_utils/rows.py libs/libcommon/tests/viewer_utils/test_rows.py`
- `uvx ruff check libs/libcommon/src/libcommon/viewer_utils/rows.py libs/libcommon/tests/viewer_utils/test_rows.py`
- `cd libs/libcommon && uv run python -m pytest -q tests/viewer_utils/test_rows.py -k agent_trace -s`
- Sanity checked live `armand0e/hermes-test` rows through the patched helper: `trace` remains untruncated and the response is ~376 KB under a 1 MB budget.

Full `tests/viewer_utils/test_rows.py` does not complete in this local environment because existing audio fixture setup requires FFmpeg libraries for TorchCodec (`libavutil.so.*` missing).

### DIFF
--- a/libs/libcommon/src/libcommon/viewer_utils/rows.py
+++ b/libs/libcommon/src/libcommon/viewer_utils/rows.py
@@ -2,13 +2,14 @@
 # Copyright 2022 The HuggingFace Authors.
 
 
+from copy import deepcopy
 from typing import Optional, Protocol
 
 import PIL
 from datasets import Audio, Features, Image, Pdf, Value, Video
 from datasets.packaged_modules.json.json import AGENT_TRACES_FEATURES
 
-from libcommon.dtos import Row, RowsContent, SplitFirstRowsResponse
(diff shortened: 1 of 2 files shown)


armand0e:
Worked with codex to implement this PR in a way that I think it should work. However I'm not super sure how I could actually test it 🤷 Hopefully this is helpful

PR #3367, opened 2026-07-13
bump fsspec and allow-list fs implementations

lhoestq:
benefit from the latest fixes and make sure only the protocols we use are active (in particular remove access to the simplecache implementation and the reference and data implementations)


rajatarya (commented):
Looks good, sorry didn't review in time 👍🏾 :)

PR #3378, opened 2026-07-20
Better optimized parquet detection

lhoestq:
Detect CDC based on the coefficient of variation of parquet page sizes, if the metadata in the parquet footer are not present

will be useful for a soon-ish cool code dataset release 👀 (dataset was created with spark and doesn't have the metadata)

cc @kszucs


libs/libviewer/src/parquet.rs
```diff
@@ -157,3 +157,219 @@ pub fn read_batch_stream(
...
+///
+/// Detection criteria:
+/// 1. Page index must be present (offset_index available)
+/// 2. At least one column must have multiple pages with variable sizes
+/// 3. Page sizes must vary significantly (not all the same within ~5% tolerance)
+///
+/// This is more robust than checking for the `content_defined_chunking` metadata key
+/// which is only added by the `datasets` library.
+pub fn is_content_defined_chunked(metadata: &ParquetMetaData) -> bool {
+    // Check if offset index is present (indicates page index was written)
```
kszucs, line 181: It can still be CDC chunked even without offset index present, we just enable both CDC and offset index in datasets but CDC chunking can be enabled separately.

kszucs: We could have two approaches:
1. detect cdc from offset index: only reads metadata
2. detect cdc from page reader: needs to seek data but could early exit

2) could be the fallback of 1) for more robust detection

Ideally we should place the metadata in the parqeut writers directly, though there isn't a standard convention for the KV metadata.

libs/libviewer/src/parquet.rs
```diff
@@ -157,3 +157,219 @@ pub fn read_batch_stream(
...
+) -> std::result::Result<Arc<ParquetMetaData>, ParquetError> {
+    if !path.starts_with("hf://datasets/") {
+        return Err(ParquetError::General("Invalid hf:// URL".to_string()));
+    }
+
+    let rest = &path["hf://datasets/".len()..];
+
+    // Parse hf://datasets/<repo_id>[@<revision>]/<file_path>
+    // where repo_id = <username>/<dataset_name>
+    let (repo_id, revision, file_path) = if let Some(at_idx) = rest.find('@') {
```
kszucs, line 286: We need this because we use a pinned older opendal here (0.54.1 + patch). If we would update to the latest opendal release we could delete this logic and use something like 

```rust
let operator = opendal::Operator::from_uri(
    "hf://datasets/username/my_dataset@main/path/to/file.parquet",
    // extra options as (key, value) pairs, e.g. token/endpoint
    vec![
        ("token", hf_token.unwrap_or("")),
        ("endpoint", hf_endpoint.unwrap_or("https://huggingface.co")),
    ],
)?;
```

though we need to double check that the unconditional stat() calling is fixed (I fixed it upstream but other prs have touched that code since then)

services/worker/src/worker/job_runners/dataset/compatible_libraries.py
```diff
@@ -950,20 +969,56 @@ def is_optimized_parquet_from_first_polars_loading_code(
...
+        try:
+            fs = HfFileSystem(token=hf_token, endpoint_url=hf_endpoint)
+            parquet_path = f"datasets/{dataset}/{first_split_pattern}"
+            with fs.open(parquet_path, "rb") as f:
                 with pq.ParquetFile(f) as pf:
-                    if b"content_defined_chunking" in pf.metadata.metadata:
-                        # TODO: also check that the page index is available
-                        # in the Parquet metadata viewer it corresponds
-                        # e.g. to "row_groups.0.columns.0.offset_index_offset"
+                    if b"use_content_defined_chunking" in pf.metadata.metadata:
```
kszucs: We could also check this in rust to avoid reading twice from the hub in case of a fallback.

services/worker/src/worker/job_runners/dataset/compatible_libraries.py
```diff
@@ -950,20 +969,56 @@ def is_optimized_parquet_from_first_polars_loading_code(
...
+        # Fall back to Rust-based CDC detection (page size variability)
+        if HAS_LIBVIEWER:
+            try:
+                return bool(is_optimized_parquet_from_hub(hf_url, hf_token, hf_endpoint))
+            except Exception as e:
+                logging.debug("Failed to check Rust CDC detection: %s", e)
     return False
 
 
+def _check_page_index_with_pyarrow(metadata: Any) -> bool:
```
kszucs: Is there a case when libviewer is not available?

libs/libviewer/src/parquet.rs
lhoestq: I see, I'd like to to try to have 1. merged at least for the dataset that will be released, and we can see later for 2.

services/worker/src/worker/job_runners/dataset/compatible_libraries.py
lhoestq: nope, this function is a leftover - let me remove it

lhoestq:
I'm keeping the old opendal syntax for now, we can update later

let me knowif you have other comments, otherwise I think we're good

kszucs:
> I'm keeping the old opendal syntax for now, we can update later
> 
> let me knowif you have other comments, otherwise I think we're good

SGTM

PR #3381, opened 2026-07-22
fix: temporarily disable raw Arrow datasets

XciD:
## Summary

- temporarily reject the packaged raw Arrow builder in `safe_load_dataset_builder`
- return a cacheable HTTP 501 response instead of reading Arrow IPC files
- cover both the blocked Arrow builder and an allowed non-Arrow builder with unit tests

## Context

This is an immediate containment measure while huggingface/datasets#8350 is reviewed and deployed. The guard is placed in the shared safe builder loader so it covers both streaming first rows and Parquet conversion workers. It can be removed after the validated datasets dependency and the hardened PyArrow runtime are deployed.

## Validation

- `poetry run python -m pytest tests/test_utils.py -q`, 25 passed
- `poetry run ruff check` and `poetry run ruff format --check` on the changed files
- mypy and Bandit on the changed files
- `git diff --check`

### DIFF
--- a/libs/libcommon/src/libcommon/exceptions.py
+++ b/libs/libcommon/src/libcommon/exceptions.py
@@ -83,6 +83,7 @@ CacheableErrorCode = Literal[
     "DatasetGenerationCastError",
     "DatasetInBlockListError",
     "DatasetNotFoundError",
+    "DatasetWithArrowFilesNotSupportedError",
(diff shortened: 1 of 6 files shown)


lhoestq (approved):

lhoestq:
You can ignore the one failure in CI worker and merge

PR #3388, opened 2026-07-27
feat(secrets): mount secrets as files instead of injecting them in the environment

XciD:
## Why

The secrets reach the pods in two hops today: the Infisical operator syncs the whole secret path into a Kubernetes Secret, then every pod mounts individual keys as environment variables through `valueFrom.secretKeyRef`.

That leaves the material in etcd, readable by anyone with `get secrets` on the namespace, present in each pod's `/proc/<pid>/environ`, inherited by every subprocess, and captured by whatever an error reporter or APM sends off-site.

## What

Each pod mounts only the secrets it needs as files in a read-only tmpfs volume, through the Secrets Store CSI driver, and reads them once at startup into process memory. No Kubernetes Secret is created and nothing is injected in the environment.

The node's CSI provider is what talks to Infisical, so **the pods hold no credential of their own**. An application-side compromise cannot ask for a different secret path or a different environment; it can only read the files that were mounted for it.

## Blast radius is now explicit

`secrets.infisical.csi.workloads` lists what each workload mounts, and the chart renders one `SecretProviderClass` per workload from it. A pod never holds the whole project: `mongodb-migration` mounts one file, `cache-metrics-collector` two, `webhook` nine.

That list is the blast radius of a compromised pod, so it is written to be read and audited rather than defaulted.

## libcommon knows nothing about Infisical

The contract is a directory of files, so the same code works behind any CSI provider or injector. `libcommon/secrets.py` is about twenty lines of logic: list the directory, read each file, cache in memory, fall back to the environment variable when nothing is mounted.

It deliberately never writes back to `os.environ`, which would move the leak from the pod spec to `/proc/self/environ` and to every subprocess. The 18 call sites did not change: `resolve_secret(env, name, key, default)` kept its signature, only the source changed.

An **empty file means "not there yet"**, not an absent secret. A provider that bounds how long secrets stay readable serves empty files once that window closes, and a freshly granted window takes a few seconds to take effect. The first read waits that out, then fails loudly rather than booting with a secret missing.

## Rollout

Off by default and mutually exclusive with the operator mode, which is kept as is for rollback; the chart refuses to render if both are enabled. With it off, `helm template` on `prod.yaml` and `staging.yaml` is unchanged apart from the new `S3_USE_IRSA` variable, a non-secret boolean mirroring the existing `s3.useIrsa` value. Merging changes nothing until a deployment opts in.

Turning it on needs two things that do not live here: the Secrets Store CSI driver and the Infisical provider installed on the cluster, and an Infisical machine identity the provider authenticates with. Its ID, the project ID and the instance URL are injected at deploy time by the private overlay, following the same convention as the IRSA role ARN.

## Optionally, a read window

Setting `windowDuration` makes the provider serve the secrets only while the consuming containers are freshly started, and empty files afterwards. That shrinks the exposure of an application-side compromise from the pod's whole life to a few minutes around startup.

It needs a provider that supports it. Measured against a self-hosted instance: expiry takes effect about two seconds after the window closes, and granting one takes about eleven, which is why five minutes is a saner floor than one and why the application retries its first read.

## Notes for review

- `S3Config.from_env` skips resolving the static keys when `S3_USE_IRSA` is true. Without that guard, enabling this alongside `s3.useIrsa: true` would pull the static credentials back and have them take precedence over the pod's role.
- A volume is the unit of enforcement while a container is the unit of decision, so a tight window wants one consumer per volume. Documented in the chart README.
- The secret lists mirror what the chart injects today, which is broader than strictly necessary in places: every service that includes `envHf` gets `WEBHOOK_SECRET`, though only `webhook` verifies webhooks. Tightening that is a follow-up, kept out of here so the change stays mechanical.


rtrompier (approved):

lhoestq (approved):

lhoestq:
e2e got 404 errors from hub-ci but I'm not sure it's related

ah and it makes me think that we want to make sure the secrets are passed to subprocesses, e.g. when using WorkerExecutor in `services/worker/src/worker/executor.py`. If I understand correctly the subprocess will also read the secrets from local files on startup so we should be fine.

EDIT: The issue may arise for uvicorn workers for the `api`, `rows`, `search` and `admin` services though, that could restart but local secret files may be gone. Especially `rows` and `search` uvicorn workers since they could be subject to OOM

lhoestq:
maybe hacky but we could pass app_config to the uvicorn.Config, and patch uvicorn.Config.configure_logging() which runs on a worker startup to set the secrets like `libcommon.config.CACHE_MONGO_URL = app_config.cache.mongo_url` (or use `os.environ` but this is not the goal iiuc)

PR #3403, opened 2026-08-17
fix(worker): add SSRF egress guard blocking internal/link-local fetches

rtrompier:
## The vulnerability

`datasets` sends a HEAD request through `huggingface_hub` (httpx) from `_prepare_path_and_storage_options` for any URL containing the substring `drive.google.com` — which a query string satisfies. Both `xopen` and `resolve_pattern` call that helper *before* they look the protocol up in fsspec, so the request is sent upstream of the allow-list in `libcommon/fsspec.py`, which never sees it. A dataset can therefore make the worker request an arbitrary host and port from inside the cluster.

Reproduced live from production on `dataset-config-names` (a README-only dataset, first job in the pipeline) and on `split-descriptive-statistics` (attacker URL in Parquet content, so no `data_files` check sees it). Blind SSRF: HEAD only, no read primitive, and the production requests carried no `Authorization` or `Cookie`.

## The fix

| Route | Load-bearing check |
|---|---|
| `dataset-config-names`, `dataset-init`, `config-split-names`, `split-first-rows`, `config-parquet-and-info`, `dataset-compatible-libraries` | `allow_only_relative_data_files()` — refuses a `data_files` pattern that is not relative to the repository, before `resolve_pattern` |
| `split-descriptive-statistics` media | the `hf://` check in `MediaColumn.open`, before `xopen` |
| parquet external-shard probe | the last-hop `hf://` check in `get_total_files_size` |
| fsspec/aiohttp transport | `libcommon/fsspec.py` now evicts the live registry as well as `known_implementations`, so the protocol removal is order-independent (previously a protocol imported before `libcommon` stayed usable — and only under the name it was imported with, so `https` worked while `http` was refused) |
| any transport, any route | `libcommon.ssrf` — refuses non-public addresses (loopback, RFC 1918, CGNAT, link-local, unique-local, multicast, reserved) on both aiohttp (fsspec, s3fs) and httpx (huggingface_hub) |

**The `hf://` path checks — not the fsspec registry eviction — are what close the live routes**, because the HEAD is sent upstream of fsspec entirely. Please don't drop them as redundant with the allow-list; the ordering comments and the `assert_not_called()` assertions in the tests exist to pin that.

## Residual risk

- DNS rebinding is closed on both transports: each connects to an address it has already checked instead of resolving the host again. TLS is unaffected, since `aiohttp` and `httpcore` take the name used for SNI and for the certificate check from the request URL, not from the address connected to.
- An `HTTP(S)_PROXY` in the environment (`trust_env`) routes httpx through `_mounts` and bypasses the guarded transport, so the proxy would be checked rather than the eventual destination. No environment configures one.
- The Rust transports cannot be hooked from Python: `libviewer` (opendal), `hf_xet`, `pylance`, and duckdb's extension installer, which runs `INSTALL 'fts'` before `SET enable_external_access=false`. Their destinations are the HF endpoint or operator configuration.
- The media path check requires `hf://` but is not scoped to the dataset's own repository, so a cross-repo read on the Hub remains possible. That is token scoping, not SSRF.
- The `"drive.google.com"` substring test is an upstream `huggingface_hub`/`datasets` bug (`datasets/utils/file_utils.py:921`): it fires a HEAD at an unvalidated host from a helper that runs before any protocol check, so any future caller of `xopen` or `resolve_pattern` re-inherits it. To be reported through the security process rather than as a public issue.

## Rollback

Removing the `_install_ssrf_guard()` call in `services/worker/src/worker/__init__.py` restores the previous transport behaviour on its own. The path checks and the `fsspec.py` change should not be reverted with it — those are what close the live routes and the allow-list bypass. No migration, no configuration change.

### DIFF
--- a/libs/libcommon/src/libcommon/fsspec.py
+++ b/libs/libcommon/src/libcommon/fsspec.py
@@ -1,5 +1,15 @@
-from fsspec.registry import known_implementations
+from fsspec.registry import _registry, known_implementations
+
+ALLOWED_PROTOCOLS = ["hf", "s3", "zip", "file", "local"]
 
 for name in list(known_implementations):
-    if name not in ["hf", "s3", "zip", "file", "local"]:
+    if name not in ALLOWED_PROTOCOLS:
         del known_implementations[name]
+
+# `get_filesystem_class` looks the protocol up in the registry of the already imported filesystems
+# before falling back to `known_implementations`, so removing it from the latter is not enough: a
+# protocol that was imported before this module runs would stay usable. Note that it only stays
+# usable under the name it was imported with, e.g. "https" keeps working while "http" is refused.
+for name in list(_registry):
+    if name not in ALLOWED_PROTOCOLS:
+        del _registry[name]
--- a/libs/libcommon/src/libcommon/ssrf.py
+++ b/libs/libcommon/src/libcommon/ssrf.py
@@ -1,0 +1,209 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 The HuggingFace Authors.
+
+"""Egress guard against SSRF.
+
+The worker fetches URLs that come from dataset content: the `data_files` patterns declared in a
+dataset card, and the `path` of the media values stored in a dataset's Parquet files. Without a
+guard, a dataset can make the worker request internal endpoints: the cloud metadata service,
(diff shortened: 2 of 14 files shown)


lhoestq (approved):
lgtm !

PR #3409, opened 2026-08-19
fix(libviewer): bump Rust dependencies flagged by Dependabot

XciD:
Fixes the open Dependabot alerts on `libs/libviewer/Cargo.lock`.

## Bumps

| Crate | From | To | Alerts |
|---|---|---|---|
| pyo3 | 0.26.0 | 0.29.2 | OOB read in `nth`/`nth_back` for `PyList`/`PyTuple` slices (high), missing `Sync` bound on `new_closure` (medium) |
| quinn-proto | 0.11.13 | 0.11.17 | QUIC transport parameter panic DoS (high), out-of-order stream memory exhaustion (high) |
| rustls-webpki | 0.103.8 | 0.103.14 | CRL BIT STRING panic DoS (high), name constraint bypasses |
| bytes | 1.10.1 | 1.12.1 | `BytesMut::reserve` integer overflow (medium) |
| lz4_flex | 0.11.5 | dropped (only >= 0.13 remains) | uninitialized memory leak (high) |
| rand | 0.9.2 | dropped (only 0.10 remains) | unsound rng with custom logger (low) |
| thrift | 0.17.0 | removed | excessive allocation (medium); it was an unused direct dependency |

## Support changes needed by the bumps

- arrow/parquet 57 -> 59.1: pyo3 0.29 needs the newer pyarrow bridge. `with_page_index(true)` becomes `with_page_index_policy(PageIndexPolicy::Optional)`, and `PyDataset` needs the `from_py_object` pyclass attribute. parquet 59 deprecates `ParquetObjectReader` (warning only, not migrated here).
- object_store 0.12 -> 0.13.2: `get()` moves to the `ObjectStoreExt` trait.
- opendal/object_store_opendal move from the `kszucs/opendal` fork branch to a pinned upstream `apache/opendal` revision that contains the hf-unconditional-stat fix, replacing the `[patch.crates-io]` section. `Operator::new` no longer needs `finish()`.

`cargo check`, `cargo clippy` and `cargo fmt --check` pass locally (4 pre-existing-style deprecation warnings from parquet 59).

### DIFF
--- a/libs/libviewer/src/dataset.rs
+++ b/libs/libviewer/src/dataset.rs
@@ -101,7 +101,11 @@ impl Dataset {
         hf_endpoint: Option<&str>,
     ) -> Result<Self> {
         // Initialize the data store (Huggingface in this case)
-        let mut builder = Huggingface::default().repo_type("dataset").repo_id(name);
+        // upstream opendal defaults to xet downloads; keep the plain HTTP path the fork used
+        let mut builder = Huggingface::default()
+            .download_mode("http")
+            .repo_type("dataset")
+            .repo_id(name);
         if let Some(token) = hf_token {
             builder = builder.token(token);
         }
--- a/libs/libviewer/src/dataset.rs
+++ b/libs/libviewer/src/dataset.rs
@@ -111,7 +111,7 @@ impl Dataset {
         if let Some(endpoint) = hf_endpoint {
             builder = builder.endpoint(endpoint);
         }
-        let operator = Operator::new(builder)?.finish();
+        let operator = Operator::new(builder)?;
         let data_store = Arc::new(OpendalStore::new(operator));
 
         // Initialize the metadata store from the given URI, usually a local directory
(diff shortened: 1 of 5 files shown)


libs/libviewer/src/dataset.rs
```diff
@@ -101,7 +101,11 @@ impl Dataset {
         hf_endpoint: Option<&str>,
     ) -> Result<Self> {
         // Initialize the data store (Huggingface in this case)
-        let mut builder = Huggingface::default().repo_type("dataset").repo_id(name);
+        // upstream opendal defaults to xet downloads; keep the plain HTTP path the fork used
+        let mut builder = Huggingface::default()
+            .download_mode("http")
```
lhoestq, line 106: we could use xet here no ? cc @kszucs

lhoestq (approved):
otherwise lgtm ! let's merge and see later if we need to improve or not

PR #3424, opened 2026-09-02
Add environment dataset snippets

burtenshaw:
This PR derives environment framework snippets from Dataset Card tags.

### DIFF
--- a/libs/libcommon/src/libcommon/processing_graph.py
+++ b/libs/libcommon/src/libcommon/processing_graph.py
@@ -684,7 +684,7 @@ specification: ProcessingGraphSpecification = {
     "dataset-compatible-libraries": {
         "input_type": "dataset",
         "triggered_by": "dataset-info",
-        "job_runner_version": 6,
+        "job_runner_version": 7,
         "difficulty": 20,
     },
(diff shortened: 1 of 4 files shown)


libs/libcommon/src/libcommon/processing_graph.py
```diff
@@ -684,7 +684,7 @@ def parse_id(id: str) -> tuple[str, str, Optional[str], Optional[str], str]:
     "dataset-compatible-libraries": {
         "input_type": "dataset",
         "triggered_by": "dataset-info",
-        "job_runner_version": 6,
+        "job_runner_version": 7,
```
lhoestq, line 687: Let's keep it at 6, otherwise it will recompute for all datasets

I can manually update a list of datasets, it will be faster
```suggestion
        "job_runner_version": 6,
```

lhoestq (commented):
Great ! Is it possible to autodetect environment datasets without using tags ? That would make the detection more robust

(and btw could an environment dataset have multiple subsets/configs ?)

I'd wait for the other PRs to get reviews before merging if you don't mind

lhoestq (approved):
ok for me !

PR #3426, opened 2026-09-03
Add support for vortex backed datasets

robert3005:
Upgrades hash of datasets to the version that supports vortex and adds necessary
plumbing to be able to work on vortex backed datasets.

Majority of the diff is pyarrow upgrade


robert3005:
@lhoestq this is the follow up to the datasets pr. Majority of the change is from the dependency bump

PR #3431, opened 2026-09-14
Recompute datasets with cached errors when switching to public

### LINKED ISSUE #3429: Switching a dataset from private to public does not trigger a recompute
When a (PRO/Enterprise) dataset is switched from private to public with no new commit, the webhook skips the update because `headSha` equals the cached revision and `updatedConfig` is `{"private": false}`:

https://github.com/huggingface/dataset-viewer/blob/0cb4515aaa0122362a30b5bf14ec3c16d08f24a0/services/webhook/src/webhook/routes/webhook.py#L95-L105

The comment there says this is intentional, but the cache can contain errors that only happen while the repo is private. Example: `Mike0021/pickup-truck-transcripts-sanitized-9d20ec42842b` failed while private (`FileNotFoundError: ... (repository not found)`), was made public at the same revision, and still shows the error.

Expected: switching from private to public recomputes the dataset even if the revision is unchanged.

cfahlgren1:
When a dataset switches from private to public without a new commit, the webhook currently skips processing even if cached results contain errors from private access. Recompute when that visibility change has cached errors at the current revision, while preserving the skip for successful datasets and conversion-branch updates.

Fixes #3429.


lhoestq:
tbh it's the first time I see this RepositoryNotFoundError and I'm not sure why it happened

I think this change is likely going to cause more trouble for people having a dataset in private that wait for the viewer to be ready before publishing. It often happens for big datasets - i.e. the ones which are likely to have at least one error among their 1000 subsets. So I would either not merge, or investigate more the original issue, or only catch RepositoryNotFoundError instead of any error

Actually I just duplicated the repo to a private repo and reproduced the error so I would rather investigate the original issue

cfahlgren1:
sounds good @lhoestq, i'm with you on that as well. closing!

